PR #7, opened 2019-08-07
updated to "key in index" from "contains" to suppress FutureWarning

olliethomas:
using FairLearn throws 2 FutureWarnings

```
fairlearn/classred.py:309: FutureWarning: The 'contains' method is deprecated and will be removed in a future version. Use 'key in index' instead of 'index.contains(key)'
```

```
fairlearn/classred.py:358: FutureWarning: The 'contains' method is deprecated and will be removed in a future version. Use 'key in index' instead of 'index.contains(key)'
```

This PR just implements the suggested changes


MiroDudik:
Merged with a small style fix.

PR #9, opened 2019-08-12
Convert tests to PyTest

riedgar-ms:
Convert the existing set of tests to PyTest
Also add a .gitignore file

### DIFF
--- a/test_fairlearn_smoke.py
+++ b/test_fairlearn_smoke.py
@@ -1,0 +1,103 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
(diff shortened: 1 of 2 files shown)


test_fairlearn_smoke.py
```diff
@@ -0,0 +1,98 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
romanlutz: So this adds the same file reformatted as pytest? What about the existing test_fairlearn.py?

riedgar-ms: We might want to remove it eventually. I think that Miro would prefer to keep the existing test script as well for now

PR #10, opened 2019-08-14
Set up CI with Azure Pipelines

romanlutz:


requirements.txt
```diff
@@ -0,0 +1,4 @@
+numpy
```
riedgar-ms: Not now, but we need to discuss version locking

azure-pipelines.yml
```diff
@@ -0,0 +1,18 @@
...
+- master
+- dev
+
+pool:
+  vmImage: 'ubuntu-latest'
+
+steps:
+- task: UsePythonVersion@0
+  inputs:
+    versionSpec: '3.5'
```
riedgar-ms: Might need to consider other python versions, but this is fine for now

azure-pipelines.yml
```diff
@@ -0,0 +1,18 @@
+trigger:
+- master
```
riedgar-ms: I like doing this on master, but surely the pytest tests don't exist there yet.....

requirements.txt
romanlutz: definitely!

azure-pipelines.yml
romanlutz: Yes, it's on my radar, but not my top priority

azure-pipelines.yml
romanlutz: next time we merge they will :-)

romanlutz: Should I remove it?

romanlutz: Actually since this is only in the dev  branch it won't affect master until we check this into master.

PR #12, opened 2019-08-16
Initial implementation of 1d grid search for binary classification

riedgar-ms:
An initial implementation of a 1d grid search for binary classification, along with some simple tests

Comments both algorithmic and pythonic welcome

### DIFF
--- a/fairlearn/grid_search.py
+++ b/fairlearn/grid_search.py
@@ -17,11 +17,16 @@ from collections import namedtuple
 def _generate_p0_p1(Y):
     """ Function to compute p0 and p1 for the given
     set of labels Y
+
+    Assume we have already checked that Y is binary with
+    values 0 and 1
     """
     unique, counts = np.unique(Y, return_counts=True)
 
-    
-    return 0,1
+    p0 = counts[0] / len(Y)
+    p1 = counts[1] / len(Y)
+
+    return p0, p1
 
 def grid_search_classification_binary_protected_1d(learner, X, Y, A):
     """Function to generate a 1d set of models for a classification problem with
--- a/fairlearn/grid_search.py
+++ b/fairlearn/grid_search.py
@@ -47,14 +47,14 @@ def _generate_weights(Y, A, L, p_ratio):
 
         W.append(w)
 
-    return W
+    return np.array(W)
 
 def classification_binary_protected_1d(learner, X, Y, A, Ls=None, num_Ls = 11):
     """Function to generate a 1d set of models for a classification problem with
     a single binary protected attribute
     """
     # Must specify Ls or num_Ls
-    if not( Ls is None ^ num_Ls is None):
+    if not( (Ls is None) ^ (num_Ls is None)):
         raise RuntimeError("Must specify either Ls or num_Ls")
 
     # Check that X, Y and A have the same number of rows
@@ -95,7 +95,7 @@ def classification_binary_protected_1d(learner, X, Y, A, Ls=None, num_Ls = 11):
 
         # Generate Y'
         f = lambda x: 1 if x>0 else 0
-        Yprime = map(f, Y)
+        Yprime = np.vectorize(f)(Y)
 
         # Run the learner
         model = learner.fit(abs(W), X, Yprime)
--- a/fairlearn/grid_search.py
+++ b/fairlearn/grid_search.py
@@ -14,6 +14,15 @@ import numpy as np
 import pandas as pd
 from collections import namedtuple
 
+def _generate_p0_p1(Y):
+    """ Function to compute p0 and p1 for the given
+    set of labels Y
+    """
+    unique, counts = np.unique(Y, return_counts=True)
+
+    
+    return 0,1
+
 def grid_search_classification_binary_protected_1d(learner, X, Y, A):
     """Function to generate a 1d set of models for a classification problem with
     a single binary protected attribute
--- a/fairlearn/grid_search.py
+++ b/fairlearn/grid_search.py
@@ -34,24 +34,44 @@ def _generate_p0_p1(Y):
 
     return p0, p1
 
-def grid_search_classification_binary_protected_1d(learner, X, Y, A):
+def classification_binary_protected_1d(learner, X, Y, A, Ls=None, num_Ls = 11):
     """Function to generate a 1d set of models for a classification problem with
     a single binary protected attribute
-
     """
+    # Must specify Ls or num_Ls
+    if not( Ls is None ^ num_Ls is None):
+        raise RuntimeError("Must specify either Ls or num_Ls")
 
     # Check that X, Y and A have the same number of rows
-
     # Check that Y and A are both 1d vectors
+    if len(Y.shape) != 1:
+        raise RuntimeError("Supplied Y not 1d vector")
+
+    if len(A.shape) != 1:
+        raise RuntimeError("Supplied A not 1d vector")
+
+    if A.shape[0] != Y.shape[0]:
+        raise RuntimeError("Supplied A and Y not same length")
+
+    if X.shape[0] != Y.shape[0]:
+        raise RuntimeError("Supplied X and Y do not have same number of rows")
 
     # Check that A only has values 0 and 1
-    
+    uniqueA = np.unique(A)
+    if not np.array_equal(uniqueA, [0,1]):
+        raise RuntimeError("Supplied A labels not 0 or 1")
+
     # Compute p0 and p1
     # This will also check that Y is binary with values
     # 0 and 1
     p0, p1 = _generate_p0_p1(Y)
 
     # If not supplied, generate array of trial Lagrangian multipliers
+    if Ls is None:
+        limit = 1
+        if p0/p1 > 1:
+            limit = p0/p1
+        Ls = np.linspace(-2*limit, 2*limit, num_Ls)
 
     for L in Ls:
         # Generate W array
--- a/fairlearn/grid_search.py
+++ b/fairlearn/grid_search.py
@@ -1,0 +1,48 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+"""
+This module implements a 1d grid search for binary classification with a single
+binary protected attribute.
+
+This is a simplification of the full exponentiated gradient search
+and produces a sequence of models, each of which corresponds to a
+different accuracy/fairness trade-off
+"""
+
+import numpy as np
+import pandas as pd
+from collections import namedtuple
+
+def grid_search_classification_binary_protected_1d(learner, X, Y, A):
+    """Function to generate a 1d set of models for a classification problem with
+    a single binary protected attribute
+
+    """
+
+    # Check that X, Y and A have the same number of rows
+
+    # Check that Y and A are both 1d vectors
+
+    # Check that Y only has values 0 and 1
+
+    # Check that A only has values 0 and 1
+
+    # If needed, check Y only has values 0 and 1
+    
+    # Compute p0 and p1
+
+    # If not supplied, generate array of trial Lagrangian multipliers
+
+    for L in Ls:
+        # Generate W array
+
+        # Generate Y'
+
+        # Run the learner
+        model = learner.fit(abs(W), X, Yprime)
+
+        # Append the new learner, along with its L value to the result
+
+    # Return the result array (tuples of (L,model))
+
--- a/fairlearn/grid_search.py
+++ b/fairlearn/grid_search.py
@@ -23,6 +23,12 @@ def _generate_p0_p1(Y):
     """
     unique, counts = np.unique(Y, return_counts=True)
 
+    if len(unique) != 2:
+        raise RuntimeError("Supplied Y not binary")
+
+    if not np.array_equal(unique, [0, 1]):
+        raise RuntimeError("Supplied Y labels not 0 or 1")
+
     p0 = counts[0] / len(Y)
     p1 = counts[1] / len(Y)
 
@@ -38,13 +44,12 @@ def grid_search_classification_binary_protected_1d(learner, X, Y, A):
 
     # Check that Y and A are both 1d vectors
 
-    # Check that Y only has values 0 and 1
-
     # Check that A only has values 0 and 1
-
-    # If needed, check Y only has values 0 and 1
     
     # Compute p0 and p1
+    # This will also check that Y is binary with values
+    # 0 and 1
+    p0, p1 = _generate_p0_p1(Y)
 
     # If not supplied, generate array of trial Lagrangian multipliers
 
--- a/fairlearn/grid_search.py
+++ b/fairlearn/grid_search.py
@@ -10,7 +10,7 @@ and produces a sequence of models, each of which corresponds to a
 different accuracy/fairness trade-off
 """
 
-import itertools
+import copy
 import numpy as np
 import pandas as pd
 from collections import namedtuple
@@ -98,10 +98,11 @@ def classification_binary_protected_1d(learner, X, Y, A, Ls=None, num_Ls = 11):
         Yprime = np.vectorize(f)(Y)
 
         # Run the learner
-        model = learner.fit(abs(W), X, Yprime)
+        mylearner = copy.deepcopy(learner)
+        mylearner.fit(X, Yprime, np.absolute(W))
 
         # Append the new learner, along with its L value to the result
-        result.append({"model":model, "lambda":L})
+        result.append({"model": mylearner, "lambda":L})
 
     # Return the result array (tuples of (L,model))
     return result
--- a/fairlearn/grid_search.py
+++ b/fairlearn/grid_search.py
@@ -10,6 +10,7 @@ and produces a sequence of models, each of which corresponds to a
 different accuracy/fairness trade-off
 """
 
+import itertools
 import numpy as np
 import pandas as pd
 from collections import namedtuple
@@ -33,6 +34,20 @@ def _generate_p0_p1(Y):
     p1 = counts[1] / len(Y)
 
     return p0, p1
+
+def _generate_weights(Y, A, L, p_ratio):
+    W = []
+
+    for y, a in zip(Y, A):
+        w = 1e128
+        if a==0:
+            w = 2*y - 1 - L * p_ratio
+        else:
+            w = 2*y - 1 + L
+
+        W.append(w)
+
+    return W
 
 def classification_binary_protected_1d(learner, X, Y, A, Ls=None, num_Ls = 11):
     """Function to generate a 1d set of models for a classification problem with
@@ -73,15 +88,21 @@ def classification_binary_protected_1d(learner, X, Y, A, Ls=None, num_Ls = 11):
             limit = p0/p1
         Ls = np.linspace(-2*limit, 2*limit, num_Ls)
 
+    result = []
     for L in Ls:
         # Generate W array
+        W = _generate_weights(Y, A, L, p1/p0)
 
         # Generate Y'
+        f = lambda x: 1 if x>0 else 0
+        Yprime = map(f, Y)
 
         # Run the learner
         model = learner.fit(abs(W), X, Yprime)
 
         # Append the new learner, along with its L value to the result
+        result.append({"model":model, "lambda":L})
 
     # Return the result array (tuples of (L,model))
+    return result
 
--- a/.gitignore
+++ b/.gitignore
@@ -102,3 +102,9 @@ venv.bak/
 
 # mypy
 .mypy_cache/
+
+# Visual Studio
+/.vs
+
+# VS Code
+/.vscode
--- a/test_fairlearn_smoke.py
+++ b/test_fairlearn_smoke.py
@@ -1,0 +1,98 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import numpy as np
+import pandas as pd
+import fairlearn.moments as moments
+import fairlearn.classred as red
+
+import pytest
+
+
+class LeastSquaresLearner:
+    def __init__(self):
+        self.weights = None
+
+    def fit(self, X, Y, W):
+        sqrtW = np.sqrt(W)
+        matX = np.array(X) * sqrtW[:, np.newaxis]
+        vecY = Y * sqrtW
+        self.lsqinfo = np.linalg.lstsq(matX, vecY, rcond=-1)
+        self.weights = pd.Series(self.lsqinfo[0], index=list(X))
+
+    def predict(self, X):
+        pred = X.dot(self.weights)
+        return 1*(pred > 0.5)
+
+
+class TestFairLearnSmoke:
+    def setup_method(self, method):
+        print("setup_method      method:%s" % method.__name__)
+        attrs = [str(x) for x in 'AAAAAAA' 'BBBBBBB' 'CCCCCC']
+        labls = [int(x) for x in '0110100' '0010111' '001111']
+        feat1 = [int(x) for x in '0110101' '0111101' '001011']
+        feat2 = [int(x) for x in '0000100' '0000011' '111111']
+        feat3 = [int(x) for x in '1111111' '1111111' '111111']
+        self.dataX = pd.DataFrame({"feat1": feat1, "feat2": feat2, "feat3": feat3})
+        self.dataY = pd.Series(labls)
+        self.dataA = pd.Series(attrs)
+        self.learner = LeastSquaresLearner()
+        self._PRECISION = 1e-6
+
+    smoke_test_data = [{"cons_class": moments.DP, "eps": 0.100, "best_gap": 0.000000,
+          "last_t": 5, "best_t": 5, "disp": 0.100000, "error": 0.250000,
+          "n_oracle_calls":  32, "n_classifiers": 3},
+         {"cons_class": moments.DP, "eps": 0.050, "best_gap": 0.000000,
+          "last_t": 5, "best_t": 5, "disp": 0.050000,
+          "error": 0.266522, "n_oracle_calls":  23, "n_classifiers": 6},
+         {"cons_class": moments.DP, "eps": 0.020, "best_gap": 0.000000,
+          "last_t": 5, "best_t": 5, "disp": 0.020000, "error": 0.332261,
+          "n_oracle_calls":  22, "n_classifiers": 5},
+         {"cons_class": moments.DP, "eps": 0.010, "best_gap": 0.000000,
+          "last_t": 5, "best_t": 5, "disp": 0.010000, "error": 0.354174,
+          "n_oracle_calls":  22, "n_classifiers": 5},
+         {"cons_class": moments.DP, "eps": 0.005, "best_gap": 0.000000,
+          "last_t": 5, "best_t": 5, "disp": 0.005000, "error": 0.365130,
+          "n_oracle_calls":  22, "n_classifiers": 5},
+         {"cons_class": moments.EO, "eps": 0.100, "best_gap": 0.000000,
+          "last_t": 5, "best_t": 5, "disp": 0.100000, "error": 0.309333,
+          "n_oracle_calls":  21, "n_classifiers": 4},
+         {"cons_class": moments.EO, "eps": 0.050, "best_gap": 0.000000,
+          "last_t": 5, "best_t": 5, "disp": 0.050000, "error": 0.378827,
+          "n_oracle_calls":  19, "n_classifiers": 6},
+         {"cons_class": moments.EO, "eps": 0.020, "best_gap": 0.000000,
+          "last_t": 5, "best_t": 5, "disp": 0.020000, "error": 0.421531,
+          "n_oracle_calls":  19, "n_classifiers": 6},
+         {"cons_class": moments.EO, "eps": 0.010, "best_gap": 0.000000,
+          "last_t": 5, "best_t": 5, "disp": 0.010000, "error": 0.435765,
+          "n_oracle_calls":  19, "n_classifiers": 6},
+         {"cons_class": moments.EO, "eps": 0.005, "best_gap": 0.000000,
+          "last_t": 5, "best_t": 5, "disp": 0.005000, "error": 0.442883,
+          "n_oracle_calls":  19, "n_classifiers": 6},]
+
+    def run_smoke_test(self, data):
+        res_tuple = red.expgrad(self.dataX, self.dataA, self.dataY, self.learner,              
+                         		constraints=data["cons_class"](), eps=data["eps"])
+
+        res = res_tuple._asdict()
+        Q = res["best_classifier"]
+        res["n_classifiers"] = len(res["classifiers"])
+
+        disp = data["cons_class"]()
+        disp.init(self.dataX, self.dataA, self.dataY)
+        error = moments.MisclassificationError()
+        error.init(self.dataX, self.dataA, self.dataY)
+        res["disp"] = disp.gamma(Q).max()
+        res["error"] = error.gamma(Q)[0]
+
+        assert res["best_gap"] == pytest.approx(data["best_gap"], abs=self._PRECISION)
+        assert res["last_t"] == data["last_t"]
+        assert res["best_t"] == data["best_t"]
+        assert res["disp"] == pytest.approx(data["disp"], abs=self._PRECISION)
+        assert res["error"] == pytest.approx(data["error"], abs=self._PRECISION)
+        assert res["n_oracle_calls"] == data["n_oracle_calls"]
+        assert res["n_classifiers"] == data["n_classifiers"]
+
+    @pytest.mark.parametrize("testdata", smoke_test_data)
+    def test_smoke(self, testdata):
+        self.run_smoke_test(testdata)
--- a/test_grid_search.py
+++ b/test_grid_search.py
@@ -66,3 +66,7 @@ class TestGridSearch:
 
         result = gs.classification_binary_protected_1d(LeastSquaresLearner(), X, Y, A, num_Ls=11)
         assert len(result) == 11
+        assert result[5]["lambda"] == 0
+        lambdas = [x["lambda"] for x in result]
+        assert len(np.unique(lambdas)) == len(result)
+        assert sorted(lambdas)
--- a/test_grid_search.py
+++ b/test_grid_search.py
@@ -1,0 +1,17 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import fairlearn.grid_search as gs
+
+import numpy as np
+
+import pytest
+
+class TestGridSearch:
+    def test_generate_p0_p1_smoke(self):
+        Y = [int(x) for x in '0110101' '0111101' '001011']
+
+        p0, p1 = gs._generate_p0_p1(Y)
+
+        assert p0 == (7/21)
+        assert p1 == (14/21)
--- a/test_grid_search.py
+++ b/test_grid_search.py
@@ -4,8 +4,24 @@
 import fairlearn.grid_search as gs
 
 import numpy as np
+import pandas as pd
 
 import pytest
+
+class LeastSquaresLearner:
+    def __init__(self):
+        self.weights = None
+
+    def fit(self, X, Y, W):
+        sqrtW = np.sqrt(W)
+        matX = np.array(X) * sqrtW[:, np.newaxis]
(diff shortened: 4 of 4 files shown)


riedgar-ms:
@romanlutz I do note that my new tests don't appear to have been run as part of the build

fairlearn/grid_search.py
```diff
@@ -0,0 +1,109 @@
...
+            w = 2*y - 1 - L * p_ratio
+        else:
+            w = 2*y - 1 + L
+
+        W.append(w)
+
+    return np.array(W)
+
+def classification_binary_protected_1d(learner, X, Y, A, Ls=None, num_Ls = 11):
+    """Function to generate a 1d set of models for a classification problem with
```
romanlutz: a 1D set of models --> you mean a list?

fairlearn/grid_search.py
```diff
@@ -0,0 +1,109 @@
...
+    a single binary protected attribute
+    """
+    # Must specify Ls or num_Ls
+    if not( (Ls is None) ^ (num_Ls is None)):
```
romanlutz: did you mean to use "and"?

fairlearn/grid_search.py
```diff
@@ -0,0 +1,109 @@
...
+        if a==0:
+            w = 2*y - 1 - L * p_ratio
+        else:
+            w = 2*y - 1 + L
+
+        W.append(w)
+
+    return np.array(W)
+
+def classification_binary_protected_1d(learner, X, Y, A, Ls=None, num_Ls = 11):
```
romanlutz: What is num_Ls? Can you either find a descriptive name or at least document it with
:param num_Ls: the number of ...
:type num_Ls: int

fairlearn/grid_search.py
```diff
@@ -0,0 +1,109 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+"""
+This module implements a 1d grid search for binary classification with a single
+binary protected attribute.
```
romanlutz: Should we add a comment on how this won't scale beyond one or two dimensions?

fairlearn/grid_search.py
```diff
@@ -0,0 +1,109 @@
...
+    if Ls is None:
+        limit = 1
+        if p0/p1 > 1:
+            limit = p0/p1
+        Ls = np.linspace(-2*limit, 2*limit, num_Ls)
+
+    result = []
+    for L in Ls:
+        # Generate W array
+        W = _generate_weights(Y, A, L, p1/p0)
```
romanlutz: can we just call it weights?

fairlearn/grid_search.py
```diff
@@ -0,0 +1,109 @@
...
+
+        # Generate Y'
```
romanlutz: What does Y' mean in this case? Y was passed to this function so it should already contain the labels 0 and 1. Mapping 0 and 1 to 0 and 1 seems unnecessary, so I'm sure I'm missing something

fairlearn/grid_search.py
riedgar-ms: No, this is XOR. I'll tweak the comment to make this explicit

fairlearn/grid_search.py
riedgar-ms: I'll expand the comment, but `W` makes sense to me. Will defer to @MiroDudik if `W` is not always weights in ML usage

fairlearn/grid_search.py
riedgar-ms: Actually, you appear to have caught a typo on my part. In @MiroDudik 's notes in OneNote, the function was supposed to be applied to `W` not `Y`

fairlearn/grid_search.py
riedgar-ms: Done

fairlearn/grid_search.py
riedgar-ms: Indeed - I was being a bit fast and loose in my terminology

fairlearn/grid_search.py
riedgar-ms: Done

azure-pipelines.yml
```diff
@@ -14,5 +14,5 @@ steps:
 - script: pip install -r requirements.txt
   displayName: 'Install required packages'
 
-- script: python -m pytest test_fairlearn_smoke.py
-  displayName: 'Run smoke tests'
+- script: python -m pytest test/test*.py
```
riedgar-ms: This doesn't work in Powershell, but does appear to work in the build

fairlearn/grid_search.py
riedgar-ms: Looking at scikit-learn, they do appear to use at least abbreviations for these:
https://scikit-learn.org/stable/modules/generated/sklearn.svm.SVC.html#sklearn.svm.SVC.fit
We should discuss the naming conventions a bit more. Not to mention whether things should be in classes or functions etc. That should probably be discussed with @mesameki and the other PMs, who are familiar with ML packages in Python

riedgar-ms: I've been updating a bunch of argument names. It appears that Keras and sklearn use `x`, `y` and `sample_weight` here, so I've changed to match

azure-pipelines.yml
romanlutz: Probably needs ./test/test*.py ?

romanlutz: Actually just running python -m pytest ./test should be sufficient

riedgar-ms: Is PyTest smart enough to cope with files which don't contain tests? I'm thinking that we might end up with test utility files in this directory.

fairlearn/grid_search.py
romanlutz: I'm not 100% convinced this is right. If I remember the paper correctly then even using a different fairness metric such as EO would mean that we get multiple dimensions, thus making it 2d (despite a single protected attribute). So if this is specific to DP then we should probably say so.
@MiroDudik might be able to comment.

azure-pipelines.yml
romanlutz: It certainly should be. We've used that all over the place.

fairlearn/grid_search.py
riedgar-ms: I mention in the actual 'work' method below that this is for demographic parity. That was based on arguing with myself that there should be more dimensions _somewhere_ if we were doing EO (not the most sophisticated argument, I know :-) ). Agreed that @MiroDudik  is the authority here :-)

azure-pipelines.yml
riedgar-ms: Presumably we just need to avoid having methods start with `test_` to get that working (ISTR the non-pytest script confuses pytest for that reason)

romanlutz: Ok I"m confused. I thought they HAVE to start with test_?

fairlearn/grid_search.py
```diff
@@ -0,0 +1,144 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+"""
+This module implements a 1d grid search for binary classification with a single
+binary protected attribute.
+
+This is a simplification of the full exponentiated gradient search
```
romanlutz: nit: it's not really related to exponentiated gradient search, so I feel like calling it a simplification seems a bit misleading if people don't have the context of the paper. We could point to the paper section (I believe 3.4) and say that it's just there as a baseline for comparison with expgrad etc. Open to suggestions, though :-)

azure-pipelines.yml
riedgar-ms: I believe that PyTest requires methods to start with `test_` in order to pick them up. So if one of our utility functions starts with `test_` we could be in trouble. It's just where we put the naming convention entirely inside the files, or also have one for the filenames themselves. On reflection, I think that the former (as you've suggested) is the right one.

fairlearn/grid_search.py
riedgar-ms: 'Simplified alternative' perhaps? I agree that it's not a direct simplification of `expgrad`. It can't be since each point is a single model, not an ensemble :-)

fairlearn/grid_search.py
MiroDudik: Yes. The current implementation is for: binary classification + demographic parity with a binary protected attribute. I think we should signal that in the name of the class below. How about something like "DemographicParityGridSearch", or even "GridSearchDemographicParity"... since I imagine we'll also have "GridSearchBoundedGroupLoss".

fairlearn/grid_search.py
```diff
@@ -0,0 +1,144 @@
...
+        set of labels
+        """
+        unique, counts = np.unique(y, return_counts=True)
+
+        if len(unique) != 2:
+            raise RuntimeError("Supplied y not binary")
+
+        if not np.array_equal(unique, [0, 1]):
+            raise RuntimeError("Supplied y labels not 0 or 1")
+
```
MiroDudik: I think our code should also work when all the labels are 0 or 1, rather than both 0 and 1.

riedgar-ms: There are definitely more checks for `None` and empty arrays to be added.

For this specific comment, if all the labels are 0 or 1, doesn't that give a trivial model? How is this case regarded by users - in other packages, would supplying a dataset with only a single label be considered an error, or would the user still expect to get a model back?

fairlearn/grid_search.py
MiroDudik: How about we drop the reference to expgrad. It's effectively a different algorithm.

fairlearn/grid_search.py
riedgar-ms: Will we also be wanting a multiclass grid search? How about the following names:
```python
class BinaryAttributeGridSearchDemographicParity:
    def binary_classification(…):

    def multiclass_classification(…):

class BinaryAttributeGridSearchBoundedGroupLoss:
    def binary_classification(…):

    def multiclass_classification(…):

class BinaryAttributeGridSearchRegression:
    def regression(…):
```
Or perhaps the 'BinaryAttributeGridSearch' "stem" should be hoisted into the filename.

fairlearn/grid_search.py
riedgar-ms: Reference to `expgrad` dropped

fairlearn/grid_search.py
```diff
@@ -0,0 +1,144 @@
...
+            w = 1e128
+            if current_a == 0:
+                w = 2*current_y - 1 - L * p_ratio
+            else:
+                w = 2*current_y - 1 + L
+
+            weights.append(w)
+
+        return np.array(weights)
+
```
MiroDudik: Should we use a vectorized version rather than an iteration? My guess is it's going to be more concise--likely easier to read.

fairlearn/grid_search.py
```diff
@@ -0,0 +1,144 @@
...
+        :rtype: List of dictionaries. Each dictionary has fields "lagrangian_multiplier" and "model." Each
+        model will correspond to the input parameter learner after calling 'fit' on it (a deep
+        copy is made). The user is responsible for converting these objects to an actual model,
+        if further processing is required.
+        """
+        # Must specify either Ls or num_Ls
+        if not (lagrangian_multipliers is None) ^ (number_lagrangian_multipliers is None):
+            raise RuntimeError("Must specify either lagrangian_multipliers or number_lagrangian_multipliers")
+
+        # Check that x, y and protected_attribute have the same number of rows
```
MiroDudik: I think we should be able to handle a and y represented as either series / 1d arrays, or also dataframes / 2d arrays (where the second dimension=1).

fairlearn/grid_search.py
```diff
@@ -0,0 +1,144 @@
...
+        use in the optimisation problem. If this is set then lagrangian_multipliers must be None.
+        The result array will have as many entries as specified here
+
+        :return: The models corresponding to each value of the Lagrangian multiplier tested
+        :rtype: List of dictionaries. Each dictionary has fields "lagrangian_multiplier" and "model." Each
+        model will correspond to the input parameter learner after calling 'fit' on it (a deep
+        copy is made). The user is responsible for converting these objects to an actual model,
+        if further processing is required.
+        """
+        # Must specify either Ls or num_Ls
```
MiroDudik: lagrangian_multiplier -> lagrange_multiplier

fairlearn/grid_search.py
riedgar-ms: Changed. It at least gets rid of that horrible append.

fairlearn/grid_search.py
romanlutz: Multiclass would also mean multiple dimensions, right?

fairlearn/grid_search.py
romanlutz: Yes, the user always expects a model back. You can print a warning, though, just in case the user isn't aware of it.

fairlearn/grid_search.py
```diff
@@ -0,0 +1,142 @@
...
+        for current_multiplier in lagrangian_multipliers:
+            # Generate weights array
+            sample_weights = self._generate_weights(y, protected_attribute, current_multiplier, p1/p0)
+
+            # Generate Y'
+            f = lambda x: 1 if x > 0 else 0
+            re_labels = np.vectorize(f)(sample_weights)
+
+            # Run the learner
+            mylearner = copy.deepcopy(learner)
```
romanlutz: I'm not a huge fan of mylearner, but I'm also struggling to come up with a better name.

Since you already used "current_" above, maybe that's an option?

test/test_grid_search.py
```diff
@@ -0,0 +1,163 @@
...
+        self.lsqinfo = np.linalg.lstsq(matX, vecY, rcond=-1)
+        self.weights = pd.Series(self.lsqinfo[0], index=list(X))
+
+    def predict(self, X):
+        pred = X.dot(self.weights)
+        return 1*(pred > 0.5)
+
+class TestGridSearch:
+    def test_generate_p0_p1_smoke(self):
+        Y = [int(x) for x in '0110101' '0111101' '001011']
```
romanlutz: What is this meant to do? Is the formatting as 0/1 sequences helpful in understanding it? If so, why are we converting it to int. I think it would be very helpful to have a comment explaining it since it's all over this file.

test/test_grid_search.py
```diff
@@ -0,0 +1,163 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import fairlearn.grid_search as gs
```
romanlutz: nit: Looks like you only ever use BinaryClassificationGridSearch, so you could just 
```
from fairlearn.grid_search import BinaryClassificationGridSearch
```
and not put gs.... everywhere because gs isn't very descriptive either.

test/test_grid_search.py
```diff
@@ -0,0 +1,163 @@
...
+
+import numpy as np
+import pandas as pd
+
+import pytest
+
+class LeastSquaresLearner:
```
romanlutz: Is this copied from a different test file? It looks familiar... If so, put it in another file and import elsewhere

test/test_grid_search.py
```diff
@@ -0,0 +1,163 @@
...
+        target = gs.BinaryClassificationGridSearch()
+        with pytest.raises(RuntimeError, match=message):
+            _ = target.grid_search_binary_protected_attribute(LeastSquaresLearner(), X, Y, A)
+
+    def test_grid_bad_A_labels(self):
+        message = r"Supplied protected_attribute labels not 0 or 1"
+
+        A = pd.Series([int(x) for x in '0000000' '2222222' '020202'])
+        Y = pd.Series([int(x) for x in '0110100' '0010111' '001111'])
+        feat1 = [int(x) for x in '0110101' '0111101' '001011']
```
romanlutz: nit: I'm generally advocating for not abbreviating names like feature since you're not really saving much in terms of length, but you do open up the possibility that it could be something other than feature.

test/test_grid_search.py
```diff
@@ -0,0 +1,163 @@
...
+        A = pd.Series([int(x) for x in '0000000' '1111111' '010101' '1'])
+        Y = pd.Series([int(x) for x in '0110100' '0010111' '001111'])
+        feat1 = [int(x) for x in '0110101' '0111101' '001011']
+        feat2 = [int(x) for x in '0000100' '0000011' '111111']
+        feat3 = [int(x) for x in '1111111' '1111111' '111111']
+        X = pd.DataFrame({"feat1": feat1, "feat2": feat2, "feat3": feat3})
+
+        target = gs.BinaryClassificationGridSearch()
+        with pytest.raises(RuntimeError, match=message):
+            _ = target.grid_search_binary_protected_attribute(LeastSquaresLearner(), X, Y, A)
```
romanlutz: nit: you don't need to assign it to anything :-) You're explicitly not assigning it to anything by using _ but you can also just running it without variable

test/test_grid_search.py
```diff
@@ -0,0 +1,163 @@
...
+
+    def test_grid_X_Y_mismatch(self):
+        message = r"Supplied x and y do not have same number of rows"
+
+        A = pd.Series([int(x) for x in '0000000' '1111111' '010101' '1'])
+        Y = pd.Series([int(x) for x in '0110100' '0010111' '001111' '0'])
+        feat1 = [int(x) for x in '0110101' '0111101' '001011']
```
romanlutz: looks like we're repeating these. Do you mind putting them as constants at the top of the file?

fairlearn/grid_search.py
riedgar-ms: I don't think so.... @MiroDudik ? I had the impression that the dimensionality was set by the protected attribute, not the number of unique labels

fairlearn/grid_search.py
riedgar-ms: For right now, I'm removing all these checks, since getting them right is going to be non-trivial

fairlearn/grid_search.py
riedgar-ms: Done

fairlearn/grid_search.py
romanlutz: I think you're right (except for the EO case in which the labels would definitely factor in since we're conditioning on them, but we don't care about EO right now I think).

test/test_grid_search.py
riedgar-ms: Refactored

test/test_grid_search.py
riedgar-ms: Refactored

test/test_grid_search.py
riedgar-ms: Is it particularly non-pythonic to do this? I sort of like the reminder that it is producing a result, even if we don't need it

test/test_grid_search.py
riedgar-ms: Have improved naming

test/test_grid_search.py
riedgar-ms: Have improved

fairlearn/grid_search.py
riedgar-ms: I've just added tests which ensure that we use regular lists, numpy or pandas for X, Y and A.  Allowing extra dimensions of length 1 looks to be rather hairer. For example, if we allow Y or A to be ndarrays of size (n,1), should we also allow them to have size (1,n). How about (1, 1, n, 1, 1, 1)?

fairlearn/grid_search.py
riedgar-ms: Done

test/test_grid_search.py
```diff
@@ -0,0 +1,152 @@
...
+import numpy as np
+import pandas as pd
+
+import simple_learners
+
+import pytest
+
+class TestGridSearch:
+    def _quick_data(self, number_samples=8):
+        feature_1 = np.random.randint(2, size=number_samples)
```
romanlutz: Could this result in tests passing randomly? I'm a little worried that we'll hit failures and then we won't be able to repro.

riedgar-ms: I was just about to say 'no' but I've just hit a case where it does. If all the attributes are randomly chosen to be 0 or 1. I'm guessing that's another 'trivial' case which we should allow..

riedgar-ms: There is another case you can see when I'm checking that '0' appears in the automatically generated set of Lagrange multipliers. The np.linspace method doesn't always land exactly on zero :-/

test/test_grid_search.py
romanlutz: Not entirely sure, just wanted to point out that you have the option. I've seen _ used more in the context of multiple return values, e.g. 
```
a, b, _ = f()
```
where we don't care about the third one. But I have no real objection to leaving it as is (as discussed in person :-) )

test/test_grid_search.py
romanlutz: Correct, so that kind of randomness is not necessarily desirable.

linspace just produces floating point numbers, so you'd probably have to put in some tolerance?

riedgar-ms: Indeed - I should have known better. Even trying to rig things by using powers of two (hence making the floating point representation exact) didn't quite work

fairlearn/grid_search.py
riedgar-ms: I have a proposed class renaming, which I think will work for now

fairlearn/grid_search.py
riedgar-ms: For now, I'm not going to make the dimension checks; the learner itself will scream if it can't coerce the arrays into the right shape. It is something to be addressed in due course, but this PR has already been through a lot

test/test_grid_search.py
riedgar-ms: I'll leave this as-is for now. I think I would like to leave something of the 'namespace' intact, at least while I'm getting used to Python

test/test_grid_search.py
riedgar-ms: Things should be reliable now

fairlearn/grid_search.py
```diff
@@ -0,0 +1,127 @@
...
+        if a_val == 0:
+            return 2*y_val - 1 - L * p_ratio
+        else:
+            return 2*y_val - 1 + L
+
+    def _generate_weights(self, y, protected_attribute, L, p_ratio):
+        weight_func = np.vectorize(self._weight_function)
+        return weight_func(y, protected_attribute, L, p_ratio)
+
+    def grid_search_binary_classification(self, learner, x, y, protected_attribute, lagrange_multipliers=None, number_lagrange_multipliers=11):
```
riedgar-ms: Looking at the OneNote again, I'm wondering if `demographic_parity` should be down here

riedgar-ms: Done

fairlearn/grid_search.py
```diff
@@ -0,0 +1,127 @@
...
+        """
+        unique_labels, counts = np.unique(protected_attribute, return_counts=True)
+        if len(unique_labels) > 2:
+            raise RuntimeError("Protected Attribute is not binary")
+
+        p0 = counts[0] / len(protected_attribute)
+        # It is possible that our data only contain one value for
+        # the protected attribute
+        # We do not consider this an error
+        if len(counts) > 1:
```
romanlutz: if we have only ones (label), then counts[0] is the count for 1, not for 0

fairlearn/grid_search.py
```diff
@@ -0,0 +1,127 @@
...
+            p1 = counts[1] / len(protected_attribute)
+        else:
+            p1 = 0
+
+        return p0, p1
+
+    def _weight_function(self, y_val, a_val, L, p_ratio):
+        if a_val == 0:
+            return 2*y_val - 1 - L * p_ratio
```
romanlutz: nit: can we do the spacing uniformly, i.e. either
```
2*y_val-1-L*p_ratio
```
or
```
2 * y_val - 1 - L * p_ratio
```

romanlutz: I like the latter

fairlearn/grid_search.py
```diff
@@ -0,0 +1,127 @@
...
+        :param y: The list of binary classes, which must be 0 or 1. Must contain the same
+        number of entries as rows in x. We do not consider it an error if only one
+        class is present
+        :type y: List or Numpy array with one dimension
+
+        :param protected_attribute: The binary protected attribute corresponding to each row of x
+        We do not consider it an error if only one attribute value is present
+        :type protected_attribute: List of Numpy array with one dimension
+        
+        :param lagrange_multipliers: User specified set of lagrange multipliers to use for
```
romanlutz: super nit: in the description do we want to write Lagrange multiplier with capital L?

fairlearn/grid_search.py
```diff
@@ -0,0 +1,127 @@
...
+        the optimisation problem. If this is set then num_Ls must be None. The result array will
```
romanlutz: num_Ls has been renamed

fairlearn/grid_search.py
```diff
@@ -0,0 +1,127 @@
...
+        be equal in length to this array
+        :type lagrange_multipliers: List of real numbers
```
romanlutz: I think people use this with actual python types, so in this case 
```
:type lagrange_multipliers: list
```

fairlearn/grid_search.py
riedgar-ms: I was just working on something related. This is getting super fiddly when we allow non-numeric labels. Basically, I'm taking
p_0 = n_{A==a_0} / n
but this case will have a_0 == 1, since there's no other value

fairlearn/grid_search.py
```diff
@@ -0,0 +1,127 @@
...
+        The result array will have as many entries as specified here
+
+        :return: The models corresponding to each value of the lagrange multiplier tested
+        :rtype: List of dictionaries. Each dictionary has fields "lagrange_multiplier" and "model." Each
+        model will correspond to the input parameter learner after calling 'fit' on it (a deep
+        copy is made). The user is responsible for converting these objects to an actual model,
+        if further processing is required.
+        """
+        # Must specify either an array of Lagrange multipliers or how many of them to generate
+        if not (lagrange_multipliers is None) ^ (number_lagrange_multipliers is None):
```
romanlutz: logical and should be "and", not ^ since this is not bit comparison

fairlearn/grid_search.py
riedgar-ms: I've just run the VS Code auto-formatter over both files

fairlearn/grid_search.py
riedgar-ms: Indeed. I hadn't notice that the find/replace hadn't been smart about that :-(

fairlearn/grid_search.py
riedgar-ms: Fixed

fairlearn/grid_search.py
riedgar-ms: What do they do when multiple types are possible? For example A, Y and X can be a variety of different things

romanlutz: Hmm good question. Maybe it's best to just leave it as is then...

fairlearn/grid_search.py
riedgar-ms: It's a logical XOR, which is what we want - one of the two must be specified, but not both

fairlearn/grid_search.py
riedgar-ms: I'm pretty sure everything is consistent now

fairlearn/grid_search/binary_protected_attribute/__init__.py
```diff
@@ -0,0 +1,2 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
romanlutz: AFAIK there's no need for __init__.py anymore since they're implicit
https://www.python.org/dev/peps/pep-0420/
That is, unless you want to support super old Python versions. In fact, Python 2.7 will be deprecated in a few months https://pythonclock.org/
I think they added this somewhere in early 3.*
Either way, if you feel like it's worth having just resolve.

fairlearn/grid_search/binary_protected_attribute/binary_classification.py
```diff
@@ -0,0 +1,123 @@
...
+    else:
+        return 2 * y_val - 1 + L
+
+
+def _generate_weights(y, protected_attribute, L, p_ratio, a0_val):
+    weight_func = np.vectorize(_weight_function)
+    return weight_func(y, protected_attribute, L, p_ratio, a0_val)
+
+
+def demographic_parity(learner,
```
romanlutz: This name seems a little bit less descriptive to me than it probably should be. I was expecting something like grid_search_for_demographic_parity, because DP is only the metric

romanlutz: Same for BGL

riedgar-ms: I was thinking that all the rest of that (not to mention binary_protected_attribute) was implicit from the path

fairlearn/grid_search/binary_protected_attribute/__init__.py
riedgar-ms: I had thought they were essential to having things discovered..... let me experiment. It's conceivable that something else which I did allowed everything to hang together

riedgar-ms: It appears that the one in the root of the `test` tree is required, but the others are not

fairlearn/grid_search/binary_protected_attribute/binary_classification.py
romanlutz: But people don't use it that way. In the code you'd have
```
demographic_parity(...)
```
so as a user I'm very confused about what that will even return. Apparently it returns models, which is funny because the function is named after a metric which doesn't indicate that there will be models.

riedgar-ms: I could prefix with `sweep_` if you feel that would be better (and I can see that making a lot of sense). But I don't like `grid_search_for_demographic_parity` because
- This is already in a `grid_search` module
- If we're not going to have the namespace be the description, we also need to add something about binary protected attributes as well

romanlutz: Definitely better. I guess the problem I'm seeing is that people don't read imports, they read code, so nobody will know what your module is unless they are completely confused and then look at your imports. We don't necessarily want them to do that. If they really care about the details they can always do that, but it shouldn't be that the code is unreadable. Adding a verb like sweep would solve that.

riedgar-ms: Change made and pushed. Adding the verb definitely distinguishes it from just the metrics named.

fairlearn/grid_search/binary_protected_attribute/binary_classification.py
```diff
@@ -0,0 +1,123 @@
...
+    multiplier tested
+    :rtype: List of dictionaries. Each dictionary has fields
+    "lagrange_multiplier" and "model." Each model will correspond to
+    the input parameter learner after calling 'fit' on it (a deep
+    copy is made). The user is responsible for converting these
+    objects to an actual model, if further processing is required.
+    """
+    # Must specify either an array of Lagrange multipliers or how many
+    # of them to generate
+    if not (lagrange_multipliers is None) ^ (number_lagrange_multipliers is None):  # noqa: E501
```
romanlutz: please use the pythonic way of expressing this
```
if lagrange_multipliers is not None and number_langrange_multipliers is None:
```

fairlearn/grid_search/binary_protected_attribute/regression.py
```diff
@@ -0,0 +1,46 @@
...
+        return (1 - trade_off) / p1
+
+
+def bounded_group_loss(learner,
+                       x, y, protected_attribute,
+                       tradeoffs=None,
+                       number_of_tradeoffs=11):
+    # Must specify either an array of tradeoffs or the number
+    # of tradeoffs to generated
+    if not (tradeoffs is None) ^ (number_of_tradeoffs is None):
```
romanlutz: please use the pythonic way of expressing this
if tradeoffs is not None and number_of_tradeoffs is None:
```

File: fairlearn/grid_search/binary_protected_attribute/regression.py [hunk 32]
```diff
@@ -0,0 +1,46 @@
...
+
+
+def _regression_weight_function(a_val, trade_off, p0, p1, a0_val):
+    if a_val == a0_val:
+        return trade_off / p0
+    else:
+        return (1 - trade_off) / p1
+
+
+def bounded_group_loss(learner,
```
romanlutz: As a user without knowledge I don't know what tradeoffs mean in this context. Can we add some documentation for this method?

Also, should it be sweep_ again?

fairlearn/grid_search/binary_protected_attribute/regression.py
riedgar-ms: That's not what I'm expressing. As the comment says this is an either/or. According to what I've read, `^` is XOR ... and then I do the `not` of that

fairlearn/grid_search/binary_protected_attribute/regression.py
riedgar-ms: It should have `sweep` already. Yes, it needs the same docs as `sweep_demographic_parity` on the classifier side

fairlearn/grid_search/binary_protected_attribute/binary_classification.py
riedgar-ms: Changed to way we agreed on Teams

fairlearn/grid_search/binary_protected_attribute/regression.py
riedgar-ms: Changed to way we agreed on teams

fairlearn/grid_search/binary_protected_attribute/regression.py
riedgar-ms: Docs added

fairlearn/grid_search/binary_protected_attribute/binary_classification.py
```diff
@@ -0,0 +1,123 @@
...
+
+    # Extract required statistics from protected_attribute
+    p0, p1, a0_val = utilities.generate_protected_attribute_info(
+        protected_attribute)
+
+    # If not supplied, generate array of trial lagrange multipliers
+    if lagrange_multipliers is None:
+        limit = 1
+        if p0 / p1 > 1:
+            limit = p0 / p1
```
MiroDudik: division by zero; when p1=0, this will be infinity

fairlearn/grid_search/binary_protected_attribute/binary_classification.py
```diff
@@ -0,0 +1,123 @@
...
+    # Must specify either an array of Lagrange multipliers or how many
+    # of them to generate
+    if bool(lagrange_multipliers is None) == bool(number_lagrange_multipliers is None):  # noqa: E501
+        raise RuntimeError("Must specify either lagrange_multipliers "
+                           "or number_lagrange_multipliers")
+
+    # Verify we have a binary classification problem
+    unique_labels = np.unique(y)
+    if len(unique_labels) > 2:
+        raise RuntimeError("Supplied Y labels are not binary")
```
MiroDudik: the first check is redundant given that you have the check below

riedgar-ms: I'd been wondering about whether I should remove that. I'd initially thought there might be some value in having two exceptions, but I've been getting less sure. I've removed this.

fairlearn/grid_search/binary_protected_attribute/binary_classification.py
riedgar-ms: Yes, this is similar to the case where all the Y labels are the same. In that case we press on and return a bunch of trivial learners. In this one... eventually an exception gets thrown. The ratio is infinity, and then np.linspace generates an array of NaNs.

I'll throw an exception in this case, since the failure isn't pretty

fairlearn/grid_search/binary_protected_attribute/binary_classification.py
```diff
@@ -0,0 +1,120 @@
...
+    else:
+        return 2 * y_val - 1 + L
+
+
+def _generate_weights(y, protected_attribute, L, p_ratio, a0_val):
+    weight_func = np.vectorize(_weight_function)
+    return weight_func(y, protected_attribute, L, p_ratio, a0_val)
+
+
+def sweep_demographic_parity(learner,
```
romanlutz: We should probably cite Miro's paper here? https://arxiv.org/abs/1803.02453
After all, this is described in there.

PR #19, opened 2019-08-28
Notebook for Grid Search Classification

riedgar-ms:
This is a draft of a notebook to show grid search for binary classification. In part it was written to explain things to myself, so some redrafting may be in order.


riedgar-ms:
This is now part of the GridSearch API change PR

PR #22, opened 2019-09-04
New API for grid search

riedgar-ms:
A rapidly put-together conversion of the classification grid search to the new API


fairlearn/reductions/grid_search_classification.py
```diff
@@ -0,0 +1,98 @@
...
+                 learner,
+                 fairness_metric=DemographicParity(),
+                 number_of_lagrange_multipliers=7):
+        self.learner = learner
+        if not isinstance(fairness_metric, DemographicParity):
+            raise RuntimeError("DemographicParity is only currently supported fairness metric")
+        self.fairness_metric = fairness_metric
+        self.number_of_lagrange_multipliers = number_of_lagrange_multipliers
+
+    def fit(self, X, Y, protected_attribute, lagrange_multipliers=None):
```
riedgar-ms: Since the Lagrange multipliers can depend on the data, I thought it appropriate to put them here. Opinions?

fairlearn/reductions/grid_search_classification.py
```diff
@@ -0,0 +1,98 @@
...
+            # Append the new model, along with its current_multiplier value
+            # to the result
+            # Note that we call it a model because it is a learner which has
+            # had 'fit' called
+            self.all_models.append({"model": current_learner,
+                                    "lagrange_multiplier": current_multiplier})
+
+        # Designate a 'best' model
+        # Selection algorithm not yet fully implemented
+        self.best_model = self.all_models[0]
```
riedgar-ms: Need to work on the means of selecting the `best_model` from `all_models` …. but are those names OK with people?

fairlearn/reductions/grid_search_classification.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def predict(self, X):
```
riedgar-ms: Implement `predict`, `predict_proba`, and their `posterior_` variants as agreed. Should the `posterior_` ones also have the information about each model attached?

test/reductions/test_grid_search_classification.py
```diff
@@ -0,0 +1,76 @@
...
+        number_a1 = 4
+
+        a0_label = 17
+        a1_label = 37
+
+        X, Y, A = self._simple_threshold_data(number_a0, number_a1,
+                                              score_threshold, score_threshold,
+                                              a0_label, a1_label)
+
+        target = GridSearchClassification(LogisticRegression(),
```
riedgar-ms: Have started using sklearn since the existing `simple_learner` wasn't quite giving its predictions as I expected.

test/reductions/test_grid_search_classification.py
```diff
@@ -0,0 +1,76 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+from fairlearn.metrics.demographic_parity import DemographicParity
+from fairlearn.reductions.grid_search_classification import GridSearchClassification
```
riedgar-ms: Can I have a pointer on how to clean these up? I know that @romanlutz has a way :-)

Basically, I'd like these to be
```python
from fairlearn.metrics import DemographicParity
from fairlearn.reductions import GridSearchClassification
```

fairlearn/reductions/grid_search_classification.py
```diff
@@ -0,0 +1,98 @@
...
+import numpy as np
+
+from fairlearn.metrics.demographic_parity import DemographicParity
+
+
+class GridSearchClassification:
+    def __init__(self,
+                 learner,
+                 fairness_metric=DemographicParity(),
+                 number_of_lagrange_multipliers=7):
```
riedgar-ms: This is the fallback if the user does not specify the Lagrange multipliers with the `fit` call

test/reductions/test_grid_search_classification.py
riedgar-ms: Found @romanlutz 's information about this in Teams

fairlearn/metrics/demographic_parity.py
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+
+class DemographicParity:
```
romanlutz: Do we want to have a base class FairnessMetric that all of these inherit from?

fairlearn/reductions/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+from fairlearn.reductions.grid_search_classification import GridSearchClassification  # noqa: F401
```
romanlutz: what about Regression?

fairlearn/reductions/grid_search_classification.py
romanlutz: this shoudl be documented

fairlearn/reductions/grid_search_classification.py
romanlutz: just models would be sufficient I thinK?

test/reductions/test_grid_search_classification.py
```diff
@@ -0,0 +1,76 @@
...
+
+from fairlearn.metrics import DemographicParity
+from fairlearn.reductions import GridSearchClassification
+
+import numpy as np
+import pandas as pd
+from sklearn.linear_model import LogisticRegression
+
+
+class TestGridSearchClassification:
```
romanlutz: just generally: pytest doesn't need these classes, so you can save yourself from indenting everything

requirements.txt
```diff
@@ -3,4 +3,5 @@ numpy
 pandas
 pytest
 scipy
-flake8
\ No newline at end of file
+flake8
+sklearn
```
romanlutz: These are test requirements, right?
For our package to install we shouldn't require sklearn. Does setup.py reflect that?

fairlearn/reductions/grid_search_classification.py
```diff
@@ -0,0 +1,97 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import copy
+import numpy as np
+
+from fairlearn.metrics import DemographicParity
+
+
+class GridSearchClassification:
```
romanlutz: This should inherit from a general class that describes the interface for fit, etc.

fairlearn/reductions/__init__.py
riedgar-ms: I think that will be a separate class - GridSearchRegression. I wanted this reviewed first, to make sure I'm not doing anything which no one was expecting.

requirements.txt
riedgar-ms: Yes, those are just for test. Not to mention pytest

fairlearn/metrics/demographic_parity.py
riedgar-ms: I have been thinking about that - particularly given the composability we've discussed as an idea. Having said that....  to what extent do base classes have meaning in Python? Unless there is common functionality, I'm not sure there's a huge amount of value.

test/reductions/test_grid_search_classification.py
riedgar-ms: It's something of a habit for me. Looking at how Python  deals with files-as-modules, I should probably remove them.

fairlearn/reductions/grid_search_classification.py
riedgar-ms: We would have to define that, and while it might match the signature of things in sklearn etc. ….. In a strongly typed language, I'd agree. Do we gain anything from doing that in Python?

requirements.txt
riedgar-ms: Can we put a separate requirements.txt in the test directory, or does pip not work that way?

fairlearn/reductions/grid_search_classification.py
riedgar-ms: Am cogitating on how to make the whole thing more general

fairlearn/reductions/simple_quality_metric.py
```diff
@@ -0,0 +1,31 @@
...
+    def set_data(self, X, Y, protected_attribute):
+        self.X = X
+        self.Y = Y
+        self.protected_attribute = protected_attribute
+
+        self.error_metric.init(X, protected_attribute, pd.Series(Y))
+        self.disparity_metric.init(X, protected_attribute, pd.Series(Y))
+
+    def get_quality(self, model):
+        current_error_metric = copy.deepcopy(self.error_metric)
```
riedgar-ms: I'm being paranoid about possible state in the two metrics here

fairlearn/reductions/grid_search/simple_quality_metrics.py
```diff
@@ -0,0 +1,52 @@
...
+    def get_quality(self, model):
+        labels = pd.Series(self.Y)
+        preds = pd.Series(model.predict(self.X))
+        attrs = pd.Series(self.protected_attribute)
+        attr_vals = attrs.unique()
+        errors = (preds-labels)**2
+        error = errors.mean()
+        error0 = errors[attrs == attr_vals[0]].mean()
+        error1 = errors[attrs == attr_vals[1]].mean()
+        return -(error+error0+error1)
```
riedgar-ms: Is this a sane thing to be doing @MiroDudik ?

fairlearn/reductions/grid_search_classification.py
MiroDudik: i think that this should be just fit(self, X, Y, **kwargs). then **kwargs should be passed to the fairness metrics and the objective.

fairlearn/reductions/grid_search_classification.py
riedgar-ms: Have switched to keyword arguments on the `fit` call

fairlearn/reductions/__init__.py
riedgar-ms: Have reworked as we agreed in the meeting.

fairlearn/reductions/grid_search_classification.py
riedgar-ms: Done

fairlearn/reductions/grid_search/simple_quality_metrics.py
MiroDudik: Let's replace this by:
return -(error+max(error0,error1))

riedgar-ms: Thanks! I had a nasty feeling I'd picked the wrong thing to do

fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -0,0 +1,189 @@
...
+class GridSearch(ReductionsLearner):
+    """Learner to perform a grid search given a blackbox algorithm.
+    The supplied algorithm must implement a method
+    fit(X, Y, sample_weight=[...])
+    At the present time, the only disparity metrics supported
+    are demographic parity (for classification) and bounded group
+    loss (for regression)
+    """
+
+    _KW_PROTECTED_ATTRIBUTE = "protected_attribute"
```
romanlutz: I usually put these outside classes, although I can't exactly find any resource that says so. Maybe I've been doing it all wrong all this time? Just wondering if you have any opinions...

riedgar-ms: Since it's for the kwargs for the `fit` method of this class, I thought it made more sense here. Having said that, we could well end up using at least some of these in many places.

romanlutz: Okay sure, maybe they'll move to a constants file eventually then. That's great.

requirements.txt
romanlutz: I'm not 100% sure on whether this is right. I feel like there should be 1 place for package dependencies, e.g. setup.py, and another with the additional ones, but it doesn't seem to work like that.... Either way, let's just proceed and tackle this whenever we figure out the right way.

PR #24, opened 2019-09-08
Add CONTRIBUTING.md and adjust it to reflect GitHub flow

romanlutz:


CONTRIBUTING.md
```diff
@@ -0,0 +1,45 @@
...
+
+When you submit a pull request, a CLA-bot will automatically determine whether you need to provide a CLA and decorate the PR appropriately (e.g., label, comment). Simply follow the instructions provided by the bot. You will only need to do this once across all repositories using our CLA.
+
+## Code of Conduct
+This project has adopted the [Microsoft Open Source Code of Conduct](https://opensource.microsoft.com/codeofconduct/). For more information see the [Code of Conduct FAQ](https://opensource.microsoft.com/codeofconduct/faq/) or contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additional questions or comments.
+
+## Development Process
+Development happens against the `master` branch following the [GitHub flow model](https://guides.github.com/introduction/flow/). Contributors create feature branches off of `master`, and their pull requests should target the `master` branch. Maintainers will review pull requests within two business days.
```
riedgar-ms: Have we agreed when we're going to make this switch? We're still doing everything in `dev` right now.

CONTRIBUTING.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Investigating automated test failures
+For every pull request to `master` with automated tests you can check the logs of the tests to find the root cause of failures. Our tests currently run through Azure Pipelines with steps for setup, test suites, and teardown. The `Checks` view of a pull request contains a link to the [Azure Pipelines Page](dev.azure.com/responsibleai/fairlearn/_build/results). All the steps are represented in the Azure Pipelines page, and you can see logs by clicking on a specific step. If you encounter problems with this workflow please reach out through the `Issues`.
+
+## API
+
+Reduction-based fairness mitigation algorithms (such as the ones under `fairlearn.reductions`) provide `fit`, `predict`, and `predict_proba` methods with the following signatures:
+
+```python
+reduction.fit(X, Y, protected_attribute)
```
riedgar-ms: I've started making the `protected_attribute` a kwarg, as we agreed

CONTRIBUTING.md
```diff
@@ -0,0 +1,45 @@
...
+
+Post-processing algorithms (such as the ones under `fairlearn.post_processing`) also provide the same functions albeit with `protected_attribute` as a required argument for `predict` and `predict_proba`
+
+```python
+post_processor.fit(X, Y, protected_attribute)
+post_processor.predict(X, protected_attribute)
+post_processor.predict_proba(X, protected_attribute)
+```
+
+Any algorithm-specific parameters are passed to the constructor. Reductions require a learner/estimator to be passed that implements the `fit` method. Post-processing algorithms require an already trained model/predictor. For consistency we also provide the option to pass a learner/estimator instead, and will call `fit` internally.
```
riedgar-ms: Also, the `fit` method must support a `sample_weight` argument. I'd also argue for putting the constructors before the methods.

CONTRIBUTING.md
romanlutz: Yes, that's for the future (as in when all of this becomes presented/"public"). If you follow the thread in the RAI channel everyone agreed there.

CONTRIBUTING.md
romanlutz: That was kind of a future feature, not an MVP feature, but sure, why not.

CONTRIBUTING.md
romanlutz: Thought about this, too, but just for chronological reasons. That's a great point as well, so I'll reorder it.

CONTRIBUTING.md
romanlutz: Will wait with this until we've settled on the actual name, i.e. protected_attribute, or A, or sensitive_data, or whatever.

CONTRIBUTING.md
```diff
@@ -0,0 +1,47 @@
...
+
+## Code of Conduct
+This project has adopted the [Microsoft Open Source Code of Conduct](https://opensource.microsoft.com/codeofconduct/). For more information see the [Code of Conduct FAQ](https://opensource.microsoft.com/codeofconduct/faq/) or contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additional questions or comments.
+
+## Development Process
+Development happens against the `master` branch following the [GitHub flow model](https://guides.github.com/introduction/flow/). Contributors create feature branches off of `master`, and their pull requests should target the `master` branch. Maintainers will review pull requests within two business days.
+
+Pull requests against `master` trigger automated tests that are run through Azure DevOps. Additional test suites are run on a rolling basis after check-ins and periodically. When adding new code paths or features consider adding tests in the `test` directory.
```
bethz: suggest change from:
tests that run through Azure DevOps 
to:
test pipelines that run on Azure DevOps

bethz: Unless there's a plan to do CI or rolling tests suggest to remove rolling basis and
Change from:
Additional test suites are run on a rolling basis after check-ins and periodically.
to:
Additional tests are run [pick one: daily, weekly, what?]

CONTRIBUTING.md
```diff
@@ -0,0 +1,47 @@
...
+
+### Investigating automated test failures
+For every pull request to `master` with automated tests you can check the logs of the tests to find the root cause of failures. Our tests currently run through Azure Pipelines with steps for setup, test suites, and teardown. The `Checks` view of a pull request contains a link to the [Azure Pipelines Page](dev.azure.com/responsibleai/fairlearn/_build/results). All the steps are represented in the Azure Pipelines page, and you can see logs by clicking on a specific step. If you encounter problems with this workflow please reach out through the `Issues`.
```
bethz: since we are using the free agent pool, no teardown is needed.

README.md
```diff
@@ -40,24 +40,7 @@ The function `expgrad` in the module `fairlearn.classred` implements the reducti
...
-
-When you submit a pull request, a CLA-bot will automatically determine whether you need to provide a CLA and decorate the PR appropriately (e.g., label, comment). Simply follow the instructions provided by the bot. You will only need to do this once across all repositories using our CLA.
-
-### Code of Conduct
-This project has adopted the [Microsoft Open Source Code of Conduct](https://opensource.microsoft.com/codeofconduct/). For more information see the [Code of Conduct FAQ](https://opensource.microsoft.com/codeofconduct/faq/) or contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additional questions or comments.
-
-### Development Process
-The `master` branch is always in a stable release version. Active development happens on the `dev` branch. Contributors create feature branches off of `dev`, and their pull requests should target the `dev` branch instead of `master`. Maintainers will review pull requests within two business days.
```
bethz: no dev, just master. do we want to guarantee two bus days?

CONTRIBUTING.md
romanlutz: I'm not entirely sure what this has to do with the agents being free. Even otherwise no teardown is needed since you'd specify a timeout (default is fairly low)

riedgar-ms: If we ever have this repo running against Azure resources, then there might be a teardown, to delete them after the tests. Not a major point, though.

CONTRIBUTING.md
```diff
@@ -0,0 +1,63 @@
...
+
+## API
+<div id="api">
+
+This section heavily relies on the definitions from our [terminology guide](TERMINOLOGY.md). Specifically, we use the terms "reduction", "group data", "moment", and "parity" in the following.
+
+For all unfairness mitigation methods algorithm-specific parameters are passed to the constructor. The methods to fit a mitigator and predict values with the resulting model shall resemble the APIs used by scikit-learn as much as possible for the sake of ease of use. Any deviations are noted below.
```
riedgar-ms: "disparity mitigation methods"

TERMINOLOGY.md
```diff
@@ -0,0 +1,53 @@
+# Terminology
+
+## Estimators and Predictors
+Fairlearn largely follows the [terminology established by scikit-learn](https://scikit-learn.org/stable/developers/contributing.html#different-objects), specifically:
+- Estimator: implements a `fit` method
+- Predictor: implements a `predict` method
+However, contrary to [scikit-learn estimators](https://scikit-learn.org/stable/glossary.html#term-estimator) we do not require determinism. Lots of unfairness mitigation techniques require randomized output to meet their objective in expectation. We do provide unrandomized versions of every method as well, but they may not be able to uphold the same theoretical guarantees.
```
riedgar-ms: Rather than "do not require determinism" would it be better to say that "in fairlearn, estimators may not be deterministic." Also rather than "contrary to scikit-learn" it might be better to say "Unlike scikit-learn" since we aren't in opposition to sklearn. The "However" isn't necessary.

TERMINOLOGY.md
```diff
@@ -0,0 +1,53 @@
...
+
+## Randomized Predictors
+
+Rather than providing deterministic output many predictors produced by unfairness mitigation techniques in the fairness literature are randomized. As a consequence, it is possible to get different output from the predictor's `predict` method when providing samples with identical features. `predict_proba` provides insights on the probabilities themselves. If randomized predictions are not allowed or undesirable, it is possible to always use the most likely prediction from `predict_proba`. Please be aware that this can affect the fairness of the resulting predictions.
```
riedgar-ms: I thought it was 'predict_posterior' which was going to give the individual predictions from each model (along with the probability of that model being chosen)?

CONTRIBUTING.md
romanlutz: Right. We can add info about that then.

TERMINOLOGY.md
romanlutz: Depends on how you look at it, but I agree that this is not specific enough, so I'll take that part out. We should write about what to expect from predict, predict_proba, etc. soon though

PR #25, opened 2019-09-11
Add Code Scanning to build pipeline

riedgar-ms:
Adding CredScan and PoliCheck, which requires a separate job since they are Windows tools

Note that while the scans run and the logs are published they have to be checked manually


azure-pipelines.yml
```diff
@@ -11,24 +11,57 @@ schedules:
       - dev
   always: true
 
-pool:
-  vmImage: 'ubuntu-latest'
+# =================================================
```
romanlutz: should this just go into a new file?

azure-pipelines.yml
```diff
@@ -11,24 +11,57 @@ schedules:
...
+  - script: python -m pytest test/ --junitxml=./TEST-TEST.xml 
+    displayName: 'Run tests'
+
+  - task: PublishTestResults@2
+    displayName: 'Publish Test Results **/TEST-*.xml'
+    condition: succeededOrFailed()
+
+# -----
+
+- job: "WindowsCodeScans"
```
romanlutz: I don't see a requirement to run it for every PR anywhere.

romanlutz: I think it's called trigger.

riedgar-ms: I'd rather do that - keep anything bad out of the codebase.

It's already flagging you for having ROC everywhere :-)

azure-pipelines.yml
riedgar-ms: I have to admit I wasn't sure how to do that. The name `azure-pipelines.yml` appears to be magic; I'm not sure how to have multiple pipelines defined and have them recognised by the Azure Pipelines infrastructure. I was hoping Beth would show us that.

romanlutz: That's set in VSO actually.

azure-pipelines.yml
romanlutz: ROC is a perfectly fine acronym, and I stand by that :D 
Maybe we can add the trigger then? We can add exceptions for ROC and potential other ones

azure-pipelines.yml
riedgar-ms: I'm not surprised :-)

azure-pipelines.yml
riedgar-ms: There are definitely ways to add exceptions; just need to read up on them

riedgar-ms: Have done so. There's now a separate pipeline

riedgar-ms: For now, the results don't actually go anywhere, so it's not relevant.

riedgar-ms:
Apparently should not be doing this

PR #28, opened 2019-09-16
Update and refactor requirements.txt

riedgar-ms:
Split requirements.txt into sections
Add missing dependencies

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,18 @@
-matplotlib
+# Required for fairlearn
 numpy
 pandas
-pytest
 scipy
+
+# Required for environment
 flake8
-sklearn
+autopep8
+
+# Required for test
+# Pin pytest due to VS Code issue
+pytest==5.0.1
+scikit-learn
+
+# Required for notebooks
+jupyter
+matplotlib
+shap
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -1,7 +1,18 @@
...
+autopep8
+
+# Required for test
+# Pin pytest due to VS Code issue
+pytest==5.0.1
+scikit-learn
+
+# Required for notebooks
+jupyter
+matplotlib
```
romanlutz: matplotlib is actually required for post processing if you run it in debug mode right now

riedgar-ms: Anything else I should add in?

romanlutz: I don't think anything's missing. pip install has no issues with comments this way, right?

riedgar-ms: I've tested on a fresh environment :-)

riedgar-ms: Are you planning on keeping the matplotlib dependency for postprocessing?

romanlutz: I thought about this a little, and I think the plots are useful. Also, 99.9% of people who do machine learning will have matplotlib installed anyway. I don't think it's unreasonable. 

FYI I've made a note about discussing the package versioning with the entire group in the next meeting.

PR #31, opened 2019-09-18
Argument Checks for GridSearch

riedgar-ms:
Adding tests for GridSearch that ensure that the agreed argument types are accepted
Add tests to look for simple problems in the input data

### DIFF
--- a/fairlearn/reductions/grid_search/grid_search.py
+++ b/fairlearn/reductions/grid_search/grid_search.py
@@ -54,32 +54,23 @@ class GridSearch(ReductionsLearner):
             number_of_lagrange_multipliers = kwargs[self._KW_NUMBER_LAGRANGE_MULTIPLIERS]
 
         # Extract the protected attribute
-        A = None
-        if isinstance(aux_data, pd.DataFrame):
-            if len(aux_data.columns) == 1:
-                A = aux_data[0]
-            else:
-                raise RuntimeError("aux_data is DataFrame with more than one column")
-        else:
-            if len(aux_data.shape) == 1:
-                A = aux_data
-            elif len(aux_data.shape) == 2 and aux_data.shape[1] == 1:
-                A = aux_data[:, 0]
-            else:
-                raise RuntimeError("aux_data was ndarray with more than one column")
+        A = self._make_vector(aux_data)
+
+        # Extract the Y values
+        Y_vector = self._make_vector(Y)
 
         # Prep the quality metric
-        self.quality_metric.set_data(X, Y, A)
+        self.quality_metric.set_data(X, Y_vector, A)
 
         # We do not yet have disparity metrics fully implemented
         # For now, we assume that if we are passed a DemographicParity
         # object we have a binary classification problem whereas
         # BoundedGroupLoss indicates a regression
         if isinstance(self.disparity_metric, DemographicParity):
-            self._fit_classification(X, Y, A,
+            self._fit_classification(X, Y_vector, A,
                                      lagrange_multipliers, number_of_lagrange_multipliers)
         elif isinstance(self.disparity_metric, BoundedGroupLoss):
-            self._fit_regression(X, Y, A,
+            self._fit_regression(X, Y_vector, A,
                                  lagrange_multipliers, number_of_lagrange_multipliers)
         else:
             raise RuntimeError("Can't get here")
@@ -213,3 +204,24 @@ class GridSearch(ReductionsLearner):
             return trade_off / p0
         else:
             return (1 - trade_off) / p1
+
+    def _make_vector(self, formless):
+        formed_vector = None
+        if isinstance(formless, pd.DataFrame):
+            if len(formless.columns) == 1:
+                formed_vector = formless[0].to_numpy()
+            else:
+                raise RuntimeError("formless is DataFrame with more than one column")
+        elif isinstance(formless, pd.Series):
+            formed_vector = formless.to_numpy()
+        elif isinstance(formless, np.ndarray):
+            if len(formless.shape) == 1:
+                formed_vector = formless
+            elif len(formless.shape) == 2 and formless.shape[1] == 1:
+                formed_vector = formless[:, 0]
+            else:
+                raise RuntimeError("formless is ndarray with more than one column")
+        else:
+            raise RuntimeError("formless not ndarray or DataFrame")
+
+        return formed_vector
--- a/fairlearn/reductions/grid_search/grid_search.py
+++ b/fairlearn/reductions/grid_search/grid_search.py
@@ -26,6 +26,7 @@ class GridSearch(ReductionsLearner):
     _MESSAGE_Y_NOT_BINARY = "Supplied Y labels are not 0 or 1"
     _MESSAGE_X_NONE = "Must supply X"
     _MESSAGE_Y_NONE = "Must supply Y"
+    _MESSAGE_X_Y_ROWS = "X and Y must have same number of rows"
 
     def __init__(self,
                  learner,
@@ -66,6 +67,10 @@ class GridSearch(ReductionsLearner):
 
         # Extract the Y values
         Y_vector = self._make_vector(Y, "Y")
+
+        X_rows, _ = self._get_matrix_shape(X, "X")
+        if X_rows != Y_vector.shape[0]:
+            raise RuntimeError(self._MESSAGE_X_Y_ROWS)
 
         # Prep the quality metric
         self.quality_metric.set_data(X, Y_vector, A)
@@ -230,6 +235,24 @@ class GridSearch(ReductionsLearner):
             else:
                 raise RuntimeError("%s is an ndarray with more than one column" % formless_name)
         else:
-            raise RuntimeError("formless not an ndarray or DataFrame")
+            raise RuntimeError("%s not an ndarray or DataFrame" % formless_name)
 
         return formed_vector
+
+    def _get_matrix_shape(self, formless, formless_name):
+        num_rows = -1
+        num_cols = -1
+
+        if isinstance(formless, pd.DataFrame):
+            num_cols = len(formless.columns)
+            num_rows = len(formless.index)
+        elif isinstance(formless, np.ndarray):
+            if len(formless.shape) == 2:
+                num_rows = formless.shape[0]
+                num_cols = formless.shape[1]
+            else:
+                raise RuntimeError(
+                    "%s is an ndarray which does not have exactly two dimensions" % formless_name)
+        else:
+            raise RuntimeError("%s not an ndarray or DataFrame" % formless_name)
+        return num_rows, num_cols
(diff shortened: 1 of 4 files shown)


riedgar-ms:
I'm not quite sure why the code coverage build got triggered too

fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -52,31 +63,43 @@ def fit(self, X, Y, **kwargs):
...
 
+        # Extract the target attribute
+        A = self._make_vector(aux_data, "aux_data")
+
+        # Extract the Y values
+        Y_vector = self._make_vector(Y, "Y")
+
+        X_rows, _ = self._get_matrix_shape(X, "X")
+        if X_rows != Y_vector.shape[0]:
+            raise RuntimeError(self._MESSAGE_X_Y_ROWS)
```
riedgar-ms: Would appreciate thoughts on whether these (and other) exceptions are the correct ones

fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -52,31 +63,43 @@ def fit(self, X, Y, **kwargs):
...
                                      lagrange_multipliers, number_of_lagrange_multipliers)
         elif isinstance(self.disparity_metric, BoundedGroupLoss):
-            self._fit_regression(X, Y, aux_data,
+            self._fit_regression(X, Y_vector, A,
                                  lagrange_multipliers, number_of_lagrange_multipliers)
         else:
             raise RuntimeError("Can't get here")
 
-    def _fit_classification(self, X, Y, protected_attribute,
+    def _fit_classification(self, X, Y, target_attribute,
```
riedgar-ms: Mindful of 'protected attribute' having legal meaning, I've been changing to `target_attribute` ….thoughts? This is for more than just binning, so I think this is more descriptive

fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -187,3 +210,41 @@ def _regression_weight_function(self, a_val, trade_off, p0, p1, a0_val):
...
         else:
             return (1 - trade_off) / p1
+
+    def _make_vector(self, formless, formless_name):
+        formed_vector = None
+        if isinstance(formless, pd.DataFrame):
+            if len(formless.columns) == 1:
+                formed_vector = formless[0].to_numpy()
+            else:
+                raise RuntimeError("%s is a DataFrame with more than one column" % formless_name)
```
riedgar-ms: More exception types for comment

fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -52,31 +63,43 @@ def fit(self, X, Y, **kwargs):
...
+
+        X_rows, _ = self._get_matrix_shape(X, "X")
+        if X_rows != Y_vector.shape[0]:
+            raise RuntimeError(self._MESSAGE_X_Y_ROWS)
+        if X_rows != A.shape[0]:
+            raise RuntimeError(self._MESSAGE_X_A_ROWS)
+
         # Prep the quality metric
-        self.quality_metric.set_data(X, Y, aux_data)
+        self.quality_metric.set_data(X, Y_vector, A)
```
romanlutz: Shouldn't we use lower case y for vectors?

test/reductions/grid_search/test_grid_search_arguments.py
```diff
@@ -0,0 +1,257 @@
...
+from fairlearn.metrics import DemographicParity, BoundedGroupLoss
+from fairlearn.reductions import GridSearch
+from fairlearn.reductions.grid_search.simple_quality_metrics import SimpleClassificationQualityMetric, SimpleRegressionQualityMetric  # noqa: E501
+
+
+def identity(X):
+    return X
+
+
+def pandasdf(X):
```
romanlutz: you don't need all these functions. You can provide them in the list below directly
```
X_transform = [lambda X: X, pd.DataFrame]

File: fairlearn/reductions/grid_search/grid_search.py (hunk 4 above)
Comment (riedgar-ms) [reply fairlearn/reductions/grid_search/grid_search.py]: True. I've just updated a bunch of these.

File: test/reductions/grid_search/test_grid_search_arguments.py (hunk 5 above)
Comment (riedgar-ms) [reply test/reductions/grid_search/test_grid_search_arguments.py]: I think it's clearer to have them all as functions grouped at the top; it is something of a matter of taste, though.

File: fairlearn/reductions/grid_search/grid_search.py (hunk 1 above)
Comment (wamartin-aml) [reply fairlearn/reductions/grid_search/grid_search.py]: Both from my experience and doing a bit of stackoverflow browsing, it seems that in cases like this both `RuntimeError` and `ValueError` are common. In the interpretability codebase, I don't think we use `RuntimeError` at all, but we are also probably too liberal with throwing around the basic `Exception`.

File: fairlearn/reductions/grid_search/grid_search.py (hunk 3 above)
Comment (wamartin-aml) [reply fairlearn/reductions/grid_search/grid_search.py]: That raises a good point, which is that if you like having the extra work of maintaining your own exception library, you can of course extend these ones to be more specific.

Comment (wamartin-aml) [inline fairlearn/reductions/grid_search/grid_search.py]: Also, as a nit, our other codebases (as well as AzureMlCli) have largely standardized around using `"".format()`

File: test/reductions/grid_search/test_grid_search_arguments.py [hunk 6]
```diff
@@ -0,0 +1,290 @@
...
+    return pd.DataFrame(X)
+
+
+def pandasseries(X):
+    # Will not work if ndarray has more than one dimension
+    return pd.Series(X)
+
+
+def ndarray2d(X):
```
wamartin-aml: See https://docs.scipy.org/doc/numpy/reference/generated/numpy.expand_dims.html

fairlearn/reductions/grid_search/grid_search.py
riedgar-ms: I wouldn't eagerly sign up for extra work. This was just highlighting more cases where I was wondering if I was throwing the right sort of exception.

test/reductions/grid_search/test_grid_search_arguments.py
```diff
@@ -0,0 +1,290 @@
...
+        feature_2 = np.random.randint(6, size=number_samples)
+        feature_3 = np.random.randint(3, size=number_samples)
+        X = np.stack((feature_1, feature_2, feature_3), -1)
+        Y = np.random.randint(2, size=number_samples)
+        A = np.random.randint(2, size=number_samples)
+        return X, Y, A
+
+    # ----------------------------
+
+    @pytest.mark.parametrize("transformA", Atransform)
```
romanlutz: nit: this looks a little funny since it's transformA once and Atransform the other time.

test/reductions/grid_search/test_grid_search_arguments.py
```diff
@@ -0,0 +1,290 @@
...
+Ytransform = [identity, pandasdf, pandasseries, ndarray2d]
+Atransform = [identity, pandasdf, pandasseries, ndarray2d]
+
+
+# Base class for tests
+# Tests which must be passed by all calls to the GridSearch
+# go here
+class ArgumentTests:
+    def _quick_data(self, number_samples=48):
+        feature_1 = np.random.randint(2, size=number_samples)
```
romanlutz: if your data is randomized how can we reproduce test failures?

fairlearn/reductions/grid_search/grid_search.py
romanlutz: I tend to use that as well, since you don't need to actually care about making the argument a string. format() takes care of that.

fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -38,7 +43,13 @@ def __init__(self,
...
-    def fit(self, X, Y, **kwargs):
+    def fit(self, X, y, **kwargs):
+        if X is None:
+            raise ValueError(self._MESSAGE_X_NONE)
+
+        if y is None:
+            raise ValueError(self._MESSAGE_Y_NONE)
+
         if self._KW_AUX_DATA in kwargs:
             aux_data = kwargs[self._KW_AUX_DATA]
```
romanlutz: To be honest I doubt anyone is going to read the code to find out that aux_data is a possible kwarg. Why not just make it part of the signature with
```
def fit(self, X, y, aux_data=None, **kwargs)

romanlutz: There's also no documentation that says that it's an option to pass it.

fairlearn/reductions/grid_search/grid_search.py
riedgar-ms: Changed to match

test/reductions/grid_search/test_grid_search_arguments.py
riedgar-ms: Which 'other time' ? I'm confused.

test/reductions/grid_search/test_grid_search_arguments.py
riedgar-ms: Changed.... I suppose it's nicer than reshaping via assignment to the `shape` property.

test/reductions/grid_search/test_grid_search_arguments.py
romanlutz: same line :-)

riedgar-ms: Well, once it's the name of the argument to a method, the other time it's the list of arguments to use. I suppose I could have 'transform_list_for_A` or something like that?

romanlutz: Sure, but the inverted naming doesn't express that either :-)
Like I wrote above it's a "nit".... just thought it's odd.

test/reductions/grid_search/test_grid_search_arguments.py
riedgar-ms: I'm using random data to make it plain that the actual values don't matter. We can get occasional failures if all the labels have the same value - that's why I bumped the number of samples up to 48.

romanlutz: Sure, none of the values matter in tests, but say it fails for some weird combination, then we can't even figure out what it was. Rerunning it might fix it then, but we've lost the combination.

romanlutz: You could fix the seed, or parametrize the seed as well with a bunch of different values.

riedgar-ms: Have hard coded some carefully selected random numbers

test/reductions/grid_search/test_grid_search_arguments.py
riedgar-ms: Have made the latter more descriptive

fairlearn/reductions/grid_search/grid_search.py
riedgar-ms: Have changed `aux_data` to being named. Documentation... waiting on details like this. Also, I understand there are multiple doc standards we could follow?

PR #32, opened 2019-09-18
Code Coverage Triggers

riedgar-ms:
Try to stop code coverage triggering on pushes to all branches


bethz:
If you only want the cron build, adding trigger:none will stop the additional ci builds and adding pr: none will stop the pr builds.

PR #35, opened 2019-09-20
Add Support for new data types & update API

romanlutz:
This includes updates to all existing tests as well as parametrization for all tests.

The API update is due to the renaming of protected_attribute to aux_data.


fairlearn/post_processing/roc_curve_based_post_processing.py
```diff
@@ -147,18 +154,27 @@ def _validate_estimator(self):
...
-            raise ValueError(INPUT_DATA_CONSISTENCY_ERROR_MESSAGE
-                             .format(type(X).__name__, type(y).__name__,
-                                     type(protected_attribute).__name__))
-
-        if len(X) == 0 or len(protected_attribute) == 0 or (y is not None and len(y) == 0):
+    def _validate_input_data(self, X, aux_data, y=None):
+        allowed_input_types = [list, np.ndarray, pd.DataFrame, pd.Series]
+        if type(X) not in allowed_input_types or \
+                type(aux_data) not in allowed_input_types or \
+                (y is not None and type(y) not in allowed_input_types):
```
wamartin-aml: Same kind of thing as I'd mentioned before - this kind of line continuation can be really hard to follow in a conditional. We've tended to set a variable for each of the sections then have a one line conditional with those variables. Same benefit of the variable name helping document everything.

fairlearn/post_processing/roc_curve_based_post_processing.py
```diff
@@ -360,42 +380,89 @@ def _roc_curve_based_post_processing_equalized_odds(attributes, labels, scores,
...
 
-def _sanity_check_and_group_data(attributes, labels, scores):
+def _convert_to_ndarray(data, dataframe_multiple_columns_error_message):
+    if type(data) == list:
+        data = np.array(data)
+    elif type(data) == pd.DataFrame:
+        if len(data.columns) > 1:
+            # TODO: extend to multiple columns for additional data
+            raise ValueError(dataframe_multiple_columns_error_message)
+        data = data[data.columns[0]].values
```
wamartin-aml: Why not just grab `values` generally from the DataFrame?

fairlearn/post_processing/roc_curve_based_post_processing.py
riedgar-ms: To what extent would we get this 'for free' if we split the multiple-type support from interpret into a common library?

fairlearn/post_processing/roc_curve_based_post_processing.py
```diff
@@ -147,18 +154,27 @@ def _validate_estimator(self):
...
 
-        if len(X) != len(protected_attribute) or (y is not None and len(X) != len(y)):
-            raise ValueError(DIFFERENT_INPUT_LENGTH_ERROR_MESSAGE)
+        if y is None:
+            if len(X) != len(aux_data) or (y is not None and len(X) != len(y)):
+                raise ValueError(DIFFERENT_INPUT_LENGTH_ERROR_MESSAGE
+                                 .format("X and aux_data"))
+        else:
+            if len(X) != len(aux_data) or (y is not None and len(X) != len(y)):
+                raise ValueError(DIFFERENT_INPUT_LENGTH_ERROR_MESSAGE
```
riedgar-ms: We will need to sit down and review all the exceptions we raise, so that we use the same exception type for the same sort of errors

fairlearn/post_processing/roc_curve_based_post_processing.py
```diff
@@ -237,7 +253,7 @@ def _roc_curve_based_post_processing_demographic_parity(attributes, labels, scor
     x_best = x_grid[i_best_DP]
 
     # create the solution as interpolation of multiple points with a separate predictor per
-    # protected attribute
+    # sensitive attribute
```
riedgar-ms: Based on the teams thread, we might want this to be 'grouping attribute' to make it even more anodyne

test/post_processing/test_roc_curve_based_post_processing.py
```diff
@@ -30,150 +31,163 @@
...
     n = len(example_attributes1)
     for permutation in [(0, 1), (1, 0)]:
-        with pytest.raises(ValueError, match=DIFFERENT_INPUT_LENGTH_ERROR_MESSAGE):
-            X, Y, A = _format_X_Y_A(formatting_function,
-                                    _format_as_list_of_lists(example_attributes1)[:n-permutation[0]],  # noqa: E501
-                                    example_labels[:n-permutation[1]],
-                                    example_attributes1)
+        with pytest.raises(ValueError, match=DIFFERENT_INPUT_LENGTH_ERROR_MESSAGE
+                           .format("X, aux_data, and y")):
+            X = X_transform(_format_as_list_of_lists(example_attributes1)[:n-permutation[0]])
```
riedgar-ms: If these transforms aren't what are under test, should they go outside the `with pytest.raises` section?

fairlearn/post_processing/roc_curve_based_post_processing.py
romanlutz: I only use ValueError when it's an error from an input, for example bad format/length, etc.; NotImplementedError when I don't want to implement something, NotFittedException when a predictor isn't fitted (and therefore not a predictor). That's consistent throughout

test/post_processing/test_roc_curve_based_post_processing.py
romanlutz: Right! I had this in a different structure before, and clearly didn't pay attention when restructuring it.

fairlearn/post_processing/roc_curve_based_post_processing.py
romanlutz: I started out thinking about this for the future when we'll have multiple attributes, at which point I want to extract them separately (because there could be some in X, some in auxiliary data) so that'll be the way to go. Making it fully compatible turned out to be too much work for right now, but probably sometime in November or so I'll get to it.

fairlearn/post_processing/roc_curve_based_post_processing.py
riedgar-ms: So no RuntimeErrors?

fairlearn/post_processing/roc_curve_based_post_processing.py
romanlutz: Oh there was one more consideration. If I directly get the values it's of shape (n,1), whereas if I get it this way it's shape (3,)

fairlearn/post_processing/roc_curve_based_post_processing.py
romanlutz: What would I use those for?

test/post_processing/test_roc_curve_based_post_processing.py
romanlutz: actually these need to be here because we're changing the data every time

fairlearn/post_processing/roc_curve_based_post_processing.py
riedgar-ms: It's what I happened on for some cases where you've gone for ValueError. I'm not arguing for one or the other, but we will need to sit down and standardise.

romanlutz: Sure, we can do that.

riedgar-ms: After this PR, though :-)

fairlearn/post_processing/roc_curve_based_post_processing.py
romanlutz: But it reads like text!
if type of X not in allowed input types or type of aux_data not in allowed input types or type of y not in allowed input types. I could not possibly put that in three variables that make it clearer. 
```
X_not_in_allowed_input_types = type(X) not in allowed_input_types
aux_data_not_in_allowed_input_types = type(aux_data) not in allowed_input_types
y_not_in_allowed_input_types = (y is not None and type(y) not in allowed_input_types)

if X_not_in_allowed_input_types or aux_data_not_in_allowed_input_types or \
    y_not_in_allowed_input_types:
    ...
```
hardly seems better.

romanlutz: I can come back to this later if anyone has a better suggestion, but I want to get this in since it's a huge change already. Everything is passing and I have Richard's approval. Still, I'm curious about @wamartin-aml 's opinion on this one.

PR #43, opened 2019-09-21
Move Exponentiated Gradient into its own directory and remove duplicate tests

romanlutz:
First step in adjusting this to the new APIs


fairlearn/reductions/__init__.py
```diff
@@ -3,3 +3,4 @@
 
 from .grid_search.grid_search import GridSearch  # noqa: F401
 from .reductions_estimator import ReductionsEstimator  # noqa: F401
+from .exponentiated_gradient.exponentiated_gradient import exponentiated_gradient  # noqa: F401
```
romanlutz: This is obviously highly suboptimal, but it will change in the next change when expgrad becomes a class, and some of the logic is distributed to multiple files.

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -11,14 +11,12 @@
...
 
-__all__ = ["expgrad"]
-
 import numpy as np
 import scipy.optimize as opt
 import pandas as pd
 import pickle
 import functools
-import fairlearn.reductions.moments as moments
+from fairlearn.reductions import moments as moments
```
wamartin-aml: `as moments` redundant here?

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -226,13 +224,13 @@ def _mean_pred(dataX, hs, weights):
...
-def expgrad(dataX, dataA, dataY, learner,
-            constraints=moments.DemographicParity(),
-            eps=0.01,
-            T=50,
-            nu=None,
-            eta_mul=2.0,
-            debug=False):
+def exponentiated_gradient(dataX, dataA, dataY, learner,
+                           constraints=moments.DemographicParity(),
+                           eps=0.01,
```
wamartin-aml: Unrelated to this PR, but maybe in general consider documenting where these default values come from

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
romanlutz: Correct! This must be some remnant from the original code.

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
romanlutz: Great idea!

romanlutz: I'll ask Miro who came up with these.

test_fairlearn.py
```diff
@@ -1,118 +0,0 @@
-# Copyright (c) Microsoft Corporation. All rights reserved.
```
MiroDudik: The reason why we kept the "redundant" test script was because it was giving more legible error messages than the new test scripts. New test scripts work nicely for automated checks, but didn't seem suitable for humans (at least not for me). My preference would be to keep this one around as a reference until we fix the legibility of new test scripts. (Maybe this was done, I'm just not aware.)

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
romanlutz: I already have a follow-up PR and will address this there. I hope that's okay for you.

romanlutz:
@Miro Dudik<mailto:mdudik@microsoft.com> I think this came after we already completed the PR. In any case, I’ve created a work item for myself, which I’ll get to this week, to get the test output to the same level of legibility. I’ll send you the PR once I have it. Thanks for pointing this out!

romanlutz:
In that case, it's good to go.

From: Roman Lutz <rolutz@microsoft.com>
Sent: Monday, September 23, 2019 3:00 PM
To: microsoft/fairlearn <reply@reply.github.com>; microsoft/fairlearn <fairlearn@noreply.github.com>; Miro Dudik <mdudik@microsoft.com>
Cc: Author <author@noreply.github.com>
Subject: RE: [microsoft/fairlearn] Move Exponentiated Gradient into its own directory and remove duplicate tests (#43)

@Miro Dudik<mailto:mdudik@microsoft.com> I think this came after we already completed the PR. In any case, I've created a work item for myself, which I'll get to this week, to get the test output to the same level of legibility. I'll send you the PR once I have it. Thanks for pointing this out!

PR #44, opened 2019-09-23
Metrics Draft

riedgar-ms:
Straw example of what our metrics API might look like

I'm quite sure the list handling is horribly inefficient for Python. This is to give us a concrete API to discuss


fairlearn/metrics/true_positive_rate.py
```diff
@@ -0,0 +1,35 @@
...
+    positives_all = 0
+    true_positives_all = 0
+
+    positives = {}
+    true_positives = {}
+    for group in groups:
+        positives[group] = 0
+        true_positives[group] = 0
+
+    for y, yhat, a in zip(y_actual, y_predict, group_id):
```
riedgar-ms: Leaving aside whether this is the best way to implemented the loop in Python (answer: almost certainly not), there's a wider point.

Calculating metrics like this (single request per metric) is going to result in lots of scans through the data. That's going to be slow; the same loop could gather up true negatives, false positives and false negatives essentially for free.

fairlearn/metrics/true_positive_rate.py
```diff
@@ -0,0 +1,35 @@
...
+
+import numpy as np
+
+from . import MetricsResult
+
+
+def true_positive_rate(y_actual, y_predict, group_id):
+    result = MetricsResult()
+
+    groups = np.unique(group_id)
```
riedgar-ms: This is another scan through one of the arrays. For very large data, it might be more efficient to do the necessary 'contains_key' calls in the main loop, rather than using this to pre-initialise to zero.

fairlearn/metrics/true_positive_rate.py
```diff
@@ -0,0 +1,35 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import numpy as np
+
+from . import MetricsResult
+
+
+def true_positive_rate(y_actual, y_predict, group_id):
```
imatiach-msft: I think this is the equivalent of this scikit learn metric:
https://scikit-learn.org/stable/modules/generated/sklearn.metrics.precision_score.html
the only thing is that it doesn't compute "group_id" metrics like this does

imatiach-msft: sorry recall score, not precision:
https://github.com/scikit-learn/scikit-learn/blob/1495f6924/sklearn/metrics/classification.py#L1577

riedgar-ms: I'm sure - that brings up another point: do we take a dependency on `scikit-learn` for this? It would speed up the initial coding at least, since I would just have to implement the wrapper which groups things up for passing to `scikit-learn`. On the other hand it's another dependency and doing things that way might introduce extra performance bottlenecks. The former concern is probably greater.

fairlearn/metrics/true_positive_rate.py
riedgar-ms: Have had a discussion with @rihorn2 and @imatiach-msft  with a conclusion that I'm worrying prematurely about Big Data in the above. For the data sizes we're going to consider, shifting this to using numpy vector operations will be far more important to performance than worrying about unnecessary scans through the data.

fairlearn/metrics/true_positive_rate.py
riedgar-ms: Per the above.... let's not worry about data that big for now.

fairlearn/metrics/metrics_result.py
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+
+class MetricsResult:
+    def __init__(self):
+        self.metric = None
+        self.group_metric = {}
```
romanlutz: similarly, what can we expect of group_metric?

fairlearn/metrics/true_positive_rate.py
riedgar-ms: For now, I'm taking the dependency.....

fairlearn/metrics/metrics_result.py
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+
+class MetricsResult:
+    def __init__(self):
+        self.metric = None
```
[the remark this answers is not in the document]
riedgar-ms: Added comment

fairlearn/metrics/metrics_result.py
riedgar-ms: Added comment

fairlearn/metrics/all_metrics.py
```diff
@@ -0,0 +1,81 @@
...
+
+    return metric_by_groups(selection_metric, y_actual, y_predict, group_id, sample_weight)
+
+
+def metric_by_groups(metric_function, y_actual, y_predict, group_id, sample_weight=None):
+    # TODO: Validate y_actual and y_predict are from {0, 1}
+    # TODO: Validate that group_id are from {0 ... n}
+    result = MetricResult()
+
+    groups = np.unique(group_id)
```
romanlutz: group_id indicates a single id, as int or str
This looks more like it's an array?

riedgar-ms: I was following the sklearn convention that the array is singular - it's `sample_weight` not `sample_weights` (I'm not wholly convinced myself....)

fairlearn/metrics/metric_result.py
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+
+class MetricResult:
+    def __init__(self):
+        # The 'metric' property is the given metric evaluated without regard to group
+        self.metric = None
+        # The 'group_metric' array contains one entry for each group passed to the
+        # evaluation routine. The groups are required to be integers, starting from
```
romanlutz: so is there a mapping from integers to groups? I'm surprised that we don't just use groups as defined by the user. What this describes could be expressed with a dictionary, too. Maybe I'm not seeing where the array comes in handy.

riedgar-ms: @rihorn2 was quite keen on having the groups always be integers with a lower bound of zero. Otherwise JS-interop gets complicated with a dictionary.

fairlearn/metrics/all_metrics.py
romanlutz: Hmm I think it's more confusing than anything else. Not convinced this is actually a convention https://stackoverflow.com/questions/395739/do-you-name-your-arrays-plurally-or-singularly

fairlearn/metrics/metric_result.py
romanlutz: So we're making the user handle all that mapping back and forth?

fairlearn/metrics/all_metrics.py
riedgar-ms: 'Convention' was restricted to `sklearn` …. and now I look a little further, even that is not consistent. Personally I would make them plural, but I thought in a conversation last week we'd said to make them singular?

fairlearn/metrics/metric_result.py
riedgar-ms: We would be requiring that mapping (although they _should_ be doing that anyway, or their data will be pointlessly large). We can discuss with @rihorn2

fairlearn/metrics/all_metrics.py
romanlutz: I may not have been part of that conversation (or maybe I didn't pay attention), because I can't imagine that I would endorse such a resolution :-)

riedgar-ms: I believe it was a somewhat offhand comment from someone; I may have taken it overly to heart.

fairlearn/metrics/metric_result.py
romanlutz: Interesting. If that's the case we need to provide some really good examples of how to do that.

fairlearn/metrics/all_metrics.py
```diff
@@ -0,0 +1,87 @@
...
+        raise NotImplementedError("selection_rate and sample_weight")
+
+    def selection_metric(y_a, y_p, sample_weight):
+        return np.mean(y_p)
+
+    return metric_by_groups(selection_metric, y_actual, y_predict, group_id, sample_weight)
+
+
+def metric_by_groups(metric_function, y_actual, y_predict, group_id, sample_weight=None):
+    _check_binary(y_actual, "y_actual")
```
imatiach-msft: you probably want to handle multiclass scenario eventually too?

fairlearn/metrics/all_metrics.py
```diff
@@ -0,0 +1,87 @@
...
+    y_a = np.array(y_actual)
+    y_p = np.array(y_predict)
+    for group in groups:
+        group_indices = [i for i, elem in enumerate(group_id) if elem == group]
+        group_actual = y_a[group_indices]
+        group_predict = y_p[group_indices]
+        result.group_metric[group] = metric_function(
+            group_actual,
+            group_predict,
+            sample_weight=sample_weight)
```
imatiach-msft: sample weight needs to be indexed by group_indices

fairlearn/metrics/metrics_engine.py
```diff
@@ -0,0 +1,93 @@
...
+
+def make_disparity_metric(metric_function, comparison):
+
+    def wrapper(y_actual, y_predict, group_id, sample_weight=None):
+        return compute_disparity(metric_function,
+                                 y_actual,
+                                 y_predict,
+                                 group_id,
+                                 comparison,
+                                 sample_weight=None)
```
imatiach-msft: I think you want to pass through the sample weight here:
sample_weight=None ->  sample_weight=sample_weight

fairlearn/metrics/true_positive_rate.py
riedgar-ms: Or maybe not... just doing the 'engine' for now

fairlearn/metrics/metric_result.py
riedgar-ms: Following comments from @MiroDudik , have gone back to the dictionary option which allows group_id to be of any type

fairlearn/metrics/all_metrics.py
riedgar-ms: Yes; just found that myself :-/

fairlearn/metrics/metrics_engine.py
riedgar-ms: Indeed

fairlearn/metrics/all_metrics.py
imatiach-msft: probably better to make it plural if it makes the code more readable, I never knew scikit-learn had that convention, strange

MiroDudik: we also discussed group_data; either group_ids or group_data is fine with me.

riedgar-ms: `group_data` it is

fairlearn/metrics/all_metrics.py
riedgar-ms: @MiroDudik  how would that fit in?

fairlearn/metrics/metrics_engine.py
```diff
@@ -0,0 +1,103 @@
...
+    elif comparison == 'diff':
+        result.disparity = np.max(metric_values) - np.min(metric_values)
+    else:
+        raise ValueError("comparison")
+
+    return result
+
+
+def make_disparity_metric(metric_function, comparison):
+
```
MiroDudik: The argument names should reflect sklearn convention, so i'd go with **y_true** and **y_pred**; we also decided to switch group_id to one of group_ids or group_data--please coordinate with @romanlutz what you want to do. I think I have a bit of a preference for group_data--it is a bit more generic and still signals that we're using this for "group by" type operation.

fairlearn/metrics/metrics_engine.py
```diff
@@ -0,0 +1,120 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import numpy as np
+from sklearn.metrics import recall_score
```
riedgar-ms: Still keen on getting rid of this.... do we really need to provide all the one line wrappers ourselves?

fairlearn/metrics/metrics_engine.py
```diff
@@ -0,0 +1,120 @@
...
+        if s_w is not None:
+            group_weight = s_w[group_indices]
+        result.group_metric[group] = metric_function(group_actual,
+                                                     group_predict,
+                                                     sample_weight=group_weight)
+
+    return result
+
+
+def make_group_metric(metric_function):
```
riedgar-ms: With this we can write
```python
true_positive_rate = make_group_metric(sklearn.metrics.recall_score)
```

fairlearn/metrics/metrics_engine.py
riedgar-ms: Changed.

fairlearn/metrics/metrics_engine.py
```diff
@@ -0,0 +1,105 @@
...
+    else:
+        raise ValueError("comparison")
+
+    return result
+
+
+def make_disparity_metric(metric_function, comparison):
+
+    def wrapper(y_true, y_pred, group_data, sample_weight=None):
+        return compute_disparity(metric_function,
```
MiroDudik: I think it would be more natural if metric_function was not the group aware kind.

fairlearn/metrics/metric_result.py
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+
+class MetricResult:
+    def __init__(self):
```
MiroDudik: GroupMetricResult?

MiroDudik: I'm a bit concerned with the proliferation of small files.

fairlearn/metrics/metrics_engine.py
```diff
@@ -0,0 +1,105 @@
...
+    result.metric = metric_function(y_true, y_pred, sample_weight=sample_weight)
+
+    # The slicing we use requires Numpy arrays
+    y_a = np.array(y_true)
+    y_p = np.array(y_pred)
+    s_w = None
+    if sample_weight is not None:
+        s_w = np.array(sample_weight)
+    for group in groups:
+        group_indices = [i for i, elem in enumerate(group_data) if elem == group]
```
MiroDudik: this looks inefficient. you should be able to do something like group_indices = (group_data==elem), which is handled in C rather than the above code that looks like the vector creation is handled by python.

MiroDudik: (i mean it's handled in C if group_data is ndarray)

fairlearn/metrics/metrics_engine.py
riedgar-ms: Now that I've had a think overnight, I agree

fairlearn/metrics/metrics_engine.py
riedgar-ms: Converted; I had suspected there might be a quicker way

fairlearn/metrics/metrics_engine.py
riedgar-ms: Removed disparity metric and put everything under group metric following Teams discussion

fairlearn/metrics/metric_result.py
riedgar-ms: Changed to GroupMetricResult

test/metrics/test_metrics_engine.py
```diff
@@ -0,0 +1,230 @@
...
+
+        assert exCtxt.value.args[0] == "Array y_true contains values other than 0 and 1"
+
+    @pytest.mark.parametrize("transform_y_p", supported_conversions)
+    def test_pred_not_0_1(self, transform_y_p):
+        y_a = [0, 1, 1, 1, 1, 0, 0, 1]
+        y_p = transform_y_p([0, 2, 0, 2, 0, 2, 2, 2])
+        gid = [0, 0, 0, 0, 1, 1, 2, 2]
+
+        with pytest.raises(ValueError) as exCtxt:
```
romanlutz: `exCtxt`.... maybe exception_context?

test/metrics/test_metrics_engine.py
```diff
@@ -0,0 +1,230 @@
...
+import pytest
+
+import fairlearn.metrics as metrics
+
+# ===========================================================
+
+# Conversions from Python lists to our supported datatypes
+
+
+def identity(X):
```
romanlutz: It seems like we defined all of these somewhere already, unless that PR was deleted or changed. Can we maybe put them in a central place?

romanlutz: Sorry, "all of these" meaning these functions "identity", "tondarray", etc.

fairlearn/metrics/metrics_engine.py
```diff
@@ -0,0 +1,95 @@
...
+
+from . import GroupMetricResult
+
+_MESSAGE_NON_BINARY = "Array {0} contains values other than 0 and 1"
+_MESSAGE_SIZE_MISMATCH = "Array {0} is not the same size as {1}"
+
+
+def metric_by_groups(metric_function, y_true, y_pred, group_data, sample_weight=None):
+    """ Applies a metric to each subgroup of a set of data
+
```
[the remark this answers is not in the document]
romanlutz: When we generate documentation pages off of this it should render nicely.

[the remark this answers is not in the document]
romanlutz: But if you think we should make a general pass over all docs instead and standardize these things then, that's fine by me. We should just make sure we don't forget :-)

test/metrics/test_metrics_engine.py
riedgar-ms: Agreed and changed

fairlearn/metrics/group_metric_result.py
```diff
@@ -0,0 +1,17 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+
+class GroupMetricResult:
+    def __init__(self):
```
romanlutz: Is the empty constructor arg list intended? Wouldn't it be easier to just provide everything in the constructor? Maybe I've missed the reason.

fairlearn/metrics/metrics_engine.py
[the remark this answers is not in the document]
riedgar-ms: I think we should hold off on too much doc writing until we're sure we've nailed down the API (we don't want to wait too long, though). In this case, putting together a notebook is on my To Do list for this PR

test/metrics/test_metrics_engine.py
riedgar-ms: They're slightly different this time, since these take lists not ndarrays. Agreed that we need to standardise all this, possibly in conjunction with the interpret library for handling different input types.

fairlearn/metrics/group_metric_result.py
riedgar-ms: Yes. I was having the constructor make sure a bunch of fields exist

romanlutz: Oh great! Let me know if I'm reviewing this too early. I can hold off :-)

test/metrics/test_metrics_engine.py
romanlutz: OK great, in that case ignore for now

fairlearn/metrics/group_metric_result.py
riedgar-ms: Sorry, the "yes" was for "empty argument list intended"

fairlearn/metrics/group_metric_result.py
```diff
@@ -0,0 +1,17 @@
...
+        # The 'metric' property is the given metric evaluated without regard to group
```
MiroDudik: The object itself is a "metric result", so i'm not sure we need to repeat "metric" in each field. I'm in favor of the following replacements... please treat them as suggestions--only if they seem clear!!!
.metric -> .overall
.group_metrics -> .by_group
.min_metric -> .min (or: min_over_groups)
.max_metric -> .max (or: max_over_groups)
.min_metric_groups -> .argmin (to reflect numpy, or possibly: argmin_groups)
.max_metric_groups -> .argmax (to reflect numpy, or possibly: argmax_groups)
.metric_range -> .range (or: .disparity_difference)
.metric_range_ratio -> .range_ratio (or: .disparity_ratio)

fairlearn/metrics/metrics_engine.py
```diff
@@ -0,0 +1,93 @@
...
+import numpy as np
+
+from . import GroupMetricResult
+
+_MESSAGE_NON_BINARY = "Array {0} contains values other than 0 and 1"
+_MESSAGE_SIZE_MISMATCH = "Array {0} is not the same size as {1}"
+
+
+def metric_by_groups(metric_function, y_true, y_pred, group_data, sample_weight=None):
+    """ Applies a metric to each subgroup of a set of data
```
MiroDudik: by_groups -> by_group

fairlearn/metrics/metrics_engine.py
```diff
@@ -0,0 +1,93 @@
...
+
+    :param group_data
+    :type Array indicating the group to which each result belongs
+
+    :param sample_weight
+    :type Array of weights to apply to each result
+    """
+    _check_binary(y_true, "y_true")
+    _check_binary(y_pred, "y_pred")
+    _check_array_sizes(y_true, y_pred, 'y_true', 'y_pred')
```
MiroDudik: we don't want to specialize to binary, we have plenty of regression based metrics required for the dashboard. for the time being, let's however restrict attention to 1d-array-like y_true and y_pred (will need to relax that once we work with multi-class classification/regression).

fairlearn/metrics/metrics_engine.py
```diff
@@ -0,0 +1,93 @@
...
+    for group in groups:
+        group_indices = (group == g_d)
+        group_actual = y_a[group_indices]
+        group_predict = y_p[group_indices]
+        group_weight = None
+        if s_w is not None:
+            group_weight = s_w[group_indices]
+        result.group_metrics[group] = metric_function(group_actual,
+                                                      group_predict,
+                                                      sample_weight=group_weight)
```
MiroDudik: my suggestion is to do an explicit if, and only include sample_weight in the call to metric_function when it's not None. this is to allow disparity for metrics that do not support sample_weight.

fairlearn/metrics/metrics_engine.py
```diff
@@ -0,0 +1,93 @@
...
+
+    result.min_metric = min(result.group_metrics.values())
+    result.max_metric = max(result.group_metrics.values())
+
+    result.min_metric_groups = set([k for k, v in result.group_metrics.items() if v == result.min_metric])  # noqa:E501
+    result.max_metric_groups = set([k for k, v in result.group_metrics.items() if v == result.max_metric])  # noqa:E501
+
+    result.metric_range = result.max_metric - result.min_metric
+    result.metric_range_ratio = result.max_metric / result.min_metric
+
```
MiroDudik: the disparity_ratio should be: min / max (assuming that both are non-negative). if min<0, we probably want to return np.nan.

fairlearn/metrics/metrics_engine.py
```diff
@@ -0,0 +1,93 @@
...
+def _check_binary(arr, arr_name):
+    unique_values = np.unique(arr)
+    if not set(unique_values).issubset({0, 1}):
+        raise ValueError(_MESSAGE_NON_BINARY.format(arr_name))
+
+
+def _check_array_sizes(a, b, a_name, b_name):
+    if len(a) != len(b):
+        raise ValueError(_MESSAGE_SIZE_MISMATCH.format(b_name, a_name))
```
MiroDudik: we need to have the handling of 1d-array-like objects not just for the metrics, but for all of our other functions. it would be nice to support lists, Series, 1d-ndarrays, and 2d-ndarrays / DataFrames of type n*1.

fairlearn/metrics/metrics_engine.py
riedgar-ms: Done

fairlearn/metrics/group_metric_result.py
riedgar-ms: I've added _over_groups to a few, but changed

fairlearn/metrics/metrics_engine.py
riedgar-ms: Removed `_check_binary`

fairlearn/metrics/metrics_engine.py
riedgar-ms: Done

fairlearn/metrics/metrics_engine.py
riedgar-ms: If you see the tests, I try
- Python lists
- 1-d ndarrays
- (n,1) ndarrays
- Pandas Series
- Pandas DataFrame
I agree that we need some unification, quite possibly with the work that interpret has been doing (Ilya and Walter mentioned that they have code they could split out as a separate library)

fairlearn/metrics/metrics_engine.py
```diff
@@ -0,0 +1,95 @@
...
+        group_actual = y_a[group_indices]
+        group_predict = y_p[group_indices]
+        group_weight = None
+        if s_w is not None:
+            group_weight = s_w[group_indices]
+        result.group_metrics[group] = metric_function(group_actual,
+                                                      group_predict,
+                                                      sample_weight=group_weight)
+
+    result.min_metric = min(result.group_metrics.values())
```
romanlutz: not sure if it's useful information, but would we want to have a `min_metric_group` that tells you which group is the one with the min value? Same for max. It would be easy to determine even otherwise.

romanlutz: Oh it's right below :-( oops

PR #45, opened 2019-09-23
Restructure Exponentiated Gradient into a class with the new API

romanlutz:
We certainly need more tests as well, although most of them won't come in this PR.

Note that I've made sure not to change any actual logic, except whatever needed changing to fit it into a class.


fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -0,0 +1,267 @@
...
+
+def _mean_pred(X, hs, weights):
+    """Return a weighted average of predictions produced by classifiers in hs"""
+    pred = pd.DataFrame()
+    for t in range(len(hs)):
+        pred[t] = hs[t](X)
+    return pred[weights.index].dot(weights)
+
+
+# TODO: decide whether we want a structured object instead of the namedtuple
```
romanlutz: Do we prefer this over a real object like ExponentiatedGradientResult right below?

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -0,0 +1,267 @@
...
+            "best_t": self._best_t,
+            "n_oracle_calls": self._n_oracle_calls
+        }
+
+
+class ExponentiatedGradient(ReductionsEstimator):
+    def __init__(self, estimator, disparity_metric, quality_metric, eps=0.01):
+        self._estimator = estimator
+        self._disparity_metric = disparity_metric
+        # TODO use quality metric
```
riedgar-ms: Do you really need this? I just had it in GridSearch as a way of selecting one model from the sweep

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -0,0 +1,267 @@
...
+
+    def posterior_predict_proba(self, X):
+        raise NotImplementedError()
+
+    def _expgrad_result(self):
+        # Just making this available since it's needed for some of the tests
+        return self._expgrad_result
+
+
+def exponentiated_gradient(X, A, y, estimator,
```
riedgar-ms: Why not make this a member method?

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -0,0 +1,267 @@
...
+            "classifiers": self._classifiers,
+            "weights": self._weights,
+            "last_t": self._last_t,
+            "best_t": self._best_t,
+            "n_oracle_calls": self._n_oracle_calls
+        }
+
+
+class ExponentiatedGradient(ReductionsEstimator):
+    def __init__(self, estimator, disparity_metric, quality_metric, eps=0.01):
```
riedgar-ms: Why the change from constraints? I suspect that's a much better word, and as I said, I'd like to move GridSearch to using moments rather than metrics.DemographicParity etc. (for starters, DemographicParity isn't really a metric, but a condition).

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -0,0 +1,267 @@
...
+        self._estimator = estimator
+        self._disparity_metric = disparity_metric
+        # TODO use quality metric
+        self._quality_metric = quality_metric
+        self._eps = eps
+        self._best_classifier = None
+        self._classifiers = None
```
riedgar-ms: What happened to T, nu and eta_mul? Shouldn't they be set in the constructor too?

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -0,0 +1,267 @@
...
+import pandas as pd
+import functools
+from fairlearn.reductions import moments
+from fairlearn.reductions import ReductionsEstimator
+from ._constants import _ACCURACY_MUL, _REGRET_CHECK_START_T, _REGRET_CHECK_INCREASE_T, \
+    _SHRINK_REGRET, _SHRINK_ETA, _MIN_T, _RUN_LP_STEP, _PRECISION, _INDENTATION
+from ._lagrangian import _Lagrangian
+from collections import namedtuple
+
+print = functools.partial(print, flush=True)
```
riedgar-ms: I assume that will be changed to logger or something similar?

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -0,0 +1,267 @@
...
+        # TODO: validate input data; unify between grid search an expgrad?
+        self._expgrad_result = exponentiated_gradient(X, aux_data, y, self._estimator,
+                                                      constraints=self._disparity_metric,
+                                                      eps=self._eps)
+        self._best_classifier = expgrad_result.best_classifier
+        self._classifiers = expgrad_result.classifiers
+        # TODO: figure out whether we should keep the remaining data of the result object
+
+    def predict(self, X):
+        return self._expgrad_result._predict_with_best_classifier(X)
```
riedgar-ms: Shouldn't this be doing the stochastic one - the one which actually minimises disparity (on average)? Unless we have a requirement to make `predict` deterministic?

riedgar-ms: @MiroDudik ?

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -0,0 +1,267 @@
...
+        self._classifiers = classifiers
+        self._weights = weights
+        self._last_t = last_t
+        self._best_t = best_t
+        self._n_oracle_calls = n_oracle_calls
+
+    def _predict_with_best_classifier(self, X):
+        return self._best_classifier(X)
+
+    def _asdict(self):
```
riedgar-ms: This appears to be a magic Python name?

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
romanlutz: yes! And it will happen in this PR

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
romanlutz: I can rename it to `_as_dict`

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
romanlutz: Interesting point. I thought we always provide disparity_metric for every one of our methods (including post processing). Would there otherwise be a guide on how to write constraints?

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
romanlutz: That's what I was wondering! Probably not then. Let me remove it

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
romanlutz: Absolutely! Just pushed really hard for this to work, and the others just worked by default. Will come!

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
romanlutz: Right, so you say I should take this output number, take a random number and if the random number's above the output number then it's 1 else 0? (or I guess the other way round?)

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
romanlutz: Good point. I've not had too much time to do that, and I want to first determine how much of this will be reusable for regression. If those are totally separate then there will be a lot of code. I'm not 100% convinced that all needs to land in that class. Maybe that's better, maybe not? I'm open to all options right now, but focused on a working solution.

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
riedgar-ms: I thought we were preferring real objects. We're not really typed, but if we declare our return objects we have more of a chance of tracking them down.

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
riedgar-ms: Fair enough.

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
romanlutz: Cool, so you prefer the class over the namedtuple?

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
romanlutz: done

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
riedgar-ms: Section 2.1 of the Reductions Approach paper, I believe :-)

The rest of this is very definitely getting into the discussion I was trying to articulate last week. I would say that GridSearch and ExponentiatedGradient are closer (in some sense) than PostProcessing, so getting the first two together is more important than getting all three similar.

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
riedgar-ms: I thought it was that you picked one of the predictors in the result at random, and called predict on that.

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
riedgar-ms: If that's the magic Python name. I was just trying to understand a little more of Python. I'm not quite sure how it's used (or even if there really is a magic Python name), and was trying to get educated :-)

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
riedgar-ms: That's what I've been doing (e.g. for GridSearchResult and now MetricsResult), but.... I suffer from heresies of types and compilation.

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
romanlutz: I may have confused best_classifier with that. It's just the mean_prediction of all classifiers. Actually, that should be equivalent since it's weighted. So that should be a float number, and I can interpret that as probability for 1, right?

fairlearn/reductions/exponentiated_gradient/_lagrangian.py
```diff
@@ -0,0 +1,174 @@
...
+from ._constants import _PRECISION, _INDENTATION, _LINE
+
+logger = logging.getLogger(__name__)
+
+
+class _Lagrangian:
+    """ Operations related to the Lagrangian"""
+
+    def __init__(self, dataX, dataA, dataY, learner, constraints, eps, B,
+                 opt_lambda=True, debug=False):
```
riedgar-ms: Do you still need debug, now there's logger?

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
riedgar-ms: I think this is a @MiroDudik  question

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -0,0 +1,264 @@
...
+        return self._expgrad_result
+
+
+def exponentiated_gradient(X, A, y, estimator,
+                           constraints=moments.DemographicParity(),
+                           eps=0.01,
+                           T=50,
+                           nu=None,
+                           eta_mul=2.0,
+                           debug=False):
```
riedgar-ms: `debug` now redundant?

test/reductions/exponentiated_gradient/test_expgrad_smoke.py
```diff
@@ -2,94 +2,88 @@
...
                         "error": 0.309333, "n_oracle_calls": 21,
                         "n_classifiers": 4},
-                       {"cons_class": moments.EO, "eps": 0.050,
+                       {"cons_class": moments.EqualizedOdds, "eps": 0.050,
                         "best_gap": 0.000000, "last_t": 5,
                         "best_t": 5, "disp": 0.050000,
                         "error": 0.378827, "n_oracle_calls": 19,
                            "n_classifiers": 6},
-                       {"cons_class": moments.EO, "eps": 0.020,
+                       {"cons_class": moments.EqualizedOdds, "eps": 0.020,
```
riedgar-ms: I thought I'd already made all those changes?

fairlearn/reductions/exponentiated_gradient/_lagrangian.py
romanlutz: I do not! Thank you.

test/reductions/exponentiated_gradient/test_expgrad_smoke.py
romanlutz: Correct, maybe I did something weird while merging?

romanlutz: Yes, there are certainly a few warnings at the bottom of this PR. Let me merge and see what happens.

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -6,256 +6,116 @@
...
+def _mean_pred(X, hs, weights):
+    """Return a weighted average of predictions produced by classifiers in hs"""
+    pred = pd.DataFrame()
+    for t in range(len(hs)):
+        pred[t] = hs[t](X)
+    return pred[weights.index].dot(weights)
 
 
-def exponentiated_gradient(dataX, dataA, dataY, learner,
+class ExponentiatedGradientResult:
```
MiroDudik: Why are we doing this instead of using the named tuples?

test/reductions/exponentiated_gradient/test_expgrad_smoke.py
```diff
@@ -2,26 +2,21 @@
...
 from test import simple_learners
+from .test_utilities import aux_data, X1, X2, X3, labels
 
 import pytest
 
 
 class TestExpgradSmoke:
     def setup_method(self, method):
         print("setup_method      method:%s" % method.__name__)
-        attrs = [str(x) for x in 'AAAAAAA' 'BBBBBBB' 'CCCCCC']
```
MiroDudik: Is there a specific reason why this has been moved into its own file? I don't care too much either way, but I'd like to stem proliferation of files where possible.

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
MiroDudik: "predict" should be a randomized prediction, i.e., 0/1, whereas best_classifier is returning a real number between 0 and 1-corresponding to "posterior_predict". you should be flipping a coin based on what best_classifier is returning and returning the result of that flip as "predict".

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
romanlutz: We had a thread about that in this PR. It's always a trade-off, do you want structure and extensibility, or something simple that "just works". namedtuple is cute and nice for a couple, or maybe three or four things, but here we had six or seven, so that becomes tricky and less readable. Richard's opinion was that we use well defined objects elsewhere so this should probably be a real class, too, and I agreed.

riedgar-ms: It was actually advice from Ilya and Walter. I know that the difference between a class and named tuple in Python is minimal.

test/reductions/exponentiated_gradient/test_expgrad_smoke.py
romanlutz: moved it into one file

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
romanlutz: It's constraints again. Please raise again if it's not as you expect it

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
romanlutz: This should be correct now.

PR #48, opened 2019-09-24
Build Definition for Notebooks

riedgar-ms:
Build which runs all notebooks through pytest


riedgar-ms:
Currently, the notebooks aren't working:
https://dev.azure.com/responsibleai/fairlearn/_build/results?buildId=435

riedgar-ms:
Once the build is working, we should probably make this build happen for PRs (and be required)

PR #50, opened 2019-09-24
add Azure NB launch button

bethz:
Added button to readme to import and launch our notebooks in Azure NB

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,6 +2,8 @@
 
 # Fairness in Machine Learning - Mitigation algorithms
 
+[![Azure Notebooks](https://notebooks.azure.com/launch.png)](https://notebooks.azure.com/import/gh/microsoft/fairlearn)
+
 A Python package that implements a variety of fairness-related algorithms to mitigate bias including:
 
 | algorithm | description | classification/regression | protected attributes | supported fairness definitions |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -2,6 +2,8 @@
 
 # Fairness in Machine Learning - Mitigation algorithms
 
+[![Azure Notebooks](https://notebooks.azure.com/launch.png)](https://notebooks.azure.com/import/gh/microsoft/fairlearn)
```
romanlutz: Actually this makes me wonder... This probably opens up notebooks with a view of all notebooks in the repo?

That's obviously awesome, but I'm also wondering whether we should add buttons for each of the algorithms (since they'll probably all get their own notebooks over time). So that's unrelated to your PR, but more of a question.

romanlutz: maybe just in the table below?

vingu: Unless Azure Notebook has solved the problem, otherwise widgets wouldn't work in Azure Notebooks. The widget has to be pre-installed by Azure Notebook on the image.

bethz: I was going to wait and see what happens with the widget before putting an azure nb button in.

bethz:
Remove az nb button for now.

PR #54, opened 2019-09-26
Move GridSearch to Moments [Incomplete]

riedgar-ms:
Make `GridSearch` use moments instead of metrics
This is very incomplete, but sketches where I think we want our API to go

### DIFF
--- a/fairlearn/reductions/grid_search/grid_search.py
+++ b/fairlearn/reductions/grid_search/grid_search.py
@@ -61,7 +61,7 @@ class GridSearch(ReductionsEstimator):
             number_of_lagrange_multipliers = kwargs[self._KW_NUMBER_LAGRANGE_MULTIPLIERS]
 
         # Extract the target attribute
-        A = self._make_vector(aux_data, "aux_data")
+        A = self._make_vector(self.constraint.extract_A(X, aux_data), "A")
 
         # Extract the Y values
         y_vector = self._make_vector(y, "y")
--- a/fairlearn/reductions/grid_search/grid_search.py
+++ b/fairlearn/reductions/grid_search/grid_search.py
@@ -49,9 +49,6 @@ class GridSearch(ReductionsEstimator):
         if y is None:
             raise ValueError(self._MESSAGE_Y_NONE)
 
-        if aux_data is None:
-            raise RuntimeError("Must specify aux_data (for now)")
-
         lagrange_multipliers = None
         if self._KW_LAGRANGE_MULTIPLIERS in kwargs:
             lagrange_multipliers = kwargs[self._KW_LAGRANGE_MULTIPLIERS]
--- a/fairlearn/reductions/grid_search/grid_search.py
+++ b/fairlearn/reductions/grid_search/grid_search.py
@@ -5,7 +5,7 @@ import copy
 import numpy as np
 import pandas as pd
 
-from fairlearn.metrics import DemographicParity, BoundedGroupLoss
+from fairlearn.reductions.moments import DemographicParity, BoundedGroupLoss
 from fairlearn.reductions.reductions_estimator import ReductionsEstimator
 from fairlearn.reductions.grid_search import QualityMetric, GridSearchResult
 
@@ -30,13 +30,13 @@ class GridSearch(ReductionsEstimator):
 
     def __init__(self,
                  learner,
-                 disparity_metric,
+                 constraint,
                  quality_metric):
         self.learner = learner
-        if (not isinstance(disparity_metric, DemographicParity) and
-                not isinstance(disparity_metric, BoundedGroupLoss)):
-            raise RuntimeError("Unsupported disparity metric")
-        self.disparity_metric = disparity_metric
+        if (not isinstance(constraint, DemographicParity) and
+                not isinstance(constraint, BoundedGroupLoss)):
+            raise RuntimeError("Unsupported constraint")
+        self.constraint = constraint
 
         if not isinstance(quality_metric, QualityMetric):
             raise RuntimeError("quality_metric must derive from QualityMetric")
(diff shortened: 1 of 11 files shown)


fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -61,7 +61,7 @@ def fit(self, X, y, aux_data=None, **kwargs):
             number_of_lagrange_multipliers = kwargs[self._KW_NUMBER_LAGRANGE_MULTIPLIERS]
 
         # Extract the target attribute
-        A = self._make_vector(aux_data, "aux_data")
+        A = self._make_vector(self.constraint.extract_A(X, aux_data), "A")
```
riedgar-ms: This is one key part; we use the moment object to get the protected column which GridSearch requires

fairlearn/reductions/moments/conditional_opportunity.py
```diff
@@ -48,6 +48,15 @@ def signed_weights(self, lambda_vec):
         )
         return signed_weights
 
+    def extract_A(self, X, aux_data):
```
riedgar-ms: This obviously needs a lot of work to cope with the different input types we support. For now, it's a sketch

fairlearn/reductions/moments/conditional_opportunity.py
```diff
@@ -56,6 +65,18 @@ class DemographicParity(ConditionalOpportunity):
     """
     short_name = "DemographicParity"
 
+    def __init__(self, from_X=True, column_name=None, column_index=None):
+        super().__init__()
+        self.from_X = from_X
+        if column_name is None and column_index is None:
+            # raise ValueError("Did not specify column_name or column_index")
```
riedgar-ms: Need to figure out why these raise clauses were giving VS Code grief. For now, I'm skipping the error check

test/reductions/grid_search/test_grid_search_classification.py
```diff
@@ -52,8 +52,9 @@ def test_demographicparity_fair_uneven_populations():
                                      score_threshold, score_threshold,
                                      a0_label, a1_label)
 
+    dp = DemographicParity(from_X=False, column_name="a_col")
```
riedgar-ms: This is now the new usage pattern

test/reductions/grid_search/test_grid_search_classification.py
```diff
@@ -157,3 +160,39 @@ def test_can_specify_and_generate_lagrange_multipliers():
         coef1 = target1.all_results[i].model.coef_
         coef2 = target2.all_results[i].model.coef_
         assert np.array_equal(coef1, coef2)
+
+
+def test_can_take_A_from_X_or_aux_data():
```
riedgar-ms: This shows the ways of specifying the constraint column which would be allowed

fairlearn/reductions/grid_search/grid_search.py
romanlutz: What if it's not a dataframe?

fairlearn/reductions/moments/conditional_opportunity.py
```diff
@@ -48,6 +48,18 @@ def signed_weights(self, lambda_vec):
...
 
+    def extract_A(self, X, aux_data):
+        if self.from_X:
+            if self.column_name is not None:
+                return X[self.column_name]
+            else:
+                raise NotImplementedError("This one is hard given range in input types")
+        else:
+            if self.column_name is not None:
+                return aux_data[self.column_name]
```
romanlutz: aux_data may not be a dataframe, though

test/reductions/grid_search/test_grid_search_classification.py
```diff
@@ -22,15 +22,15 @@ def _simple_threshold_data(number_a0, number_a1,
     a1_scores = np.linspace(0, 1, number_a1)
     score_feature = np.concatenate((a0_scores, a1_scores), axis=None)
 
-    A = np.concatenate((a0s, a1s), axis=None)
+    A = pd.DataFrame({"a_col": np.concatenate((a0s, a1s), axis=None)})
```
romanlutz: do we want to standardize what we call these columns? In PP I've set up a constants file and keep referring to the column name constants from there. I think it's a good pattern if we ever change the name. Whether we want to share these names is a different question, of course. Maybe not...

fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -30,13 +30,13 @@ class GridSearch(ReductionsEstimator):
 
     def __init__(self,
                  learner,
-                 disparity_metric,
+                 constraint,
```
romanlutz: This worries me. We'll have more constraints in the future, right? So this should be plural.

fairlearn/reductions/moments/conditional_opportunity.py
riedgar-ms: Hence [Incomplete]. I was just trying to get a happy path working so we had something concrete to discuss

test/reductions/grid_search/test_grid_search_classification.py
riedgar-ms: You mean standardise the names we use in test?

fairlearn/reductions/grid_search/grid_search.py
riedgar-ms: As noted in the title, this is incomplete. Dataframes were the easiest to get working, so I did those.

fairlearn/reductions/grid_search/grid_search.py
riedgar-ms: It should certainly be consistent with what `ExpGrad` ends up with. Although there, even if we combine different constraining moments, we should still end up with another moment object.

fairlearn/reductions/moments/conditional_opportunity.py
romanlutz: What grief?

riedgar-ms: As I recall flake8 was raising issues because it thought the exceptions would always be thrown. Basically, something I didn't want to deal with while trying to get a happy path working. I may have been overly paranoid about what it was saying.

riedgar-ms:
This is going to be superseded by @MiroDudik 's latest PR

PR #55, opened 2019-09-26
Create PR Gated build in Azure DevOps directory

riedgar-ms:
First stage of retiring `azure-pipelines.yml`
Create a version of the gated build in the `azuredevops` directory


azuredevops/pull-request-gate.yml
```diff
@@ -0,0 +1,42 @@
...
+pool:
+  vmImage: 'ubuntu-latest'
+
+jobs:
+- job: Linux
+  pool:
+    vmImage: 'ubuntu-16.04'
+  strategy:
+    maxParallel: 2
+    matrix: # Would like to grab this from the Library variables
```
riedgar-ms: I would like to be able to generate this matrix from a variable. The variable would be set to something like "3.5,3.6"

azuredevops/pull-request-gate-steps.yml
```diff
@@ -0,0 +1,19 @@
...
+    versionSpec: '$(PyVer)' 
+    addToPath: true
+
+- script: pip install -r requirements.txt
+  displayName: 'Install required packages'
+
+- script: flake8 .
+  displayName: "Run flake8"
+
+- script: python -m pytest test/ --junitxml=./TEST--TEST.xml -o junit_suite_name="$(Agent.JobName)"
```
riedgar-ms: It appears that Windows is not respecting the `junit_suite_name` argument

azuredevops/pull-request-gate.yml
bethz: I see what you're doing.

A few things:
- I usually end up splitting windows and Linux into their own pipelines for maintenance convenience. But, I suggest to keep them in the same pipeline for now until one or the other fails at a different rate.

- I believe if you do something like this:
jobs:
- template: jobs/build.yml  #Template ref
  parameters:
     name: Linux
    Pool: ubuntu-xxxx

then in the template, use a matrix for python versions, that that may simplify.
since the template can be used for both Linux and windows, that will make it cleaner.
If there are differences needed btw windows and Linux in the template, I can show you an example.

then in the template, set up it's own parameters and jobs
     vmImage: 'ubuntu-16.04'

riedgar-ms: Not quite set in a variable, but the latest iteration takes the above idea of a job template and is able to generate its matrix

PR #63, opened 2019-10-02
Mirodudik/grid search via moments

MiroDudik:
Add support for grid search via moments


fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -8,6 +8,38 @@
 from fairlearn.metrics import DemographicParity, BoundedGroupLoss
```
riedgar-ms: I'm guessing that the `metrics` versions of these can be removed? Everything should be handled as a moment.

fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -75,6 +108,57 @@ def fit(self, X, y, aux_data=None, **kwargs):
...
+                quality = self.quality_metric.get_quality(current_learner)
+
+                nxt = GridSearchResult(current_learner, lambda_vec, quality)
+                self.all_results.append(nxt)
+
+            # Designate a 'best' model
+            self.best_result = max(self.all_results, key=lambda x: x.quality_metric_value)
+            return
+            
         # We do not yet have disparity metrics fully implemented
```
riedgar-ms: Presumably the rest of the function from this point can be cut?

fairlearn/reductions/moments/conditional_opportunity.py
```diff
@@ -27,15 +54,16 @@ def gamma(self, predictor):
...
         expect_grp = self.tags.groupby("grp").mean()
         expect_attr_grp = self.tags.groupby(
-            ["grp", "protected_attribute"]).mean()
+            ["grp", _GROUP_ID]).mean()
         expect_attr_grp["diff"] = expect_attr_grp["pred"] - expect_grp["pred"]
         g_unsigned = expect_attr_grp["diff"]
         g_signed = pd.concat([g_unsigned, -g_unsigned],
                              keys=["+", "-"],
-                             names=["sign", "grp", "protected_attribute"])
+                             names=["sign", "grp", _GROUP_ID])
```
riedgar-ms: Making `sign` and `grp` into constants as well would be nice

fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -8,6 +8,38 @@
...
+    def __init__(self, limit, n_units, dim, neg_allowed, force_L1_norm):
+        # static variables for integer grid accumulation
+        self.dim = dim
+        self.neg_allowed = neg_allowed
+        self.force_L1_norm = force_L1_norm
+
+        # working variables for integer grid accumulation
+        self.entry = np.zeros(self.dim)
+        self.accumulator = []
+        self.accumulate_integer_grid(0, n_units)
```
riedgar-ms: It's a smaller thing, but given that `accumulate_integer_grid` is a recursive method, do we really want to call it in the constructor?

fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -8,6 +8,38 @@
 from fairlearn.metrics import DemographicParity, BoundedGroupLoss
 from fairlearn.reductions.reductions_estimator import ReductionsEstimator
 from fairlearn.reductions.grid_search import QualityMetric, GridSearchResult
+from fairlearn.reductions.moments.moment import Moment, _REDUCTION_TYPE_CLASSIFICATION
+
+class _GridGenerator:
+    """A generator of a grid of points with a bounded L1 norm.
```
riedgar-ms: This appears to be a multi-dimensional grid? So what you're constructing is a list of points with a "Manhattan distance" to the origin of less than some given bound?

fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -34,7 +66,8 @@ def __init__(self,
                  quality_metric):
         self.learner = learner
         if (not isinstance(disparity_metric, DemographicParity) and
-                not isinstance(disparity_metric, BoundedGroupLoss)):
+                not isinstance(disparity_metric, BoundedGroupLoss) and
+                not isinstance(disparity_metric, Moment)):
```
riedgar-ms: This should become the only check we need?

fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -75,6 +108,57 @@ def fit(self, X, y, aux_data=None, **kwargs):
         # Prep the quality metric
         self.quality_metric.set_data(X, y_vector, A)
 
+        if isinstance(self.disparity_metric, Moment):
+            # Prep the disparity metric and objective
+            self.disparity_metric.init(X, A, y_vector)
```
riedgar-ms: Should `self.disparity_metric` be turning into `self.constraint`  (although that can happen later)?

fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -75,6 +108,57 @@ def fit(self, X, y, aux_data=None, **kwargs):
...
+            is_classification_reduction = (self.disparity_metric.reduction_type == _REDUCTION_TYPE_CLASSIFICATION)
+
+            # Basis information
+            pos_basis = self.disparity_metric.pos_basis
+            neg_basis = self.disparity_metric.neg_basis
+            neg_allowed = self.disparity_metric.neg_basis_present
+            objective_in_the_span = (self.disparity_metric.default_objective_lambda_vec is not None)
+            dim = len(pos_basis.columns)
+
+            # Grid parameters: these should be provided as arguments, but explicitly here for the time being
```
riedgar-ms: Constructor arguments to the `GridSearch` object, right?

fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -75,6 +108,57 @@ def fit(self, X, y, aux_data=None, **kwargs):
...
+                    y_reduction = 1 * (weights > 0)
+                    weights = weights.abs()
+                else:
+                    y_reduction = y
+
+                current_learner = copy.deepcopy(self.learner)
+                current_learner.fit(X, y_reduction, sample_weight=weights)
+
+                # Evaluate the quality metric
+                quality = self.quality_metric.get_quality(current_learner)
```
riedgar-ms: Do we want to keep a separate `quality_metric` or is this just a convenience for now, so that not too much breaks at once?

fairlearn/reductions/moments/bounded_group_loss.py
```diff
@@ -0,0 +1,106 @@
...
+            self.neg_basis[i] = 0 + zero_vec
+            self.pos_basis[i][attr] = 1
+            self.neg_basis_present.at[i] = False
+            i += 1
+
+    def gamma(self, predictor):
+        """ Calculates the degree to which constraints are currently violated by
+        the predictor.
+        """
+        self.tags["pred"] = predictor(self.X)
```
riedgar-ms: Would be good to make all of these strings (they're column names, I believe?) into constants like `_GROUP_ID`

fairlearn/reductions/grid_search/grid_search.py
MiroDudik: But first, we should cross-check that the new implementation works on all the tests for the old implementation! I didn't have a chance to do it and there isn't an easy one-to-one map--because of how the lagrange multipliers are represented between the two implementations.

fairlearn/reductions/grid_search/grid_search.py
MiroDudik: Not yet...

fairlearn/reductions/moments/conditional_opportunity.py
MiroDudik: Yes that's the idea... I just ran out of time on train and wanted to get this PR out for review.

fairlearn/reductions/grid_search/grid_search.py
riedgar-ms: Goes with the previous comment :-)

fairlearn/reductions/grid_search/grid_search.py
MiroDudik: I went for a more concise version that allows me to say things like "grid = _GridGenerator(...).grid". But if that rings hard to read, we can refactor.

fairlearn/reductions/grid_search/grid_search.py
MiroDudik: yes. the one twist is that some coordinates are only allowed to be non-negative and some both positive and negative.

fairlearn/reductions/grid_search/grid_search.py
riedgar-ms: If I've figured out what this is doing correctly, then it's probably not too bad. In general you want minimal logic in a constructor, but it's not like this is doing network calls to create AzureML workspaces (not that I'd have any idea where I got that as an example...). And if this is used once and the grid never needs regeneration, then having a slightly conciser version for use is also desirable.

fairlearn/reductions/grid_search/grid_search.py
riedgar-ms: I was guessing that was the reason for the extra Booleans :-)

fairlearn/reductions/grid_search/grid_search.py
MiroDudik: Yes.

PR #66, opened 2019-10-02
Post processing notebook: use real valued scores and add charts

romanlutz:
This adds
- corrections for changes from the last PR (renaming post processing and adjusting API), mainly removing last occurrences of `aux_data`.
- use real valued scores by wrapping logistic regression into an object that returns `predict_proba` output (only the probability for predicting 1). This improves the results because the algorithm can threshold at many more points.
- added charts to visualize DP and EO since the fairness dashboard isn't ready yet


notebooks/Threshold Optimization Post-Processing for Binary Classification.ipynb
```diff
@@ -12,17 +12,153 @@
    "cell_type": "markdown",
```
wamartin-aml: Are you guys checking in notebooks with outputs?

romanlutz: Nope, definitely not! In this particular case someone desperately needed a working version of this, and I didn't want to erase my notebook a few minutes before giving a demo :-)

Issue #64, opened 2019-10-02
Moments use dataX/A/Y instead of X, group_data, and y


romanlutz:
e.g.
```python
def init(self, dataX, dataA, dataY):
```
For consistency this should be adjusted.

romanlutz:
seems like all occurrences are gone

PR #74, opened 2019-10-08
update terminology

MiroDudik:


TERMINOLOGY.md
```diff
@@ -1,53 +1,70 @@
...
 
-### Parity/Disparity vs. Fairness/Unfairness vs. Bias
+- _Equalized odds_: A classifier _h_ satisfies equalized odds under a distribution over (_X, A, Y_) if its prediction _h_(_X_) is conditionally independent of the group membership _A_ given the label _Y_. This is equivalent to E[_h_(_X_) | _A_=_a_, _Y_=_y_] = E[_h_(_X_) | _Y_=_y_] for all _a, y_. [[Agarwal et al.]](https://arxiv.org/pdf/1803.02453.pdf)
 
-There are a number of terms in the literature used to describe concepts related to disparity and fairness. As noted above, some can be particularly fraught since in legal circles they have precise meanings. We shall use the following definitions:
+- _Equal opportunity_: a relaxed version of equalized odds that only considers conditional expectations with respect positive labels, i.e., _Y_=1. [[Hardt et al.]]( https://ttic.uchicago.edu/~nati/Publications/HardtPriceSrebro2016.pdf)
 
-- *Disparity* is something that we can _measure_ and potentially _mitigate_. For example, if we see that a credit scoring model is favoring one group over another.
```
riedgar-ms: These definitions not well liked?

MiroDudik: I guess the issue is that bias, fairness, and disparity can all be appropriately defined, measured and "mitigated", but disparity best matches what we are doing. Fairness works too, but it is less precise: we capture some notion of fairness--specifically group fairness--but do not capture other notions of fairness--such as individual fairness or counterfactual fairness.

riedgar-ms: That's what I'd been trying to get across - and I admit that I was trying to pre-settle terminology which might still be in flux in the literature :-)

TERMINOLOGY.md
```diff
@@ -1,53 +1,70 @@
...
+
+## Fairness of AI systems
+
+AI systems can behave unfairly for a variety of reasons. Sometimes it is because of societal biases reflected in the training data and in the decisions made during the development and deployment of these systems. In other cases, AI systems behave unfairly not because of societal biases, but because of characteristics of the data (e.g., too few data points about some group of people) or characteristics of the systems themselves. Because it can be hard to distinguish between these reasons (and these reasons are not mutually exclusive and often exacerbate one another), we define whether an AI system is behaving unfairly in terms of its impact on people – i.e., in terms of harms – and not in terms of specific causes, such as societal biases, or in terms of intent, such as prejudice.
```
romanlutz: didn't we want to stay away from the term bias?

romanlutz: oh, below you're talking about that. Did you use it here because this is not related to fairlearn yet?

TERMINOLOGY.md
```diff
@@ -1,53 +1,70 @@
...
+
+**Usage of the word _bias_.** Since we define fairness in terms of harms rather than specific causes (such as societal biases), we avoid the usage of the words _bias_ or _debiasing_ in describing the functionality of fairlearn.
+
+### Types of harms
+
+There are many types of harms (see, e.g., the [keynote by K. Crawford at NIPS 2017](https://www.youtube.com/watch?v=fMym_BKWQzk)). Fairlearn is most applicable to two kinds of harms:
```
romanlutz: NIPS or NeurIPS? Not sure if they had the new name already in 2017.

romanlutz: We need to standardize the way we write fairlearn:
- fairlearn
- Fairlearn
- FairLearn
are all possible, but it should be uniform. I'm actually for `fairlearn` since that's consistent with the code, but open to anything.

@bethz @riedgar-ms FYI

TERMINOLOGY.md
```diff
@@ -1,53 +1,70 @@
...
+
+### Fairness assessment and unfairness mitigation
+
+In fairlearn, we provide tools to assess fairness of predictors for classification and regression. We also provide tools that mitigate unfairness in classification and regression. In both assessment and mitigation scenarios, fairness is quantified using disparity metrics as we describe below.
+
+#### Group fairness, group-membership features
+
+There are many approaches to conceptualizing fairness. In fairlearn, we follow the approach known as group fairness, which asks: _Which groups of individuals are at risk for experiencing harms?_
+
+The relevant groups (also called subpopulations) are defined using **group-membership features**, which represented as a data frame, an array or a vector called `group_membership`. Alternative terminology for group-membership features:
```
romanlutz: Are you sure `group_membership` makes sense for the arg? I would be very confused as a user, and I'd think that I have to preprocess my matrix (for multiple attributes) into a vector that represents groups. At least we should do a great job in documenting it if we go with it. Should confirm asap.

Adding @bethz  @riedgar-ms  @vingu to settle this.

TERMINOLOGY.md
```diff
@@ -1,53 +1,70 @@
...
-- It may signify the presence of multiple sensitive attributes (as a matrix of values), as opposed to other choices such as `sensitive_attribute`.
-- It does not falsely imply ordering between `X`, `y`, and `group_data`, as opposed to, for example, `Z`.
-- It does not falsely imply that it is always part of the training features, as opposed to variable names that subscript `X`.
+- _Statistical parity_ (also known as _demographic parity_): A classifier _h_ satisfies statistical parity under a distribution over (_X, A, Y_) if its prediction _h_(_X_) is statistically independent of the group membership _A_. This is equivalent to E[_h_(_X_) | _A_=_a_] = E[_h_(_X_)] for all _a_. [[Agarwal et al.]](https://arxiv.org/pdf/1803.02453.pdf)
 
-### Parity/Disparity vs. Fairness/Unfairness vs. Bias
+- _Equalized odds_: A classifier _h_ satisfies equalized odds under a distribution over (_X, A, Y_) if its prediction _h_(_X_) is conditionally independent of the group membership _A_ given the label _Y_. This is equivalent to E[_h_(_X_) | _A_=_a_, _Y_=_y_] = E[_h_(_X_) | _Y_=_y_] for all _a, y_. [[Agarwal et al.]](https://arxiv.org/pdf/1803.02453.pdf)
```
romanlutz: nit: let's try to be uniform and always write group-membership or group membership (with or without dash)

TERMINOLOGY.md
```diff
@@ -1,53 +1,70 @@
...
-There is no definitive standard yet, but for this repository, we shall use the variable name `group_data` for a matrix of these attributes. This choice provides a few advantages:
+_Binary classification_:
 
-- The term ‘protected attribute’ has a specific meaning in legal contexts which does not always match the data science usage of the term
-- It may signify the presence of multiple sensitive attributes (as a matrix of values), as opposed to other choices such as `sensitive_attribute`.
-- It does not falsely imply ordering between `X`, `y`, and `group_data`, as opposed to, for example, `Z`.
-- It does not falsely imply that it is always part of the training features, as opposed to variable names that subscript `X`.
+- _Statistical parity_ (also known as _demographic parity_): A classifier _h_ satisfies statistical parity under a distribution over (_X, A, Y_) if its prediction _h_(_X_) is statistically independent of the group membership _A_. This is equivalent to E[_h_(_X_) | _A_=_a_] = E[_h_(_X_)] for all _a_. [[Agarwal et al.]](https://arxiv.org/pdf/1803.02453.pdf)
```
romanlutz: Are we supposed to rename demographic parity to statistical parity? I don't mind doing that, just want to use the right term.

TERMINOLOGY.md
MiroDudik: Yes... I wanted to get across why we are not talking about debiasing, but also acknowledge that societal biases exist.

TERMINOLOGY.md
MiroDudik: first neurips was 2018. the youtube link says "NIPS 2017" so i decided no to retrofit, but i'll check with hanna what she thinks (she's the general chair for neurips 2019)

riedgar-ms: I'd vote for `fairlearn` too

MiroDudik: I'm in favor of treating it as a noun, but not a proper noun, so "fairlearn" in the middle of the sentence, but "Fairlearn" at the beginning of the sentence. Inside URLs and file paths, I think we should write "fairlearn". Thoughts?

TERMINOLOGY.md
riedgar-ms: What about allowing the grouping data to be taken from `X`? The above implies that the `group_membership` argument will always have to be present, even if everything is in `X`

TERMINOLOGY.md
riedgar-ms: But `fairlearn` is a proper noun - our "brand" (for want of a better word). So even if starting a sentence, it should not be capitalised.

We could cheat, and make sure we never begin sentences with it...

TERMINOLOGY.md
MiroDudik: Given that currently we only support vectors for group_membership, your conclusion would be correct! But jokes aside--we could expand to something like group_membership_features if that would make things clearer. After talking with a bunch of folks, I'd go for clarity over conciseness.

TERMINOLOGY.md
```diff
@@ -1,53 +1,70 @@
...
 
-### Parity/Disparity vs. Fairness/Unfairness vs. Bias
+- _Equalized odds_: A classifier _h_ satisfies equalized odds under a distribution over (_X, A, Y_) if its prediction _h_(_X_) is conditionally independent of the group membership _A_ given the label _Y_. This is equivalent to E[_h_(_X_) | _A_=_a_, _Y_=_y_] = E[_h_(_X_) | _Y_=_y_] for all _a, y_. [[Agarwal et al.]](https://arxiv.org/pdf/1803.02453.pdf)
 
-There are a number of terms in the literature used to describe concepts related to disparity and fairness. As noted above, some can be particularly fraught since in legal circles they have precise meanings. We shall use the following definitions:
+- _Equal opportunity_: a relaxed version of equalized odds that only considers conditional expectations with respect positive labels, i.e., _Y_=1. [[Hardt et al.]]( https://ttic.uchicago.edu/~nati/Publications/HardtPriceSrebro2016.pdf)
```
riedgar-ms: Side thing - can we get all of these maths pieces into (pseudo)\LaTeX?

TERMINOLOGY.md
riedgar-ms: That's orthogonal to my comment. I had thought that we wanted that `group_membership_features` (aka `aux_data` aka....) argument to be optional? We were going to modify the `Moment` constructors so that they could specify whether their data were to be taken from `X` or from this other arg.

TERMINOLOGY.md
MiroDudik: I'm following a standard convention here about compound adjectives. In "group membership", group acts an adjective and membership is a noun. In "group-membership features", group-membership is an adjective and features is a noun: https://en.wikipedia.org/wiki/Compound_modifier#Compound_adjectives
(it's the same logic we apply when we wrie "estimator as a black box" vs. "a black-box approach").

TERMINOLOGY.md
MiroDudik: that would be great--what's the way to render that in markdown?

TERMINOLOGY.md
MiroDudik: [I was replying to @romanlutz ]... that's correct. I was going to take a stab at the initial rewrite of Moment once you update GridSearch.

TERMINOLOGY.md
MiroDudik: Let me ask around if there's a sentiment that one is better than the other.

TERMINOLOGY.md
riedgar-ms: Hmmm.... it looks like Markdown-for-Notebooks lets you just put in '$' and away you go. GitHub may not :-(

TERMINOLOGY.md
MiroDudik: see #64

TERMINOLOGY.md
MiroDudik: @romanlutz : does this need to be moved or edited?

MiroDudik: keeping as is.

TERMINOLOGY.md
MiroDudik: resolving to NeurIPS

TERMINOLOGY.md
MiroDudik: leaving as is for now.

TERMINOLOGY.md
MiroDudik: resolving to demographic parity

TERMINOLOGY.md
MiroDudik: resolving to all lowercase fairlearn

TERMINOLOGY.md
MiroDudik: resolving to sensitive_features

TERMINOLOGY.md
```diff
@@ -1,53 +1,68 @@
...
+#### Group fairness, sensitive features
 
-### Fairness assessment
+There are many approaches to conceptualizing fairness. In fairlearn, we follow the approach known as group fairness, which asks: _Which groups of individuals are at risk for experiencing harms?_
 
-Fairness assessment is the process of determining and evaluating the fairness of a model through a variety of metrics in multiple configurations:
+The relevant groups (also called subpopulations) are defined using **sensitive features** (or sensitive attributes). This term suggests that the system designer should be sensitive to these features when assessing group fairness. Although these features may sometimes have privacy implications (e.g., gender or age) in other cases they may not (e.g., whether or not someone is a native speaker of language). Moreover, the word sensitive does not imply that these features should not be used to make predictions – indeed, in some cases it may be better to include them.
```
romanlutz: do we want to explicitly say that we use the name `sensitive_features`?

TERMINOLOGY.md
```diff
@@ -1,53 +1,68 @@
...
 
-## Randomized predictors
+The fairlearn package largely follows the [terminology established by scikit-learn](https://scikit-learn.org/stable/developers/contributing.html#different-objects), specifically:
+- _Estimators_ implement a `fit` method.
+- _Predictors_ implement a `predict` method.
 
-Rather than providing deterministic output many predictors produced by disparity mitigation techniques in the fairness literature are randomized. As a consequence, it is possible to get different output from the predictor's `predict` method when providing samples with identical features.
 
-## Group Fairness
+**Randomization.** Unlike [scikit-learn estimators](https://scikit-learn.org/stable/glossary.html#term-estimator), estimators in fairlearn can produce randomized predictors. Randomization of predictions is required to satisfy many definitions of fairness. Because of randomization, it is possible to get different outputs from the predictor's `predict` method on identical data. For each of our methods, we provide explicit access to the probability distribution used for randomization.
```
riedgar-ms: Nitpick: this will read "Unlike scikit-learn estimators, estimators in fairlearn" …. I don't think having the same noun on either side of a comma looks good.

TERMINOLOGY.md
riedgar-ms: Also, I think it should be "native speaker of _a particular_ language" - except in rare cases, people are generally a "native speaker of language."

PR #77, opened 2019-10-09
Update GridSearch tests

riedgar-ms:
Update the GridSearch tests to target the new `Moment` based functionality

Removal of the older `metrics` functionality will come in a future changeset


fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -181,7 +194,7 @@ def fit(self, X, y, aux_data=None, **kwargs):
                     y_reduction = 1 * (weights > 0)
                     weights = weights.abs()
                 else:
-                    y_reduction = y
+                    y_reduction = y_vector
```
riedgar-ms: This was an actual bug which my tests found :-)

MiroDudik: I noticed! nice!

fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -135,6 +136,11 @@ def fit(self, X, y, aux_data=None, **kwargs):
         # Extract the target attribute
         A = self._make_vector(aux_data, "aux_data")
```
romanlutz: aux_data?

PR #82, opened 2019-10-10
change moment api and switch to sensitive_features

MiroDudik:
refactor moments and reductions


fairlearn/reductions/moments/moment.py
```diff
@@ -3,30 +3,32 @@
...
 
-_REDUCTION_TYPE_CLASSIFICATION = "classification"
-_REDUCTION_TYPE_LOSS_MINIMIZATION = "loss_minimization"
-
 _GROUP_ID = "group_id"
+_EVENT = "event"
 _LABEL = "label"
 _LOSS = "loss"
 _PREDICTION = "pred"
+_EVENT = "event"
```
riedgar-ms: Why the redefinition? See line 7

test/unit/reductions/grid_search/test_grid_search_regression.py
```diff
@@ -154,8 +154,8 @@ def test_bgl_lagrange_specifications():
 
     tradeoffs = [0, 0.25, 0.5, 0.75, 1]
 
-    target1.fit(X, y, aux_data=A, number_of_lagrange_multipliers=len(tradeoffs))
-    target2.fit(X, y, aux_data=A, lagrange_multipliers=tradeoffs)
+    target1.fit(X, y, group_membership=A, number_of_lagrange_multipliers=len(tradeoffs))
+    target2.fit(X, y, group_membership=A, lagrange_multipliers=tradeoffs)
```
riedgar-ms: These extra arguments (Lagrange multipliers) to `fit` shouldn't be there. I must have missed them on my trawl :-(

fairlearn/reductions/moments/conditional_selection_rate.py
```diff
@@ -0,0 +1,104 @@
...
+    def gamma(self, predictor):
+        """ Calculates the degree to which constraints are currently violated by
+        the predictor.
+        """
+        pred = predictor(self.X)
+        self.tags[_PREDICTION] = pred
+        expect_event = self.tags.groupby(_EVENT).mean()
+        expect_group_event = self.tags.groupby(
+            [_EVENT, _GROUP_ID]).mean()
+        expect_group_event["diff"] = expect_group_event[_PREDICTION] - expect_event[_PREDICTION]
```
riedgar-ms: Looks like I missed `diff` when extracting constants :-/

fairlearn/reductions/exponentiated_gradient/_lagrangian.py
```diff
@@ -42,10 +42,10 @@ def eval_from_error_gamma(self, error, gamma, lambda_vec):
...
         """
-        lambda_signed = self.constraints.lambda_signed(lambda_vec)
+        lambda_projected = self.constraints.project_lambda(lambda_vec)
         if self.opt_lambda:
-            L = error + np.sum(lambda_vec * gamma) \
-                - self.eps * np.sum(lambda_signed.abs())
+            L = error + np.sum(lambda_projected * gamma) \
+                - self.eps * np.sum(lambda_projected)
         else:
             L = error + np.sum(lambda_vec * gamma) \
```
riedgar-ms: Given that this first line is common, should it go outside the `if` statement?

fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -137,24 +136,24 @@ def fit(self, X, y, aux_data=None, **kwargs):
...
+        if isinstance(self.disparity_metric, ClassificationMoment):
             # We have a classification problem
             # Need to make sure that y is binary (for now)
             unique_labels = np.unique(y_vector)
             if not set(unique_labels).issubset({0, 1}):
                 raise RuntimeError(self._MESSAGE_Y_NOT_BINARY)
 
         # Prep the disparity metric and objective
-        self.disparity_metric.init(X, A, y_vector)
+        self.disparity_metric.load_data(X, y_vector, **kwargs)    # group_membership=A)
```
riedgar-ms: Don't need the end of this line?

MiroDudik: This is intentional. "fit" shouldn't be doing anything special with kwargs and just pass them on to the constraints (disparity_metric) and the objective. for the time being, we are doing some checks of sensitive_features inside "fit", but those kinds of checks should happen inside disparity_metric.

fairlearn/reductions/exponentiated_gradient/_lagrangian.py
MiroDudik: the first L is based on lambda_projected (both lines), the second L is based on lambda_vec (both lines).

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -62,9 +63,10 @@ def __init__(self, estimator, constraints, eps=0.01, T=50, nu=None, eta_mul=2.0)
         self._best_classifier = None
         self._classifiers = None
 
-    def fit(self, X, y, group_data=None, **kwargs):
+    def fit(self, X, y, **kwargs):
         # TODO: validate input data; unify between grid search and expgrad?
-        self._expgrad_result = exponentiated_gradient(X, group_data, y, self._estimator,
+        self._expgrad_result = exponentiated_gradient(X, kwargs[_KW_GROUP_MEMBERSHIP], y,
```
romanlutz: oh this is rough. There's no documentation of the argument, but then we'll fail right here when using it because kwargs won't have an entry named like that.

fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -107,18 +106,18 @@ def __init__(self,
...
             raise ValueError(self._MESSAGE_Y_NONE)
 
-        if aux_data is None:
-            raise RuntimeError("Must specify aux_data (for now)")
+        if _KW_GROUP_MEMBERSHIP not in kwargs:
+            raise RuntimeError("Must specify {0} (for now)".format(_KW_GROUP_MEMBERSHIP))
 
         # Extract the target attribute
-        A = self._make_vector(aux_data, "aux_data")
+        A = self._make_vector(kwargs[_KW_GROUP_MEMBERSHIP], _KW_GROUP_MEMBERSHIP)
```
romanlutz: Going back to `A` will only confuse people.

fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -137,24 +136,24 @@ def fit(self, X, y, aux_data=None, **kwargs):
...
+        if isinstance(self.disparity_metric, ClassificationMoment):
             # We have a classification problem
             # Need to make sure that y is binary (for now)
             unique_labels = np.unique(y_vector)
             if not set(unique_labels).issubset({0, 1}):
                 raise RuntimeError(self._MESSAGE_Y_NOT_BINARY)
 
         # Prep the disparity metric and objective
-        self.disparity_metric.init(X, A, y_vector)
+        self.disparity_metric.load_data(X, y_vector, **kwargs)    # sensitive_features=A)
```
romanlutz: Is there something you planned to do here? There's a comment at the end

test/unit/reductions/grid_search/test_grid_search_regression.py
```diff
@@ -154,8 +154,8 @@ def test_bgl_lagrange_specifications():
 
     tradeoffs = [0, 0.25, 0.5, 0.75, 1]
 
-    target1.fit(X, y, aux_data=A, number_of_lagrange_multipliers=len(tradeoffs))
-    target2.fit(X, y, aux_data=A, lagrange_multipliers=tradeoffs)
+    target1.fit(X, y, sensitive_features=A)
```
romanlutz: We should get rid of all usage of A to avoid confusion

fairlearn/metrics/metrics_engine.py
```diff
@@ -8,7 +8,7 @@
 _MESSAGE_SIZE_MISMATCH = "Array {0} is not the same size as {1}"
 
 
-def metric_by_group(metric_function, y_true, y_pred, group_data, sample_weight=None):
+def metric_by_group(metric_function, y_true, y_pred, sensitive_features, sample_weight=None):
```
riedgar-ms: Does `metric_by_group` need to change, now that we no longer have an argument `group_data` ?

riedgar-ms: Also, in this case, `sensitive_features` has to be singular.

fairlearn/reductions/exponentiated_gradient/_lagrangian.py
riedgar-ms: Sorry, missed that :-(

fairlearn/reductions/grid_search/grid_search.py
riedgar-ms: There are these in a few places. I think they're just conversion leftovers, and should just be removed.

fairlearn/post_processing/threshold_optimizer.py
```diff
@@ -376,18 +376,18 @@ def _vectorized_prediction(function_dict, group_data, scores):
...
+    :param sensitive_features: grouping attributes for each sample
+    :type sensitive_features: list, numpy.ndarray, pandas.DataFrame, or pandas.Series
     :param scores: vector of predicted values
     :type scores: list, numpy.ndarray, pandas.DataFrame, or pandas.Series
     """
     # handle type conversion to ndarray for other types
-    group_data_vector = _convert_to_ndarray(group_data, MULTIPLE_DATA_COLUMNS_ERROR_MESSAGE
-                                            .format("group_data"))
+    sensitive_features_vector = _convert_to_ndarray(
+        sensitive_features, MULTIPLE_DATA_COLUMNS_ERROR_MESSAGE.format("sensitive_features"))
```
riedgar-ms: `_KW_GROUP_MEMBERSHIP` ?

fairlearn/post_processing/threshold_optimizer.py
```diff
@@ -429,7 +429,7 @@ def _reformat_data_into_dict(key, data_dict, additional_data):
                                               additional_data.shape[1] > 1):
             # TODO: extend to multiple columns for additional_group data
             raise ValueError(
-                MULTIPLE_DATA_COLUMNS_ERROR_MESSAGE.format("group_data"))
+                MULTIPLE_DATA_COLUMNS_ERROR_MESSAGE.format("sensitive_features"))
```
riedgar-ms: `_KW_GROUP_MEMBERSHIP` ? And on line 446 below.

notebooks/Grid Search for Binary Classification.ipynb
```diff
@@ -60,7 +60,7 @@
     "    Y = np.concatenate((Y_a0, Y_a1), axis=None)\n",
     "\n",
     "    X = pd.DataFrame({\"credit_score_feature\": score_feature,\n",
-    "                      \"aux_data_feature\": A})\n",
+    "                      \"sensitive_features\": A})\n",
```
riedgar-ms: I don't think that the name of the column in `X` should be the same as the keyword argument. I made the previous version `aux_data_feature` for this reason. Otherwise, we might give the impression that it's a 'magic' column name.

fairlearn/reductions/moments/conditional_selection_rate.py
MiroDudik: addressed.

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
MiroDudik: [the truth is even more confusing... _KW_GROUP_MEMBERSHIP="sensitive_features"; so the constant is being renamed to _KW_SENSITIVE_FEATURES]

fairlearn/reductions/grid_search/grid_search.py
MiroDudik: Agreed. i'll fix this specific instance (A -> sensitive), and leave the additional fixing for a future PR... since this PR is blocking too many files.

fairlearn/reductions/grid_search/grid_search.py
MiroDudik: good catch. removing.

test/unit/reductions/grid_search/test_grid_search_regression.py
MiroDudik: Agreed. Shall we rename A -> sensitive?

fairlearn/metrics/metrics_engine.py
MiroDudik: Argh... i think that metric_by_group is actually okay to leave as is. I was thinking about leaving sensitive_features as a plural--since we are leaving it as a plural elsewhere even though we only support a single sensitive_feature everywhere. Thoughts?

fairlearn/post_processing/threshold_optimizer.py
MiroDudik: it's not a keyword argument here, so i think it would be a bit confusing. @romanlutz ?

fairlearn/metrics/metrics_engine.py
riedgar-ms: I had actually thought that was why we were using `group_data` here rather than `aux_data`  - because conceptually, we are just producing metrics by subgroups, without trying to treat them as sensitive, protected or whatever

fairlearn/post_processing/threshold_optimizer.py
MiroDudik: again... not sure.

fairlearn/post_processing/threshold_optimizer.py
riedgar-ms: If it's not a keyword argument, then it should be left as a string.

notebooks/Grid Search for Binary Classification.ipynb
MiroDudik: good idea. replacing by "example_sensitive_feature". reasonable?

test/unit/reductions/grid_search/test_grid_search_regression.py
riedgar-ms: I'd go for something a bit more descriptive. `grouping_data` or something like that.

fairlearn/post_processing/threshold_optimizer.py
riedgar-ms: If they aren't kwargs, leave them. I'd lost track of which file I was looking at.

notebooks/Grid Search for Binary Classification.ipynb
riedgar-ms: That works.

fairlearn/metrics/metrics_engine.py
MiroDudik: I see--very good point. I think I would go with group_membership here then. Apparently, group_data is confusing. Is that fine with you?

test/unit/reductions/grid_search/test_grid_search_regression.py
MiroDudik: it's a test, so we can call it example_sensitive_feature or test_sensitive_feat or something like that?

fairlearn/metrics/metrics_engine.py
riedgar-ms: `group_membership` is fine with me

riedgar-ms: Updated to `group_membership` and pushed to branch

PR #88, opened 2019-10-18
Mirodudik/remove quality metric

MiroDudik:


fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -88,20 +90,23 @@ class GridSearch(ReductionsEstimator):
     _MESSAGE_X_SENSITIVE_ROWS = "X and the sensitive features must have same number of rows"
 
     def __init__(self,
-                 learner,
+                 estimator,
                  disparity_metric,
-                 quality_metric,
+                 selection_rule=_TRADEOFF_OPTIMIZATION,
```
riedgar-ms: If it's a default value, the constant shouldn't be notionally private (i.e. start with underscore)

fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -179,30 +181,37 @@ def fit(self, X, y, **kwargs):
...
+            current_estimator = copy.deepcopy(self.estimator)
+            current_estimator.fit(X, y_reduction, sample_weight=weights)
+            def predict_fct(X): return current_estimator.predict(X)
 
-            nxt = GridSearchResult(current_learner, lambda_vec, quality)
+            nxt = GridSearchResult(current_estimator,
+                                   lambda_vec,
+                                   objective.gamma(predict_fct)[0],
+                                   self.disparity_metric.gamma(predict_fct)
+                                   )
```
riedgar-ms: Nitpick: would probably put closing parenthesis on previous line.

fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -88,20 +90,23 @@ class GridSearch(ReductionsEstimator):
     _MESSAGE_X_SENSITIVE_ROWS = "X and the sensitive features must have same number of rows"
 
     def __init__(self,
-                 learner,
+                 estimator,
                  disparity_metric,
-                 quality_metric,
+                 selection_rule=_TRADEOFF_OPTIMIZATION,
+                 constraint_weight=0.5,
```
riedgar-ms: Are there limits on the range of sensible values for `constraint_weight` ? I'm guessing [0, 1] ?

riedgar-ms: In fact, reading more of the code below, I'm pretty certain that that's the required range.

fairlearn/reductions/grid_search/grid_search.py
MiroDudik: I was thinking about "tradeoff_optimization" as something that the users should be inputting rather than grid_search.TRADEOFF_OPTIMIZATION, that's why the underscore, but I can drop the underscore.

PR #89, opened 2019-10-19
Extra classification metrics

riedgar-ms:
Add the following classification metrics:
- Specificity (aka TNR)
- miss (aka FNR)
- fallout (aka FPR)


fairlearn/metrics/extra_metrics.py
```diff
@@ -0,0 +1,28 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import sklearn.metrics as skm
+
+
+def specificity_score(y_true, y_pred, sample_weight=None):
```
riedgar-ms: I'm slightly concerned that we should be providing the richness of the `scikit-learn` metrics here - see for example:
https://scikit-learn.org/stable/modules/generated/sklearn.metrics.recall_score.html
They provide a lot of support for multiclass metrics

fairlearn/metrics/__init__.py
```diff
@@ -6,12 +6,17 @@
 from .group_metric_result import GroupMetricResult  # noqa: F401
 from .metrics_engine import metric_by_group, make_group_metric  # noqa: F401
 
+from .extra_metrics import specificity_score, miss_score, fallout_score
+
 # Classification metrics
 group_accuracy_score = make_group_metric(skm.accuracy_score)
 group_confusion_matrix = make_group_metric(skm.confusion_matrix)
 group_precision_score = make_group_metric(skm.precision_score)
```
riedgar-ms: Now I look at these again, I have a nasty feeling that wrapping the `scikit-learn` metrics like this is implicitly restricting their use to binary classification on {0, 1}

fairlearn/metrics/extra_metrics.py
```diff
@@ -0,0 +1,28 @@
...
+    tpr = skm.recall_score(y_true, y_pred, sample_weight=sample_weight)
+
+    # FNR == 1 - TPR
+    return 1 - tpr
+
+
+def fallout_score(y_true, y_pred, sample_weight=None):
+    # aka False Positive Rate
+    # Since we use specificity, also restricted to binary classification
+    return 1 - specificity_score(y_true, y_pred, sample_weight)
```
rihorn2: Would you consider including fpr, fnr? I can write them in the widget otherwise. Similarly, we need the outcomes = class 0 binned by group and overall.

MiroDudik: "miss rate" and "fallout rate" are not scores (score is something that you want to maximize). if anything, they are error metrics--but the names are somewhat obscure, so I'm not sure we want to define them. I would suggest just defining specificity_score as you do, and then metrics for fpr, fnr, tpr, tnr, etc.

riedgar-ms: I can rename to `fallout_rate` per @MiroDudik 's comment, but... this is the false positive rate. I understood from @mesameki 's interviews that these were the names data scientists preferred?

In any case, this needs revision to cope with non-binary classification.

fairlearn/metrics/extra_metrics.py
```diff
@@ -0,0 +1,53 @@
...
+    cm = skm.confusion_matrix(y_true, y_pred, sample_weight=sample_weight)
+    # Taken from
+    # https://scikit-learn.org/stable/modules/generated/sklearn.metrics.confusion_matrix.html
+    # This restricts us to binary classification
+    tn, fp, fn, tp = cm.ravel()
+    return tn / (tn+fp)
+
+
+def miss_rate(y_true, y_pred, sample_weight=None):
+    # aka False Negative Rate
```
romanlutz: Can we make this "real" documentation?
```
def f():
    """ doc
    """
```

fairlearn/metrics/skm_wrappers.py
```diff
@@ -0,0 +1,95 @@
...
+                        sample_weight=None):
+
+    def internal_zol_wrapper(y_true, y_pred, sample_weight=None):
+        return skm.zero_one_loss(y_true, y_pred,
+                                 normalize=normalize,
+                                 sample_weight=sample_weight)
+
+    return metric_by_group(internal_zol_wrapper, y_true, y_pred, group_membership, sample_weight)
+
+# --------------------------------------------------------------------------------------
```
romanlutz: what's this for?

test/unit/metrics/test_extra_metrics.py
```diff
@@ -0,0 +1,156 @@
...
+
+def test_specificity_some_correct_with_false_negative():
+    y_true = [0, 0, 0, 0, 1]
+    y_pred = [0, 0, 1, 0, 0]
+
+    result = metrics.specificity_score(y_true, y_pred)
+    assert result == 0.75
+
+
+# ========================================
```
romanlutz: if there should be a logical separation, can we just put it in a different file?

fairlearn/metrics/skm_wrappers.py
riedgar-ms: The separator line?

test/unit/metrics/test_group_sklearn_wrappers.py
```diff
@@ -2,46 +2,187 @@
...
 # ======================================================
 
+a = "a"
+b = "b"
+c = "c"
+
 Y_true = [0, 0, 0, 1, 1, 1, 0, 0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1]
 Y_pred = [1, 0, 1, 0, 1, 0, 0, 1, 0, 0, 0, 0, 0, 1, 1, 1, 1, 1]
+
+Y_t3 = [a, b, c, c, c, b, b, b, c, c, a, a, a, a, a, b, c, c]
```
romanlutz: I'm assuming t is short for true, and p for pred? Can we not collapse that?

test/unit/metrics/test_extra_metrics.py
riedgar-ms: If the files grow much bigger, certainly. I'm trying to balance keeping things separate with having lots of tiny files scattered all over the place.

test/unit/metrics/test_group_sklearn_wrappers.py
riedgar-ms: Fair point. Done

fairlearn/metrics/skm_wrappers.py
romanlutz: Yes :-) never seen this in python before. You can use new files if you want separation.

riedgar-ms: In this case, they're all simple wrappers for `scikit-learn` and I'm just flagging where the regression ones begin.

test/unit/metrics/test_extra_metrics.py
riedgar-ms: I've pulled `selection_rate` into its own file, since that is a bit more sophisticated.

PR #92, opened 2019-10-22
GridSearch alternate constraint specification

riedgar-ms:
Allow the constraint to be specified as a string as well as a `Moment` object


fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -74,6 +76,10 @@ def accumulate_integer_grid(self, index, max_val):
                 self.accumulate_integer_grid(index+1, max_val-abs(current_value))
 
 
+_constraint_strings = {"demographic_parity": DemographicParity(),
```
riedgar-ms: We should standardise the strings. @romanlutz  has used CamelCase, but I think this is more like `sklearn` usage. For example:
https://scikit-learn.org/stable/modules/generated/sklearn.metrics.mean_squared_error.html#sklearn.metrics.mean_squared_error

fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -98,9 +104,14 @@ def __init__(self,
                  grid_limit=2.0,
                  grid=None):
         self.estimator = estimator
-        if not isinstance(constraints, Moment):
-            raise RuntimeError("Unsupported disparity metric")
-        self.constraints = constraints
+        if isinstance(constraints, Moment):
+            self.constraints_ = constraints
```
riedgar-ms: I think this trailing underscore is following `sklearn` - AIUI, the convention for this is `{ get: private set;}` in divers other languages

fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -98,9 +104,14 @@ def __init__(self,
...
+        elif isinstance(constraints, str) and constraints in _constraint_strings:
+            self.constraints_ = copy.deepcopy(_constraint_strings[constraints])
+        else:
+            msg = ("Unsupposed constraint. Please supply a Moment object or "
+                   "string from {0}").format(list(_constraint_strings.keys()))
+            raise RuntimeError(msg)
```
riedgar-ms: ToDo: A test to check this is raised OK

fairlearn/reductions/grid_search/grid_search.py
romanlutz: It should definitely be the same, not two different ways of specifying it.
Let's also fix the name in a separate variable 
```
_demographic_parity = "demographic_parity"
```
so that it can be used in tests without having to make 100 replacements

fairlearn/reductions/grid_search/grid_search.py
romanlutz: It's generally used in python to avoid naming conflicts with keywords, https://stackoverflow.com/questions/21075208/whats-the-advantage-of-a-trailing-underscore-in-python-naming

Therefore I would not suggest using this. If you want to restrict get/set there are other ways of achieving that.

fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -98,9 +104,14 @@ def __init__(self,
...
         self.estimator = estimator
-        if not isinstance(constraints, Moment):
-            raise RuntimeError("Unsupported disparity metric")
-        self.constraints = constraints
+        if isinstance(constraints, Moment):
+            self.constraints_ = constraints
+        elif isinstance(constraints, str) and constraints in _constraint_strings:
+            self.constraints_ = copy.deepcopy(_constraint_strings[constraints])
+        else:
+            msg = ("Unsupposed constraint. Please supply a Moment object or "
```
romanlutz: what's the parenthesis for?

riedgar-ms: Multi-line string

fairlearn/reductions/grid_search/grid_search.py
riedgar-ms: I was going on the link from `sklearn` :
https://scikit-learn.org/stable/glossary.html#term-attributes
which states "A public attribute may have the same name as a constructor parameter, with a _ appended. This is used to store a validated or estimated version of the user’s input." Which is not quite what I wrote above :-)

fairlearn/reductions/grid_search/grid_search.py
riedgar-ms: Done - can you take a look and tell me if you like it :-)

fairlearn/constants.py
```diff
@@ -0,0 +1,5 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+demographic_parity = "demographic_parity"
+bounded_group_loss = "bounded_group_loss"
```
MiroDudik: why not DEMOGRAPHIC_PARITY and BOUNDED_GROUP_LOSS to match our convention for constants elsewhere?

fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -74,6 +77,10 @@ def accumulate_integer_grid(self, index, max_val):
                 self.accumulate_integer_grid(index+1, max_val-abs(current_value))
 
 
+_constraint_strings = {flc.demographic_parity: DemographicParity(),
+                       flc.bounded_group_loss: GroupLossMoment(ZeroOneLoss())}
+
```
MiroDudik: I don't think that ZeroOneLoss is a good default--and I'm not sure there's a good default here. I'd prefer to require an additional keyword argument "loss" defaulted to None, and would throw an error when constraints="bounded_group_loss", but loss is not set.

fairlearn/constants.py
riedgar-ms: Yes, they should be (having just had a look at https://www.python.org/dev/peps/pep-0008/#constants)

fairlearn/reductions/grid_search/grid_search.py
riedgar-ms: I don't think that adding another keyword argument to the constructor to allow one of the other arguments to possibly be a string is a good pattern.

riedgar-ms:
We have agreed that this is not a good way forward due to the issue of supplying the loss function to bounded_group_loss

PR #97, opened 2019-10-23
Add performance test scaffolding using tempeh, UCI adult dataset, and COMPAS dataset

romanlutz:
There are still a few TODOs in there which will be addressed in subsequent PRs.


devops/nightly.yml
```diff
@@ -34,5 +33,9 @@ jobs:
     pyVersions: [3.6, 3.7] 
 
 - template: notebook-job-template.yml
+  parameters:
+    vmImage: 'ubuntu-16.04'
+
+- template: perf-tests-job-template.yml
```
riedgar-ms: Should we have a separate "nightly performance" build?

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -65,7 +65,20 @@ def __init__(self, estimator, constraints, eps=0.01, T=50, nu=None, eta_mul=2.0)
 
     def fit(self, X, y, **kwargs):
         # TODO: validate input data; unify between grid search and expgrad?
-        self._expgrad_result = exponentiated_gradient(X, kwargs[_KW_SENSITIVE_FEATURES], y,
+
+        if type(X) in [np.ndarray, list]:
```
riedgar-ms: Not necessary right now, but this (and the next if block) should probably be in a `utils` library and used in other methods

test/perf/test_performance.py
```diff
@@ -0,0 +1,86 @@
...
+        # sensitive feature is either race or sex
+        # TODO add another case where we use sex as well, or both (?)
+        sensitive_features_train = dataset.race_train
+        sensitive_features_test = dataset.race_test
+    else:
+        raise ValueError("Sensitive features unknown for dataset {}"
+                         .format(perf_test_configuration.dataset))
+
+    print("Fitting estimator")
+    estimator = models[perf_test_configuration.predictor]()
```
riedgar-ms: Might be interesting to know how long the 'raw' estimator takes.... are nested timers available?

test/perf/test_performance.py
```diff
@@ -0,0 +1,86 @@
...
+    unconstrained_predictor = models[perf_test_configuration.predictor]()
+    unconstrained_predictor.fit(dataset.X_train, dataset.y_train)
+    print("Done fitting estimator")
+
+    start_time = time()
+    if perf_test_configuration.mitigator == ThresholdOptimizer.__name__:
```
riedgar-ms: If we're only able to time one thing, then the constructors aren't that interesting

test/perf/test_performance.py
```diff
@@ -0,0 +1,86 @@
...
+                                          constraints=DemographicParity())
+    elif perf_test_configuration.mitigator == GridSearch.__name__:
+        mitigator = GridSearch(estimator=estimator,
+                               constraints=DemographicParity())
+    else:
+        raise Exception("Unknown mitigation technique.")
+
+    print("Fitting mitigator")
+
+    mitigator.fit(dataset.X_train, dataset.y_train, sensitive_features=sensitive_features_train)
```
riedgar-ms: As a follow up, it would be good if our fairlearn estimators could also keep track of how many times they called `fit` on the supplied estimator. Even the actual time spent there as well. For performance testing of fairlearn, we just want to make sure that we're not doing anything stupid; most of the time should be spend in the raw estimator.

test/perf/test_performance.py
```diff
@@ -0,0 +1,86 @@
...
+        mitigator.predict(dataset.X_test, sensitive_features=sensitive_features_test)
+    else:
+        mitigator.predict(dataset.X_test)
+
+    # TODO evaluate accuracy/fairness tradeoff
+
+    total_time = time() - start_time
+    assert total_time <= perf_test_configuration.max_time_consumption
+    print("Total time taken: {}s".format(total_time))
+    print("Maximum allowed time: {}s".format(perf_test_configuration.max_time_consumption))
```
riedgar-ms: Does the perf framework you're using have some way of reporting these up? Having to check build logs is not ideal.

devops/nightly.yml
```diff
@@ -10,7 +10,6 @@ schedules:
   branches:
     include:
       - master
-      - dev
```
riedgar-ms: I can take care of these bits in a separate PR. I need to delete the 'schedule test' YAML anyway

devops/nightly.yml
romanlutz: I thought nightly contains everything that runs nightly? :D What do you see as the benefits of separating it?

fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
romanlutz: Absolutely! Just needed a couple of things here to get it running. I want to tackle this validation separately (or you, whoever gets to it first)

test/perf/test_performance.py
romanlutz: I mean right now it's all fairly "dumb" in the sense that we just count from start to end. We can think about creating a way to expose that. Don't want to do it in this PR, though.

test/perf/test_performance.py
romanlutz: Yes! I'll count all your great suggestions as follow-up tasks/feature requests, because this needs to go in quickly so that I can use the same for notebook examples.

test/perf/test_performance.py
romanlutz: No! It's an issue I've created on the framework itself. Ideally I want some sort of persistent storage for the results, too.

devops/nightly.yml
romanlutz: Sure, whatever works. It's not critical, I just stumbled over it and there's no harm in getting rid of it. Whoever goes first...

test/perf/test_performance.py
romanlutz: Unless we do something complicated there!

devops/nightly.yml
riedgar-ms: It keeps the current Nightly build focussed on correctness. Performance would be separate from that

devops/nightly.yml
riedgar-ms: I've already done these :-)

test/perf/test_performance.py
romanlutz: will tackle that through tempeh

romanlutz: or through pytest config here later

devops/nightly.yml
romanlutz: OK! Separated it out

PR #100, opened 2019-10-23
Sphinx onboarding

riedgar-ms:
Initial onboarding to Sphinx
This is to allow everyone to run documentation builds locally and see how things look; it is not a complete onboarding.

To use:
```
pip install -r requirements.txt
cd docs/
sphinx-apidoc  --module-first --no-headings --no-toc --implicit-namespaces ../fairlearn -o source/
sphinx-build . _build
```
The `sphinx-build` step currently has warnings, but documentation is produced


docs/Makefile
```diff
@@ -0,0 +1,20 @@
+# Minimal makefile for Sphinx documentation
```
riedgar-ms: This was generated by `sphinx-quickstart`

docs/conf.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+# -- Project information -----------------------------------------------------
+
+project = 'fairlearn'
+copyright = '2019, Microsoft'
+author = 'Microsoft'
+
+# The full version, including alpha/beta/rc tags
+release = '0.3.0-pre-alpha-interim'
```
romanlutz: oh no, another place where we need to keep a consistent version number :-(

fairlearn/metrics/__init__.py
```diff
@@ -17,13 +20,34 @@
...
+"""A grouped metric for the fallout rate
+"""
 
 # Regression metrics
 group_max_error = make_group_metric(skm.max_error)
+"""A grouped wrapper around the sklearn.metrics.max_error routine
+"""
+
 group_mean_absolute_error = make_group_metric(skm.mean_absolute_error)
+"""A grouped wrapper around the sklearn.metrics.mean_absolute_error
```
romanlutz: Interesting that the doc comes *after* the definition and not before :O

fairlearn/metrics/metrics_engine.py
```diff
@@ -11,20 +11,29 @@
 def metric_by_group(metric_function, y_true, y_pred, group_membership, sample_weight=None):
     """ Applies a metric to each subgroup of a set of data
 
-    :param metric_function
-    :type Function with signature (y_true, y_pred, sample_weight=None) which returns a scalar
+    Parameters
```
romanlutz: Are you sure this is right? I'm 99% sure we use a different standard for the AML SDK.

romanlutz: https://stackoverflow.com/questions/3898572/what-is-the-standard-python-docstring-format
Pretty sure we want to stay with reST as opposed to numpy doc

romanlutz: if we do want to switch we can use pyment

requirements.txt
```diff
@@ -17,4 +17,8 @@ pytest-cov
 # Required for notebooks
 jupyter
 nbval
-shap
\ No newline at end of file
+shap
```
romanlutz: ?

romanlutz: What happened here?

docs/conf.py
riedgar-ms: There may well be a way of reading this from the python

fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -30,7 +30,7 @@ def __init__(self, grid_size, grid_limit, pos_basis, neg_basis, neg_allowed, for
             true_dim = self.dim
 
         # a conservative lower bound on the scaling parameter of the grid
-        n_units = (float(grid_size) / (2.0**neg_allowed.sum())) ** (1.0/true_dim) - 1
+        n_units = (float(grid_size) / (2.0**neg_allowed.sum())) ** (1.0 / true_dim) - 1   # noqa: E501
```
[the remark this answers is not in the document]
riedgar-ms: I couldn't see an obvious place to break this one, and wanted something quickly :-)

requirements.txt
riedgar-ms: Newline added perhaps?

fairlearn/metrics/__init__.py
riedgar-ms: Indeed.

fairlearn/metrics/metrics_engine.py
riedgar-ms: Reading:
http://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#module-sphinx.ext.autodoc
This is because I was using the Napoleon extension. Which suggests that the markup is not properly separated from the presentation :-(

PR #106, opened 2019-10-24
Metrics Reorganisation

riedgar-ms:
A major reworking of `metrics` after looking at the documentation generated by sphinx:
- Many files have gained an underscore on the start of their names, to make them private
- Updated `import` directives and added `__all__` settings to get things in the correct place
- Converted `GroupMetricResult` to using properties
- Various smaller documentation tweaks


fairlearn/metrics/extra_metrics.py
```diff
@@ -1,9 +1,29 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
+"""This module contains a variety of extra metrics which are useful for assessing fairness
```
riedgar-ms: I'm open to other names beyond the rather generic `extra_metrics`

romanlutz: extra? As in additional? Or should we just have a separate file for each of these.

riedgar-ms: Well, the naming question is no longer relevant....

PR #107, opened 2019-10-25
GridSearch Properties and Protection

riedgar-ms:
Minor improvements for `GridSearch`:
- `GridSearchResult` now uses properties for everything
- `GridSearch` now uses properties for `all_results` and `best_result`, allowing them to be documented
- Add a check on the `predict` and `predict_proba` methods of `GridSearch` to make sure that `fit` has been called previously

### DIFF
--- a/fairlearn/reductions/grid_search/grid_search.py
+++ b/fairlearn/reductions/grid_search/grid_search.py
@@ -120,6 +120,25 @@ class GridSearch(Reduction):
         self.grid_limit = float(grid_limit)
         self.grid = grid
 
+        self._all_results = []
+        self._best_result = None
+
+    @property
+    def all_results(self):
+        """A list of :class:`GridSearchResult` from each
+        point in the grid
+        """
+        return self._all_results
+
+    @property
+    def best_result(self):
+        """The best result found from the grid search.
+        The predictor contained in this instance of
+        :class:`GridSearchResult` is used in calls to
+        ``predict`` and ``predict_proba``.
+        """
+        return self._best_result
+
     def fit(self, X, y, **kwargs):
         """Runs the grid search. This will result in multiple copies of the
         estimator being made, and the `fit` method of each one called.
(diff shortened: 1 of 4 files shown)


fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -233,17 +253,21 @@ def predict(self, X):
...
         best model found by the grid search. The underlying
-        estimator must support predict_proba for this
+        estimator must support ``predict_proba`` for this
         to work.
 
         :param X: The data for which predictions are required
         :type X: Array
         """
+        if self.best_result is None:
+            raise NotFittedException(_NO_PREDICT_BEFORE_FIT)
```
MiroDudik: should this be self._best_result?

fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -233,17 +253,21 @@ def predict(self, X):
             the result will be a scalar. Otherwise the result will be an
         :rtype: Scalar or array
         """
+        if self.best_result is None:
+            raise NotFittedException(_NO_PREDICT_BEFORE_FIT)
```
MiroDudik: should this be self._best_result? does the property magic take care of this somehow?

fairlearn/reductions/grid_search/grid_search.py
riedgar-ms: It could be - this will read it via the property

fairlearn/reductions/grid_search/grid_search.py
riedgar-ms: Yes. Magic in the property makes this work. However, by only having the 'getter' property public, users are dissuaded from fiddling with the results.

fairlearn/reductions/grid_search/grid_search.py
```diff
@@ -120,6 +121,25 @@ def __init__(self,
...
+    @property
+    def all_results(self):
+        """A list of :class:`GridSearchResult` from each
+        point in the grid
+        """
+        return self._all_results
+
+    @property
+    def best_result(self):
+        """The best result found from the grid search.
```
romanlutz: you have 99 characters, no need to fit into 50 :-)

PR #108, opened 2019-10-25
Move exponentiated gradient logic into class and reformat existing documentation

romanlutz:
adjusting to reST format


fairlearn/reductions/exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -113,162 +227,42 @@ def _pmf_predict(self, X):
...
-
-    return result
+    def posterior_predict(self, X):
+        # TODO provide implementation
+        raise NotImplementedError()
+
+    def posterior_predict_proba(self, X):
+        # TODO provide implementation
+        raise NotImplementedError()
+
```
MiroDudik: let's also drop posterior_predict and posterior_predict_proba

PR #109, opened 2019-10-25
Reorganise Reductions

riedgar-ms:
Updates to `reductions` so that everything appears under that top level only. Several directories have acquired an underscore prefix to their names as a result.


fairlearn/reductions/_exponentiated_gradient/__init__.py
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+from .exponentiated_gradient import ExponentiatedGradient  # noqa: F401
+from .exponentiated_gradient import ExponentiatedGradientResult  # noqa: F401
+
+__all__ = [
+    "ExponentiatedGradieht",
```
romanlutz: typo

riedgar-ms: Hmmm... I wonder how this was working then :-/

PR #110, opened 2019-10-25
Unified input validation for reductions

romanlutz:


fairlearn/_input_validation.py
```diff
@@ -0,0 +1,93 @@
...
+
+_MESSAGE_X_NONE = "Must supply X"
+_MESSAGE_Y_NONE = "Must supply y"
+_MESSAGE_X_Y_ROWS = "X and y must have same number of rows"
+_MESSAGE_X_SENSITIVE_ROWS = "X and the sensitive features must have same number of rows"
+
+_KW_SENSITIVE_FEATURES = "sensitive_features"
+
+
+def _validate_and_reformat_reductions_input(X, y, enforce_binary_sensitive_feature=False,
```
riedgar-ms: A minor thing, but if the filename has an underscore, you don't need an underscore on the methods therein :-)

romanlutz: This one is so that people don't start relying on it. The filename was for doc, right?

riedgar-ms: Well, I don't think people should be looking in an underscored file. It's not a big thing, and does no harm to leave it. I didn't go through everything in reductions and metrics to make sure that files with underscored names only contained underscored things.

romanlutz: I don't think that's necessary either. I just worry about methods moving between files, so if these suddenly go into a different file then they're not "private" anymore. I don't see any harm in leaving it the way it is right now. If you feel very strongly that it should change I can adjust.

riedgar-ms: My feelings are definitely not strong on the matter :-)

PR #112, opened 2019-10-25
Reorganise postprocessing

riedgar-ms:
Modify the `postprocessing` module so that only `Postprocessing` and `ThresholdOptimizer` are exported. Update tests and notebooks accordingly


fairlearn/postprocessing/__init__.py
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+
+from ._post_processing import PostProcessing  # noqa: F401
+from ._threshold_optimizer import ThresholdOptimizer  # noqa: F401
```
MiroDudik: I think the subdirectory should also be called _postprocessing. For the class name I'm indifferent between Postprocessing (maybe more logical given the directory and package name) and PostProcessing.

PR #113, opened 2019-10-25
Requirements pinning

riedgar-ms:
Several enhancements to improve our backwards-compatibility story
- Add minimum versions of packages (and Python) to `setup.py`
- Add the same minimum versions to `requirements.txt` (only for packages listed in `setup.py`)
- Add a `requirements-fixed.txt` which pins packages to the minimums in `requirements.txt`
- Update the build pipelines to make the choice of file for `pip install -r` a parameter
- Add a "Nightly-Requirements-Fixed" build which uses `requirements-fixed.txt`

Unfortunately, the fixed requirements (specifically `matplotlib`) are not playing well with MacOS, leaving a slight hole in the new nightly build


riedgar-ms:
The new build definition:
https://dev.azure.com/responsibleai/fairlearn/_build/results?buildId=1503

PR #115, opened 2019-10-25
Consistency for Requirements

riedgar-ms:
Script to check that two requirements files for pip are consistent. Use in the builds to compare `requirements.txt` and `requirements-fixed.txt`

There is no support for `setup.py` at the present time.


riedgar-ms:
Sample build:
https://dev.azure.com/responsibleai/fairlearn/_build/results?buildId=1527
Look at the "Run check on requirements file" step

PR #118, opened 2019-10-27
Use setup.cfg

riedgar-ms:
Move the data embedded in `setup.py` into the `setup.cfg` file. This stops us mixing code and data, and will allow the consistency checking script for requirements to be extended to cover `setup.cfg` as well

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,5 +1,5 @@
 [metadata]
-name = attr: fairlearn.__name__
+name = fairlearn
 version = attr: fairlearn.__version__
 author = Miroslav Dudik
 author_email = mdudik@gmail.com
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -1,2 +1,43 @@
+[metadata]
+name = fairlearn
+version = attr: fairlearn.__version__
+author = Miroslav Dudik
```
romanlutz: Time to adjust that? I think this can also have authors instead of singular author?

setup.cfg
```diff
@@ -1,2 +1,43 @@
...
+author_email = mdudik@gmail.com
+description = Algorithms for fair classification
```
romanlutz: also should be adjusted...

romanlutz:
I hope this won't break other things. I think a lot of tools are used to checking setup.py

riedgar-ms:
@romanlutz  which tools check `setup.py` as well?

riedgar-ms:
Closing for now. I still think this might be a good idea, but we need to sort out what's happening with both this and requirements.txt

PR #120, opened 2019-10-28
random_state on predict instead of ThresholdOptimizer constructor

romanlutz:
`random_state` is currently on the constructor, but the method itself (i.e. the logic in `fit`) isn't randomized, so it could lead the user to believe that it is. Instead, we add it to `predict` now where it is actually used.

The other changes are merely to get everything to work, i.e. a few follow-up corrections from earlier changes (post-processing -> postprocessing, test corrections)


test/perf/test_performance.py
```diff
@@ -46,6 +46,8 @@ def test_perf(perf_test_configuration, request):
         # TODO add another case where we use sex as well, or both (?)
         sensitive_features_train = dataset.race_train
         sensitive_features_test = dataset.race_test
+        dataset.y_train = dataset.y_train.astype(int)
```
riedgar-ms: Is this change relevant?

test/perf/test_performance.py
```diff
@@ -6,10 +6,10 @@
 from time import time
 from tempeh.configurations import models, datasets
 
-from fairlearn.post_processing import ThresholdOptimizer
-from fairlearn.post_processing.threshold_optimizer import DEMOGRAPHIC_PARITY
+from fairlearn.postprocessing import ThresholdOptimizer
+from fairlearn.postprocessing._threshold_optimizer import DEMOGRAPHIC_PARITY
```
riedgar-ms: These look like leftovers from the file reorg :-)

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -104,7 +100,14 @@ def fit(self, X, y, *, sensitive_features, **kwargs):
         self._post_processed_predictor_by_attribute = threshold_optimization_method(
             sensitive_features, y, scores, self._grid_size, self._flip, self._plot)
 
-    def predict(self, X, *, sensitive_features):
+    def predict(self, X, *, sensitive_features, random_state=None):
+        """
+        :param random_state: set to a constant for reproducibility
+        :type random_state: int
+        """
+        if random_state:
```
riedgar-ms: This is rather less sophisticated than `sklearn` allows, but it is at least a subset

test/perf/test_performance.py
romanlutz: Yes. Not sure what you mean by that, but it doesn't run otherwise

test/perf/test_performance.py
romanlutz: correct

test/perf/test_performance.py
riedgar-ms: Why does changing where the randomisation happens require this change?

fairlearn/postprocessing/_threshold_optimizer.py
romanlutz: Right, that's because I'm not using it with the same sophistication as of now. Didn't want to waste time on that yet

test/perf/test_performance.py
romanlutz: It doesn't. It is currently broken, so to get it to work I need to make this change. I want to get this particular change in asap because it actually changes the API, so I would prefer not to do multiple PRs for minimal changes like this. That's why this is included here.

PR #122, opened 2019-10-29
GridSearch with COMPAS

riedgar-ms:
Add a notebook for `GridSearch` based on the COMPAS dataset. Also uses the new widget


riedgar-ms:
Will use Law School instead

PR #124, opened 2019-10-29
Handle more input cases

rihorn2:
Support sensitive_features as 1d, and predicted_ys as 1d. Convert pd.Series to python list. warn if mismatch in size with non-needed parameter, and throw for needed parameter.

### DIFF
--- a/fairlearn/widget/fairlearnDashboard.py
+++ b/fairlearn/widget/fairlearnDashboard.py
@@ -25,8 +25,7 @@ class FairlearnDashboard(object):
             self, *,
             sensitive_features,
             true_y, predicted_ys,
-            class_names=None,
-            feature_names=None,
+            sensitive_feature_names=None,
             is_classifier=None):
 
         """Initialize the fairlearn Dashboard.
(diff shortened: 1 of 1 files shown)


fairlearn/widget/fairlearnDashboard.py
```diff
@@ -126,20 +121,34 @@ def __init__(
...
+        
+        if np.shape(self._true_y)[0] != np.shape(dataset)[0]:
+            raise ValueError("Sensitive features shape does not match true y shape")
+
 
         dataArg = {
-            "true_y": self._convertToList(true_y),
-            "predicted_ys": self._convertToList(predicted_ys),
-            "dataset": self._convertToList(sensitive_features),
+            "true_y": self._true_y,
```
romanlutz: why do we call it true_y, but predicted_ys with s at the end?

romanlutz: oh because it's literally multiple ... got it

fairlearn/widget/fairlearnDashboard.py
```diff
@@ -178,12 +187,20 @@ def _on_request(self, change):
     def _show(self):
         display(self._widget_instance)
 
+    def _sanitizeDataShape(self, dataset):
```
romanlutz: nit: python doesn't do camel case, but 
_sanitize_data_shape

same in filenames btw

fairlearn/widget/fairlearnDashboard.py
rihorn2: this is a good terminology question, true y is 1d, where predicted ys is 2d, one per model. If a 1d value is passed in for predicted ys, we will wrap it to make it 2d. If the terminology is bad, I'll change the names

fairlearn/widget/fairlearnDashboard.py
```diff
@@ -126,20 +121,34 @@ def __init__(
...
 
-        if feature_names is not None:
-            dataArg["features"] = self._convertToList(feature_names)
-
-        if class_names is not None:
-            dataArg["classes"] = self._convertToList(class_names)
+        if sensitive_feature_names is not None:
+            sensitive_feature_names =  self._convertToList(sensitive_feature_names)
+            if np.shape(dataset)[1] != np.shape(sensitive_feature_names)[0]:
+                raise Warning("Feature names shape does not match dataset, ignoring")
```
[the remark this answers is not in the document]
rihorn2: just the variable name? I can change dataset to _sensitive_features?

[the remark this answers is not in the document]
riedgar-ms: I'm guessing that it is just the variable name. As I said, a very small point :-)

PR #125, opened 2019-10-29
Documentation Fixes and related improvements

riedgar-ms:
More documentation tweaks:
- Move everything from `_group_metrics.py` into corresponding `__init__.py` so that we can get the documentation showing for all the automatic wrappers
- Fill out some more `sklearn` links in the metrics documentation
- Update `ExponentiatedGradientResult` with properties, so that the contents can be documented
- Move the `Postprocessing` documentation into the correct locations


fairlearn/metrics/_extra_metrics.py
```diff
@@ -30,7 +31,10 @@ def miss_rate(y_true, y_pred, sample_weight=None):
     """
     The miss rate is also known as the False Negative Rate.
     At the present time, this routine only supports binary
-    classifiers with labels taken from {0, 1}
+    classifiers with labels taken from {0, 1}.
+    By definition, this is the complement of the True Positive
```
romanlutz: nit: you have 99 characters, no need to wrap so early

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -211,6 +231,10 @@ def fit(self, X, y, **kwargs):
         # TODO: figure out whether we should keep the remaining data of the result object
 
     def predict(self, X):
+        """Provide a prediction for the given input data.
+        Note that this is non-deterministic, due to the nature of the
+        exponentiated gradient algorithm
+        """
```
romanlutz: no description of X?

riedgar-ms: OK, OK......

PR #126, opened 2019-10-30
text updates, fix regression bug, allow disparity by outputs

rihorn2:


fairlearn/widget/fairlearnDashboard.py
```diff
@@ -75,6 +75,14 @@ def __init__(
                 "model_type": ["classification"],
                 "function": group_fallout_rate
             },
+            "false_positive_over_total": {
+                "model_type": [],
+                "function": group_fallout_rate
```
riedgar-ms: Should this now be `group_mean_overpredict` ?

rihorn2: Yes, based on our discussion yesterday, the use of this method was replaced with mean over/under prediction. This method is never called or presented, but I'd rather not remove code from the branch I'd like to have ready for demos. A TODO will be to remove all the methods I added from metrics that are not actually needed in the dashboard

riedgar-ms: OK. I was assuming that was what was giving the nicely summing bar charts

PR #134, opened 2019-10-31
Notebook for GridSearch for Admissions

riedgar-ms:
A new notebook for `GridSearch` based on an Admissions scenario
Uses synthetic data, with the sensitive feature (gender) held out from the training data


riedgar-ms:
Strange things happening with signoff

PR #145, opened 2019-11-01
tempeh upgrade to v0.1.7 to fix notebooks in python 3.5

romanlutz:
The issue before was that we were catching a ModuleNotFoundError for not having tensorflow/pytorch installed, but ModuleNotfoundError itself doesn't exist in py3.5, so that was generalized in tempeh release v0.1.7 to catch ImportError. Adding notebook test for py3.5 back.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,7 +14,7 @@ requirements-parser
 # Pin pytest due to VS Code issue
 pytest==5.0.1
 pytest-cov
-tempeh==0.1.6
+tempeh==0.1.7
 
 # Required for notebooks
 jupyter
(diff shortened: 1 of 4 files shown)


requirements.txt
```diff
@@ -14,7 +14,7 @@ requirements-parser
 # Pin pytest due to VS Code issue
 pytest==5.0.1
 pytest-cov
-tempeh==0.1.6
+tempeh==0.1.7
```
riedgar-ms: Should we start going with '>=' here?

romanlutz: No, because tempeh is in no way backwards compatible and can break the tests here anytime.

PR #148, opened 2019-11-04
New notebook with Admissions example

riedgar-ms:
A new notebook for `GridSearch` based on an Admissions scenario
Uses synthetic data, with the sensitive feature (gender) held out from the training data


riedgar-ms:
Will do the Law School example instead, since it has real data

PR #151, opened 2019-11-04
Exponentiated Gradient documentation updates

romanlutz:

Just some improvements in terms of documentation, and a few methods are prefixed with underscore if they're just helpers within a class.

### DIFF
--- a/fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
+++ b/fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
@@ -13,15 +13,34 @@ logger = logging.getLogger(__name__)
 
 
 class _Lagrangian:
-    """ Operations related to the Lagrangian"""
+    """ Operations related to the Lagrangian
 
-    def __init__(self, X, A, y, learner, constraints, eps, B, opt_lambda=True):
+    :param X: the training features
+    :type X: Array
+    :param sensitive_features: the sensitive features to use for constraints
+    :type sensitive_features: Array
+    :param y: the training labels
+    :type y: Array
+    :param estimator: the estimator to fit in every iteration of best_h
+    :type estimator: an estimator that has a `fit` method with arguments X, y, and sample_weight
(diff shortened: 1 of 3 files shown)


fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -35,53 +54,50 @@ def __init__(self, X, A, y, learner, constraints, eps, B, opt_lambda=True):
...
+
+        :return: tuple `(L, L_high)` where `L` is the value of the Lagrangian, and
+            `L_high` is the value of the Lagrangian under the best response of the lambda player
+        :rtype: tuple of two floats
         """
         lambda_projected = self.constraints.project_lambda(lambda_vec)
         if self.opt_lambda:
-            L = error + np.sum(lambda_projected * gamma) \
-                - self.eps * np.sum(lambda_projected)
+            L = error + np.sum(lambda_projected * gamma) - self.eps * np.sum(lambda_projected)
```
riedgar-ms: I take it we're allowing more text per line that when this originally formatted?

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -13,15 +13,34 @@
 
 
 class _Lagrangian:
```
riedgar-ms: Whole class is prefixed with underscore, so none of this will appear in main documentation ;-)

Still useful to start getting these described, though.

romanlutz: It's just not possible to understand this without documentation. Even with all of this it's not much better IMO

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
romanlutz: Right, and equations like this should never go across lines

PR #152, opened 2019-11-05
Postprocessing documentation updates

romanlutz:


fairlearn/postprocessing/_postprocessing.py
```diff
@@ -58,15 +69,27 @@ def predict(self, X, *, sensitive_features):
...
     def _pmf_predict(self, X, *, sensitive_features):
         """ Probabilistic mass function
+
+        :param X: Feature matrix
+        :type X: numpy.ndarray or pandas.DataFrame
+        :param sensitive_features: Sensitive features to identify groups by, currently allows
+            only a single column
+        :type sensitive_features: Currently 1D array as numpy.ndarray, list, pandas.DataFrame,
+            or pandas.Series
+        :return: array of tuples with probabilities for predicting 0 or 1, respectively
```
riedgar-ms: Wouldn't we have $p(0) = 1 - p(1)$? I know we're not really supporting `_pmf_predict` but that's not a very clear explanation of the return type.

fairlearn/postprocessing/_roc_curve_utilities.py
```diff
@@ -8,11 +8,21 @@
 from ._threshold_operation import ThresholdOperation
 
 
-def _get_roc(data, x_grid, attribute, flip=True):
+def _get_roc(data, sensitive_feature_value, flip=True):
```
riedgar-ms: This looks like more than a doc change - you're going from two arguments to one.

fairlearn/postprocessing/_postprocessing.py
romanlutz: That's correct. What would you suggest to make it clearer?

fairlearn/postprocessing/_roc_curve_utilities.py
romanlutz: x_grid wasn't used as it turns out!

fairlearn/postprocessing/_roc_curve_utilities.py
```diff
@@ -8,11 +8,21 @@
...
     """Get ROC curve's convex hull based on data columns 'score' and 'label'
     Scores represent output values from the predictor.
+
+    :param data: the DataFrame containing scores and labels
+    :type data: pandas.DataFrame
```
riedgar-ms: Side note, with an update to `conf.py` we should be able to have things like this link to the Pandas documentation (like we already can for `sklearn`). In part, the task I created for documentation consistency is about things like that.

romanlutz: what's conf.py?

fairlearn/postprocessing/_postprocessing.py
riedgar-ms: I'm not sure, since I don't know the actual return type? Isn't it an array of predictors with a label "chose this one with this probability?"

fairlearn/postprocessing/_roc_curve_utilities.py
riedgar-ms: The file which configures sphinx. It's in the docs directory. The above is more a note for the future, and needn't delay this PR.

fairlearn/postprocessing/_roc_curve_utilities.py
```diff
@@ -88,11 +117,25 @@ def _interpolate_curve(data, x_col, y_col, content_col, x_grid):
...
+        weights; if False does not allow flipping; default True
+    :type flip: bool
+    :return: the ROC curve points with their corresponding threshold operations
+    :rtype: pandas.DataFrame
+    """
     scores, labels, n, n_positive, n_negative = _get_scores_labels_and_counts(data)
 
     if n_positive == 0 or n_negative == 0:
-        raise ValueError("Degenerate labels for attribute value {}".format(attribute))
+        raise ValueError("Degenerate labels for sensitive feature value {}" \
```
riedgar-ms: Should this string be a constant up top?

romanlutz: added a test case for this case too which was missing so far :O

fairlearn/postprocessing/_roc_curve_utilities.py
romanlutz: OK great!

fairlearn/postprocessing/_postprocessing.py
romanlutz: It's an array of tuples where the first entry is the prob for 0 and the second one is the prob for 1

riedgar-ms: OK. I suppose those have to sum to 1?

romanlutz: Right. I can add that

romanlutz: done

PR #156, opened 2019-11-11
add regression metrics

rihorn2:
Add the metrics for regression and probit in the agreed order. Awaiting text descriptions, but the code is not going to change


fairlearn/metrics/_skm_wrappers.py
```diff
@@ -2,7 +2,7 @@
 # Licensed under the MIT License.
 
 import sklearn.metrics as skm
-
+import math
```
riedgar-ms: from math import sqrt
?
Just a thought

fairlearn/widget/fairlearnDashboard.py
```diff
@@ -87,29 +88,37 @@ def __init__(
...
-                "function": group_mean_squared_log_error
-            },
             "median_absolute_error": {
-                "model_type": ["regression"],
-                "function": group_median_absolute_error
-            },
-            "balanced_root_mean_squared_error": {
                 "model_type": [],
-                "function": group_balanced_root_mean_squared_error
+                "function": group_median_absolute_error
```
riedgar-ms: This doesn't match the name above

fairlearn/metrics/_skm_wrappers.py
```diff
@@ -135,3 +135,40 @@ def internal_mse_wrapper(y_true, y_pred, sample_weight=None):
 
     return metric_by_group(internal_mse_wrapper,
                            y_true, y_pred, group_membership, sample_weight=sample_weight)
+
+def group_root_mean_squared_error(y_true, y_pred, group_membership, *,
```
riedgar-ms: These should be going in to `test_group_sklearn_wrappers.py`

At the very list, they should be added to the appropriate "supported_metrics" things-which-aren't-dictionaries-like-the-comment-says at the top. Since they also take that `multioutput` argument, could you also add a test setting a different value of that, so that we have a test confirming that the argument is passed through OK? e.g. `test_group_mean_squared_error_multioutput()`

rihorn2: Thank you for showing me the tests. Done.

test/unit/metrics/test_group_sklearn_wrappers.py
```diff
@@ -36,7 +36,8 @@
 supported_metrics_weighted_binary = [(skm.precision_score, metrics.group_precision_score),
                                      (skm.recall_score, metrics.group_recall_score),
                                      (skm.roc_auc_score, metrics.group_roc_auc_score),
-                                     (skm.mean_squared_error, metrics.group_mean_squared_error)]
+                                     (skm.mean_squared_error, metrics.group_mean_squared_error),
+                                     (skm.r2_score, metrics.group_r2_score)]
```
riedgar-ms: I have a feeling that 'binary' isn't a good name for this variable...

Also, what about `group_root_mean_squared_error` ?

PR #157, opened 2019-11-11
Binning of numeric sensitive features

rihorn2:
A simple control for numeric feature binning.
Allow for treating integer valued features as categorical or numeric values
Treat empty bins as not impacting disparity calculation. Present as NaN.


fairlearn/widget/fairlearnDashboard.py
```diff
@@ -25,7 +25,7 @@ class FairlearnDashboard(object):
     def __init__(
             self, *,
             sensitive_features,
-            true_y, predicted_ys,
+            y_true, predicted_ys,
```
riedgar-ms: Should that be `ys_pred` as well?

riedgar-ms:
When this goes in, please be sure to squash it.

fairlearn/widget/fairlearnDashboard.py
rihorn2: I asked Miro for his opinion, going with y_pred singular, and also accepting a dictionary as input, mapping it to a 2D array and an matching 1D array of model names

PR #160, opened 2019-11-15
Initial Pipeline for PyPI releases

riedgar-ms:
An initial attempt at a pipeline for performing PyPI releases. The pipeline does the following
- Prevalidation
- For PyPI-Test
  - Creates the wheels
  - Uploads to PyPI-Test
  - Fetches the package using `pip install` and runs our tests
- Repeats the above for PyPI


bethz:
Suggest to consider having one pipeline for PyTest and a separate pipeline for the final pypi release to give you more control on when to do the final release.

riedgar-ms:
> Suggest to consider having one pipeline for PyTest and a separate pipeline for the final pypi release to give you more control on when to do the final release.

How do you mean? A signoff is required before each upload to PyPI. Having a separate pipeline means a risk of different commits being deployed to PyPI-Test and PyPI itself.

devops/pypi-release.yml
```diff
@@ -0,0 +1,73 @@
...
+      pyVersions: [3.6, 3.7] 
+
+  - template: notebook-job-template.yml
+    parameters:
+      pyVersions: [3.6]
+
+
+# =================================================================================================================
+
+# We rely on the implicit dependence on the previous stage
```
bethz: Suggest using a stage with explicit dependence on success of prior stage.

riedgar-ms: I had thought about that. The issue is that since I generate things out of templates, the names for the `dependsOn` clauses would have to be duplicated as well. I felt it was better to leave everything as implicitly dependent rather than duplicate the name generation logic. I'm persuadable on that point, though.

PR #174, opened 2019-11-22
Include yarn build in deployment pipeline

riedgar-ms:
Update the deployment pipeline so that it runs `yarn` to build the widget code

This enables the removal of some automatically generated files from our repository


devops/testing-build-package.yml
```diff
@@ -0,0 +1,17 @@
+variables:
```
riedgar-ms: This is only so I have a quick means of testing. Not for checkin

fairlearn/__init__.py
```diff
@@ -11,7 +11,7 @@
 # Finesse the version
 _FAIRLEARN_DEV_VERSION_ENV_VAR = "FAIRLEARN_DEV_VERSION"
 
-_base_version = "0.4.0a1"
+_base_version = "0.4.0a1.post1"
```
riedgar-ms: Just to stop version collisions. Not for checkin

riedgar-ms:
Any other files to delete?

PR #175, opened 2019-11-22
update logging to use FileHandler instead of basicConfig

imatiach-msft:


bethz:
@imatiach-msft  I made a change in the devops ui so that forks in PRs auto-trigger the status check. Can you amend your github checkin to see if the trigger occurs? thanks.

PR #177, opened 2019-11-25
Update ReadMe with Yarn instructions

riedgar-ms:
Now that the dashboard tarball is no longer checked in, provide instructions on creating it in a cloned repo

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,15 @@ or you can clone the repository locally via
 ```python
 git clone git@github.com:fairlearn/fairlearn.git
 ```
+If you clone from git and wish to use the Fairness dashboard, then you will need to [install Yarn](https://yarnpkg.com/lang/en/docs/install), and then do the following:
+```bash
+cd fairlearn/widget/js
+yarn install
+yarn build:all
+rm -rf dist
+rm -rf lib
+rm -rf node_modules
+```
 
 To verify that the cloned repository works (the pip package does not include the tests), run
 
--- a/README.md
+++ b/README.md
@@ -62,7 +62,7 @@ or you can clone the repository locally via
 ```python
 git clone git@github.com:fairlearn/fairlearn.git
 ```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -62,6 +62,15 @@ or you can clone the repository locally via
 ```python
 git clone git@github.com:fairlearn/fairlearn.git
 ```
+If you clone from git and wish to use the Fairness dashboard, then you will need to [install Yarn](https://yarnpkg.com/lang/en/docs/install), and then do the following:
```
rihorn2: Maybe make clear that this is for using a local install, that you can always clone and tehn just pip install fairlearn without this? Unless that's too pedantic.

README.md
```diff
@@ -62,6 +62,15 @@ or you can clone the repository locally via
...
+```bash
+cd fairlearn/widget/js
+yarn install
+yarn build:all
+rm -rf dist
```
rihorn2: I can write a yarn task to do these removes. These are OS specific, no?

README.md
riedgar-ms: If people cloned and did the pip install, which would actually run? Or did you mean do a `pip install -e .` ?

README.md
riedgar-ms: I believe that Powershell has the appropriate aliases to make them work like they do in *nix. However, if it could happen automatically in `build:all`, that would be better.

README.md
riedgar-ms: I think we agreed that this was clear enough. It's likely that only developers would be doing the clone.

README.md
riedgar-ms: As discussed, it's probably safer to just leave these commands there. If the clean ups don't work as-is on another platform, a developer should be able to adapt.

However, per your other suggestion, I have made it plain that the yarn build doesn't have to be done very often

PR #178, opened 2019-11-26
Standardise input convertors for test

riedgar-ms:
Create and use a standard set of convertors for use with our 'argument type' tests. This has required adding several `__init__.py` files to the `test` directory to enable `pytest` to find the common code.

Also add an 'argument type' test to `ExponentiatedGradient`

### DIFF
--- a/test/unit/reductions/exponentiated_gradient/test_expgrad_arguments.py
+++ b/test/unit/reductions/exponentiated_gradient/test_expgrad_arguments.py
@@ -1,0 +1,60 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import pandas as pd
+import pytest
+
+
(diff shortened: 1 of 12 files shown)


test/unit/reductions/exponentiated_gradient/test_expgrad_arguments.py
```diff
@@ -0,0 +1,60 @@
...
+
+# Ways of transforming the data
+candidate_X_transforms = [ensure_ndarray, ensure_dataframe]
+candidate_Y_transforms = conversions_for_1d
+candidate_A_transforms = conversions_for_1d
+
+# ================================================================
+
+
+class TestExpGradArguments:
```
romanlutz: nit: we should probably try not to use "expgrad" since the method is fully spelled out elsewhere, just for consistency and to avoid confusing users.

PR #180, opened 2019-11-27
perf test through Azure ML

romanlutz:
This PR introduces performance tests run through Azure ML in order to be able to track the runtime over time. Any kind of storage would be sufficient, but we might as well take advantage of Azure ML's capabilities for visualizing this.

still missing:
- adjustment of devops configurations
- simplifications to the script generation
- I would like to move some of the code to tempeh if possible, but need to look into that a bit more.
- getting the secrets (SP password mostly), ideally from a keyvault (?), but this requires a subscription
- more elaborate assertions in addition to just total time (potentially in a future PR)


devops/perf-tests-job-template.yml
```diff
@@ -25,8 +25,19 @@ jobs:
...
+  - script: pip install azureml-sdk
+    displayName: 'Install azureml-sdk'
+
   - script: python -m pytest -s test/perf --junitxml=./TEST--TEST.xml -o junit_suite_name="$(Agent.JobName)"
     displayName: 'Run Perf tests'
+    env:
+      SERVICE_PRINCIPAL_PASSWORD: $(fairlearnautomationspsecret)
+      SERVICE_PRINCIPAL_ID: $(fairlearnautomationspid)
+      TENANT_ID: $(fairlearnautomationsptenantid)
+      SUBSCRIPTION_ID: cecafb73-04ae-4432-9f96-d96925d28058
```
riedgar-ms: Can the subscription/rg/ws etc. be made into template parameters?

devops/nightly-perf.yml
```diff
@@ -4,6 +4,9 @@ trigger: none # No CI build
 
 pr: none # Not for pull requests
 
+variables:
+- group: "Key Vault"
```
riedgar-ms: I would be interested in how this is working. It would not be hard to concoct something better than what I did for the releases, and we should standardise

requirements-fixed.txt
```diff
@@ -19,7 +19,7 @@ requirements-parser
 # Pin pytest due to VS Code issue
 pytest==5.0.1
 pytest-cov
-tempeh==0.1.8
```
riedgar-ms: If you resync with master, this file is gone

test/perf/environment_setup.py
```diff
@@ -0,0 +1,43 @@
...
+    shutil.rmtree("build", True)
+    print('removing fairlearn.egg-info')
+    shutil.rmtree("fairlearn.egg-info", True)
+    print('removing dist directory')
+    shutil.rmtree("dist", True)
+
+    widget_path = os.path.join(os.getcwd(), 'fairlearn', 'widget', 'js')
+    os.chdir(widget_path)
+    try:
+        print('running yarn install')
```
riedgar-ms: Would it be better to convert `build-widget.ps1` so that we don't have the same code in two languages?

test/perf/environment_setup.py
```diff
@@ -0,0 +1,43 @@
...
+        subprocess.Popen(["yarn", "install"], shell=True).wait()
+        print('removing yarn build:all')
+        subprocess.Popen(["yarn", "build:all"], shell=True).wait()
+    except Exception as ex:
+        print(ex.message)
+    finally:
+        os.chdir(os.path.join('..', '..', '..'))
```
riedgar-ms: Rather than that rather unpleasant path, could you push/pop?

test/perf/script_generation.py
```diff
@@ -0,0 +1,98 @@
...
+from fairlearn.reductions import ExponentiatedGradient, GridSearch
+
+
+def generate_script(request, perf_test_configuration, script_name, script_directory):
+    if not os.path.exists(script_directory):
+        os.makedirs(script_directory)
+
+    script_lines = []
+    # imports
+    script_lines.append('from time import time')
```
riedgar-ms: Would it be easier to have a `script.template_py` and do a search/replace on that (even if some of the replaces a multiple lines)?

devops/nightly-perf.yml
romanlutz: Sure, let's discuss in person!

devops/perf-tests-job-template.yml
romanlutz: They sure can!

devops/nightly-perf.yml
romanlutz: Outcome from discussion: I will change this to using the keyvault task, that way everything is in the YAML as opposed to obscure ADO definitions of things.

test/perf/script_generation.py
romanlutz: I'm hoping to eventually get to a point where this looks nicer, but I'm scared of "replace". It reads well to have it in order at least. Using replace might mean that we make just a few replacements and nobody understands the logic anymore. I think the next step will have something like:
```
add_imports(script_lines)
add_dataset(script_lines)
add_estimator(script_lines)
add_mitigator(script_lines)
add_analysis(script_lines)

romanlutz: I should have added that these will in part come from actual files, so we don't need to write it like this, but there might be a file with all the imports and we can just read it at test time.

romanlutz: not doing more than what I did now in this PR though

test/perf/environment_setup.py
romanlutz: Right, there was also another issue there that prevented this from working in ADO. I decided that I don't need the widget in perf tests as of right now and removed the code that does this. So there's no duplication right now, but we should think about how to do this. Maybe build-widget.ps1 can just be a pre-step so we don't need to build the wheel here?

PR #185, opened 2019-12-02
Build the widget

riedgar-ms:
Add a job template which builds the widget to the PR-Gate, Nightly and Nightly-Fixed builds. Note that this does not run any tests, but just ensures that the widget builds successfully


riedgar-ms:
This is just a build - the minimal sanity check. Yes, we will want to do more testing along the lines you mention in the future. For example, right now the notebook tests are passing, even though those which reference the widget should probably fail.

PR #186, opened 2019-12-02
Markdown updates based on doc bash

romanlutz:
Mostly small changes based on Richard's feedback from the doc bash. The main change is using LaTeX for the formulas. To view it you may have to switch to that branch and open the terminology file, see: https://github.com/fairlearn/fairlearn/blob/rolutz/markdown_updates/TERMINOLOGY.md


CONTRIBUTING.md
```diff
@@ -46,7 +48,7 @@ All of the currently supported parity constraints (subclasses of `Moment`) are b
 
 #### Post-processing algorithms
 
-The constructors of post-processing algorithms require an already trained predictor as an argument. As an alternative, it is possible to provide an estimator, and the estimator is then fitted on the data at the beginning of the execution of `fit`. For post-processing algorithms, the `constraints` argument is provided as a string.
+The constructors of post-processing algorithms require an already trained predictor as an argument or, alternatively, an estimator that is subsequently fitted on the data when executing `fit`. For post-processing algorithms, the `constraints` argument is provided as a string.
```
riedgar-ms: How about "algorithms require either a trained predictor or an estimator (which is trained on the data by executing :code:`fit`" ?

TERMINOLOGY.md
```diff
@@ -11,17 +11,17 @@ The fairlearn package largely follows the [terminology established by scikit-lea
...
+AI systems can behave unfairly for a variety of reasons. Sometimes societal biases are reflected in the training data and in the decisions made during the development and deployment of these systems. In other cases, AI systems behave unfairly because of characteristics of the data (e.g., too few data points about some group of people) or characteristics of the systems themselves. It can be hard to distinguish between these reasons, especially since they are not mutually exclusive and often exacerbate one another. Therefore we define whether an AI system is behaving unfairly in terms of its impact on people – i.e., in terms of harms – and not in terms of specific causes, such as societal biases, or in terms of intent, such as prejudice.
```
riedgar-ms: I don't think there's a comma after "e.g."

CONTRIBUTING.md
romanlutz: I took your suggestion with a small change ("when" isntead of "by")

CONTRIBUTING.md
```diff
@@ -18,6 +18,8 @@ Pull requests against `master` trigger automated tests that are run through Azur
 ### Investigating automated test failures
 For every pull request to `master` with automated tests, you can check the logs of the tests to find the root cause of failures. Our tests currently run through Azure Pipelines with steps for setup, testing, and teardown. The `Checks` tab of a pull request contains a link to the [Azure Pipelines page](dev.azure.com/responsibleai/fairlearn/_build/results), where you can review the logs by clicking on a specific step in the automated test sequence. If you encounter problems with this workflow, please reach out through [GitHub issues](https://github.com/fairlearn/fairlearn/issues).
 
+To run the same tests locally find the corresponding pipeline definition (a `yml` file) in the `devops` directory. It either directly contains the command to execute the tests (usually starting with `python -m pytest`) or it refers to a template file with the command.
+
 ## API conventions
```
MiroDudik: comma after "locally"

TERMINOLOGY.md
```diff
@@ -11,17 +11,17 @@ The fairlearn package largely follows the [terminology established by scikit-lea
...
+AI systems can behave unfairly for a variety of reasons. Sometimes societal biases are reflected in the training data and in the decisions made during the development and deployment of these systems. In other cases, AI systems behave unfairly because of characteristics of the data (e.g. too few data points about some group of people) or characteristics of the systems themselves. It can be hard to distinguish between these reasons, especially since they are not mutually exclusive and often exacerbate one another. Therefore we define whether an AI system is behaving unfairly in terms of its impact on people – i.e. in terms of harms – and not in terms of specific causes, such as societal biases, or in terms of intent, such as prejudice.
```
MiroDudik: Let's revert the edits at the beginning of the paragraph, because they subtly change the meaning away from the intended--the goal was to highlight that unfairness may arise due to societal biases or in their absence--that's lost right now.

I like the rewrite that starts with "It can be hard" but a few issues there need correcting. Please use the following:

It can be hard to distinguish between these reasons, especially since they are not mutually exclusive and often exacerbate one another. Therefore, we define whether an AI system is behaving unfairly in terms of its impact on people&mdash;i.e., in terms of harms&mdash;and not in terms of specific causes, such as societal biases, or in terms of intent, such as prejudice.

As a convention, we should write "i.e.," and "e.g.," with the comma:
https://www.chicagomanualofstyle.org/qanda/data/faq/topics/Abbreviations/faq0047.html

TERMINOLOGY.md
```diff
@@ -11,17 +11,17 @@ The fairlearn package largely follows the [terminology established by scikit-lea
...
 
 ### Types of harms
 
 There are many types of harms (see, e.g., the [keynote by K. Crawford at NeurIPS 2017](https://www.youtube.com/watch?v=fMym_BKWQzk)). The fairlearn package is most applicable to two kinds of harms:
 
-- _Allocation harms_. These harms can occur when AI systems extend or withhold opportunities, resources, or information. Some of the key applications are in hiring, school admissions, and lending.
+- _Allocation harms_ can occur when AI systems extend or withhold opportunities, resources, or information. Some of the key applications are in hiring, school admissions, and lending.
 
-- _Quality-of-service harms_. Quality of service refers to whether a system works as well for one person as it does for another, even if no opportunities, resources, or information are extended or withheld.
+- _Quality-of-service harms_ can occur when a system does not work as well for one person as it does for another, even if no opportunities, resources, or information are extended or withheld. Examples include varying accuracy in face recognition or hate speech detection.
```
MiroDudik: Let's replace by some less hot-button applications. How about "Examples include varying accuracy in face recognition, document search, or product recommendation."

TERMINOLOGY.md
romanlutz: Interesting, this is in direct conflict with Richard's feedback :-D

riedgar-ms: I won't argue with the style manual.

PR #187, opened 2019-12-03
Add more flake8 checks

riedgar-ms:
Add a number of extra flake8 checks:
- flake8-blind-except
- flake8-builtins
- flake8-docstrings
- flake8-logging-format
- flake8-rst-docstrings

Since these create a huge number of issues, suppress a lot of these for now in `setup.cfg` (plus a handful of special cases done inline). Put in fixes for the simpler complaints, such as:
- Separate summaries in docstrings
- Spacing within and around docstrings
- Deferring string interpolation in `logging` calls

### DIFF
--- a/fairlearn/metrics/_balanced_root_mean_squared_error.py
+++ b/fairlearn/metrics/_balanced_root_mean_squared_error.py
@@ -16,7 +16,7 @@ def balanced_root_mean_squared_error(y_true, y_pred, sample_weight=None):
     Used for binary logistic regression, this computes the error as
 
     .. math::
-       \frac{RMSE(Y=0) + RMSE(Y=1)}{2}
+       \frac{\text{RMSE}(Y=0) + \text{RMSE}(Y=1)}{2}
 
     The classes are constrained to be :math:`\in {0, 1}`. The :code:`y_true` values must
     always be one of these, while :code:`y_pred` can be a continuous probability
--- a/fairlearn/metrics/_balanced_root_mean_squared_error.py
+++ b/fairlearn/metrics/_balanced_root_mean_squared_error.py
@@ -11,17 +11,20 @@ _Y_TRUE_NOT_0_1 = "Only 0 and 1 are allowed in y_true and both must be present"
 
 
 def balanced_root_mean_squared_error(y_true, y_pred, sample_weight=None):
-    """
+    r"""The mean of the root mean squared error (RMSE) for the positive and negative cases.
+
     Used for binary logistic regression, this computes the error as
-    [RMSE(Y=0) + RMSE(Y=1)]/2
-    The classes are constrained to be {0, 1}. The `y_true` values must
-    always be one of these, while `y_pred` can be a continuous probability
+
+    .. math::
+       \frac{RMSE(Y=0) + RMSE(Y=1)}{2}
+
+    The classes are constrained to be :math:`\in {0, 1}`. The :code:`y_true` values must
+    always be one of these, while :code:`y_pred` can be a continuous probability
     (which could be thresholded to get a predicted class).
 
     Internally, this builds on the
     :any:`sklearn.metrics.mean_squared_error` routine.
     """
-
     y_ta = _convert_to_ndarray_and_squeeze(y_true)
     y_pa = _convert_to_ndarray_and_squeeze(y_pred)
     s_w = np.ones(len(y_ta))
--- a/fairlearn/postprocessing/_threshold_optimizer.py
+++ b/fairlearn/postprocessing/_threshold_optimizer.py
@@ -3,9 +3,9 @@
 
 """Threshold Optimization Post Processing algorithm.
 
-This is based on M. Hardt,
-E. Price, N. Srebro's paper "`Equality of Opportunity in Supervised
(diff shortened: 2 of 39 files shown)


fairlearn/metrics/_balanced_root_mean_squared_error.py
```diff
@@ -11,17 +11,20 @@
 
 
 def balanced_root_mean_squared_error(y_true, y_pred, sample_weight=None):
-    """
+    r"""The mean of the root mean squared error (RMSE) for the positive and negative cases.
```
romanlutz: what does this r do?

riedgar-ms: It allows you to have single `\` characters in the \LaTeX blocks rather than having to put `\\` throughout. I only found out from `flake8` itself complaining :-)

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -1,7 +1,9 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
-""" Threshold Optimization Post Processing algorithm based on M. Hardt,
+"""Threshold Optimization Post Processing algorithm.
+
+This is based on M. Hardt,
```
romanlutz: this line break is odd

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -267,7 +275,7 @@ def _threshold_optimization_demographic_parity(sensitive_features, labels, score
             selection_error_curve[sensitive_feature_value]['error']
 
         logger.debug(OUTPUT_SEPARATOR)
-        logger.debug("Processing " + str(sensitive_feature_value))
+        logger.debug("Processing {0}", str(sensitive_feature_value))
```
romanlutz: why suddenly {0} here? Above it looks different, just
logger.debug("some string ", value)

Which one is right?

romanlutz: if str(...) is required then it's required above as well

bethz: for future, you can use {} so the 0 is not needed.

riedgar-ms: It's one where there was a complaint about not deferring the formatting of the log string. The basic idea being that it's best to wait until you're sure that a log message is going to be written to interpolate the string. I do recognise that the `str(..)` call does go somewhat against that....

fairlearn/postprocessing/_threshold_optimizer.py
riedgar-ms: I'll fix.

fairlearn/widget/fairlearnWidget.py
```diff
@@ -9,8 +9,7 @@
 
 @widgets.register
 class FairlearnWidget(widgets.DOMWidget):
-    """The python widget definition for the Fairlearn dashboard.
-    """
+    """The python widget definition for the Fairlearn dashboard."""
```
romanlutz: lowercase fairlearn!

setup.cfg
```diff
@@ -1,2 +1,15 @@
 [flake8]
-max-line-length = 99
\ No newline at end of file
+max-line-length = 99
```
bethz: the norm for max-line-length is usuall 79. why 99?

riedgar-ms: I believe that @romanlutz  is too young to have used F77, and I only did so under duress :-)

fairlearn/widget/fairlearnWidget.py
riedgar-ms: Thanks - fixed

fairlearn/metrics/_balanced_root_mean_squared_error.py
```diff
@@ -11,17 +11,20 @@
...
+
     Used for binary logistic regression, this computes the error as
-    [RMSE(Y=0) + RMSE(Y=1)]/2
-    The classes are constrained to be {0, 1}. The `y_true` values must
-    always be one of these, while `y_pred` can be a continuous probability
+
+    .. math::
+       \frac{RMSE(Y=0) + RMSE(Y=1)}{2}
```
romanlutz: if this is actual LaTeX then it should probably have \text{} around RMSE or it will look like the product of four single letter variables.

setup.cfg
romanlutz: @bethz if you only have a screen that accommodates 79 characters I'll be happy to talk to facilities for you :D 
It's simply annoying to have this restriction. Code becomes unreadable if you constantly have to wrap it around. I'm actually for upping this to 119

fairlearn/metrics/_balanced_root_mean_squared_error.py
riedgar-ms: Good point. Fixed

setup.cfg
riedgar-ms: I'm not sure I'd want 119 for everything, but allowing a few longer lines per source file (without having to put a special exception in) would be a nice feature for the line-length check.

PR #190, opened 2019-12-04
Pin scikit-learn in setup.py

riedgar-ms:
Had forgotten that this needed doing as well as the `requirements` files


riedgar-ms:
Looks like we're going to fix the notebook to work around the issue instead

PR #191, opened 2019-12-04
Fix for Law School Notebook

riedgar-ms:
Tweak the Law School notebook so that it works with the latest `scikit-learn`

This enables us to unpin the version of `scikit-learn` in our `requirements.txt` file

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@
 matplotlib>=3.0.3
 numpy>=1.17.2
 pandas>=0.25.1
-scikit-learn==0.21.3
+scikit-learn>=0.21.3
 scipy>=1.3.1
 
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -2,7 +2,7 @@
 matplotlib>=3.0.3
 numpy>=1.17.2
 pandas>=0.25.1
-scikit-learn==0.21.3
+scikit-learn>=0.21.3
 scipy>=1.3.1
```
MiroDudik: also requirements-fixed.txt? (I think that setup.py is fine though.)

riedgar-ms: The point of `requirements-fixed.txt` is that versions are pinned.

PR #192, opened 2019-12-04
Reorganise documentation

riedgar-ms:
Reorganising how the documentation is presented, since the default style from `sphinx-apidoc` assumed we had lots of individual modules rather than larger packages

The effects can be seen at:
https://fairlearn.readthedocs.io/en/riedgar-ms-doc-reorg/index.html


fairlearn/widget/__init__.py
```diff
@@ -1,9 +1,9 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
-"""Module for Fairlearn Dashboard widget."""
```
bethz: should Fairlearn be fairlearn?

riedgar-ms: I'm seeing that it is lower case?

PR #204, opened 2019-12-11
Use pip freeze during releases

riedgar-ms:
Adjusts the release pipeline so that `pip freeze` is called at the beginning. The freeze files generated are stored in Artifacts and used in subsequent steps.

This avoids the problem of another group releasing an update when we have a release in flight.


riedgar-ms:
However, as I noted on Teams, I'm somewhat ambivalent about this. It's quite a bit of complication to fix the problem I mentioned - and it's not like we put those third party packages into our wheel. It simply stops our release trains being halted because a just-released package from a third party has an issue.

riedgar-ms:
Not convinced this is actually useful

PR #210, opened 2019-12-18
Re-enable Linux 3.5

riedgar-ms:
Roman figured out a workaround for getting `shap` installed with Linux and Python 3.5. Put this into `fairlearn`

### DIFF
--- a/devops/pypi-release.yml
+++ b/devops/pypi-release.yml
@@ -37,7 +37,7 @@ stages:
     parameters:
       name: Linux
       vmImage: 'ubuntu-16.04'
-      pyVersions: [3.6,3.7]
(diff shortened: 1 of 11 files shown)


devops/pypi-release.yml
```diff
@@ -59,7 +59,7 @@ stages:
 
   - template: notebook-job-template.yml
     parameters:
-      pyVersions: [3.6]
+      pyVersions: [3.5, 3.6]
```
romanlutz: why no 3.7 here?

riedgar-ms: Can't remember offhand. But I also have a task for getting the NB tests running on more platforms.

PR #217, opened 2019-12-27
Basic unit tests for EqualizedOdds and DemographicParity moments

riedgar-ms:
Some very basic unit tests for the `EqualizedOdds` and `DemographicParity` moment classes. These are 'pinning' tests two establish the behaviour of these classes. The `gamma` method is not yet included in these tests, since that requires a trained model.

### DIFF
--- a/test/unit/reductions/moments/data_generator.py
+++ b/test/unit/reductions/moments/data_generator.py
@@ -19,6 +19,33 @@ def simple_binary_threshold_data(number_a0, number_a1,
     The uniformly distributed score is set in the 'example_feature'
     of the resultant X array; the sensitive feature is included in
     this array under the label 'example_sensitive_feature'
+
+    :param number_a0: Number of samples to generate with the label a0
+    :type number_a0: int
+
+    :param number_a1: Number of samples to generate with the label a1
+    :type number_a1: int
+
+    :param a0_threshold: Threshold value for samples with label a0 to
+        get a result of '1' in the Y array
+    :type a0_threshold: float
+
+    :param a1_threshold: Threshold value for samples with label a1 to
+        get a result of '1' in the Y array
+    :type a1_threshold: float
+
+    :param a0_label: The label value for the a0 class
+    :type a0_label: int, string
+
+    :param a1_label: The label value for the a1 class
+    :type a1_label: int, string
+
+    Returns X, Y, A:
+        X is the feature array (containing the uniformly distributed feature
+        and the sensitive feature labels)
+        Y is the classification array (either 0 or 1)
+        A is the array of sensitive features (which will duplicate a column
+        in the X array)
     """
     a0s = np.full(number_a0, a0_label)
     a1s = np.full(number_a1, a1_label)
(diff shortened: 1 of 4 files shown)


test/unit/reductions/moments/data_generator.py
```diff
@@ -0,0 +1,28 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import numpy as np
+import pandas as pd
+
+
+def simple_binary_threshold_data(number_a0, number_a1,
```
romanlutz: I didn't find this self-explanatory in case that was intended. 
Still not quite sure what "a" stands for. I'll have to look at how it's used.
Then further below it suddenly says credit score. That seems oddly specific for such a generally named function.

test/unit/reductions/moments/test_moments_demographic_parity.py
```diff
@@ -0,0 +1,171 @@
...
+from fairlearn.reductions._moments.moment import _EVENT, _GROUP_ID, _SIGN
+
+from .data_generator import simple_binary_threshold_data
+
+
+def test_construct_and_load():
+    dp = DemographicParity()
+    assert dp.short_name == "DemographicParity"
+
+    num_samples_a0 = 10
```
romanlutz: so "a" is sensitive feature?
I can only assume this confuses other people like it confused me. At least add a comment that that's what it means, or let's just name it num_samples_sensitive_feature0

test/unit/reductions/moments/test_moments_equalized_odds.py
```diff
@@ -0,0 +1,203 @@
...
+    lambda_a0_T = 4000 - 700
+    num_a0_F = int(a0_threshold * num_samples_a0)
+    num_a0_T = num_samples_a0 - num_a0_F
+
+    lambda_a1_F = 1000 - 100
+    lambda_a1_T = 5000 - 900
+    num_a1_F = int(a1_threshold * num_samples_a1)
+    num_a1_T = num_samples_a1 - num_a1_F
+
+    sw_a0_F = (lambda_a0_F + lambda_a1_F) / (1 - sum(Y) / len(Y)) - \
```
romanlutz: I've been called out for my glorious multi-line formulas, so I feel the duty to spread the wisdom :-)

test/unit/reductions/moments/test_moments_equalized_odds.py
```diff
@@ -0,0 +1,203 @@
...
+        lambda_a0_F * (num_samples / num_a0_F)
+    sw_a1_F = (lambda_a0_F + lambda_a1_F) / (1 - sum(Y) / len(Y)) - \
+        lambda_a1_F * (num_samples / num_a1_F)
+    sw_a0_T = (lambda_a0_T + lambda_a1_T) / (sum(Y) / len(Y)) - \
```
romanlutz: sw_a0_T is a wonderfully incomprehensible name :D

test/unit/reductions/moments/data_generator.py
riedgar-ms: Fair point. I've added some documentation

test/unit/reductions/moments/test_moments_demographic_parity.py
riedgar-ms: Documentation added

test/unit/reductions/moments/test_moments_equalized_odds.py
riedgar-ms: "signed weights for a0 label which have value True in the Y array" would blow through the limit on line length just for the variable name ;-)

PR #218, opened 2019-12-30
Refactor and document Moments object

riedgar-ms:
Refactor some of the `Moment` objects, creating some properties to replace fields. Add some documentation to go with them.


riedgar-ms:
Please see:
https://fairlearn.readthedocs.io/en/riedgar-ms-moments-refactor/fairlearn.reductions.html
For how the documentation now looks

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -16,12 +16,102 @@ class ConditionalSelectionRate(ClassificationMoment):
...
+    def default_objective_lambda_vec(self):
+        """Return the default objective lambda vector."""
+        return self._default_objective_lambda_vec
+
+    @default_objective_lambda_vec.setter
+    def default_objective_lambda_vec(self, value):
+        self._default_objective_lambda_vec = value
+
+    @property
+    def pos_basis(self):
```
riedgar-ms: I'd like to understand what this property and the following couple of properties are for. I can describe mechanistically what happens, but I'm not sure _why_.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -16,12 +16,102 @@ class ConditionalSelectionRate(ClassificationMoment):
...
 
     def load_data(self, X, y, event=None, **kwargs):
-        """Load the specified data into this object."""
+        """Load the specified data into this object.
+
+        This adds a column named `event` to the :meth:`Moment.tags`
+        property.
+        """
         super().load_data(X, y, **kwargs)
         self.tags[_EVENT] = event
```
riedgar-ms: We should probably consider promoting `_EVENT` and a couple of other constants to being public

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -67,7 +157,26 @@ def gamma(self, predictor):
...
+        make :math:`\lambda_{+}=2` and :math:`\lambda_{-}=0`. Similarly if
+        :math:`\lambda_{+}=2` and :math:`\lambda_{-}=3` then this method will
+        yield :math:`\lambda_{+}=0` and :math:`\lambda_{-}=1`.
+
+        The supplied :class:`pandas:pandas.DataFrame` must have the same
+        :class:`pandas:pandas.MultiIndex` as `index` and the resultant
+        :class:`pandas:pandas.DataFrame` will have that `index` as well.
+
+        :param lambda_vec: The set of Lagrange multipliers indexed by `index`
+        :type lambda_vec: :class:`pandas:pandas.DataFrame`
```
riedgar-ms: It's a little odd that (assuming I didn't mess up my testing) this method requires `lambda_vec` to be a DataFrame, while the next one requires a Series.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -67,7 +157,26 @@ def gamma(self, predictor):
...
+        constrained to be positive. The 'true' Lagrange multiplier is then
+        :math:`\lambda = \lambda_{+} - \lambda_{-}`. Since it makes little
+        sense to the positive and negative constraints violated at the same time,
+        this routine rebalances between them. For example, if we have
+        :math:`\lambda_{+}=3` and :math:`\lambda_{-}=1` then this method will
+        make :math:`\lambda_{+}=2` and :math:`\lambda_{-}=0`. Similarly if
+        :math:`\lambda_{+}=2` and :math:`\lambda_{-}=3` then this method will
+        yield :math:`\lambda_{+}=0` and :math:`\lambda_{-}=1`.
+
+        The supplied :class:`pandas:pandas.DataFrame` must have the same
```
riedgar-ms: Is this constraint on `index` correct? It would seem to make sense, since otherwise this method would never reference `self`. If so, I should add the check.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -78,7 +187,17 @@ def project_lambda(self, lambda_vec):
         return lambda_projected
 
     def signed_weights(self, lambda_vec):
-        """Return the signed weights."""
+        """Compute the signed weights.
+
+        Uses the equations for :math:`C_i^0` and :math:`C_i^1` as defined
+        in Section 3.2 of `Agarwal et al. (2018) <https://arxiv.org/abs/1803.02453>`_
```
riedgar-ms: Please tell me @MiroDudik  if I've got these references to the paper correct.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -78,7 +187,17 @@ def project_lambda(self, lambda_vec):
...
+        in the 'best response of the Q-player' subsection to compute the
+        signed weights to be applied to the data by the next call to the underlying
+        Estimator.
+
+        :param lambda_vec: The set of Lagrange multipliers indexed by :meth:`index`
```
riedgar-ms: Again, please confirm that I've got this constraint on `index` correct.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -16,12 +16,102 @@ class ConditionalSelectionRate(ClassificationMoment):
...
+
+    @neg_basis_present.setter
+    def neg_basis_present(self, value):
+        self._neg_basis_present = value
+
     def default_objective(self):
         """Return the default objective for moments of this kind."""
         return ErrorRate()
 
     def load_data(self, X, y, event=None, **kwargs):
```
riedgar-ms: I've not spotted anything which actually _requires_ `y` to be binary... did I miss something? Or if it has to be binary, then we should add a check in this method

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -100,6 +219,17 @@ class DemographicParity(ConditionalSelectionRate):
...
+
+    This implementation of :class:`ConditionalSelectionRate` defines
+    a single event, `all`. Consequently, the `prob_event`
+    :class:`pandas:pandas.DataFrame`
+    will only have a single entry, which will be equal to 1.
+    Similarly, the `index` property will have twice as many entries
+    (corresponding to the positive and negative Lagrange multipliers)
+    as there are unique values for the sensitive feature.
+    The :meth:`signed_weights` method will compute the costs according
+    to Example 3 of
```
riedgar-ms: Please confirm this is the correct reference @MiroDudik

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -116,6 +246,22 @@ class EqualizedOdds(ConditionalSelectionRate):
...
+    fraction of the samples corresponding to each unique value in
+    the `Y` array.
+
+    The `index` MultiIndex will have a number of entries equal to
+    the number of unique values for the sensitive feature, multiplied by
+    the number of unique values of the `Y` array, multiplied by two (for
+    the positive and negative Lagrange multipliers).
+
+    With these definitions, the :meth:`signed_weights` method
+    will calculate the costs according to Example 4 of
```
riedgar-ms: Another reference to be double-checked.

fairlearn/reductions/_moments/moment.py
```diff
@@ -33,20 +33,61 @@ def load_data(self, X, y, **kwargs):
...
+    def X(self):
+        """Return the array of features."""
+        return self._X
+
+    @X.setter
+    def X(self, value):
+        self._X = value
+
+    @property
+    def n(self):
```
romanlutz: maybe this is a good time to rename this from `n` to something more descriptive like `n_samples`

riedgar-ms: Agreed and done.

fairlearn/reductions/_moments/conditional_selection_rate.py
MiroDudik: all of the moments work for categorical y

MiroDudik: i'll need to double check. while there might be no runtime errors generated for non-binary y, the correctness might require binary y.

fairlearn/reductions/_moments/conditional_selection_rate.py
MiroDudik: this explanation is too operational. the right way is to go from the moment-level definition of project_lambda.

riedgar-ms: I've simplified the text somewhat, although I'm not quite sure what you mean by the "moment-level definition of `project_lambda`. I've had a look at the OneNote, and this `project_lambda` doesn't return `argmax` of anything - although it's only used by `_eval_from_error_gamma` in the `_Lagrangian` and `max()` values are taken there. Also, this rebalancing between the positive and negative multipliers is going to be necessary to discover the true largest $\lambda$

riedgar-ms:
Have made alternative changeset with fewer changes.

PR #219, opened 2019-12-31
Add more time metrics to performance tests

romanlutz:
Specifically, this PR adds 
- mitigation time (explicitly named as such compared to total time before)
- estimator fit time (without mitigation)
- overhead from mitigation as absolute number and relative to estimator fit


test/perf/script_generation.py
```diff
@@ -68,43 +73,50 @@ def add_dataset_setup(script_lines, perf_test_configuration):
...
+            script_lines.append('mitigator.predict('
+                                'X_test, '
+                                'sensitive_features=sensitive_features_test, '
+                                'random_state=1)')
+        else:
+            script_lines.append('predictions = mitigator.predict(X_test)')
+
+
+def add_additional_metric_calculation(script_lines):
+    # We need to know how much overhead fairlearn adds on top of the basic estimator fit.
```
riedgar-ms: Do we record how many times `estimator.fit()` is called? Because that's part of the measure too.

romanlutz: expgrad does that actually! It's called n_oracle_calls and saved in the expgrad result. I can log that! Thanks for the suggestion.

riedgar-ms: …. and record the time for each call (you can't assume that each is the same as the unmitigated estimator)

fairlearn/reductions/_exponentiated_gradient/_exponentiated_gradient_result.py
```diff
@@ -60,6 +60,11 @@ def n_oracle_calls(self):
         """Return the number of times the estimator was called."""
         return self._n_oracle_calls
 
+    @property
+    def oracle_calls_execution_time(self):
+        """Return the number of times the estimator was called."""
```
riedgar-ms: I'm confused as to what this is. The name implies it's the time spent in the oracle (aka estimator.fit()) and will be a `TimeSpan`. However the the documentation is implying that it's an integer count of the number of occasions on which the oracle was called. I suspect the former is correct, but this should be cleared up.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -6,6 +6,7 @@
 import pandas as pd
 import pickle
 import scipy.optimize as opt
+from time import time
```
riedgar-ms: Rather than changing these methods, would it be better to wrap the estimators themselves? I can see arguments either way.

fairlearn/reductions/_exponentiated_gradient/_exponentiated_gradient_result.py
romanlutz: Oh copy-paste...

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
romanlutz: Not entirely sure what you mean here

riedgar-ms: Basically, do we think that the profiling facility should be added to `fairlearn` itself, or should the performance framework wrap the supplied estimators and do the profiling that way. You'd have something like:
```python
class PerformanceEstimator:
   def __init__(estimator):
      self.real_estimator = estimator
      self.n_fit_calls = 0
      self.n_fit_call_elapsed = TimeSpan()


    def fit(…..):
       self.n_fit_calls++
      Stopwatch.start()
      self.real_estimator.fit()
      Stopwatch.stop()
      self.n_fit_call_elapsed += Stopwatch.elapsed
```
Like I said, there are arguments either way; I'm just noting the alternative.

romanlutz: Got it!
I think this may be something users are interested in tracking as well, and it's not really a lot of overhead to keep a few floats around.
What's the benefit of Stopwatch over the time package?

riedgar-ms: `Stopwatch` was pseudo-code :-)

As for what accumulates the profiling information.... arguments both ways as I said. If we are going to put it into the shipped `fairlearn` code, can I suggest creating a `ProfileInformation` (or similar class), and returning that as a property on the Result object. That way, we don't have to worry about steadily adding more fields in multiple places.

romanlutz: The complication there is that even expgrad and grid search aren't consistent since expgrad just has a series of calls to fit while grid search has separate result objects. The takeaway from our in-person discussion was that we can always adjust this later if we feel strongly either way. This is not exposed in any way that would make users think they can rely on it existing since we used underscores.

PR #223, opened 2020-01-06
Convert Notebook tests to papermill

riedgar-ms:
Rather than using `nbval`, convert our notebook tests to use `papermill`. With the help of `nteract-scrapbook` we can then examine the contents of particular variables from the notebooks to ensure that we're getting the expected results.

Explicit `scrapbook` commands are required to save out values for future examination, but we don't want to include these when our users look at the notebooks. Accordingly, we include machinery for adding the necessary cells to the notebooks dynamically.

### DIFF
--- a/devops/all-tests-job-template.yml
+++ b/devops/all-tests-job-template.yml
@@ -44,7 +44,7 @@ jobs:
   - script: flake8 .
     displayName: "Run flake8"
 
-  - script: python -m pytest test/ --ignore=test/perf --junitxml=./TEST--TEST.xml -o junit_suite_name="$(Agent.JobName)"
+  - script: python -m pytest test/unit --junitxml=./TEST--TEST.xml -o junit_suite_name="$(Agent.JobName)"
     displayName: 'Run tests'
 
   - task: PublishTestResults@2
--- a/devops/all-tests-job-template.yml
+++ b/devops/all-tests-job-template.yml
@@ -44,7 +44,7 @@ jobs:
   - script: flake8 .
     displayName: "Run flake8"
 
-  - script: python -m pytest test/unit --junitxml=./TEST--TEST.xml -o junit_suite_name="$(Agent.JobName)"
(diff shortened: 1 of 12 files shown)


riedgar-ms:
I'll wait for comments before trying to convert the rest of the notebooks

devops/notebook-job-template.yml
```diff
@@ -35,7 +35,7 @@ jobs:
   - script: pip install -e .
     displayName: 'Install fairlearn in edit mode'
 
-  - script: python -m pytest notebooks/ --nbval --junitxml=./TEST-TEST.xml -o junit_suite_name="$(Agent.JobName)"
+  - script: python -m pytest test/notebooks/ --junitxml=./TEST-TEST.xml -o junit_suite_name="$(Agent.JobName)"
```
riedgar-ms: Also need to change the templates used in checking the PyPI releases

devops/all-tests-job-template.yml
```diff
@@ -44,7 +44,7 @@ jobs:
   - script: flake8 .
     displayName: "Run flake8"
 
-  - script: python -m pytest test/ --ignore=test/perf --junitxml=./TEST--TEST.xml -o junit_suite_name="$(Agent.JobName)"
+  - script: python -m pytest test/unit --junitxml=./TEST--TEST.xml -o junit_suite_name="$(Agent.JobName)"
```
romanlutz: why this change? Explicitly ignoring means we'd catch issues in new tests that are in neither directory, while this now would ignore test/integration or whatever one might add

riedgar-ms: Have changed, but we're getting to the point where these need to be regularised some more.

devops/notebook-job-template.yml
riedgar-ms: Done

bethz:
looks good!

PR #230, opened 2020-01-09
[WIP] split into fairlearn-core and fairlearn

romanlutz:
fairlearn-core contains everything other than the widget right now and provides extensions to add matplotlib or required test packages.

This raised an issue with requirements.txt: we could either have multiple such files and duplicate dependencies even further (they were already partially duplicate between setup.py and requirements.txt), but I decided to follow the same approach as interpret which is to include various extensions for test, devenv etc.
This also meant that we're splitting fairlearn-core into various extensions, so that we can add them back by saying
```
pip install fairlearn-core[postprocessingplot]
```
for example.

To be done (which is why it's WIP right now):
- doc build is not succeeding
- validate tests work


README.md
```diff
@@ -4,20 +4,20 @@
...
 
-- [Current release](#current-release)
-- [What we mean by _fairness_](#what-we-mean-by-fairness)
-- [Overview of fairlearn](#overview-of-fairlearn)
-- [Install fairlearn](#install-fairlearn)
-- [Usage](#usage)
-- [Contributing](#contributing)
-- [Maintainers](#maintainers)
-- [Issues](#issues)
+- [Current release](https://github.com/fairlearn/fairlearn/tree/v0.4.0#current-release)
```
riedgar-ms: Expanding these URLs will break the script I added to the release process.

devops/all-tests-job-template.yml
```diff
@@ -29,12 +38,9 @@ jobs:
...
   - template: python-infra-upgrade-steps-template.yml
 
-  - script: python scripts/requirements_pin.py --input requirements.txt --output requirements-fixed.txt --loglevel INFO
-    displayName: "Pin Requirements file"
-    condition: eq('${{parameters.pinRequirements}}', True)
+  - template: requirements-pin-steps-template.yml
 
-  - script: pip install -r ${{ parameters.requirementsFile }}
-    displayName: 'Install required packages specified in ${{ parameters.requirementsFile }}'
+  - temaplte: pip-install-steps-template.yml
```
riedgar-ms: `template` ?

python/fairlearn-core/fairlearn/postprocessing/_curve_plotting_utilities.py
```diff
@@ -3,15 +3,19 @@
...
+    import matplotlib.cm as cm
+    import matplotlib.colors
+
+    tab10_norm = matplotlib.colors.Normalize(vmin=0, vmax=7)
+    tab10_scalarMap = cm.ScalarMappable(norm=tab10_norm, cmap='Dark2')
+    debug_colors = [tab10_scalarMap.to_rgba(x) for x in range(10)]
+    debug_ncolors = len(debug_colors)
+    debug_colormap = {}
+except ImportError:
+    print("Please install fairlearn-core[postprocessingplot] to get the dependencies for postprocessing plots.")
```
riedgar-ms: Should you set a flag which can be checked in all the routines which would have called matplotlib, and print this error message again?

python/fairlearn-core/setup.py
```diff
@@ -0,0 +1,68 @@
...
+        "Programming Language :: Python :: 3.6",
+        "Programming Language :: Python :: 3.7",
+        "License :: OSI Approved :: MIT License",
+        "Operating System :: OS Independent",
+        "Development Status :: 3 - Alpha"
+    ],
+    include_package_data=True,
+    zip_safe=False,
+    extras_require={
+        "postprocessingplot": [
```
riedgar-ms: Rather than something so specific, could we have `enable-matplotlib` or similar?

devops/all-tests-job-template.yml
romanlutz: Wow, well spotted!

python/fairlearn-core/setup.py
```diff
@@ -0,0 +1,68 @@
...
+            "flake8-logging-format",
+            "flake8-rst-docstrings",
+            "requirements-parser"
+        ],
+        "test": [
+            "pytest==5.0.1",  # pinned due to VSCode issue
+            "pytest-cov",
+            "tempeh==0.1.11",
+            "wheel",
+            # notebook test requirements
```
riedgar-ms: Need papermill and nteract-scrapbook too

python/fairlearn-core/setup.py
```diff
@@ -0,0 +1,68 @@
...
+        ],
+        "devenv": [
+            "autopep8",
+            "flake8",
+            "flake8-blind-except",
+            "flake8-builtins",
+            "flake8-docstrings",
+            "flake8-logging-format",
+            "flake8-rst-docstrings",
+            "requirements-parser"
```
riedgar-ms: `requirements-parser` might not be needed any more

devops/all-tests-job-template.yml
riedgar-ms: Not that I've ever made that particular typo myself, you understand......

README.md
romanlutz: Yeah that was because I ran that script and didn't realize it had overwritten the file. Undid all of these

python/fairlearn-core/fairlearn/postprocessing/_curve_plotting_utilities.py
romanlutz: There is a flag called "plot" that is True if we're supposed to plot the ROC curves. I can extend the checks to not only check for this flag, but also add a check to see if matplotlib is installed.

python/fairlearn-core/setup.py
romanlutz: 1. I don't know whether dashes are allowed here.
2. This is very specific and tells the user why they'd need this, whereas enable-matplotlib doesn't really tell you anything about why you'd want this.

Given this, do you still think it's better? I don't claim to have the perfect answer to this, but it feels better to me :-)

python/fairlearn-core/setup.py
romanlutz: Right, haven't synced with latest master yet

python/fairlearn-core/setup.py
romanlutz: We can check for that!

adrinjalali:
Interesting approach. So you're creating both packages using the same repo?

doesn't it make sense to have a separate repo for `fairlearn-core` and one for the other stuff? The current proposal kinda seems not so clean and introduces quite significant technical debt. It's also kinda hard to maintain if you ask me.

romanlutz:
Closing this for now since the way this PR imports fairlearn_core isn't nice. WIP PR 240 represents the current version of where we'd like to go and is looking for input/feedback. https://github.com/fairlearn/fairlearn/pull/240

PR #240, opened 2020-01-13
[WIP] create extensions to install custom plots separately & check in generated files

romanlutz:
This PR includes:
- adding extensions for customplots (matplotlib for postprocessing plots) and widget (ipywidgets for dashboard)
- tests to ensure the error message shows if that functionality is required but not installed (see test/install)
- devops pipelines for those tests
- moving yml templates into a separate directory because it got hard to find files otherwise
- check-in of generated js files to avoid requiring a yarn build


README.md
```diff
@@ -57,21 +57,13 @@ The package can be installed via
 pip install fairlearn
 ```
 
+and optionally with extensions such as `pip install fairlearn[widget]` to get the fairlearn dashboard.
```
rihorn2: so its required to specify that we want to include this?

romanlutz: Yeah in the current version that's true, but that won't work for us because it'll break users like you said :-(

CONTRIBUTING.md
```diff
@@ -15,6 +15,17 @@ Development happens against the `master` branch following the [GitHub flow model
...
 
+To build the fairlearn dashboard after making changes to it, [install Yarn](https://yarnpkg.com/lang/en/docs/install), and then do the following:
+
+```bash
+cd fairlearn/widget/js
+yarn install
+yarn build:all
+rm -rf dist
+rm -rf lib
+rm -rf node_modules
```
adrinjalali: alternatively you could put these in a makefile and tell users of the make command

devops/templates/limited-installation-job-template.yml
```diff
@@ -0,0 +1,46 @@
+parameters:
```
adrinjalali: it'd be nice to have a comment on top of these yml files and say what each one is supposed to be doing

CONTRIBUTING.md
romanlutz: Good idea. Isn't makefile something platform specific, though? We've seen this many times in the past that there are people who write powershell scripts, others write bash scripts, etc. and others have to rewrite it for their OS or set of tools. The only good answer to this problem that I've seen is to write everything in python since it runs cross-platform.

romanlutz: Ah, indeed I wasn't far from the truth! This already exists as a powershell file under scripts. I guess I don't need this description at all then and can just refer to the script. Unfortunately powershell won't help people other than Windows users :-(

romanlutz: Turns out that isn't quite true :O https://docs.microsoft.com/en-us/powershell/scripting/install/installing-powershell-core-on-linux?view=powershell-7
Pretty sure most people wouldn't go that route though :D

devops/templates/limited-installation-job-template.yml
romanlutz: I could not agree more!

CONTRIBUTING.md
riedgar-ms: Powershell is available on Linux, and ADO does some sort of magic to make scripts like this run (well, PowerShell itself aliases `ls`, `rm` and `cd` to its own internal commands). And I believe `make` is able to drive people mad on multiple platforms too ;-) It's all a question of what's likely to be installed by default.

devops/templates/limited-installation-job-template.yml
romanlutz: Some of them do already, but I'll add descriptions where there are none yet

CONTRIBUTING.md
romanlutz: I added a link to the script that does this.

romanlutz:
> 
> 
> I guess you see the algorithms and the viz together as one package (which they have been so far), but I see them as two completely different set of tools. That may be why in my mind it makes sense to have them in two different repos.
> 
> And I don't think you'd need to duplicate README files, since I'd put the relevant info in each respective README file. Each of those files would focus on their own repo, and if you want, they can have links to examples which use both libs, and that example can be in either of the repos, or a third one, or just on a different website showing example usages.
> 
> Also, as I mentioned, the directory structure of the repo in this PR seems somewhat confusing, but that's not a problem if your maintainers are used to it, and you have it well documented in a _contributing guide_.
> 
> Have you tried writing a recipe for this?
> 
> Also, could you explain how the user should be working/installing the lib after this PR?

This is a philosophical question on whether fairlearn = mitigation + dashboard or just one of the pieces. Since we've shipped it as the combination of both of them in the past I think it'd be a tough sell to now say it doesn't include the dashboard anymore. @mesameki  @vingu @MiroDudik  should let us know what they think.

README.md
romanlutz: this is not relevant anymore since fairlearn as of now will include the widget

romanlutz:
> 
> 
> > I guess you see the algorithms and the viz together as one package (which they have been so far), but I see them as two completely different set of tools. That may be why in my mind it makes sense to have them in two different repos.
> > And I don't think you'd need to duplicate README files, since I'd put the relevant info in each respective README file. Each of those files would focus on their own repo, and if you want, they can have links to examples which use both libs, and that example can be in either of the repos, or a third one, or just on a different website showing example usages.
> > Also, as I mentioned, the directory structure of the repo in this PR seems somewhat confusing, but that's not a problem if your maintainers are used to it, and you have it well documented in a _contributing guide_.
> > Have you tried writing a recipe for this?
> > Also, could you explain how the user should be working/installing the lib after this PR?
> 
> This is a philosophical question on whether fairlearn = mitigation + dashboard or just one of the pieces. Since we've shipped it as the combination of both of them in the past I think it'd be a tough sell to now say it doesn't include the dashboard anymore. @mesameki @vingu @MiroDudik should let us know what they think.

Looks like the ease of use/understanding is an important factor, so the idea from Miro is to keep it all in one package, and add an extension for matplotlib. The PR is updated to reflect that now. Let me know if this works for you and for conda-forge

CONTRIBUTING.md
```diff
@@ -3,19 +3,25 @@
...
+
 This project has adopted the [GitHub community guidelines](https://help.github.com/en/github/site-policy/github-community-guidelines).
 
 ## Development process
+
 Development happens against the `master` branch following the [GitHub flow model](https://guides.github.com/introduction/flow/). Contributors create feature branches off of `master`, and their pull requests should target the `master` branch. Maintainers are responsible for prompt review of pull requests.
 
 Pull requests against `master` trigger automated tests that are run through Azure DevOps. Additional test suites are run periodically. When adding new code paths or features, tests are a requirement to complete a pull request. They should be added in the `test` directory.
 
+To build the fairlearn dashboard after making changes to it, [install Yarn](https://yarnpkg.com/lang/en/docs/install), and then run the [widget build script](scripts/build-widget.ps1).
```
adrinjalali: are people supposed to have windows? I assume ps is for power shell?

MiroDudik: agreed with @adrinjalali . we should provide linux/unix instructions/scripts too.

romanlutz: I'm not going to duplicate scripts. We can write them in python since that's cross-plat, but that'll be a couple days work and I'll leave it up to @vingu to schedule that. powershell exists on linux so this is not a blocker

romanlutz: That said, I do agree that it's best not to use powershell or bash or make or whatever your tool of choice is. We should use python.

romanlutz: > 
> 
> I'm still not clear on how users are supposed to make from source.
> 
> There should either be enough info in the setup.py that users can do `python setup.py sdist` or `python setup.py bdist`, and do `pip install -e path` from whatever path they're supposed to call it. If the instructions are not as clear, then they could be in some Makefile. Can't windows handle makefiles?
> 
> Or should I try to make it and create new build instructions for people on POSIX platforms?

Nothing special is required for users. In fact the commands you wrote above are all possible, with "path" being "." in the root directory (where setup.py is). You just install fairlearn like any other package. This is just if you change the widget code you need to do extra steps. 

Why do we need makefiles here? That's just one more tool like powershell that people on some platforms have and not on others.

adrinjalali: Sure. I feel like that's a very microsoft centered position, but given this package is mainly developed by microsoft, it makes sense(?).

I guess I'll post a PR regarding development under linux once I actually start working on the package. I don't think python is a desired tool to do what you'd do in makefiles.

Out of my own curiosity, I checked and saw that scikit-learn, pandas, pytoch, tensorflow, numpy, and scipy (the latter 3 not in their root) have Makefiles.

I thought that's how people make :P

devops/templates/pypi-deployment-stages-template.yml
```diff
@@ -51,37 +51,38 @@ stages:
...
-              secretsFilter: '${{parameters.kvUsername}},${{parameters.kvPassword}}'
-
-          - script: pip install twine
-            displayName: 'Install twine'
-
-          # Quote the password argument in case it contains characters the shell might try to parse
-          - script: 'twine upload --verbose --repository-url $(pypiUrl) -u $(${{parameters.kvUsername}}) -p "$(${{parameters.kvPassword}})" $(Pipeline.Workspace)/$(PackageDirectory)/*'
-            displayName: "Run twine for upload"
-    
+      matrix:
```
riedgar-ms: I hadn't spotted this before. Having this in here.... we should definitely do a 0.4.2 release to make sure everything still works

docs/fairlearn.exceptions.rst
```diff
@@ -1,7 +0,0 @@
-fairlearn\.exceptions module
-----------------------------
-
-.. automodule:: fairlearn.exceptions
```
riedgar-ms: Where did the contents go?

scripts/build-wheels.ps1
```diff
@@ -43,7 +43,7 @@ if( $targetType -eq "Test" )
 }
 
 # Store fairlearn version (including FAIRLEARN_DEV_VERSION) in the specified file
-Write-Host "Storing fairlearn version i $versionFilename"
+Write-Host "Storing fairlearn version in $versionFilename"
```
riedgar-ms: Nice catch :-)

devops/pypi-release.yml
```diff
@@ -33,48 +33,50 @@ stages:
...
       pyVersions: [3.5, 3.6, 3.7]
       freezeArtifactStem: "PreDeployment"
       freezeFileStem: $(FreezeFileStem)
       
-  - template: all-tests-job-template.yml
+  - template: templates/all-tests-job-template.yml
     parameters:
       name: MacOS
       vmImage:  'macos-latest'
       pyVersions: [3.6, 3.7]
```
riedgar-ms: Can we add 3.5 back in here, and run tests without matplotlib?

docs/fairlearn.exceptions.rst
romanlutz: This isn't a directory and is automatically included. I got a warning saying it's duplicate. If we want it to be its own module I can move it into a directory. What do you think?

devops/pypi-release.yml
romanlutz: I was going to do it afterwards but you're right, why not...

docs/fairlearn.exceptions.rst
riedgar-ms: It is a producing output, though:
https://fairlearn.readthedocs.io/en/latest/fairlearn.exceptions.html

devops/pypi-release.yml
adrinjalali: note that according to NEP18 (I think) py3.5 is no longer supported by the pydata ecosystem. matplotlib and the upcoming pandas release have already dropped it, and sklearn is dropping it for the next release.

CONTRIBUTING.md
romanlutz: What exactly are you referring to? I think not liking tools like powershell, or bash is very much company-agnostic :-) 
I had assumed make to be of a similar flavor but that may just as well be due to my ignorance (and I'm willing to fix my ignorance!).

What I'm trying to express is that I don't want duplication per OS because it leads to outdated scripts. That's usually uncontroversial. I'm not very familiar with makefiles, but if they work on every OS and are generally what people like to use then we should do that, no question!

I'll open an issue and we'll address that shortly. I don't want to club it with this change, though. Does that work for you?

romanlutz: ... forgot to mention: in various internal projects I've found python to be a fantastic tool for any kind of scripting because it's a lot more robust than bash and people who work on such projects are familiar with python already. 

Regardless, I'm the first person to abandon powershell, so let's get it done :-)

romanlutz: voila https://github.com/fairlearn/fairlearn/issues/242

docs/fairlearn.exceptions.rst
romanlutz: Oh I see why I got that impression. readthedocs doesn't show the top-level module with the class. Anyway, not a major thing, I'll leave it unchanged (i.e. revert my change)

romanlutz:
The failing builds happen on master as well. @riedgar-ms  is investigating, and they are unrelated to this change. Completing!

Issue #242, opened 2020-01-15
Create makefile and replace powershell


romanlutz:
All popular repos provide makefiles, so we should, too.

Apart from that we should move away from powershell since we want to be inclusive to all kinds of users & contributors.

romanlutz:
As much as possible this has been replaced for the moment. @adrinjalali and others may add to the Makefile as they see fit.

PR #249, opened 2020-01-16
Replace powershell scripts with python and add Makefile

romanlutz:
This addresses #242 
We'd like to remove the usage of powershell to encourage non-Windows users to participate. Python can run on any OS and is much more flexible in that regard.

This PR also adds a Makefile as requested by @adrinjalali --> please let me know if this is useful for you and feel free to update it. `make` doesn't run on Windows without additional overhead that I haven't gone through yet, so I really need feedback on that.


scripts/build_wheels.py
```diff
@@ -0,0 +1,65 @@
...
+
+from _utils import _ensure_cwd_is_fairlearn_root_dir, _LogWrapper
+from process_readme import process_readme
+
+
+_logger = logging.getLogger(__file__)
+logging.basicConfig(level=logging.INFO)
+
+
+_fairlearn_dev_version_env_var_name = 'FAIRLEARN_DEV_VERSION'
```
adrinjalali: what I usually see is that the version is set in the main `__init__.py`. Any reason you have it set in the build process here?

Makefile
```diff
@@ -0,0 +1,9 @@
+# simple makefile to simplify repetitive build env management tasks
+
+test-coverage: python -m pytest test -m "not notebooks" --ignore=test/perf --ignore=test/install --cov=fairlearn --cov-report=xml --cov-report=html
+
+test-unit: python -m pytest ./test/unit
```
adrinjalali: Somewhere in your scripts you're cleaning the build directories before running the build.

Usually, when we call a make command, the expectation is that it should only run the changed files.

If the user wants to start from scratch, they do `make clean`, and the `clean` command in the makefile takes care of removing all temp and build directories.

There are other common commands that we can add, but I can test and add them once you get this one in, in a separate PR.

adrinjalali:
I forgot to mention, in the description of the PR, if it includes "closes or fixes #number", it'll automatically close that pr or issue once it gets merged.

scripts/build_wheels.py
riedgar-ms: Because PyPI doesn't make upload and publish separate steps (each upload 'burns' the version number). This allows us to have multiple 'dev' versions on Test-PyPI and then send the final changeset to PyPI itself.

CONTRIBUTING.md
```diff
@@ -79,4 +79,4 @@ The pipeline requires sign offs immediately prior to the deployments to Test-PyP
...
 
-As part of the release process, the `build-wheels.ps1` script uses `process_readme.py` to turn all the relative links in the ReadMe file into absolute ones (this is the reason why the applied tag has be of the form `v[_base_version]`). The `process_readme.py` script is slightly fragile with respect to the contents of the ReadMe, so after significant changes its output should be verified.
\ No newline at end of file
+As part of the release process, the `build_wheels.py` script uses `process_readme.py` to turn all the relative links in the ReadMe file into absolute ones (this is the reason why the applied tag has be of the form `v[_base_version]`). The `process_readme.py` script is slightly fragile with respect to the contents of the ReadMe, so after significant changes its output should be verified.
```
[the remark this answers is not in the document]
riedgar-ms: Mainly because the links in the checked-in ReadMe are all relative. So when it goes to PyPI, we change all those to full URLs pointing at GitHub. The script should also make sure that the 'latest version' URL is properly set (that's the bit most likely to be fragile).

scripts/build_docs.py
```diff
@@ -0,0 +1,34 @@
+import logging
```
riedgar-ms: I'm not sure we actually need this anymore. ReadTheDocs does the necessary.

In particular `sphinx-apidoc` is not necessary, since I checkin the skeleton `.rst` files.

scripts/process_readme.py
```diff
@@ -3,126 +3,100 @@
...
-    logger.debug("Writing file %s", args.output)
-    with open(args.output, 'w') as f:
-        f.writelines(result_lines)
-    logger.debug("Completed")
+        with open(input_file_name, 'r') as input_file:
+            text_lines = input_file.readlines()
 
+    result_lines = [_process_line(line, target_version) for line in text_lines]
 
-if __name__ == "__main__":
```
riedgar-ms: No longer a standalone script?

Makefile
```diff
@@ -0,0 +1,9 @@
...
+
+test-perf: python -m pytest ./test/perf
```
riedgar-ms: Does this work? To me that looks like the dependencies for target `test-perf` are the files "python", "-m", "pytest" and I'm not sure what `make` does when confronted with a directory as a dependency.

Makefile
romanlutz: That would be fantastic. I guess I do that with `git clean -xdf` manually, but some of our scripts make sure that we start with clean directories, that's true

scripts/process_readme.py
romanlutz: It could be used as such actually, let me fix that

Makefile
romanlutz: I understood this to be an alias for a command, like it's used here: https://github.com/scikit-learn/scikit-learn/blob/master/Makefile

CONTRIBUTING.md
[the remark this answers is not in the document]
romanlutz: @adrinjalali how do other projects handle this?

scripts/build_docs.py
romanlutz: you mean the entire file?

riedgar-ms: Yes. ReadTheDocs just cares about the docs directory, and that it has a `conf.py`

Makefile
romanlutz: Can we put additional changes off, and you'll just adjust/enhance this file as and when you need it? I think you're the only user of this right now.

Makefile
romanlutz: @adrinjalali

scripts/build_docs.py
romanlutz: Wonderful...

romanlutz: deleted

scripts/build_wheels.py
romanlutz: I'll consider this closable unless there's any further suggestion on how to reconcile these goals @adrinjalali ?

CONTRIBUTING.md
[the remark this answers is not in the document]
adrinjalali: I don't see why those URLs shouldn't be absolute in the first place, see https://github.com/scikit-learn/scikit-learn/blob/master/README.rst as an example maybe.

scripts/build_wheels.py
adrinjalali: I'm not sure if I understand why you need to upload several copies of the same release to pypi. I haven't seen this issue since we upload only once per release.

Makefile
adrinjalali: yes, I can enhance it later.

scripts/build_wheels.py
riedgar-ms: We would prefer to only upload once per release too. However, it's not unknown for us to discover bugs after uploading to Test-PyPI. We try to remedy issues which we find so that they cannot reoccur, but they can still elude us.

adrinjalali: ah I see. That's why we do at least one RC before each release, and if things go wrong and the release needs a quick update, we do either a post release if the changes are in the docs only, and if there are code changes, we do a quick minor release afterwards. [PEP440](https://www.python.org/dev/peps/pep-0440/) specifies how the version numbers should work, in case you haven't seen it.

CONTRIBUTING.md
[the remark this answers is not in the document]
romanlutz: That's because they should reference the files of that version of the package. If CONTRIBUTING.md undergoes major changes from v0.5 to v0.6 then we wouldn't want the old version to point to the new markdown file. Even worse, some files may disappear/get renamed, which would mean the link gets lost.

Makefile
adrinjalali: I'm testing this branch locally to check the makefile.

scripts/build_wheels.py
romanlutz: We use RC/alpha releases as well. @riedgar-ms maybe we're overdoing it here?

CONTRIBUTING.md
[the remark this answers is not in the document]
romanlutz: In reality the main files shouldn't really disappear, though. Curious what @riedgar-ms  and @MiroDudik  think, but I think it doesn't need to be addressed in this PR.

scripts/build_wheels.py
riedgar-ms: There was a reason why I used `dev` instead of `rc` but I can't remember it now. It may have because `dev` could be appended to an `rc` but obviously on cannot stack up two `rc`s

CONTRIBUTING.md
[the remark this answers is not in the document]
romanlutz: started a thread in Teams, resolving this one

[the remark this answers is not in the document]
adrinjalali: I see. I still find it very odd. But we can talk about the build process maybe another time :D I really believe it can be simplified. One thing I noticed checking this, is that the pypi README links to the one on github, instead of somewhere in the docs.

Ideally the readthedocs documentation should include these, and the links should point to that one, not github.

scripts/build_wheels.py
adrinjalali: there's a clear distinction between `dev` and `rc` AFAIK. `dev` is what you have on master, and you usually don't upload them. You could have nightly builds for instance, with `devN` suffix, but the `rc` is for people out there to test the release, and it usually identifies a feature freeze for the release.

adrinjalali:
The `Makefile` should be more like this:

``` bash
# simple makefile to simplify repetitive build env management tasks

test-coverage:
	python -m pytest test -m "not notebooks" --ignore=test/perf --ignore=test/install --cov=fairlearn --cov-report=xml --cov-report=html

test-unit:
	python -m pytest ./test/unit

test-perf:
	python -m pytest ./test/perf

doc:
	python scripts/build_docs.py
```

But there is no `scripts/build_docs.py` anyway.

scripts/build_wheels.py
romanlutz: Why would you have multiple indicators, @riedgar-ms ?

riedgar-ms: That's the way the spec is written. I think (but I don't recall exactly) that the thing which steered me in that direction was that `dev` sorts at the lowest priority.

romanlutz: FYI I consider this orthogonal to this PR, but we should have this discussion, potentially on a Teams thread.

adrinjalali: Or on an issue :P

adrinjalali:
Not sure why that CI job is failing. But I'd say we should merge once/if it's mergable, and iteratively improve in future PRs.

scripts/build_wheels.py
romanlutz: Sure

romanlutz: https://github.com/fairlearn/fairlearn/issues/258

PR #250, opened 2020-01-17
Implement GroupMetricSet

riedgar-ms:
Create a `GroupMetricSet` class for holding collections of grouped metrics. This is to help with AzureML integration.

For Pythonic reasons, this has required some slight reorganisation of the `fairlearn.metrics` module internals; the exported interface should be unchanged.


fairlearn/metrics/_group_metric_set.py
```diff
@@ -0,0 +1,169 @@
...
+        self.model_type = model_type
+        if group_names is not None:
+            self.group_names = group_names
+
+        computed_metrics = {}
+        if self.model_type == GroupMetricSet.BINARY_CLASSIFICATION:
+            for k, f in GroupMetricSet.BINARY_CLASSIFICATION_METRICS.items():
+                computed_metrics[k] = f(self.y_true, self.y_pred, self.groups)
+        elif self.model_type == GroupMetricSet.REGRESSION:
+            for k, f in GroupMetricSet.REGRESSION_METRICS.items():
```
romanlutz: so k and f are metric_name and metric_function? Can we name it that?

fairlearn/metrics/_group_metric_set.py
```diff
@@ -0,0 +1,169 @@
...
+    GROUP_PRECISION_SCORE = "d2dcd0a6-3aba-4d91-8715-2655fb37674b"
+    GROUP_R2_SCORE = "445c1bec-9c6c-4085-bdea-26e37cea6194"
+    GROUP_RECALL_SCORE = "1e72002d-449d-40d9-b94c-5f053fd26032"
+    GROUP_ROC_AUC_SCORE = "5322378d-1839-4d91-8bb3-e54d295ed3cd"
+    GROUP_ROOT_MEAN_SQUARED_ERROR = "e8abde62-5511-4a67-b2c0-08b344227d7e"
+    GROUP_SELECTION_RATE = "adad0b7b-55b6-4e2f-9583-fdecc2ad9cb9"
+    GROUP_SPECIFICITY_SCORE = "53649e54-3d2b-495c-827e-bf4b201d7597"
+    GROUP_ZERO_ONE_LOSS = "3c2805a7-04b5-4b19-baa9-33928fe3d121"
+
+    BINARY_CLASSIFICATION_METRICS = {}
```
romanlutz: what about multiclass classification?

fairlearn/metrics/_input_manipulations.py
```diff
@@ -17,3 +17,16 @@ def _convert_to_ndarray_and_squeeze(target):
...
+
+def _convert_to_ndarray_1d(target):
+    """Convert the input to a `numpy.ndarray` with a single dimension.
+
+    If the input is multi-dimension, this will raise an exception
+    """
+    result = _convert_to_ndarray_and_squeeze(target)
+
+    if len(result.shape) > 1:
+        raise ValueError("Input not 1D")
```
romanlutz: can we make this string a constant and import it in the test? Otherwise we'll have to keep two strings in two places consistent

fairlearn/metrics/_group_metric_set.py
```diff
@@ -0,0 +1,169 @@
...
+    @property
+    def group_names(self):
+        """Return the group_names dictionary."""
+        return self._group_names
+
+    @group_names.setter
+    def group_names(self, value):
+        key_types = set(type(k) for k in value.keys())
+        if key_types != {int}:
+            raise ValueError("Keys not integers")
```
romanlutz: can we have full sentences in these error messages?
e.g. "The type of the argument `value` for `group_names` needs to be `int`. The provided type was ...."

fairlearn/metrics/_group_metric_set.py
riedgar-ms: I don't believe we have metrics for that yet. If I'm wrong, I can add that as another option.

fairlearn/metrics/_group_metric_set.py
riedgar-ms: Exception messages all improved

fairlearn/metrics/_input_manipulations.py
riedgar-ms: I've made the string a more descriptive constant. I'm not referencing from the test, since I think that's the sort of change which should trigger a test update.

fairlearn/metrics/_group_metric_set.py
riedgar-ms: Fair point. Have taken the opportunity to do a bit of other refactoring

fairlearn/metrics/_group_metric_set.py
romanlutz: Maybe we should? It's possible to do multiclass classification after all

riedgar-ms: Which metrics are appropriate for that?

test/unit/metrics/test_group_metric_set.py
```diff
@@ -0,0 +1,134 @@
...
+        target.group_names = {0: 1, 2: 3}
+    expected = "Values for group_names dictionary must be strings"
+    assert exception_context.value.args[0] == expected
+
+
+def test_metrics():
+    target = GroupMetricSet()
+    my_metric = GroupMetricResult()
+    my_metric.overall = 10222
+    target.metrics = {"60bdb14d-83fb-4374-ab2e-4c371f22b21e": my_metric}
```
romanlutz: I assume this is a copied GUID from the file where it's defined? Can we just import it so that these constants aren't all over the place?

riedgar-ms: It's a special GUID for this test. This is to test the get/set of the property

PR #259, opened 2020-01-22
MNT use sklearn's NotFittedError instead of NotFittedException

### LINKED ISSUE #253: Removed NotFittedException
Our `NotFittedException` should be retired in favour of the `NotFittedError` from `sklearn`; having our own type doesn't achieve anything.

It would also be good to make sure that all our Estimators implement this check (and are tested for it).

adrinjalali:
Fixes #253.

This PR replaces `NotFittedException` with `sklearn.exceptions.NotFittedError`, and removes the `exceptions.py` file, since that exception was the only on present in it.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Changes
 
+### v0.4.2
+
+* If methods such as `predict` are called before `fit`, `sklearn`'s
+  `NotFittedError` is raised instead of `NotFittedException`, and the latter
+  is now removed.
+
 ### v0.4.2, 2020-01-24
 * Separated out matplotlib dependency into an extension that can be installed via `pip install fairlearn[customplots]`.
 * Added a `GroupMetricSet` class to hold collections of `GroupMetricResult` objects
--- a/fairlearn/exceptions.py
+++ b/fairlearn/exceptions.py
@@ -1,8 +1,0 @@
-# Copyright (c) Microsoft Corporation. All rights reserved.
-# Licensed under the MIT License.
-
-"""Module holding specialised exceptions for fairlearn."""
-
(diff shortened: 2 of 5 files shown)


adrinjalali:
Could you remind me what the decision was regarding the changelog? Should I add an entry to the changelog?

Also, I'm not sure what your process is regarding merging the PRs. What I'm used to is that the person who creates a PRs almost never merges their own PR, and another person who approves does it (most contributors like me don't have permissions to merge any PRs anyway).

romanlutz:
> 
> 
> Could you remind me what the decision was regarding the changelog? Should I add an entry to the changelog?
> 
> Also, I'm not sure what your process is regarding merging the PRs. What I'm used to is that the person who creates a PRs almost never merges their own PR, and another person who approves does it (most contributors like me don't have permissions to merge any PRs anyway).

I always merge my own PRs because others would never know if there's anything missing. To some extent one can use WIP or similar kinds of markers, but once you have approvals why not complete it yourself? Permission-wise we obviously can't give the general public permission to do everything. However, I will do some research into the GitHub permissions model... lmk when you want us to compelte this one.

You can certainly add it to the changelog. Anything that has any relevance to users should be in there. For this one it's probably not affecting anyone, but we might as well be thorough about it. I'll leave it up to you.

adrinjalali:
> once you have approvals why not complete it yourself?

Having another person merging the PR, makes us certain that there have been at least two sets of eyes on a piece of code before it being merged. It makes things a bit slower, but avoids a single point of failure, and raises code quality in my experience. Of course it only works if there are enough people working on it :)

I also added the entry to the changelog. I guess it's ok now.

PR #260, opened 2020-01-22
Fix pypi release yaml

romanlutz:
Broken from recent restructuring, fixing now for v0.4.2 release.

### DIFF
--- a/devops/templates/pypi-deployment-stages-template.yml
+++ b/devops/templates/pypi-deployment-stages-template.yml
@@ -51,37 +51,36 @@ stages:
     pool:
       vmImage: ${{parameters.poolImage}}
 
+    variables:
+      packageDirectory: '${{parameters.packageArtifactStem}}${{parameters.targetType}}'
+      ${{ if eq(parameters.targetType, 'Test') }}:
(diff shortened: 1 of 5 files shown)


devops/templates/pypi-deployment-stages-template.yml
```diff
@@ -51,37 +51,36 @@ stages:
...
 
+    variables:
+      packageDirectory: '${{parameters.packageArtifactStem}}${{parameters.targetType}}'
+      ${{ if eq(parameters.targetType, 'Test') }}:
+        pypiUrl: https://test.pypi.org/legacy/
+      ${{ if eq(parameters.targetType, 'Prod') }}:
+        pypiUrl: https://upload.pypi.org/legacy/
+
     strategy:
-      matrix:
```
riedgar-ms: Why the removal?

romanlutz: I added that back when we thought it's going to be two package names (fairlearn, fairlearn-core), but that's not the case anymore. Apparently I did a bad job undoing that change, and didn't notice because there's no sanity checking the pipelines.

PR #261, opened 2020-01-22
fix yaml release 0.4.2

romanlutz:
pulling in changes from PR #260 into the release branch


romanlutz:
Turns out we don't need specific branches for releases anymore, closing!

PR #262, opened 2020-01-23
add CHANGES.md for v0.4.2

romanlutz:


adrinjalali:
Yesterday I was also thinking it's a good idea to add the changes to the changelog, in each PR which makes those changes. This way you don't have to go back and see what has happened since last release.

PR #276, opened 2020-01-29
DOC contributing: trim lines and add notes on signoff

adrinjalali:
This PR trims the lines in the contributing guide, and adds this note:

```
It basically means you need to sign off your commits with your `gpg` key,
meaning you need to have a key in your `gpg` setup, and then you can set that
key to be the default using the guidelines in [this
post](https://stackoverflow.com/questions/10161198/is-there-a-way-to-autosign-commits-in-git-with-a-gpg-key).
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -4,39 +4,81 @@ This project welcomes contributions and suggestions.
 
 ## Developer certificate of origin
 
-Contributions require you to sign a _developer certificate of origin_ (DCO) declaring that you have the right to, and actually do, grant us the rights to use your contribution. For details, visit https://developercertificate.org/.
+Contributions require you to sign a _developer certificate of origin_ (DCO)
+declaring that you have the right to, and actually do, grant us the rights to
+use your contribution. For details, visit https://developercertificate.org/.
 
-When you submit a pull request, a DCO-bot will automatically determine whether you need to provide a DCO and decorate the PR appropriately (e.g., label, comment). Simply follow the instructions provided by the bot. You will only need to do this once across all repositories using our DCO.
+When you submit a pull request, a DCO-bot will automatically determine whether
+you need to provide a DCO and decorate the PR appropriately (e.g., label,
+comment). Simply follow the instructions provided by the bot. You will only
+need to do this once across all repositories using our DCO.
+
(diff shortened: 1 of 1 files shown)


adrinjalali:
lol, apparently my note is not working on my own commit. Let me see

adrinjalali:
This solution works now, changed the note to point to the right git hook.

romanlutz:
> 
> 
> This PR trims the lines in the contributing guide, and adds this note:
> 
> ```
> It basically means you need to sign off your commits with your `gpg` key,
> meaning you need to have a key in your `gpg` setup, and then you can set that
> key to be the default using the guidelines in [this
> post](https://stackoverflow.com/questions/10161198/is-there-a-way-to-autosign-commits-in-git-with-a-gpg-key).
> ```

Pardon the question but why are we trimming lines? :-)

adrinjalali:
Kinda for the same reason the code is trimmed at 80 according to PEP8. It makes it much easier to review future changes to the doc, review code side by side, and also much easier to read.

At least on my editor, and the community I know, the lines are not wrapped at the end of the screen. When I opened this file, most of the text was outside the screen and I had to scroll to see it.

And since the rendered version is not affected with the trimming, it doesn't seem to be a high risk one :)

romanlutz:
> 
> 
> Kinda for the same reason the code is trimmed at 80 according to PEP8. It makes it much easier to review future changes to the doc, review code side by side, and also much easier to read.
> 
> At least on my editor, and the community I know, the lines are not wrapped at the end of the screen. When I opened this file, most of the text was outside the screen and I had to scroll to see it.
> 
> And since the rendered version is not affected with the trimming, it doesn't seem to be a high risk one :)

Sidenote: we upped the limit to 99, see setup.cfg

I just fear that, unlike with code where we have flake8, there's no validation for this so it's going to get ignored in the future unless there's an actual check. I'm not fundamentally opposed to it, of course, but if you know of some validation for this that may be cheap to add lmk.

adrinjalali:
> Sidenote: we upped the limit to 99, see setup.cfg

Oh, interesting :D 

It's not crucial to validate, since it's just a convenience on the documentation, and we can just violate and fix it with no consequences.

Issue #237, opened 2020-01-13
Allow grouping by multiple sensitive features


rihorn2:
Multiple sensitive features are a common scenario. The only other way to capture intersectionality is by creating a 1-dimensional grouping vector manually and providing that to Fairlearn. We want to make this easier and allow users to provide an m-dimensional matrix. This needs to be enabled on three levels:

- [x] mitigation techniques
- [x] metrics
- [ ] dashboard

The current implementation for mitigation techniques simply considers all combinations from the `sensitive_features` matrix, e.g. if column A has 4 distinct values and column B has 3 distinct values and all combinations occur we have 4*3=12 groups.

It is possible that users want more fine-grained control, but we'll wait for that feedback. This could mean specifying multiple constraints (demographic parity w.r.t. column A and demographic parity w.r.t. column B, but not intersections of A and B, for example).

Dref360:
Hello,
I'm interested in this feature for Metrics.

If we take `group_summary` for example,

would the change be as simple as calling `np.unique(s_f, axis=0)` instead of `np.unique(s_f)`?
https://github.com/fairlearn/fairlearn/blob/master/fairlearn/metrics/_metrics_engine.py#L66

romanlutz:
Sorry for missing your message, @Dref360 ! As you probably observed this went along with a complete redesign of the metrics API.

I'm closing this issue now since the remaining portion (dashboard integration) will not happen in this repo, but in the `raiwidgets` package's repo which will be released at the end of November.

PR #280, opened 2020-01-30
Support for multiple sensitive features

romanlutz:
Since that meant creating this twice (once for reductions, once for postprocessing), I merged the input validation by making it slightly more generic. This meant adjusting A LOT of tests.

The tests for multi-column sensitive feature matrices are now gone since we allow that explicitly.

Missing:
- new tests for this feature
- undo notebook changes
- not in this PR:
  - tempeh updates to support multiple sensitive features (PR ready)
  - notebook to show this feature (need a good example since lawschool doesn't lend itself to this)

related to #237


fairlearn/_input_validation.py
```diff
@@ -5,70 +5,133 @@
...
     elif isinstance(formless, pd.Series):
         formed_vector = formless
     elif isinstance(formless, np.ndarray):
         if len(formless.shape) == 1:
             formed_vector = pd.Series(formless)
         elif len(formless.shape) == 2 and formless.shape[1] == 1:
             formed_vector = pd.Series(formless[:, 0])
+        elif compress_multiple_columns:
+            formed_vector = np.apply_along_axis(
+                lambda row: "-".join([str(row[i]).replace("-", "_") for i in range(len(row))]),
```
riedgar-ms: Could there be a safer character than `-` for joining the values?

fairlearn/reductions/_grid_search/grid_search.py
```diff
@@ -179,21 +177,17 @@ def fit(self, X, y, **kwargs):
...
         if isinstance(self.constraints, ClassificationMoment):
             logger.debug("Classification problem detected")
             is_classification_reduction = True
-            # We have a classification problem
-            # Need to make sure that y is binary (for now)
-            unique_labels = np.unique(y_train)
-            if not set(unique_labels).issubset({0, 1}):
-                raise RuntimeError(self._MESSAGE_Y_NOT_BINARY)
         else:
             logger.debug("Regression problem detected")
```
riedgar-ms: Don't regression problems require an instance of `ConditionalLossMoment`?

fairlearn/_input_validation.py
romanlutz: Maybe, but this will show up in the dashboard so I don't want # or $ or something weird, right? @rihorn2  @vingu @MiroDudik thoughts?

romanlutz: Also, we're replacing those dashes first, so it's not like columns with dashes won't work

fairlearn/_input_validation.py
```diff
@@ -5,70 +5,133 @@
...
     if isinstance(formless, list):
         formed_vector = pd.Series(formless)
     elif isinstance(formless, pd.DataFrame):
         if len(formless.columns) == 1:
             formed_vector = formless.iloc[:, 0]
+        elif compress_multiple_columns:
+            formed_vector = formless.apply(
+                lambda row: "-".join(
+                    [str(row[col]).replace("-", "_") for col in formless.columns]),
+                axis=1)
```
MiroDudik: The use of `replace` can map different values to the same value. My suggestion would be to avoid replace and use a standard separator, either "," or ";". If you're paranoid, you could throw a warning if the separator appears within sensitive features.

fairlearn/_input_validation.py
MiroDudik: see the comment above.

fairlearn/_input_validation.py
MiroDudik: Also--we should generate the new column name using the same mechanism, e.g., we should end up with column named `race,gender` containing values like `black,male`.

romanlutz: Yes, but if they used comma or semicolon in a feature name then you can also map to an already existing feature name. It seems unlikely enough and at least is has a strategy for dealing with such characters. Even if we want comma or semicolon it should be
`replace(';',',')` followed by `';'.join`

MiroDudik: To wrap this up: in the replace phase, let's use the escape character "\", so:
* replace "\" -> "\\"
* replace "," -> "\,"
(in that order)
After that we can join with "," and all the original information is preserved.

fairlearn/_input_validation.py
romanlutz: changed to comma as pointed out by Miro

fairlearn/_input_validation.py
romanlutz: Made this change. The column name replacement has not happened because we lose column names through the sklearn input validation. I'll see what we can do about that.

fairlearn/reductions/_grid_search/grid_search.py
romanlutz: The logic should be identical. I didn't change anything about moments here AFAICT

riedgar-ms: Fair point.

test/unit/input_convertors.py
```diff
@@ -34,8 +47,10 @@ def ensure_ndarray(X):
 def ensure_ndarray_2d(X):
     assert X is not None
     tmp = ensure_ndarray(X)
+    if len(tmp.shape) == 2:
+        return tmp
     if len(tmp.shape) != 1:
-        raise ValueError("Requires 1d array")
+        raise ValueError("Requires 1d or 2d array")
```
riedgar-ms: Can you do this with an `if … elif … else` construct? Right now, the error message doesn't match the if statement unless you look carefully at the preceding `return` logic.

test/unit/reductions/grid_search/test_grid_search_arguments.py
```diff
@@ -21,29 +27,40 @@
 candidate_Y_transforms = conversions_for_1d
 candidate_A_transforms = conversions_for_1d
 
+def _skip_if_invalid_transformation(A, transform):
+    if len(A.shape) > 1 and A.shape[1] > 1 and transform in [ensure_list, ensure_series]:
+        pytest.skip()
```
riedgar-ms: This is executable code in the middle of a test? Does this mean we end up with lots of skipped tests in our results?

test/unit/input_convertors.py
romanlutz: Agreed, this is kind of weird right now.

romanlutz: I moved this check before the other one which makes it more obvious (I think)

test/unit/reductions/grid_search/test_grid_search_arguments.py
romanlutz: Yes we do skip some, but they're marked as such. This is from running unit tests locally
```
 5014 passed, 526 skipped in 398.52 seconds
```

riedgar-ms: I was afraid of that. I think it's strange to flag a scenario as skipped when it's never intended to run.

romanlutz: I don't think it's problematic, especially if the reason for skipping is mentioned. I agree that it's nicer if they never get generated in the first place but I doubt it's worth holding off on this right now. There are many more things we should do to make our tests nicer and maybe built on the same foundation, and that will hopefully address this issue as well.

test/unit/postprocessing/conftest.py
```diff
@@ -0,0 +1,173 @@
...
+          sensitive_features_ex3,
+          X_ex,
+          labels_ex,
+          scores_ex)])
+def data(request):
+    return request.param
+
+
+@pytest.fixture(params=candidate_A_transforms)
+def data_sf(data, request):
```
riedgar-ms: 'sf' ?

test/unit/postprocessing/test_threshold_optimization.py
```diff
@@ -320,144 +314,123 @@ def _average_prediction_for_label(label, sensitive_feature_value,
...
-                          (sensitive_features_ex2, sensitive_feature_names_ex2)])
-@pytest.mark.parametrize("X_transform", ALLOWED_INPUT_DATA_TYPES)
-@pytest.mark.parametrize("y_transform", ALLOWED_INPUT_DATA_TYPES)
-@pytest.mark.parametrize("sensitive_features_transform", ALLOWED_INPUT_DATA_TYPES)
+    for a in data_X_y_sf.feature_names:
+        pos_indices = (mapped_sensitive_features == a) * (labels_ex == 1)
+        neg_indices = (mapped_sensitive_features == a) * (labels_ex == 0)
+        average_probs_positive_indices = np.average(predictions[pos_indices], axis=0)
+        average_probs_negative_indices = np.average(predictions[neg_indices], axis=0)
+        assert np.isclose(average_probs_positive_indices[0], expected_ps[_POS_P0])
```
riedgar-ms: What are the differences between `np.isclose` and `pytest.approx.` ?

test/unit/postprocessing/test_threshold_optimization.py
```diff
@@ -320,144 +314,123 @@ def _average_prediction_for_label(label, sensitive_feature_value,
...
-                          (sensitive_features_ex2, sensitive_feature_names_ex2)])
-@pytest.mark.parametrize("X_transform", ALLOWED_INPUT_DATA_TYPES)
-@pytest.mark.parametrize("y_transform", ALLOWED_INPUT_DATA_TYPES)
 @pytest.mark.parametrize("constraints", [DEMOGRAPHIC_PARITY, EQUALIZED_ODDS])
-def test_predict_multiple_sensitive_features_columns_error(
-        sensitive_features, sensitive_feature_names, X_transform, y_transform, constraints):
-    X = X_transform(_format_as_list_of_lists(sensitive_features))
-    y = y_transform(labels_ex)
-    sensitive_features_ = pd.DataFrame({"A1": sensitive_features, "A2": sensitive_features})
+def test_predict_different_argument_lengths(data_X_y_sf, constraints):
```
riedgar-ms: Hmmm looking and the 'old' vs 'new' test names, I'm note sure `diff` is being very helpful here :-(

riedgar-ms:
I am a little concerned that this is a monster PR going in lots of different directions at once. Are there any pieces which can be split off relatively easily, or is it too tangled together?

test/unit/postprocessing/conftest.py
romanlutz: sensitive features
Didn't want it to be super long because then I have to wrap every line 15 times... :-(
It makes more sense when there's data_X_y_sf I guess

test/unit/postprocessing/test_threshold_optimization.py
romanlutz: > 
> 
> I am a little concerned that this is a monster PR going in lots of different directions at once. Are there any pieces which can be split off relatively easily, or is it too tangled together?

I keep going back to this but 90% of the code changes are in tests. I cold apply the improved input validation only to postprocessing, or only to reductions, and then only adjust their tests, but it mainly saves on changed test lines, not in actual logic.

romanlutz:
> 
> 
> I am a little concerned that this is a monster PR going in lots of different directions at once. Are there any pieces which can be split off relatively easily, or is it too tangled together?

I keep going back to this but 90% of the code changes are in tests. I cold apply the improved input validation only to postprocessing, or only to reductions, and then only adjust their tests, but it mainly saves on changed test lines, not in actual logic.

adrinjalali:
If the change is backward compatible (which I guess it is), and changing one module wouldn't break the other modules, you could break it into modules and have separate PRs for each one.

test/unit/postprocessing/test_threshold_optimization.py
romanlutz: I don't think I understand. Can you elaborate?

romanlutz:
> 
> 
> If the change is backward compatible (which I guess it is), and changing one module wouldn't break the other modules, you could break it into modules and have separate PRs for each one.

Yes, we're not breaking anything that worked before. I'll take a look and maybe make it two changes, one for reductions and one for postprocessing.

fairlearn/_input_validation.py
```diff
@@ -3,91 +3,88 @@
...
+    # compress multiple sensitive features into a single column
+    if len(sensitive_features.shape) > 1 and sensitive_features.shape[1] > 1:
+        sensitive_features = np.apply_along_axis(
+                lambda row: _SENSITIVE_FEATURE_COMPRESSION_SEPARATOR.join(
+                    [str(row[i])
+                     .replace("\\", "\\\\")  # escape backslash and separator
+                     .replace(_SENSITIVE_FEATURE_COMPRESSION_SEPARATOR,
+                              "\\" + _SENSITIVE_FEATURE_COMPRESSION_SEPARATOR)
+                     for i in range(len(row))]),
+                axis=1,
```
MiroDudik: I know it's a small piece of code, but I wonder if we can pull out the "string joining" functionality into a little utility function, so you could directly pass it to the apply function. This will make it easier if in future we want to apply the same strategy to the column names.

fairlearn/_input_validation.py
MiroDudik: reasonable. re. joining of the column names, we should think where/whether we need the joined column names. for the time being, i'm not sure we do. if they get surfaced in the dashboard, the naming is probably handled separately anyway.

fairlearn/_input_validation.py
adrinjalali: we could also make it a function in the future if it's needed again ;)

test/unit/postprocessing/test_threshold_optimization.py
riedgar-ms: The two tests being matched together by `diff` are `test_predict_multiple_sensitive_features_columns_error` and `test_predict_different_argument_lengths` which are probably not testing remotely the same thing.... and hence looking at the differences is unlikely to be edifying. This is a complaint about `diff` and not the PR :-/

fairlearn/_input_validation.py
romanlutz: It may also be needed in widget code, although I don't know details there. I'll make it its own function

fairlearn/_input_validation.py
romanlutz: I agree. This can be addressed if that need arises

test/unit/postprocessing/test_threshold_optimization.py
romanlutz: isclose: https://github.com/numpy/numpy/blob/v1.17.0/numpy/core/numeric.py#L2179-L2292
approx: https://docs.pytest.org/en/latest/_modules/_pytest/python_api.html#approx

The second link says this:
```

    If you're thinking about using ``approx``, then you might want to know how
    it compares to other good ways of comparing floating-point numbers.  All of
    these algorithms are based on relative and absolute tolerances and should
    agree for the most part, but they do have meaningful differences:

...

    - ``numpy.isclose(a, b, rtol=1e-5, atol=1e-8)``: True if the difference
      between ``a`` and ``b`` is less that the sum of the relative tolerance
      w.r.t. ``b`` and the absolute tolerance.  Because the relative tolerance
      is only calculated w.r.t. ``b``, this test is asymmetric and you can
      think of ``b`` as the reference value.  Support for comparing sequences
      is provided by ``numpy.allclose``.  `More information...`__

test/unit/postprocessing/test_threshold_optimization.py
romanlutz: Ah I see. Sometimes the side-by-side view helps, although copying out of it is a horrific experience :-(

test/unit/reductions/grid_search/test_grid_search_arguments.py
romanlutz: fixing this in the other PR. Essentially they'll be "deselected" instead of skipped. I hope that's sufficient. Closing this as duplicate of the other thread.

test/unit/postprocessing/test_threshold_optimization.py
romanlutz: I don't see this as a major issue. We can make it the same throughout our tests at some point if we feel like it

fairlearn/_input_validation.py
romanlutz: done in the other PR!

test/unit/postprocessing/conftest.py
romanlutz: Any other suggestions on how to name this? I can add a comment on what it means if you like? Or make it A instead of sf?

riedgar-ms: Other than perhaps putting in a comment, I don't have a better suggestion.

romanlutz: Since I have everything passing right now I"ll complete this PR and add the comment in the next PR since it's very minor

PR #282, opened 2020-02-02
WIP sklearn compatible ThresholdOptimizer

adrinjalali:
This PR moves the `ThresholdOptimizer` towards a sklearn compatible one.

Still working on fixing the tests, and adding the sklearn's common tests for the estimator.

The changes are substantial, but I hope none of them are controversial.

p.s. the line limit is 80 here, since longer ones don't easily fit on my screen :D


fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -427,6 +260,167 @@ def _threshold_optimization_equalized_odds(sensitive_features, labels, scores, g
...
+    elif hasattr(estimator, "decision_function"):
+        return estimator.decision_function(X)
+    else:
+        raise ValueError("{} provides neither a `predict_proba` nor a "
+                            "`decision_function` method.".format(
+                                estimator.__class__.__name__
+                            ))
+
+
+class ThresholdOptimizer(ClassifierMixin, BaseEstimator):
```
romanlutz: @MiroDudik may have thoughts on this. It should probably be a new kind of "RandomizedClassifierMixin"?

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -427,6 +260,167 @@ def _threshold_optimization_equalized_odds(sensitive_features, labels, scores, g
     return predicted_EO_by_sensitive_feature
 
 
+_SUPPORTED_CONSTRAINTS = {
```
romanlutz: This slightly confused me in `fit` where it's used because the keys here are the names of the constraints, but the values are the actual mitigation methods. Maybe `_SUPPORTED_CONSTRAINT_MITIGATIONS` ? Or something like that?

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -427,6 +260,167 @@ def _threshold_optimization_equalized_odds(sensitive_features, labels, scores, g
...
+        """
+        check_is_fitted(self)
+        X, _, sensitive_features = self._validate_input_data(
+            X, sensitive_features)
+        positive_probs = _vectorized_prediction(
+            self._post_processed_predictor_by_sensitive_feature, sensitive_features,
+            self._unconstrained_predictor.predict(X))
+        return np.array([[1.0 - p, p] for p in positive_probs])
+
+    def _validate_input_data(self, X, sensitive_features, y=None):
```
romanlutz: We'll have some major merge conflicts here with my PR #280 . Just an FYI. I'll probably take until tomorrow to finish it even in the current state, but maybe it's worth incorporating your changes first and then I'll adjust mine since I've unified the input validation for reductions and postprocessing. I love that this allows us to not reinvent the wheel and just use sklearn's input validation which should be more obvious to users anyway. So in summary: I'd like your change to go in first, if possible, and I'll unify them then

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -427,6 +260,167 @@ def _threshold_optimization_equalized_odds(sensitive_features, labels, scores, g
...
+        if y:
+            X, y = check_X_y(X, y)
```
romanlutz: I've looked at this a little bit, but it's not entirely clear to me what types are allowed. We currently support list, dataframe, series, and ndarray for y and sensitive_features, X.

fairlearn/postprocessing/_threshold_optimizer.py
MiroDudik: Agreed that it might be good to change, but to something less verbose. Maybe `_SUPPORTED_MITIGATIONS` or `_SUPPORTED_APPROACHES`

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -427,6 +260,167 @@ def _threshold_optimization_equalized_odds(sensitive_features, labels, scores, g
...
+    `Hardt et al. (2016) <https://arxiv.org/abs/1610.02413>`_.
+
+    :param unconstrained_predictor: The trained predictor whose output will be
+    post processed
+    :type unconstrained_predictor: A trained predictor
+
+    :param estimator: An untrained estimator that will be trained, and
+        subsequently its output will be post processed
+    :type estimator: An untrained estimator
+
```
MiroDudik: let's fix "post processed" -> "postprocessed"  in the above two instances (this was decided in a discussion a while back, but did not propagate to the code yet)

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -427,6 +260,167 @@ def _threshold_optimization_equalized_odds(sensitive_features, labels, scores, g
...
+    :param plot: Show ROC/selection-error plot if True
+    :type plot: bool
+
+    :param random_state: set to a constant for reproducibility
+    :type random_state: int, np.RandomState, or None
+    """
+
+    def __init__(self, *, estimator=None, constraints=DEMOGRAPHIC_PARITY,
+                 grid_size=1000, flip=True, plot=False, random_state=None):
+
```
MiroDudik: We have lost `unconstrained_predictor` and all of the related logic. Was this intentional? The idea is that if this argument is present, we should use it in `fit` to initialize `self._estimator` (without calling `self.unconstrained_predictor.fit(...)`).

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -427,6 +260,167 @@ def _threshold_optimization_equalized_odds(sensitive_features, labels, scores, g
...
+        :param sensitive_features: sensitive features to identify groups by, currently allows
+            only a single column
+        :type sensitive_features: currently 1D array as numpy.ndarray, list, pandas.DataFrame,
+            or pandas.Series
+
+        :return: predictions in numpy.ndarray
+        """
+        check_is_fitted(self)
+        random_state = check_random_state(self.random_state)
+
```
MiroDudik: I would prefer having
`self.random_state_ = check_random_state(self.random_state)` in `fit` and then relying on it in `predict`.
That seems to be a better way to implement a randomized predictor.

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -427,6 +260,167 @@ def _threshold_optimization_equalized_odds(sensitive_features, labels, scores, g
...
+    def _validate_input_data(self, X, sensitive_features, y=None):
+        if y:
+            X, y = check_X_y(X, y)
+            y = check_array(y, ensure_2d=False, dtype=int)
+        else:
+            X = check_array(X)
+        check_consistent_length(X, sensitive_features)
+
+        if set(np.unique(y)) > set([0, 1]):
+            raise ValueError(NON_BINARY_LABELS_ERROR_MESSAGE)
```
MiroDudik: this should be `not (set(np.unique(y)) <= set([0, 1])`.

fairlearn/postprocessing/_threshold_optimizer.py
MiroDudik: I think we can hold off on this for the time being. But since you mention this: the idea would be to have a uniform approach to how we return randomized classifiers that are represented as mixtures over deterministic classifiers. This currently pops up only here and in `reductions.ExponentiatedGradient`

fairlearn/postprocessing/_threshold_optimizer.py
romanlutz: I like "supported mitigations"

fairlearn/postprocessing/_threshold_optimizer.py
romanlutz: By "hold off on this" do you mean "leave as is in this PR" or "don't change what was there before the PR"? I think you mean the former, but just trying to confirm

fairlearn/postprocessing/_threshold_optimizer.py
romanlutz: @adrinjalali can elaborate but from what I understand scikit-learn doesn't take already fitted models, but rather fits them itself. We can still have another version where we take an already fitted one, but it wouldn't be scikit-learn compatible. Is that about accurate?

fairlearn/postprocessing/_threshold_optimizer.py
MiroDudik: Sorry for confusion. Leave as is in this PR!

fairlearn/postprocessing/_threshold_optimizer.py
romanlutz: I think this was just copied from the existing validation which I am working through at the moment myself, and it seems off. E.g. the labels [1,2] are not ">" than [0,1] because that's superset. What it should be IMO
```
not set(np.unique(y)).issubset(set([0, 1]))
```
which should be identical to yours, Miro, but I find that way of writing more intuitive

MiroDudik: I don't have a strong preference between `<=` and `issubset`.

fairlearn/postprocessing/_threshold_optimizer.py
adrinjalali: Yes. `GridSearchCV` _clones_ the estimators before fitting them, and that clears all the state they have. If we want to allow the user to re-fit on using a pre-fitted estimator, we can implement a `warm_start` strategy for the estimator to use the previously fitted one. But by default all of these should be fit from scratch, since grid search is fitting them on a new part of the data each time anyway.

fairlearn/postprocessing/_threshold_optimizer.py
adrinjalali: What we could do, is to put both this PR and #280 on hold, and have a separate PR which does the input validation change only. It'll also make #280 probably much smaller, since that's a pretty big one, and ideally could be broken into a few PRs, which makes the review also easier. WDYT?

fairlearn/postprocessing/_threshold_optimizer.py
adrinjalali: yes, all of those are cast into an ndarray, through `np.asarray`.

fairlearn/postprocessing/_threshold_optimizer.py
adrinjalali: yeah this is a copy/paste from existing code, we can fix it in a different PR. I think there are better ways to handle the [categorical] output anyway. Just didn't wanna tackle that in this PR.

fairlearn/postprocessing/_threshold_optimizer.py
romanlutz: That sounds reasonable. Would be sad to lose the existing functionality if we're planning to add it back in anyway... any chance we can add that warm start in this PR, or is that a major change?

fairlearn/postprocessing/_threshold_optimizer.py
romanlutz: My change influences input validation pretty heavily since I'm allowing extra inputs, namely multi-dimensional sensitive features. I'd like to get mine in asap, so depending whether you think yours can go in I can resolve merge conflicts in my branch after you merge, or I can resolve them on yours after I merge.

romanlutz:
Should there be a set of test cases to ensure we remain sklearn compatible?

adrinjalali:
I'll add the sklearn tests once the existing tests pass.

I'm finding it difficult to make them pass since they're a bit obscure to me, but getting there.

One question: the output of `predict_proba` is of shape `(n_samples, n_classes)`. How should we use them?

fairlearn/postprocessing/_threshold_optimizer.py
adrinjalali: added `warm_start`

romanlutz:
> 
> 
> I'll add the sklearn tests once the existing tests pass.
> 
> I'm finding it difficult to make them pass since they're a bit obscure to me, but getting there.
> 
> One question: the output of `predict_proba` is of shape `(n_samples, n_classes)`. How should we use them?

I don't think I understand your question properly. `predict_proba` has the probabilities, so each first entry is the probability for class 0 and the second entry the probability for class 1, etc. I think most of our work is binary classification right now, so `n_classes = 2`.

Please don't hesitate to ask if tests seem weird. They need to be made less obscure or explained with a little comment if they are a huge obstacle.

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -292,7 +293,7 @@ class ThresholdOptimizer(ClassifierMixin, BaseEstimator):
 
     :param estimator: An untrained estimator that will be trained, and
         subsequently its output will be postprocessed
-    :type estimator: An untrained estimator
+    :type estimator: An untrained estimator, default=SGDClassifier(loss='log')
```
romanlutz: Do we want to provide a default? If so, why that one? Just generally wondering, but I don't have a strong opinion

fairlearn/postprocessing/__init__.py
```diff
@@ -7,10 +7,8 @@
 learn how to adjust the predictor's output from the training data.
 """
 
-from ._postprocessing import PostProcessing  # noqa: F401
```
romanlutz: this probably warrants a description in the CHANGES.md as well :-)

fairlearn/postprocessing/_threshold_optimizer.py
romanlutz: I'm happy with how everything looks here. Any chance this can be checked in this week? That way I could resolve the conflicts in my branch which will be major if we make more adjustments to the tests.

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -427,6 +261,178 @@ def _threshold_optimization_equalized_odds(sensitive_features, labels, scores, g
     return predicted_EO_by_sensitive_feature
 
 
+_SUPPORTED_MITIGATIONS = {
+    DEMOGRAPHIC_PARITY: _threshold_optimization_demographic_parity,
+    EQUALIZED_ODDS: _threshold_optimization_equalized_odds}
+
+
+def _get_soft_predictions(estimator, X):
```
romanlutz: right now we're using `predict`, not `predict_proba` for this. However, when the scores are binary this method doesn't work as well as when using real-value scores so for examples (such as one of the notebooks) we've actually done this manually by wrapping the estimator and mapping `predict_proba` to `predict`. I'm not familiar with `decision_function`, but would we just use `predict` if the others aren't available?

romanlutz: Also adding @MiroDudik  for awareness

fairlearn/postprocessing/_threshold_optimizer.py
riedgar-ms: I thought we couldn't, since this was mutually exclusive with `unconstrained_predictor` ?

riedgar-ms: Hmmmm…. would be related to the `warm_start` below.

fairlearn/postprocessing/_threshold_optimizer.py
romanlutz: I may go ahead and check mine in first, but if that happens I'll create a branch with test changes required for yours, so you can merge with it to avoid having to fix all the merge conflicts.

adrinjalali: Sure, go ahead. I may break this PR to a few different ones to make it easier to review anyway.

fairlearn/postprocessing/_threshold_optimizer.py
MiroDudik: Sorry for a delay on this, but I think that `warm_star` is confusing in this context. 

Just to state the obvious: the high-level semantics of `ThresholdOptimizer` (after it is fitted) is really of a two-stage pipeline. The first step is implemented by `self._estimator.predict()`. It takes as input `X` and returns `y`. The returned `y` is then transformed into `y'` (inside `self.predict()`).

We would like to enable two styles of fitting this pipeline: (1) fitting both steps from the provided data; or (2) fitting only the second step from the provided data (and using the provided `unconstrained_predictor` as the first step). Now if I think about a warm start, I would expect re-fitting a previously fitted pipeline, using the previous pipeline as a warm start. This to me is quite different from (2), which is just fitting the second stage of the pipeline (from scratch) while keeping the first stage of the pipeline untouched.

I'm not quite sure I understand the issue with the interaction between the cloning behavior and the `unconstrained_predictor` argument.

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -427,6 +261,178 @@ def _threshold_optimization_equalized_odds(sensitive_features, labels, scores, g
...
+    :param warm_start: Avoid refitting the underlying estimator if it's already
+        done in a previous fit.
+    :type warm_start: bool, default=False
+
+    :param random_state: set to a constant for reproducibility
+    :type random_state: int, np.RandomState, default=None
+    """
+
+    def __init__(self, *, estimator=SGDClassifier(loss='log'),
+                 constraints=DEMOGRAPHIC_PARITY,
```
MiroDudik: If we're using SGDClassifier as a default, shall we run it with `random_state=0` (or something of the sort) to make it replicate? Alternatively, I don't think we necessarily need to have a default here.

fairlearn/postprocessing/_threshold_optimizer.py
MiroDudik: Interesting. As @romanlutz says, previously we were only using `predict` as the first stage of the pipeline before applying thresholding. I think it's reasonable to consider the back-off sequence `predict_proba` -> `decision_function` -> `predict`, where in `predict_proba` we only look at the predicted probability of the positive class.

So maybe two fixes to the below:
* for `predict_proba`, just extract the second column
* after `else`, don't throw an error, but instead return `predict`

fairlearn/postprocessing/_threshold_optimizer.py
romanlutz: Since this has happened now I can close this thread since it was just meant as a warning

fairlearn/postprocessing/_threshold_optimizer.py
romanlutz: I believe I'm fixing this in my PR so feel free to close this thread

fairlearn/postprocessing/_threshold_optimizer.py
adrinjalali: `predict` has very different semantics than `predict_proba` or `decision_function`. The probabilities `predict_proba` gives usually come from the raw values that the `decision_function` gives. You can think of the `decision_function` values as the distances from the hyperplanes in the case of an SVM.

`predict` always gives categorical results, and are not smooth. Either the method (`ThresholdOptimizer`) knows how to deal with that or it doesn't. If it doesn't, I don't think falling back to `predict` is a good idea. The `predict` function of a classifier shouldn't be giving soft outputs.

Alternatively, we could ask the user to either pass a classifier with a `predict_proba` or a `decision_function` provided, or a regression on a binary classifier which gives soft outputs as the predictions. One of our classifiers uses a regressor under the hoods in sklearn.

fairlearn/postprocessing/_threshold_optimizer.py
adrinjalali: I'm not very set on the `SGDClassifier`, really no hard feelings there. We could instead use `HistGradientBoostingClassifier` which is a much more modern one and people seem to like it.

Regarding the random state, since the user can set the `estimator` parameter, they can set it while passing a random state to it. It's the same in GridSearch and the underlying estimator for example. If the user needs a deterministic result, they should pass an rng to both.

romanlutz: Should we have a default at all? Is this commonly done?

fairlearn/postprocessing/_threshold_optimizer.py
romanlutz: I'm lacking relevant sklearn knowledge here: does every classifier have either `predict_proba` or `decision_function` or are there any that don't?
I don't understand what you mean by regression on a binary classifier. How would that be called? In some examples we use `predict_proba` from `LogisticRegression` and take the second entry (assuming binary classification) which is the probability for label 1.

MiroDudik: `ThresholdOptimizer` has no problem handling hard 0/1 predictions and that's in fact what the original paper was assuming. That's also the reason why we used `predict` in our original implementation. However, in those cases when the classifier is also providing a soft prediction (in the form of `decision_function` or `predict_proba`), it is obviously better to use that.

So I definitely think that we should allow users to run `ThresholdOptimizer` based on `predict` of the underlying classifier. What I wonder about is whether we should allow folks to force `ThresholdOptimizer` to use `predict` even in those cases when the classifier also implements the soft versions.

PROS: it allows you to run the "textbook version" of the algorithm
CONS: it's essentially guaranteed to lead to inferior performance

fairlearn/postprocessing/_threshold_optimizer.py
MiroDudik: I'm in favor of dropping the default value for the estimator. It might even make sense to drop a default on constraints.

fairlearn/postprocessing/_threshold_optimizer.py
MiroDudik: maybe you're referring to this?
https://scikit-learn.org/stable/glossary.html#term-metaestimators

fairlearn/postprocessing/_threshold_optimizer.py
adrinjalali: sklearn's common tests assume all estimators can be constructed using default values, except for the `estimator` or a meta-estimator. So we could leave this as required, and follow the [private] protocol to make the common tests understand how it should construct it. It'll use a `Ridge` if the meta-estimator is a regressor and a `LinearDiscriminantAnalysis` if a classifier.

Alternatively, the tests on the class can be disabled/skipped and only test instances.

fairlearn/postprocessing/_threshold_optimizer.py
adrinjalali: > I don't understand what you mean by regression on a binary classifier.

one example is that the `GradientBoostingClassifier` uses `DecisionTreeRegressor` at each step instead of a `DecisionTreeClassifier`. In some cases people use a regressor on the binary classification problem with 0/1 as the target, and they use the soft predicted output as a proxy for _probability_. 

We can have a `predict_method` parameter for the users to set it to whatever they want, and have it as `auto` by default.

MiroDudik: I like the final suggestion! Let's go with `predict_method` parameter defaulted to `auto`. For `auto`, we can try `predict_proba` -> `decision_function` -> `predict` (also I don't mind dropping the final `predict`).

The only weirdness is that whenever `predict_method` is `predict_proba`, we will need to extract the second column of the output.

fairlearn/postprocessing/_threshold_optimizer.py
MiroDudik: I think that the right approach is to go the meta-estimator route; this is something we will also want to use in our reductions, so if we implement it here, we can propagate it there.

On the API level, I think that `ThresholdOptimizer` should be similar to `CalibratedClassifierCV` [link](https://github.com/scikit-learn/scikit-learn/blob/b194674c42d54b26137a456c510c5fdba1ba23e0/sklearn/calibration.py#L32)

Re. defaults, in `CalibratedClassifierCV`, the `base_estimator` is defaulted to `None`, which becomes `LinearSVC(random_state=0)` during `fit`. I'm happy to go with whatever plays better with sklearn; if we need to have some default, I think that any linear model would work (LinearSVC, LinearDiscriminantAnalysis, or SGDClassifier), I just thought it would be nicer if the default was deterministic.

fairlearn/postprocessing/_threshold_optimizer.py
MiroDudik: To help resolve this. I suggest that we also implement the pattern from `CalibratedClassifierCV` [link](https://github.com/scikit-learn/scikit-learn/blob/b194674c42d54b26137a456c510c5fdba1ba23e0/sklearn/calibration.py#L32)

That would mean, using `prefit=False` instead of `warm_Start=False`.

adrinjalali: Sorry took me a while to concentrate enough to understand what you mean.

I understand that `warm_start` is different from what we have here. I'm just not sure why it's really needed. In a usual pipeline, we usually don't have pre-fitted estimators to feed to meta-estimators. Let me open a separate issue to talk about it there.

romanlutz:
@adrinjalali this is closable because you did this in multiple smaller PRs, right?

adrinjalali:
#342 should close this one :)

romanlutz:
> 
> 
> #342 should close this one :)

You probably mean the issue. I meant this PR :-) Are you still using this or is the PR closable?

TamaraAtanasoska:
@adrinjalali can I support the completion of this PR in any way? or is it at this point ok for it to be closed?

adrinjalali:
The code has changed quite a bit now, I think fresh PRs as you're doing is better.

Issue #257, opened 2020-01-22
Metrics - Alternative API


riedgar-ms:
We should consider changing the metrics API. Rather than having individual metrics be functions, we should consider creating a `GroupedData` object which would accept `y_true`, `y_pred` and `group_membership` as constructor arguments (and optionally `sample_weights`). Individual metrics would then become methods on that (still returning a `GroupMetricResult` object). Users would specify their own metrics via an `apply_function(my_metric_function)` method (where `my_metric_function` took arguments of `y_true`, `y_pred` and `sample_weights`).

adrinjalali:
Could you maybe post a pseudo code with what you'd like the API look like?

I'm not sure how that'll work integrating it with sklearn gridsearch though.

MiroDudik:
Interesting point. That's why should discuss this.

What I was thinking about was a pattern similar to pandas.DataFrame.groupby(). Specifically something like:

`# Initialize from a data frame df`
gd = GroupedData(y_true=df['label'], y_pred=df['prediction'], by=df['sex'], sample_weight=df['sample_weight'])

`# Use one of the built-in metrics`
f2_disparity_diff = gd.fbeta_score(beta=2).disparity_difference()
f2_disparity_ratio = gd.fbeta_score(beta=2).disparity_ratio()
f2_worst_case = gd.fbeta_score(beta=2).min()
f2_within_specific_group = gd.fbeta_score(beta=2)['male']
`# or:`
f2_within_specific_group = gd.fbeta_score(beta=2).by_group['male']
f2_overall = gd.fbeta_score(beta=2).overall()`

`# User-supplied metric`
custom_disparity_diff = gd.apply(custom_score).disparity_difference()

`# An alternate pattern`
f2_disparity_diff = gd.disparity_difference(fbeta_score, beta=2)
f2_worst_case = gd.min_metric(fbeta_score, beta=2)
f2_within_specific_group = gd.metric_by_group(fbeta_score, beta=2)['male']

`# Or we can go back to a current pattern (renaming "group_metric" to "disparity_summary")`
summary = disparity_summary(fbeta_score, beta=2, y_true=df['label'], y_pred=df['prediction'], by=df['sex'], sample_weight=df['sample_weight'])
f2_disparity_diff = summary.disparity_difference()
f2_worst_case = summary.min_metric()
f2_within_specific_group = summary.metric_by_group['male']

It would be also nice to enable easy evaluation of disparities of multiple metrics and/or multiple models.

[...]

PR #287, opened 2020-02-05
Remove GroupMetricResult

riedgar-ms:
Sample PR to remove the `GroupMetricResult` type and replace it with a `Bunch`. This does not yet fix the notebooks


fairlearn/metrics/_group_metric_set.py
```diff
@@ -1,6 +1,7 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
+from bunch import Bunch
```
adrinjalali: You could alternatively just use sklearn's bunch (`sklearn.utils`). The bunch package is not really maintained anyway.

fairlearn/metrics/_metrics_engine.py
```diff
@@ -57,6 +58,7 @@ def metric_by_group(metric_function,
         result.overall = metric_function(y_a, y_p, **kwargs)
 
     groups = np.unique(group_membership)
+    result.by_group = dict()
```
adrinjalali: Bunch _is_ a `dict`, and it seems the only thing we're putting in it here is another `dict` inside `by_group`. Doesn't it then make sense to put those results directly into result instead of under by_group?

fairlearn/metrics/_group_metric_set.py
riedgar-ms: Can you point me at the docs? The `sklearn` FAQ isn't keen on that:
https://scikit-learn.org/stable/faq.html?highlight=sklearn%20utils%20bunch#how-can-i-create-a-bunch-object
and I can't find 'bunch' on:
https://scikit-learn.org/stable/modules/classes.html#module-sklearn.utils
Thanks for the note about the maintenance of the 'bunch' package - I hadn't noticed the release date on PyPI.

fairlearn/metrics/_metrics_engine.py
riedgar-ms: That would be a problem if someone decides to have an 'overall' entry in their `group_membership` vector. Not sure why someone would, but that's either something else to check, or a charming pitfall for some user.

fairlearn/metrics/_group_metric_set.py
adrinjalali: haha, interesting, hadn't seen that part of the doc. From what I understand it's talking about using `Bunch` as an input, instead of output. As an output, it's used in a few places, for example: https://github.com/scikit-learn/scikit-learn/blob/76de8833195e8f4ebeaefd65e29e9b429fa2a4e7/sklearn/inspection/_permutation_importance.py#L123

I'll open an issue to discuss the docs about `Bunch` with the other developers.

riedgar-ms: I've updated the PR to use the `Bunch` in `sklearn.utils`. However, I would be happier if there were documentation about it.

fairlearn/metrics/_metrics_engine.py
romanlutz: But we can check for that, can't we?

riedgar-ms: Yes - as I said it's something else to check. I just think that it would be a bit odd to restrict the keys a user could use in their data. Especially as they most likely won't notice that bit of the documentation, but will instead end up with a sudden exception.

MiroDudik: I had a same thought as @adrinjalali . I think we should try to make the result shallow. Let's say that the two groups are "male" and "female". Then we could use keys like:
"group_male", "group_female", "overall", "min", "max", "range", "range_ratio"

riedgar-ms: I've updated this again. Please take a look at `test_metrics_engine.py` and let me know if you like the new syntax. I would like to have this agreed before going through and changing all of the other tests.

@MiroDudik @adrinjalali

fairlearn/metrics/_metrics_engine.py
```diff
@@ -3,9 +3,17 @@
...
 from ._input_manipulations import _convert_to_ndarray_and_squeeze
 
+_OVERALL = 'overall'
+_BY_GROUP_FORMAT = 'group_{0}'
+_MIN = 'min'
+_MAX = 'max'
+_RANGE = 'range'
+_RANGE_RATIO = 'range_ratio'
+_ARGMIN = 'argmin'
+_ARGMAX = 'argmax'
```
adrinjalali: I personally find this less readable than having those literals directly in the code. Especially when they're used pretty much only once.

fairlearn/metrics/_metrics_engine.py
```diff
@@ -49,29 +55,57 @@ def metric_by_group(metric_function,
...
+            range_ratio = 1
+        else:
+            range_ratio = minimum/maximum
+
+        result[_MIN] = minimum
+        result[_ARGMIN] = argmin
+        result[_MAX] = maximum
+        result[_ARGMAX] = argmax
+        result[_RANGE] = result_range
+        result[_RANGE_RATIO] = range_ratio
```
adrinjalali: how often are these two used? If not often, the user could compute them from the other values.

fairlearn/metrics/_metrics_engine.py
```diff
@@ -49,29 +55,57 @@ def metric_by_group(metric_function,
...
-                                                     group_predict,
-                                                     **kwargs)
+            result[group_key] = metric_function(group_actual,
+                                                group_predict,
+                                                **kwargs)
+
+    # Compute all the statistics, taking care not to stomp on our dictionary
+    try:
+        minimum = min(result.values())
+        argmin = [k for k, v in result.items() if v == minimum]
```
adrinjalali: use `numpy.argmin` and `numpy.argmax` instead?

fairlearn/metrics/_metrics_engine.py
```diff
@@ -49,29 +55,57 @@ def metric_by_group(metric_function,
...
+            range_ratio = minimum/maximum
+
+        result[_MIN] = minimum
+        result[_ARGMIN] = argmin
+        result[_MAX] = maximum
+        result[_ARGMAX] = argmax
+        result[_RANGE] = result_range
+        result[_RANGE_RATIO] = range_ratio
+    except ValueError:
+        # Nothing to do if the result type is not amenable to 'min' etc.
```
adrinjalali: when does this happen?

test/unit/metrics/test_group_metric_set.py
```diff
@@ -85,27 +86,20 @@ def test_group_names_values_not_string(self):
...
 
-    def test_metrics_values_not_groupmetricresult(self):
-        target = GroupMetricSet()
-        with pytest.raises(ValueError) as exception_context:
-            target.metrics = {"a": 0}
-        expected = "Values for metrics dictionary must be of type GroupMetricResult"
-        assert exception_context.value.args[0] == expected
-
 
 class TestConsistencyCheck:
```
adrinjalali: curious, why are the tests in classes?

test/unit/metrics/test_metrics_engine.py
```diff
@@ -188,13 +200,13 @@ def negative_results(y_true, y_pred):
...
-        assert result.maximum == -6
-        assert result.range == 4
-        assert np.isnan(result.range_ratio)
+        assert result['overall'] == -16
+        assert result['group_0'] == -10
+        assert result['group_1'] == -6
+        assert result['min'] == -10
+        assert result['max'] == -6
+        assert result['range'] == 4
+        assert np.isnan(result['range_ratio'])
```
adrinjalali: if the result is a `Bunch`, you could still access them as `result.min` etc.

fairlearn/metrics/_metrics_engine.py
riedgar-ms: I was discussing this with @rihorn2  at lunch, and I like this even less. Going to this level of change will require changes to the dashboard too. I do not see the benefits.

fairlearn/metrics/_metrics_engine.py
riedgar-ms: I don't believe those return the keys of a dict, but indices into an array?

fairlearn/metrics/_metrics_engine.py
riedgar-ms: When the metric is something like a confusion matrix, for which `min()` is not defined

test/unit/metrics/test_group_metric_set.py
riedgar-ms: Can't remember offhand why these ones are.

test/unit/metrics/test_metrics_engine.py
riedgar-ms: I decided against using that.

fairlearn/metrics/_metrics_engine.py
adrinjalali: it'd be nice if people who are not there in your office could see the result of your discussions in a better way :)

fairlearn/metrics/_metrics_engine.py
adrinjalali: I guess in that case, it should also be tested?

test/unit/metrics/test_metrics_engine.py
adrinjalali: also, for some of these you could keep the old _key_ name, to keep it more backward compatible.

adrinjalali: > I decided against using that.

care to elaborate why?

fairlearn/metrics/_metrics_engine.py
riedgar-ms: The `test_matrix_metric` test perhaps?

fairlearn/metrics/_metrics_engine.py
riedgar-ms: The result of the discussion: what is the benefit of doing this ('this' being _any_ of this change)?

test/unit/metrics/test_group_metric_set.py
adrinjalali: they may be remnants of `nose` or some other testing tool. `pytest` doesn't need them.

test/unit/metrics/test_metrics_engine.py
riedgar-ms: Thinking it over, I wasn't keen on a dependency on undocumented behaviour.

fairlearn/metrics/_metrics_engine.py
riedgar-ms: I would actually be happy to see all of these extra fields eliminated. Since there's no constraint on changing the other members, they could easily be inconsistent. That's why I changed `GroupMetricResult` to computing them dynamically.

test/unit/metrics/test_group_metric_set.py
riedgar-ms: We've always been `pytest`, but haven't been consistent in whether we use test classes or not. It might be down to the fact that the `_metrics_engine` module had two routines, so I had a class for each of them. That would be a remnant from before we reorganised the namespaces.

fairlearn/metrics/_metrics_engine.py
MiroDudik: the hope was to better support the use cases outlined in: https://github.com/fairlearn/fairlearn/issues/257#issuecomment-580811906
and also remove frictions that come about when you introduce new data types:
https://github.com/fairlearn/fairlearn/issues/257#issuecomment-581414697

test/unit/metrics/test_metrics_engine.py
adrinjalali: But we talked about it and the conclusion was to document it. That's why I've created  https://github.com/scikit-learn/scikit-learn/pull/16404

Also, if it's supposed to not be used, you could use a `dict` instead of a `Bunch`.

riedgar-ms: I _am_ using a `dict`.

fairlearn/metrics/_metrics_engine.py
adrinjalali: right, sorry, here `min(result, key=result.get)` would work I believe.

riedgar-ms: That appears to return only one of the minima.

riedgar-ms:
Per discussions on #257 this is not needed right now

PR #288, opened 2020-02-10
multiple sensitive features - postprocessing

romanlutz:
This is the trimmed down version of #280  for just postprocessing. After this is merged I'll update #280 which will then only be about reductions.


fairlearn/_input_validation.py
```diff
@@ -3,18 +3,96 @@
...
+        distinct values in the `sensitive_features` data from `kwargs`; default False
+    :type enforce_binary_sensitive_feature: bool
+    :param enforce_binary_labels: if True raise exception if there are more than two distinct
+        values in the `y` data; default False
+    :type enforce_binary_labels: bool
+    """
+    if y is not None:
+        # calling check_X_y with a 2-dimensional y causes a warning, so ensure it is 1-dimensional
+        if isinstance(y, np.ndarray) and len(y.shape) == 2 and y.shape[1] == 1:
+            y = y.reshape(-1)
```
riedgar-ms: How about np.squeeze?

romanlutz: Sure we can do that!

test/unit/postprocessing/conftest.py
```diff
@@ -0,0 +1,173 @@
...
+
+
+@pytest.fixture(params=candidate_A_transforms)
+def data_sf(data, request):
+    sensitive_feature_transform = request.param
+    # Skip combinations where the multi-column sensitive features would have to be compressed
+    # into a one-dimensional data structure.
+    if (data.sensitive_features == sensitive_features_ex3).all() and \
+            sensitive_feature_transform in [ensure_list_1d, ensure_series]:
+        pytest.skip(MULTIPLE_SENSITIVE_FEATURE_COMPRESSION_SKIP_REASON)
```
riedgar-ms: Is there no way we can prevent the invalid combinations being generated? Seeing skipped tests makes me think we're either got unfixed bugs (and hence are suppressing the test until the fix is in) or unimplemented features (similarly).

fairlearn/_input_validation.py
riedgar-ms: Just be aware that `squeeze` on a single element array results in a scalar.

test/unit/postprocessing/conftest.py
romanlutz: It's not super straightforward but I'll see what I can do

romanlutz: Is this acceptable to you?
```
1488 passed, 192 deselected in 285.46 seconds

Issue #252, opened 2020-01-22
Postprocessing - Separate plotting


riedgar-ms:
Having the decision on plotting done in the constructor of the `ThresholdOptimizer` object is rather odd. The plotting should be moved to a separate routine

adrinjalali:
We had very long discussions regarding the plotting API, and this is what we came up with, and will expand on: https://scikit-learn.org/stable/auto_examples/plot_roc_curve_visualization_api.html

romanlutz:
Thanks @adrinjalali ! Ours would be based on the mitigation method and the data it trained on, so we wouldn't actually pass X, y. My PR hopefully makes it clear how to use it. Any comments/feedback appreciated!

romanlutz:
This was obviously completed a long time ago. Closing the issue.

PR #289, opened 2020-02-11
Separate plotting code out of ThresholdOptimizer into functions

romanlutz:
addresses #252 
goal: don't have `plot` flag on ThresholdOptimizer anymore, instead have a function that plots based on the ThresholdOptimizer code

PR includes
- separation
- moved DP/EO code into ThresholdOptimizer because we need to save certain results on the ThresholdOptimizer object as opposed to not saving them at all
- adjusted tests based on moved DP/EO code
- added tests for the plotting functions using pytest-mpl which compares plots against baseline plots from the `baseline` directory


fairlearn/postprocessing/plotting.py
```diff
@@ -0,0 +1,118 @@
...
+
+    tab10_norm = matplotlib.colors.Normalize(vmin=0, vmax=7)
+    tab10_scalarMap = cm.ScalarMappable(norm=tab10_norm, cmap='Dark2')
+    debug_colors = [tab10_scalarMap.to_rgba(x) for x in range(10)]
+    debug_ncolors = len(debug_colors)
+    debug_colormap = {}
+
+    highlight_color = [0.95, 0.90, 0.40]
+except ImportError:
+    raise RuntimeError(_MATPLOTLIB_IMPORT_ERROR_MESSAGE)
```
riedgar-ms: This only happens when someone explicitly runs something from this file, and not just when they `import fairlearn.postprocessing` ?

requirements.txt
```diff
@@ -21,6 +21,7 @@ papermill
 # Pin pytest due to VS Code issue
 pytest==5.0.1
 pytest-cov
+pytest-mpl>=0.11
```
riedgar-ms: We should devise a way of testing our packages with-and-without the plotting extras

fairlearn/postprocessing/plotting.py
```diff
@@ -0,0 +1,118 @@
...
+             label='sensitive feature value ' + str(sensitive_feature))
+
+
+def _raise_if_not_threshold_optimizer(obj):
+    if not isinstance(obj, ThresholdOptimizer):
+        raise ValueError("Argument {} needs to be of type {}."
+                         .format(obj.__name__, ThresholdOptimizer.__name__))
+
+
+def plot_selection_error_curve(threshold_optimizer, show_plot=True):
```
riedgar-ms: I'm not an expert in matplotlib, but isn't there some sort of 'plotting' object which could be passed in, rather than just using the default? I'm thinking of cases where a user might want to make these a subplot in one of their larger plots. We would have the `plt` object you use as the default.

test/unit/postprocessing/baseline/README.md
```diff
@@ -0,0 +1,7 @@
+To regenerate the baseline plots run the following command from the root directory of the repository
```
riedgar-ms: Thanks for including this. Does it belong in CONTRIBUTING, though?

fairlearn/postprocessing/plotting.py
adrinjalali: you can pass an `axis` around, and if it's `None`, use the default device. Plotting on an `axis` makes it easier for the user to plot on a part of a larger plot as well.

test/unit/postprocessing/test_threshold_optimization.py
```diff
@@ -252,49 +266,49 @@ def test_threshold_optimization_equalized_odds(score_transform, y_transform,
...
+    assert base_value == prob_pred([sensitive_feature_names_ex1[2]], [0.5])
     assert np.isclose(value_between_0_5_and_1_5,
-                      adjusted_predictor([sensitive_feature_names_ex1[2]], [0.51]))
+                      prob_pred([sensitive_feature_names_ex1[2]], [0.51]))
     assert np.isclose(value_between_0_5_and_1_5,
-                      adjusted_predictor([sensitive_feature_names_ex1[2]], [1]))
+                      prob_pred([sensitive_feature_names_ex1[2]], [1]))
     assert np.isclose(value_between_0_5_and_1_5,
-                      adjusted_predictor([sensitive_feature_names_ex1[2]], [1.5]))
+                      prob_pred([sensitive_feature_names_ex1[2]], [1.5]))
```
adrinjalali: are all these changes related to plotting? Making each PR strictly limited to what it's supposed to do makes it easier to review and go back to see what's changed and why.

requirements.txt
adrinjalali: in the test configs, you can `SkipTest` a test or a file if a certain criteria is not satisfied. As an example you can see: https://github.com/scikit-learn/scikit-learn/pull/14675

fairlearn/postprocessing/plotting.py
romanlutz: Yes! In fact we are importing fairlearn.postprocessing earlier in this file

requirements.txt
romanlutz: I've added a pytest.skipif, but are we okay with leaving it in requirements.txt?

adrinjalali: it's not the heaviest and the worst item in the requirement.txt :P I'd leave it there :)

But you may want to add a CI to run the tests w/o the plotting deps installed, if it's not already there.

test/unit/postprocessing/baseline/README.md
romanlutz: I honestly don't know. At some point CONTRIBUTING becomes too large and people won't read it anymore. I'd rather have multiple guides at that point, maybe one for tests, and one for code, or some other way of splitting it up.

test/unit/postprocessing/test_threshold_optimization.py
romanlutz: Yes. If I don't make this adjustment then the tests fail. Root cause is that the threshold optimizer code used to be static functions as opposed to methods on the ThresholdOptimizer class. Now that we need to store some of the data during fit that needs to change which means that these tests needed a slight adjustment because the static functions are gone. It looks like a lot, but most of the changes here are added whitespace from indentation if you look at it side by side

requirements.txt
romanlutz: We have a pipeline running without matplotlib, but not without pytest-mpl. I'll look into it

romanlutz:
> 
> 
> do all those `.png` files have to be added to the repo?

How else will you check that it didn't change? :-)

requirements.txt
adrinjalali: ok then, I don't think one w/o `pytest-mpl` is needed then.

adrinjalali:
Testing the plot functions are tricky, but keeping the binary files in the repo is not a good solution IMO.

One solution is to have those plot functions used in the user guides (which we don't have yet, working on it), and if they go south, somebody will notice. We can check in the tests that they successfully run w/o any errors.

riedgar-ms:
Checking in comparison PNGs as binaries certainly isn't great, but so long as they don't change very often, I don't think it's too bad. Likely the main thing would be if new versions of `matplotlib` tweaked the display and wrecked the comparison.

I don't like the idea of waiting until 'someone' notices that a userguide has broken. Things which aren't automated can go a long time without being noticed.

romanlutz:
> 
> 
> Testing the plot functions are tricky, but keeping the binary files in the repo is not a good solution IMO.
> 
> One solution is to have those plot functions used in the user guides (which we don't have yet, working on it), and if they go south, somebody will notice. We can check in the tests that they successfully run w/o any errors.

I'm not sure I understand the concern. They're small in size and there are just six of them. They're not shipped with the package since they're under `test` either. Right now I can run `pytest --mpl` and it actually compares the figure, and just `pytest` simply runs the test and fails only if there's an exception (unless you don't have matplotlib or pytest-mpl installed)

romanlutz:
> 
> 
> Checking in comparison PNGs as binaries certainly isn't great, but so long as they don't change very often, I don't think it's too bad. Likely the main thing would be if new versions of `matplotlib` tweaked the display and wrecked the comparison.
> 
> I don't like the idea of waiting until 'someone' notices that a userguide has broken. Things which aren't automated can go a long time without being noticed.

I should add here:
`pytest --mpl` does the image comparison
`pytest` by itself just runs the test without comparison

Right now I haven't set `--mpl` in the CI pipelines because it didn't succeed yesterday. I'll look into that today, potentially related to different `matplotlib` versions. I'm thinking of turning it on only during the nightly builds to avoid disrupting PR gates, but that's just a thought.

fairlearn/postprocessing/plotting.py
romanlutz: Sounds good. Adjusting...

fairlearn/postprocessing/plotting.py
riedgar-ms: So long as it's severable :-)

test/unit/postprocessing/baseline/README.md
riedgar-ms: OK. Let's leave it here for now. If we adopt the pattern more widely, we might want to rethink where this information goes

riedgar-ms:
We need the plot tests running regularly, or there's no point having them. If they're finicky over platforms and versions, I wouldn't mind having a separate build job for them.

@adrinjalali 's concerns are very valid, so we should review how these work once they've been in the automation a while.

fairlearn/postprocessing/plotting.py
```diff
@@ -0,0 +1,124 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
```
MiroDudik: Maybe this was discussed somewhere. Why not just have a method `plot` on ThresholdOptimizer?

romanlutz: @adrinjalali  recommended following the same pattern that users might be used to from sklearn. I don't think it's better one way or the other, but if that's what users are used to then that's how it should be. What do you think @MiroDudik ?

adrinjalali: Yeah, `plot` is not what people expect to see as a method on an estimator

MiroDudik: I can see that. I don't have a strong preference, so happy to go with the standalone function.

However, I would still suggest a minor refactoring:
* create just the function `plot_threshold_optimizer` since the plotting is highly specific to plotting the solution of a *fitted* threshold optimizer; and only one kind of a plot applies to any instance anyway.
* don't put `plot_threshold_optimizer` into a standalone module, but instead put it in `fairlearn.postprocessing`. I think it's better to have `ThresholdOptimizer` and `plot_threshold_optimizer` in the same module (in our case, `fairlearn.postprocessing`). This is similar to `roc_curve` and `plot_roc_curve` being both in `sklearn.metrics`.

romanlutz: I can't put it under fairlearn.postprocessing because that doesn't necessarily guarantee the installation of matplotlib.

I like your rationale for a `plot_threshold_optimizer` method, although I might adjust the name slightly.

MiroDudik: it's a bit awkward to have modules with singleton functions. can we try something like this?
https://github.com/scikit-learn/scikit-learn/blob/b194674c42d54b26137a456c510c5fdba1ba23e0/sklearn/metrics/_plot/roc_curve.py#L84

romanlutz: Adjusted to `plot_threshold_optimizer`, but still in a separate module. @MiroDudik if you really want that we need to add the matplotlib dependency back

romanlutz:
> 
> 
> We need the plot tests running regularly, or there's no point having them. If they're finicky over platforms and versions, I wouldn't mind having a separate build job for them.
> 
> @adrinjalali 's concerns are very valid, so we should review how these work once they've been in the automation a while.

Couldn't get them to work across platforms. The results look identical to me, but it fails saying that it's outside of the tolerance. I guess the underlying matplotlib code various by platform. We could either just keep running the generation code to validate that it does fail (that's what the PR does right now) or have per-platform pictures. However, even the pictures of the platform that I generated the picture on didn't work (Windows in my case) so maybe there's more to it. I would suggest just keeping the simple validation that nothing fails for now.

riedgar-ms:
Can we have a single platform test, where the plots can be compared? It might have to be a separate build pipeline.

fairlearn/postprocessing/plotting.py
MiroDudik: Agreed that we shouldn't require matplotlib to always install. In sklearn, matplotlib is not automatically installed, but `plot_*` might then throw `ImportError`. See:
[sklearn dependencies](https://pypi.org/project/scikit-learn/)
[how ImportError is invoked](https://github.com/scikit-learn/scikit-learn/blob/d4a3ac4252401a20368c405976ff3db6f30995ac/sklearn/utils/__init__.py#L1094)
Is that too much hassle?

romanlutz:
> 
> 
> Can we have a single platform test, where the plots can be compared? It might have to be a separate build pipeline.

Those tests don't seem to be passing on any platform (I tried earlier when I had hoped it would "just work"). I'll create another issue to address this, though.

fairlearn/postprocessing/plotting.py
romanlutz: Done!

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -62,15 +62,12 @@ class ThresholdOptimizer(ClassifierMixin, BaseEstimator):
...
-    :type plot: bool
     """
 
     def __init__(self, *, unconstrained_predictor=None, estimator=None,
-                 constraints=DEMOGRAPHIC_PARITY, grid_size=1000, flip=True, plot=False):
+                 constraints=DEMOGRAPHIC_PARITY, grid_size=1000, flip=True):
         self.grid_size = grid_size
         self.flip = flip
-        self.plot = plot
         self.post_processed_predictor_by_sensitive_feature = None
```
MiroDudik: this should be moved to `fit` and renamed to `self.postprocessed_predictor_by_sensitive_feature_` (trailing underscore)

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -119,15 +116,15 @@ def fit(self, X, y, *, sensitive_features, **kwargs):
...
-                _threshold_optimization_demographic_parity
+                self._threshold_optimization_demographic_parity
         elif self.constraints == EQUALIZED_ODDS:
             threshold_optimization_method = \
-                _threshold_optimization_equalized_odds
+                self._threshold_optimization_equalized_odds
         else:
             raise ValueError(NOT_SUPPORTED_CONSTRAINTS_ERROR_MESSAGE)
 
         self._post_processed_predictor_by_sensitive_feature = threshold_optimization_method(
```
MiroDudik: is this intended to be `postprocessed_predictor_by_sensitive_feature_`?

MiroDudik: update: to expedite things, i'm fine leaving this as it is currently. we can expose this as a fitted attribute in future.

fairlearn/postprocessing/_threshold_optimizer.py
romanlutz: I'll have another PR for this since I finally got everything passing. Thanks for spotting it!

fairlearn/postprocessing/_threshold_optimizer.py
romanlutz: Indeed. That will be part of my follow-up PR then. Thank you!

Issue #265, opened 2020-01-24
test_bgl_unfair fails consistently on Windows with Python3.5


romanlutz:
Example: https://dev.azure.com/responsibleai/fairlearn/_build/results?buildId=3657&view=logs&jobId=a846d25a-e32c-5640-1b53-e815fab94407&j=f2ae15cd-30d2-5fb2-07f7-558e9be6b891&s=778b61a2-25e7-5600-2cb0-dd2dc8e23627

The affected code:
```
assert logging_all_close([[3.2, 11.2],
                              [-3.47346939, 10.64897959],
                              [-2.68, 10.12],
                              [-1.91764706, 9.61176471],
                              [-1.18461538,  9.12307692],
                              [-0.47924528,  8.65283019],
                              [0.2, 0.7]],
                             all_predict)
```
and only the first entry fails.
```
E        +  where False = logging_all_close([[3.2, 11.2], [-3.47346939, 10.64897959], [-2.68, 10.12], [-1.91764706, 9.61176471], [-1.18461538, 9.12307692], [-0.47924528, 8.65283019], ...], [array([ 3.03010885, 11.2       ]), array([-3.47346939, 10.64897959]), array([-2.68, 10.12]), array([-1.91764706,  9.61176471]), array([-1.18461538,  9.12307692]), array([-0.47924528,  8.65283019]), ...])

test\unit\reductions\grid_search\test_grid_search_regression.py:72: AssertionError
---------------------------- Captured stdout call -----------------------------
a mismatches:  [3.2]
b mismatches:  [3.03010885]
mismatch indices:  (array([0], dtype=int64), array([0], dtype=int64))
```

romanlutz:
According to Miro this may be because it's at the edge of the grid where one of the classes is ignored, and that may lead to imbalance resulting in multiple solutions with the same least squares loss. So that means it's actually undetermined. I'm putting in some tolerance to accept both to unblock myself, but we'll need to see whether the hypothesis can be verified.

riedgar-ms:
That looks pretty similar to the transient a week or so ago. Same item, same rough tolerance required. Which is rather large (on the order of 10%). Even if this is a case of two equally good solutions, it seems odd that we've only seen it recently. AFAICT, the `LinearRegression` solver is supposed to be fully deterministic (it wraps SciPy, which in turn wraps LAPACK), so even if there are two solutions which are equally good it should still always chose the same one of them.

MiroDudik:
[...]

PR #295, opened 2020-02-12
Fix Windows notebook tests for  python 3.5 & 3.6 build by adding special case for BGL test case back

romanlutz:
This was there previously but we thought it was fixed. Will need to validate with nightly builds to make sure it works.

Related to #265 but does not fix it :-(

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,7 +17,8 @@ requirements-parser
 
 # Required for test
 nbformat
-papermill
+# Pin papermill due to notebook test issue on Windows with python 3.5 & 3.6
+papermill<2.0.0
 # Pin pytest due to VS Code issue
 pytest==5.0.1
 pytest-cov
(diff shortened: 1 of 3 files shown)


adrinjalali:
just a reminder that according to NEP29 python3.5 is no more supported by the scipy ecosystem. Pandas, matplotlib, and the coming scikit-learn have already all dropped support for 3.5

riedgar-ms:
I would not be averse to dropping py3.5 (and probably adding py3.8...). @MiroDudik ? @romanlutz ?

romanlutz:
There's another issue in the nightly builds. The Windows py3.5 and py3.6 notebook pipelines fail due to a tolerance issue. I've upped the tolerance to 0.08 from 0.04 and they pass now. I wonder whether the base value 0.35 is off by a little or whether something on these platforms just works differently....

romanlutz:
> 
> 
> I would not be averse to dropping py3.5 (and probably adding py3.8...). @MiroDudik ? @romanlutz ?

I don't mind that at all, but you may want to factor in that you're trying to use this package in other packages that may rely on having py3.5? Not sure to what extent that should be a consideration for an open source project, but I want to point this out at least

romanlutz:
> 
> 
> > I would not be averse to dropping py3.5 (and probably adding py3.8...). @MiroDudik ? @romanlutz ?
> 
> I don't mind that at all, but you may want to factor in that you're trying to use this package in other packages that may rely on having py3.5? Not sure to what extent that should be a consideration for an open source project, but I want to point this out at least

This was also failing for python 3.6 on Windows (notebooks only), so the pinning of papermill is required regardless.

PR #296, opened 2020-02-13
Remove GroupMetricSet

riedgar-ms:
Remove the `GroupMetricSet` class

Put the relevant functionality into a new `create_group_metric_set` method, which returns a dictionary matching the dashboard's cache layout


test/unit/metrics/test_create_group_metric_set.py
```diff
@@ -0,0 +1,394 @@
...
+        assert gmr.overall == pytest.approx(accuracy['global'])
+        assert isinstance(accuracy['bins'], list)
+        assert len(accuracy['bins']) == 3
+        # Use the fact that the groups are integers
+        for j in range(3):
+            assert gmr.by_group[j+4] == pytest.approx(accuracy['bins'][j])
+
+
+@pytest.mark.parametrize("transform_y_true", conversions_for_1d)
+@pytest.mark.parametrize("transform_y_pred1", conversions_for_1d)
```
riedgar-ms: I had wanted to do the argument type testing on all the sub-lists, but it turns out that exponential growth is quite fast.

PR #297, opened 2020-02-14
Metrics Updates

riedgar-ms:
Some small metrics changes:

- Add a `version` field to the dictionary created by `create_group_metric_set()`
- Add `to_dict()`, `to_pandas_series()` and `__repr__()` methods to `GroupMetricResult`

Update the CHANGES document and bump the version to `0.4.4`.


CHANGES.md
```diff
@@ -1,5 +1,12 @@
 # Changes
 
+### v0.4.4
+
+* Remove `GroupMetricSet` in favour of a `create_group_metric_set` method
+* Add `to_dict()` and `__repr__()` methods to `GroupMetricResult`
+* Add basic support for multiple sensitive features
```
romanlutz: Thank you! I thought I had written that somehow...

fairlearn/metrics/_group_metric_result.py
```diff
@@ -91,6 +99,24 @@ def range_ratio(self):
...
+        if include_derived:
+            result[_MAX] = copy.deepcopy(self.maximum)
+            result[_MIN] = copy.deepcopy(self.minimum)
+            result[_RANGE] = copy.deepcopy(self.range)
+            result[_RANGE_RATIO] = copy.deepcopy(self.range_ratio)
+        return result
+
+    def __repr__(self):
+        """Generate string representation of a `GroupMetricResult`."""
+        return repr(self.to_dict())
```
romanlutz: would that be equal to `self.__dict__` ?

riedgar-ms: Depends on exactly what that does, if you look at the arguments to `to_dict()`

CHANGES.md
riedgar-ms: However, I think that your build fix needs to go in first....

fairlearn/metrics/_group_metric_result.py
```diff
@@ -1,8 +1,16 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
```
MiroDudik: This class right now feels like an overkill. I think I'm in favor of using the lighterweight approach that we played with a while back. Also, my proposal below should handle vector-valued metrics (but we can drop those and simplify things even further):

```python
from sklearn.utils import Bunch
import numpy as np

class GroupMetricResult(Bunch):

    def __init__(self, overall, by_group):
        self.overall = overall
        self.by_group = by_group

        self.min = np.min(list(by_group.values()), axis=0)   # or: min(by_group.values())
        self.max = np.max(list(by_group.values()), axis=0)   # or: max(by_group.values())
        self.disparity_diff = self.max - self.min
        self.disparity_ratio = _vec_positive_divide(self.min, self.max) # or: _positive_divide(self.min, self.max)

def _positive_divide(small, large):
    if (small<0):
        return np.nan
    elif (large==0):
        return 1.0
    else:
        return small/large

_vec_positive_divide = np.frompyfunc(_positive_divide, 2, 1)
```

riedgar-ms: Wouldn't deriving from `Bunch` make this a _heavier_ class?

Also in the case of vector-valued metrics, isn't `numpy.minimum` going to mix together elements from different subgroups? If so, does that really leave the `minimum` meaning anything?

MiroDudik: why was this merged? let's make sure to resolve the comments before merging in future.

PR #299, opened 2020-02-17
Report: Fix copy bug on overprediction explanation in binary classifier

kevinrobinson:
hello!  The code that points to the copy for explaining what overprediction and underprediction are has a typo, check out the part of the example UCI notebook highlighted in purple:

image

Here's this diff for the test app in the project (the NaN is intentional in the setup for the test app, and some data is random):

### copy change
Screen Shot 2020-02-17 at 2 11 11 PM

Separately, this PR adds in a background color for the `presentationArea` to fix this in the test app.  This isn't a visible issue in Jupyter since the cell background is white, but this is what that same screen looks like in the test app without the background color:

### presentationArea backgroundColor
Screen Shot 2020-02-17 at 2 04 02 PM

Also, I assume to merge this we need to build the widget assets and check them in, but I didn't do any of that here yet.

### DIFF
--- a/visualization/FairnessDashboard/src/WizardReport.tsx
+++ b/visualization/FairnessDashboard/src/WizardReport.tsx
@@ -119,7 +119,8 @@ export class WizardReport extends React.PureComponent<IReportProps, IState> {
         presentationArea: {
             display: "flex",
             flexDirection: "row",
-            padding: "20px 0 30px 90px"
+            padding: "20px 0 30px 90px",
+            backgroundColor: 'white'
         },
         chartWrapper: {
             flex: "1 0 40%",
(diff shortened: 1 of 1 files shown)


visualization/FairnessDashboard/src/WizardReport.tsx
```diff
@@ -344,7 +345,7 @@ export class WizardReport extends React.PureComponent<IReportProps, IState> {
                     <div className={WizardReport.classNames.colorBlock} style={{backgroundColor: ChartColors[0]}}/>
                     <div>
                         <div>{localization.Report.overestimationError}</div>
-                        <div>{localization.Report.underpredictionExplanation}</div>
+                        <div>{localization.Report.overpredictionExplanation}</div>
```
romanlutz: I'm not super familiar with this code, but why did it switch from under to over? @rihorn2

kevinrobinson: @romanlutz sorry, I missed all these notifications for some reason.  This change is fixing an existing bug, these were mislabeled.  You can check out the screenshots in the PR description to see the impact of the change.

Issue #222, opened 2020-01-06
try to pass sklearn's check_estimator


adrinjalali:
[This page](https://scikit-learn.org/dev/developers/develop.html#rolling-your-own-estimator) explains how you can write a test to pass tests included in `check_estimator`. We can see which ones fail and see if we can fix them.

TamaraAtanasoska:
Does this one still need to be open? We have recently used the new way to test estimators to test `AdversarialFairnessClassifier`, `AdversarialFairnessRegressor` and `_AdversarialFairness` (example in #1384). Failing tests have either been resolved or moved under the `xfail` flags to be resolved soon.

hildeweerts:
I agree I think this can be closed. @adrinjalali are you happy with the changes?

PR #304, opened 2020-02-19
FIX ThresholdOptimizer: move validations to fit

adrinjalali:
This PR moves validations from `__init__` to `fit`, and adds an `estimator_` attribute to store the fitted estimator. Towards #222. Was a part of #282.

### DIFF
--- a/fairlearn/postprocessing/_threshold_optimizer.py
+++ b/fairlearn/postprocessing/_threshold_optimizer.py
@@ -15,7 +15,6 @@ import pandas as pd
 import random
 
 from sklearn.base import BaseEstimator, ClassifierMixin
-from sklearn.exceptions import NotFittedError
 from sklearn.utils.validation import check_is_fitted
 
 from fairlearn._input_validation import _validate_and_reformat_input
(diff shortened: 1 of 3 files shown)


fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -69,25 +70,13 @@ class ThresholdOptimizer(ClassifierMixin, BaseEstimator):
...
-        if self._constraints not in _SUPPORTED_CONSTRAINTS:
-            raise ValueError(NOT_SUPPORTED_CONSTRAINTS_ERROR_MESSAGE)
-
-        self._grid_size = grid_size
-        self._flip = flip
-        self._plot = plot
-        self._post_processed_predictor_by_sensitive_feature = None
+        self.grid_size = grid_size
+        self.flip = flip
+        self.plot = plot
```
romanlutz: FYI plot is gone with my PR which I hope to complete today provided Richard/Miro manage to get a look

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -114,25 +111,25 @@ def fit(self, X, y, *, sensitive_features, **kwargs):
         else:
             y = [int(y_val) for y_val in y]
 
-        if self._estimator:
+        if self.estimator:
             # train estimator on data first
-            self._validate_estimator()
-            self._estimator.fit(X, y, **kwargs)
-            self._unconstrained_predictor = self._estimator
+            self.estimator_ = self.estimator.fit(X, y, **kwargs)
```
romanlutz: This is mostly for my own understanding: doesn't this mean that `self.estimator` and `self.estimator_` are the same thing?

romanlutz: Oh I guess we did the same thing before as well... nvm

fairlearn/postprocessing/_threshold_optimizer.py
romanlutz: ... unless this goes in first of course

romanlutz: Played around with this and managed to fix the flake8 issues. If the pipelines/tests pass I'll complete

fairlearn/postprocessing/_threshold_optimizer.py
adrinjalali: yeah this is what was done before. It should probably be instead cloning before fitting, but didn't wanna include that in this PR.

adrinjalali:
Thanks for the final touches @romanlutz

Issue #298, opened 2020-02-17
Support python 3.8


romanlutz:
... and have validation for the support through test pipelines

romanlutz:
This is completed with PR #305

PR #305, opened 2020-02-19
add python 3.8 pipelines/testing

romanlutz:
Also adjust a few existing ones where we've spread it across 3.5-3.7

Addresses #298 


romanlutz:
> 
> 
> Also adjust a few existing ones where we've spread it across 3.5-3.7
> 
> Addresses #298
> 
> Signed-off-by: Roman Lutz [rolutz@microsoft.com](mailto:rolutz@microsoft.com)

@adrinjalali before you ask: we'll keep python3.5 around now since it's working and it's not causing any overhead. As soon as that changes we can drop it if it's too much effort to fix. We do understand that numpy/scipy etc. are on 3.6+, but lots of people still use 3.5, so we don't want to put up additional barriers for them to use fairlearn. As long as these pipelines run we at least know whether it's working or not. If you ever do run into issues please shout at me and I'll be happy to remove it ;-)

fairlearn/metrics/_group_metric_result.py
```diff
@@ -69,7 +69,7 @@ def argmax_set(self):
         return set([k for k, v in self.by_group.items() if v == self.maximum])
 
     @property  # noqa: A003
-    def range(self):
+    def range(self):  # noqa: A003
```
riedgar-ms: Thank you. In view of the discussions about this whole type, I think that's the best option :-)

requirements.txt
```diff
@@ -1,9 +1,9 @@
 # Required for fairlearn
-matplotlib>=3.0.3
+matplotlib>=3.1.3
```
riedgar-ms: Does this work? I have memories of this being one of the issues with Python 3.5 on MacOS

romanlutz: You're right, but here's the issue: 3.1.3 is for  >=py3.6, but 3.0.3 doesn't work for py3.8
I could exclude python 3.8 from the requirements-fixed build, because that pins it down to the minimum version in this case, or we can drop python 3.5

adrinjalali:
Another way of handling deps is:

- document the dependencies, instead of having them in a single requirements file.
  People who want to install/test/etc, can do so by knowing the deps from the README
  or the contributing guide.
- The min deps apply to the oldest supported Python, it doesn't have to be the
  same for all python versions, cause it can't be anyway.
- The CI matrix defines the deps per python version, and not something as global as
  `requirements.txt`.
- For the CI, the deps are included in the scripts preparing the envs, taking the versions
  from the matrix environment variables or equivalent.

devops/templates/requirements-installation-steps-template.yml
```diff
@@ -0,0 +1,25 @@
+parameters:
+  pyVer:
+  requirementsFile:
+  pinRequirements:
+
+steps:
+- script: cp requirements.txt ${{requirementsFile}}
+    displayName: "Copy requirements file"
+    condition: ne('${{pyVer}}', '3.5')
```
riedgar-ms: I hope we don't have to do too many of these. Have you also run the deployment pipeline (as far as Test-PyPI) to check this works?

requirements-3.5.txt
```diff
@@ -0,0 +1,36 @@
+# Required for fairlearn
+matplotlib>=3.0.3
```
riedgar-ms: I wonder if there's a way of making sure that the `requirements` files remain usefully diffable

setup.py
```diff
@@ -23,17 +23,18 @@
         "numpy>=1.17.2",
         "pandas>=0.25.1",
         "scikit-learn>=0.22.1",
-        "scipy>=1.3.1"
+        "scipy>=1.4.1"
     ],
     extras_require={
         "customplots": [
-            "matplotlib>=3.0.3"
+            "matplotlib>=3.1.3"
```
riedgar-ms: This works when someone installs on py3.5/MacOS?

devops/templates/requirements-installation-steps-template.yml
romanlutz: It seems inevitable to have many of these to be honest. Unless we're happy to drop py35

requirements-3.5.txt
romanlutz: What do you mean?

setup.py
romanlutz: That will have to be set to the lower one, you're right

requirements-3.5.txt
riedgar-ms: Basically, we want to make sure that any differences between the requirements files are intentional due to issues like this. And not accidental due to us forgetting to update one of the files.

Ideally this would be done with a `diff requirements.txt requirements-3.5.txt` command (adding PowerShell bits if required). However, unless we're also very careful about how we add new dependencies to the files, then we could trip ourselves up by listing new dependencies in different orders.

I'm worrying in advance here; for now we might just take the note to 'be careful.'

devops/templates/requirements-installation-steps-template.yml
riedgar-ms: The 'interesting' nature of conditional syntax in the templates certainly makes the concern greater. I agree that it's pretty much inevitable for now :-(

Did the release pipeline run OK with this in place?

requirements-3.5.txt
romanlutz: Yes it works for now...
```
diff (Get-Content .\requirements.txt) (Get-Content .\requirements-3.5.txt)

InputObject       SideIndicator
-----------       -------------
matplotlib>=3.0.3 =>
scipy>=1.3.1      =>
matplotlib>=3.1.3 <=
scipy>=1.4.1      <=
```
I am somewhat hopeful that we're not adding too many new dependencies so it shouldn't ever get to that stage.

devops/templates/requirements-installation-steps-template.yml
romanlutz: Notice that I've moved the logic into a python script instead because the conditions didn't evaluate the pyVer parameter. Apparently they only evaluate it during the execution but not the generation or at least that's what I saw a few people complain about. I don't like the condition syntax a lot anyway, so let me know how you feel about the script. I'll run the release pipeline as well, without releasing, though

devops/templates/requirements-installation-steps-template.yml
```diff
@@ -0,0 +1,20 @@
+parameters:
```
riedgar-ms: Could you add a comment giving an overview of what this template does? The condition syntax isn't the nicest, so a brief guide would be good.

scripts/requirements_override.py
```diff
@@ -0,0 +1,46 @@
+"""Script for copying python version-specific requirements files to override requirements.txt.
+
+This is for pipelines only and should not be run by developers on their machines.
```
riedgar-ms: I believe that this looks for a `requirements-{m}.{n}.txt` file and matches m & n to the version of Python running this script? Again, this is another place where an extra comment would help (or better.... `desc` in the `parser` :-)

devops/PR-Gate.yml
```diff
@@ -16,15 +16,15 @@ jobs:
   parameters:
```
riedgar-ms: Not for this PR, but I'm going to add an issue to get our Pipeline logic better documented. These are getting rather involved.

scripts/requirements_override.py
romanlutz: I'll tackle that in a follow-up since I don't want to keep this one lingering for longer.

devops/templates/requirements-installation-steps-template.yml
romanlutz: Yes, but I'll delay to a further PR. This wasn't created in this PR anyway, just moved here.

romanlutz:
PyPI Release succeeded until the pypi-prod stage which I obviously didn't approve, so I consider this as green and will complete now

PR #306, opened 2020-02-22
Rename arguments to create_group_metric_set

riedgar-ms:
Bring the argument names for `create_group_metric_set` into line with the Dashboard:

- `group_memberships` becomes `sensitive_features`
- `group_titles` becomes `sensitive_feature_names`

This change propagates through the tests as well


fairlearn/metrics/_group_metric_set.py
```diff
@@ -98,16 +96,16 @@
 def create_group_metric_set(model_type,
```
riedgar-ms: Do we want to keep calling this `create_group_metric_set`? I can't think of a concise alternative, though. This would also affect what we put in the `schemaType` field.

PR #309, opened 2020-02-26
ENH make ThresholdOptimizer a meta-estimator

### LINKED ISSUE #308: ThresholdOptimizer as a scikit-learn meta-estimator
`ThresholdOptimizer`'s current implementation looks like a meta-estimator, but does not implement it as meta-estimators are implemented in sklearn.

One of the issues right now, is that it implements two different cases, one is when the child estimator is already fitted, one is when it's not.

As mentioned in https://github.com/fairlearn/fairlearn/pull/282#discussion_r377255359

> Sorry for a delay on this, but I think that `warm_star` is confusing in this context.
> 
> Just to state the obvious: the high-level semantics of `ThresholdOptimizer` (after it is fitted) is really of a two-stage pipeline. The first step is implemented by `self._estimator.predict()`. It takes as input `X` and returns `y`. The returned `y` is then transformed into `y'` (inside `self.predict()`).
> 
> We would like to enable two styles of fitting this pipeline: (1) fitting both steps from the provided data; or (2) fitting only the second step from the provided data (and using the provided `unconstrained_predictor` as the first step). Now if I think about a warm start, I would expect re-fitting a previously fitted pipeline, using the previous pipeline as a warm start. This to me is quite different from (2), which is just fitting the second stage of the pipeline (from scratch) while keeping the first stage of the pipeline untouched.
> 
> I'm not quite sure I understand the issue with the interaction between the cloning behavior and the `unconstrained_predictor` argument.

This makes it tricky to have a usual sklearn meta-estimator, since cloning the estimator (`ThresholdOptimizer` in this case) does not keep the fitted child estimator. For example, `Pipeline`:

``` python
In [1]: from sklearn.pipeline import Pipeline                   

In [2]: from sklearn.svm import SVC                              

In [3]: from sklearn.datasets import make_classification

In [4]: X, y = make_classification()

In [5]: clf = SVC().fit(X, y)

In [6]: Pipeline([('svm', clf)]).predict(X)
Out[6]: 
array([..., 0, 1, 1])

In [7]: from sklearn import clone

In [8]: clone(Pipeline([('svm', clf)])).predict(X)
...

NotFittedError: This SVC instance is not fitted yet. Call 'fit' with appropriate arguments before using this estimator.

```

And that's important since `GridSearchCV` would `clone` the estimators before fitting them.

Another point regarding `warm_start`, is that right now the user has to call `fit` even if the given estimator is pre-fitted. In `sklearn`, we usually assume `warm_start` means the user has called a fit on the estimator already; but here we could use it with a different meaning and assume that the user has passed a pre-fitted estimator. The `warm_start` semantics of sklearn are not very well-defined anyway, and we're still working on them.

__EDIT__

So my proposal is to implement a usual meta-estimator as is in sklearn, and include the `warm_start` parameter to support a pre-fitted estimator.

WDYT @MiroDudik @romanlutz

adrinjalali:
Fixes #308 

This PR makes the `ThresholdOptimizer` a meta-estimator, removes the `unconstrained_estimator` parameter, and introduces a `prefit` parameter to use a prefitted estimator.


adrinjalali:
pining @MiroDudik , @romanlutz , how does this look?

adrinjalali:
I'm getting a `                raise Expired("Credentials have expired due to inactivity.")` in the CI, which I don't think I can do much about.

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -34,44 +36,43 @@
...
+    def __init__(self, *, estimator=None,
+                 constraints=DEMOGRAPHIC_PARITY, grid_size=1000, flip=True,
+                 refit=True):
         self.grid_size = grid_size
         self.flip = flip
         self.post_processed_predictor_by_sensitive_feature = None
         self.constraints = constraints
-        self.unconstrained_predictor = unconstrained_predictor
         self.estimator = estimator
+        self.refit = refit
```
MiroDudik: I have a slight preference for `prefit` pattern (`prefit=True` to indicate when things should not be refitted) over the `refit` pattern.

The main reason is that `prefit` pattern already exists in [SelectFromModel](https://scikit-learn.org/stable/modules/generated/sklearn.feature_selection.SelectFromModel.html#sklearn-feature-selection-selectfrommodel) and somewhat in [CallibratedClassifierCV](https://scikit-learn.org/stable/modules/generated/sklearn.calibration.CalibratedClassifierCV.html#sklearn-calibration-calibratedclassifiercv). On the other hand, the usage of `refit` in [GridSearchCV](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.GridSearchCV.html#sklearn-model-selection-gridsearchcv) is a bit different.

test/install/test_no_matplotlib.py
```diff
@@ -4,12 +4,17 @@
...
 
+from sklearn.base import BaseEstimator, ClassifierMixin
+
 from fairlearn.postprocessing import ThresholdOptimizer, plot_threshold_optimizer
 from fairlearn.postprocessing._threshold_optimizer import _SUPPORTED_CONSTRAINTS
 from fairlearn.postprocessing._constants import _MATPLOTLIB_IMPORT_ERROR_MESSAGE
 
 
-class FakePredictor:
+class FakePredictor(BaseEstimator, ClassifierMixin):
```
romanlutz: nit: should this be FakeEstimator, instead of FakePredictor? I'm going purely by sklearn definitions here.

test/perf/script_generation.py
```diff
@@ -75,17 +75,11 @@ def add_dataset_setup(script_lines, perf_test_configuration):
...
     with TimedExecution(_ESTIMATOR_FIT, script_lines):
         script_lines.append('estimator = models["{}"]()'.format(perf_test_configuration.predictor))
-        script_lines.append('unconstrained_predictor = models["{}"]()'.format(perf_test_configuration.predictor))
-        script_lines.append('unconstrained_predictor.fit(X_train, y_train)')
 
 
 def add_mitigation(script_lines, perf_test_configuration):
     with TimedExecution(_MITIGATION, script_lines):
-        if perf_test_configuration.mitigator == ThresholdOptimizer.__name__:
-            script_lines.append('mitigator = ThresholdOptimizer('
```
romanlutz: This removes thresholdoptimizer from perf tests :-(

notebooks/Binary Classification on COMPAS dataset.ipynb
```diff
@@ -199,10 +208,11 @@
...
     "\n",
-    "unconstrained_predictor_wrapper = LogisticRegressionAsRegression(unconstrained_predictor)\n",
+    "estimator_wrapper = LogisticRegressionAsRegression(estimator).fit(X_train, y_train)\n",
     "postprocessed_predictor_EO = ThresholdOptimizer(\n",
-    "    unconstrained_predictor=unconstrained_predictor_wrapper,\n",
-    "    constraints=\"equalized_odds\")\n",
+    "    estimator=estimator_wrapper,\n",
+    "    constraints=\"equalized_odds\",\n",
+    "    refit=False)\n",
     "\n",
```
MiroDudik: prefit=True

notebooks/Mitigating Disparities in Ranking from Binary Data.ipynb
```diff
@@ -393,12 +394,12 @@
...
     "\n",
-    "class LogisticRegressionAsRegression:\n",
+    "class LogisticRegressionAsRegression(BaseEstimator, ClassifierMixin):\n",
     "    def __init__(self, logistic_regression_estimator):\n",
     "        self.logistic_regression_estimator = logistic_regression_estimator\n",
     "    \n",
     "    def fit(self, X, y):\n",
-    "        self.logistic_regression_estimator.fit(X, y)\n",
+    "        return self.logistic_regression_estimator.fit(X, y)\n",
     "    \n",
```
MiroDudik: We should probably return `self` here. However, note that as soon as we introduce `predict_method` parameter as suggested in https://github.com/fairlearn/fairlearn/pull/282/files#r380340267 we will have no need for `LogisticRegressionAsRegression`.

notebooks/Mitigating Disparities in Ranking from Binary Data.ipynb
```diff
@@ -411,7 +412,7 @@
     "balanced_index = balanced_index_pass0.union(balanced_index_pass1)\n",
     "\n",
     "pp_estimator = ThresholdOptimizer(\n",
-    "    unconstrained_predictor=LogisticRegressionAsRegression(unmitigated_predictor),\n",
+    "    estimator=LogisticRegressionAsRegression(unmitigated_predictor),\n",
     "    constraints=\"demographic_parity\")\n",
```
MiroDudik: prefit=True

notebooks/Mitigating Disparities in Ranking from Binary Data.ipynb
romanlutz: Oh wow I love that!

test/install/test_no_matplotlib.py
adrinjalali: I didn't want to touch the tests much. I'm not happy with them and I think they can use some restructuring in regard to these tiny estimators. But I rather leave that to another PR.

test/perf/script_generation.py
adrinjalali: I really wish we'd move this whole beast out of the repo and have readable "benchmarks" which don't have to run in the CI lol

notebooks/Mitigating Disparities in Ranking from Binary Data.ipynb
adrinjalali: Yes, but I've kept that for another PR.

notebooks/Binary Classification on COMPAS dataset.ipynb
MiroDudik: Did you mean to check in an updated notebook or leave this for a future PR? There's another place in this notebook, where we should use `prefit=True`.

test/install/test_no_matplotlib.py
romanlutz: I'm very open to suggestions on all of these, but that's probably a discussion for "issues" or at least elsewhere.

test/perf/script_generation.py
romanlutz: Sure, but can we just keep it until that's done?

notebooks/Binary Classification on COMPAS dataset.ipynb
romanlutz: You need to fix the notebook. There are (gated) tests that won't succeed unless the notebooks are up to date.

romanlutz:
> 
> 
> I'm getting a ` raise Expired("Credentials have expired due to inactivity.")` in the CI, which I don't think I can do much about.

I've been checking this. It's quite bizarre. I ran it myself through your PR thinking that you maybe lack permissions, but that wasn't it (same error). I ran it off master now in case something just broke, but that's not it either (succeeded).
It could be that it struggles to run this with PRs that are created from forks (as opposed to within the repo). I'm playing around with moving certain bits of the CI into GitHub Actions, so maybe I'll look into that. In the meanwhile don't concern yourself with that. If everything else passes I'm happy to complete the PR.

adrinjalali:
I'm still not sure what your workflow for merging PRs is. It'd be nice to make that transparent and explain in the contributing guide maybe. Do PRs need at least two "yes"s by maintainers, or is one enough? Or is that if the author is a maintainer, one other "yes" is enough, otherwise two?

adrinjalali:
@MiroDudik you happy with this now?

Issue #318, opened 2020-03-05
matplotlib 3.2.0 breaks linux tests


romanlutz:
https://dev.azure.com/responsibleai/fairlearn/_build/results?buildId=4508&view=logs&j=d0d954b5-f111-5dc4-4d76-03b6c9d0cf7e&t=2a79e97e-143e-5ca8-c651-c67f5c62e229

PR #319, opened 2020-03-05
use TKAgg only on MacOS

romanlutz:
addresses #318 

### DIFF
--- a/test/unit/postprocessing/test_plots.py
+++ b/test/unit/postprocessing/test_plots.py
@@ -3,7 +3,7 @@
 
 # On MacOS we need to use TKAgg before importing matplotlib.pyplot.
 # This used to work on all platforms until matplotlib 3.2.0 broke it on Linux.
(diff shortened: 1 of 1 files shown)


test/unit/postprocessing/test_plots.py
```diff
@@ -1,8 +1,13 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
-import matplotlib
-matplotlib.use('TkAgg')
+# On MacOS we need to use TKAgg before importing matplotlib.pyplot.
+# This used to work on all platforms until matplotlib 3.2.0 broke it on Linux.
+# Consider removing the OS-based if after a future release of matplotlib.
```
riedgar-ms: Can you file as issue and link in the comment. I'm really not a fan of code like this.

romanlutz: done #320

Issue #312, opened 2020-02-28
GridSearch incompatible with models such as LightGBM that rely on dataframe metadata


moprescu:
The `GridSearch` estimator removes the metadata from input dataframes via `_validate_and_reformat_input` which calls the following methods:
```Python 
X = check_array(X)
...
return pd.DataFrame(X) 
```
Estimators such as `LightGBM` rely on metadata for categorical columns (column type = `category`) which they process internally more efficiently than by user crafted dummies. The preprocessing function will erase the metadata causing `LightGBM` to process the columns as numerical rather than categorical.

Furthermore, execution of GridSearch fails if there is metadata because of the following two calls:
```
objective.load_data(X_train, y_train, **kwargs) # ln 197
...
current_estimator.fit(X, y_reduction, sample_weight=weights) # ln 237
```
The current estimator is trained on `X` which contains the metadata and the ` objective.gamma` function calls predict on `X_train` which doesn't contain the metadata, which leads to the following error:

```Python
# General imports
import numpy as np
import pandas as pd
import lightgbm as lgb
from fairlearn.reductions import GridSearch, EqualizedOdds

# Data
n = 100
X0 = np.random.normal(size=n)
X1 = np.random.choice([1, 2, 3], size=n)
Y = np.random.choice([0, 1], size=n)
A = np.random.choice([0, 1], size=n)
df = pd.DataFrame({"X0": X0, "X1": X1})
# Set X1 as categorical
df['X1'] = df['X1'].astype('category')

# Model
model = lgb.LGBMClassifier()

# Fairlearn
sweep = GridSearch(model,
                   constraints=EqualizedOdds())
sweep.fit(df, Y, sensitive_features=A)
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-2-e41612319198> in <module>
     22 sweep = GridSearch(model,
     23                    constraints=EqualizedOdds())
---> 24 sweep.fit(df, Y, sensitive_features=A)

~\AppData\Local\Continuum\anaconda3\lib\site-packages\fairlearn\reductions\_grid_search\grid_search.py in fit(self, X, y, **kwargs)
    242             nxt = GridSearchResult(current_estimator,
    243                                    lambda_vec,
--> 244                                    objective.gamma(predict_fct)[0],
    245                                    self.constraints.gamma(predict_fct),
    246                                    oracle_call_execution_time)

[...]

PR #321, opened 2020-03-05
Keep feature metadata through input validation

romanlutz:
So far we simply run scikit-learn input validation (`check_array` or `check_X_y`) and thus lose metadata, e.g. whether a column is categorical or not. This change helps in keeping this metadata and applying the data types back after scikit-learn input validation. I've added a test case to make sure this works out right.

Addresses #312


test/unit/test_input_validation.py
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+import numpy as np
+import pandas as pd
+from fairlearn._input_validation import _validate_and_reformat_input
+
+
+def test_input_column_metadata_retained():
```
riedgar-ms: Could we also have a test with @moprescu 's repro case? That would also have us test learners from somewhere other than scikit-learn

romanlutz: @adrinjalali  mentioned that we should move perf tests out of this repo, and I don't necessarily disagree (but it's a little bit of work). That would mean the tempeh dependency isn't required anymore. Lightgbm used to be part of tempeh (but isn't anymore I think due to some breaking change). 
The question is whether we really want to add dependencies on specific packages like lightgbm. Then what is the threshold? Do we add pytorch? What about tensorflow?
I think you see where I'm going with this. I'd much rather have the perf tests run all these things, and have the functional parts being checked here. What do you think?

riedgar-ms: It would be a dependency for the test, not fairlearn.

I do take your point about keeping testing focussed, but I'm not talking about a full battery of tests for each ML package. I just think that we should perhaps have a smoke test or two for each of them, in an effort to make sure that things like this don't happen.

Or if we're only going to check that as part of the perf tests, then those need to be inserted into the release.

romanlutz: requirements.txt has all dependencies, though, not just fairlearn's.

I actually WANT to have a full battery of tests for each ML package :-)

Anyway, I'll add some smoke tests like you suggested.

adrinjalali:
Another way to approach the issue, is to either not do input validation in meta estimators (which is what pipeline does in sklearn), or have a parameter in your input validation routine which tells it to skip certain validations steps. I would kinda prefer that to what we have in this PR right now I guess.

fairlearn/_input_validation.py
```diff
@@ -48,6 +48,16 @@ def _validate_and_reformat_input(X, y=None, expect_y=True, enforce_binary_sensit
         values in the `y` data; default False
     :type enforce_binary_labels: bool
     """
+    # scikit-learn's check_array or check_X_y lose the column types. Some estimators rely on this
```
MiroDudik: See #312. The bottom line: I don't think we need this--hopefully we can just rely on sklearn`s validation / canonicalization routines in our code. Our wrappers should never transform `X`--but some of them may need to get canonical versions of `y` and `sensitive_features`.

romanlutz: I've adjusted this accordingly. Let me know what you think @MiroDudik

test/unit/reductions/test_smoke.py
```diff
@@ -0,0 +1,35 @@
...
+
+from fairlearn.reductions import ExponentiatedGradient, GridSearch, DemographicParity, \
+    EqualizedOdds
+
+
+@pytest.mark.parametrize("Mitigator", [ExponentiatedGradient, GridSearch])
+@pytest.mark.parametrize("Constraints", [DemographicParity, EqualizedOdds])
+@pytest.mark.parametrize("Estimator", [LGBMClassifier, LogisticRegression, SVC,
+                                       DecisionTreeClassifier])
+def test_smoke(Mitigator, Constraints, Estimator):
```
riedgar-ms: Thanks!

romanlutz: MacOS of course doesn't just work with the pip install lightgbm, so I've excluded that case there. This specific behavior shouldn't be different on different OSes, so I think it's fine. I think at some point we'll have to expand this to other types (pytorch, tensorflow, xgboost, etc.) as well, so I'll leave the installation fixes for then....

test/unit/reductions/test_smoke.py
```diff
@@ -0,0 +1,42 @@
...
+from fairlearn.reductions import ExponentiatedGradient, GridSearch, DemographicParity, \
+    EqualizedOdds
+
+
+_ESTIMATORS = [LogisticRegression, SVC, DecisionTreeClassifier]
+
+if platform.system() != "Darwin":
+    # MacOS requires extra steps to install lightgbm properly, skipping for now
+    from lightgbm import LGBMClassifier
+    _ESTIMATORS.append(LGBMClassifier)
```
adrinjalali: You could also do `pytest.importorskip` in your test, to have it skipped if the env doesn't have the library. Looks nicer to me I guess.

fairlearn/_input_validation.py
```diff
@@ -47,6 +47,11 @@ def _validate_and_reformat_input(X, y=None, expect_y=True, enforce_binary_sensit
     :param enforce_binary_labels: if True raise exception if there are more than two distinct
         values in the `y` data; default False
     :type enforce_binary_labels: bool
+    :return: the validated and reformatted X, y, and sensitive_features; note that certain
+        estimators rely on metadata encoded in X which may be stripped during the reformatting
+        process, so mitigation methods should ideally use the input X instead of the returned X
+        for training estimators and leave potential reformatting of X to the estimator.
+    :rtype: (pandas.DataFrame, pandas.Series, pandas.Series)
```
adrinjalali: but now it returns the numpy ndarray, doesn't it?

romanlutz: No, it returns a dataframe as before.

test/unit/reductions/test_smoke.py
romanlutz: But then I can't parametrize the same way :-(

romanlutz: If you find a way to solve this in a more elegant way using parametrization I can make a follow-up PR

Issue #285, opened 2020-02-04
Automatic reviewer addition to PRs


romanlutz:
I've seen this in other repositories, e.g. https://github.com/pytorch/pytorch/blob/master/CODEOWNERS

and we've previously set up the CODEOWNERS file, so this should be possible.

@adrinjalali if you've seen a different way of handling this lmk!

adrinjalali:
yeah that's one approach which some projects take, and it works, especially when the project is kinda massive and it's hard for people to follow all conversations. For the time being, I'm not sure if it changes much in this repo though. That said, I don't see that as a bad process anyway.

PR #323, opened 2020-03-08
add auto-assign-reviewers action

romanlutz:
Adding a simple GitHub action to always assing me and Richard as reviewers. I'm planning to refine this in the future, ideally based on interest/specialization, but the current GitHub action doesn't have that functionality yet AFAICT. This PR also demonstrates a simple example of how to use GitHub actions in case we want to replace Azure Pipelines at some point.

addresses #285 


romanlutz:
Fantastic! It was even smart enough to make me the assignee and Richard the reviewer!

PR #325, opened 2020-03-10
Equal Opportunity implementation + sample notebooks

hannanabdul55:
## What's there in this PR? 
- Initial implementation of Equal Opportunity, which is a slightly modified version of Equalized Odds (specifically for `y=1` case only). 
- Sample notebooks for Demographic Parity constraint evaluation on Law School dataset. Check these notebooks [[1]](https://github.com/UMass-Responsible-AI/fairlearn/blob/master/notebooks/Exp%20Gradient%20for%20DP(adultl).ipynb) [[2]](https://github.com/UMass-Responsible-AI/fairlearn/blob/master/notebooks/Exp%20Gradient%20for%20DP(law%20school).ipynb) [[3]](https://github.com/UMass-Responsible-AI/fairlearn/blob/master/notebooks/Exp%20Gradient%20for%20DP.ipynb).


.gitignore
```diff
@@ -114,3 +114,10 @@ docbuild/
 
 # generated perf test scripts
 test/perf/scripts/**
+
+# idea specific stuff
+.idea/**/workspace.xml
```
riedgar-ms: Can we just ignore the entire `.idea` directory?

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -145,6 +157,49 @@ def load_data(self, X, y, **kwargs):
...
+                          **kwargs)
+
+class EqualOpportunity(ConditionalSelectionRate):
+    short_name = "EqualOpportunity"
+
+    def load_data(self, X, y, **kwargs):
+        """Load the specified data into the object."""
+        super().load_data(X, y,
+                          event=pd.Series(y).apply(lambda y: _LABEL + "=" + str(y)),
+                          filter_value=0,
```
riedgar-ms: One small tweak: can we make the 'good value' a parameter to the constructor or this method? That might fold into @MiroDudik 's API review, though

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -33,13 +34,21 @@ def default_objective(self):
...
+    def load_data(self, X, y, event=None, filter_value=None, **kwargs):
         """Load the specified data into this object.
 
         This adds a column `event` to the `tags` field.
         """
         super().load_data(X, y, **kwargs)
         self.tags[_EVENT] = event
+        if filter_value is not None:
+            self.dropped_tags = self.tags.where(self.tags['label'] == 0).dropna()
+            self.tags = self.tags.where(self.tags['label'] == 1).dropna()
```
riedgar-ms: Should these two equalities be checking against `filter_value` ?

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -5,6 +5,7 @@
 from .moment import ClassificationMoment
 from .moment import _GROUP_ID, _LABEL, _PREDICTION, _ALL, _EVENT, _SIGN
 from .error_rate import ErrorRate
+import numpy as np
```
riedgar-ms: Is this for `dropna()` ?

.gitignore
hannanabdul55: Yeah, sure I will ignore the idea folder completely.

fairlearn/reductions/_moments/conditional_selection_rate.py
hannanabdul55: Yes, this was the original intention of the flag, but I also need the 'bad value' then so that they can be filtered out. But I think there is a workaround for that which I can do with some pandas API.  I will try it out. Thank you!

fairlearn/reductions/_moments/conditional_selection_rate.py
hannanabdul55: Yes, that is correct. I will make that change too

fairlearn/reductions/_moments/conditional_selection_rate.py
hannanabdul55: No, this can be removed. This was originally added because of some NumPy array manipulation that I was trying for the DataFrame, which I am not using now.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -145,6 +157,49 @@ def load_data(self, X, y, **kwargs):
         """Load the specified data into the object."""
         super().load_data(X, y, event=_ALL, **kwargs)
 
+class EqualOpportunity(ConditionalSelectionRate):
```
romanlutz: As mentioned earlier in person, we may want to make this True Positive Parity instead.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -145,6 +157,49 @@ def load_data(self, X, y, **kwargs):
...
+    r"""Implementation of Equalized Odds as a moment.
```
romanlutz: Documentation needs update

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -145,6 +157,49 @@ def load_data(self, X, y, **kwargs):
...
+    the number of unique values for the sensitive feature, multiplied by
+    the number of unique values of the `Y` array, multiplied by two (for
+    the Lagrange multipliers for positive and negative constraints).
+
+    With these definitions, the :meth:`signed_weights` method
+    will calculate the costs according to Example 4 of
+    `Agarwal et al. (2018) <https://arxiv.org/abs/1803.02453>`_.
+    """
+
+    short_name = "EqualizedOdds"
```
romanlutz: please update the short_name as well

fairlearn/reductions/_moments/conditional_selection_rate.py
romanlutz: Wait, did you name both of them EqualOpportunity, i.e. did you rename EqualizedOdds as well?

fairlearn/reductions/_moments/conditional_selection_rate.py
romanlutz: unless this is meant to be Equalized Odds. The class name threw me off

fairlearn/reductions/_moments/conditional_selection_rate.py
romanlutz: unless this is indeed EO. The class name threw me off

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -145,6 +157,49 @@ def load_data(self, X, y, **kwargs):
...
+
+    def load_data(self, X, y, **kwargs):
+        """Load the specified data into the object."""
+        super().load_data(X, y,
+                          event=pd.Series(y).apply(lambda y: _LABEL + "=" + str(y)),
+                          **kwargs)
+
+class EqualOpportunity(ConditionalSelectionRate):
```
romanlutz: Can we add a short doc section like with EqualizedOdds? Should be fairly similar, and in fact you can refer to Equalized ODds as well

fairlearn/reductions/_moments/conditional_selection_rate.py
romanlutz: I agree with Richard here. If you know the "good value" then the "bad value" is simply the other one (using some sort of unique/set magic to determine that). This may come in handy if/when we decide to move to multi-class classification, because suddenly it's not one or the other, but one of n, and then specifying the "good value" is way easier than specifying n-1 "bad values".

notebooks/Binary Classification on COMPAS dataset.ipynb
```diff
@@ -19,9 +19,30 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 1,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
```
romanlutz: Before committing notebooks we always clear the output. That should reduce the number of changes to a minimum here :-)

test/unit/reductions/exponentiated_gradient/test_exponentiatedgradient_smoke.py
```diff
@@ -103,7 +109,7 @@ def test_smoke(self, testdata):
 
     def test_simple_fit_predict(self):
         estimator = LeastSquaresBinaryClassifierLearner()
-        constraints = DemographicParity()
+        constraints = EqualOpportunity()
```
romanlutz: You may add another test case, but please leave the existing ones in place :-)

fairlearn/reductions/_moments/conditional_selection_rate.py
hannanabdul55: Oops yeah, good catch sorry about this. I think I messed up a merge before raising this PR because of which it got added twice!

fairlearn/reductions/_moments/conditional_selection_rate.py
hannanabdul55: I have made this change. Looks cleaner than the current implementation for sure, but I don't think that would work for the multi-class classification.

fairlearn/reductions/_moments/conditional_selection_rate.py
hannanabdul55: Fixing all this in the next commit. 👍

test/unit/reductions/exponentiated_gradient/test_exponentiatedgradient_smoke.py
hannanabdul55: Sorry about this. Yes, I will do this.

fairlearn/reductions/_moments/conditional_selection_rate.py
hannanabdul55: Sorry about that. I think all this should be resolved now!

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -146,6 +159,43 @@ def load_data(self, X, y, **kwargs):
...
+    fraction of the samples corresponding to each unique value of `y = 1` in
+    the `Y` array.
+
+    The `index` MultiIndex will have a number of entries equal to
+    the number of unique values for the sensitive feature, multiplied by
+    the number of unique values of the `Y` array, multiplied by two (for
+    the Lagrange multipliers for positive and negative constraints).
+
+    With these definitions, the :meth:`signed_weights` method
+    will calculate the costs like how it is calculated in Example 4 of
```
riedgar-ms: Minor thing but how about "will calculate the costs similarly to Example 4 ….. but only for all `y=1` samples "

fairlearn/reductions/_moments/conditional_selection_rate.py
romanlutz: There are likely more things that won't work for multi-class, but setting this up the right way to start with should prevent some future pains ;-)
Still says filter_value=0 above, though?

fairlearn/reductions/_moments/conditional_selection_rate.py
romanlutz: No worries!

.gitignore
```diff
@@ -114,3 +114,4 @@ docbuild/
 
 # generated perf test scripts
 test/perf/scripts/**
+.idea
```
romanlutz: nit: can you make this a new section with a comment on what .idea files are? I presume they're related to your IDE in some way?

hannanabdul55: Yes, it is related to the IDE I use. should I remove it altogether and just keep it in my local instead?

fairlearn/reductions/_moments/conditional_selection_rate.py
hannanabdul55: Yes, I have written that in line number 186 below. I can rephrase it to this though!

fairlearn/reductions/_moments/conditional_selection_rate.py
hannanabdul55: Yes, this is the value that has to be filtered. Oh sorry, I think I misinterpreted the earlier comments then, Should I rename this parameter to `good_value`? I can do that!

riedgar-ms: If we're going to call this true_positive_parity or similar, then `positive_value` might be better (also sounds less like a value judgement)

.gitignore
riedgar-ms: I imagine we will acquire these over time anyway. I think a comment line above to note that this is to ignore files created by IDEA would be fine. We already have things for VS and VS Code, after all.

fairlearn/reductions/_moments/conditional_selection_rate.py
hannanabdul55: Done 👍

.gitignore
hannanabdul55: Done 👍

fairlearn/reductions/_moments/conditional_selection_rate.py
hannanabdul55: Yeah, we could call this `positive_value`. I thought `filter_value` is generic and could be used for positive as well as, possibly, negative values too. Anyways, I can change it to `positive_value`. Let me know if I should.

fairlearn/reductions/_moments/conditional_selection_rate.py
MiroDudik: Actually, let's rename `EqualOpportunity` -> `TruePositiveRateDifference` to match the [metric API](https://github.com/fairlearn/fairlearn-proposals/blob/master/api/METRICS.md). The same needs to propagate to the above and to the doc string below. There you can mention that this criterion is also referred to as "equal opportunity".

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -33,13 +33,21 @@ def default_objective(self):
         """Return the default objective for moments of this kind."""
         return ErrorRate()
 
-    def load_data(self, X, y, event=None, **kwargs):
+    def load_data(self, X, y, event=None, filter_value=None, **kwargs):
         """Load the specified data into this object.
```
MiroDudik: I think that the following proposal will be better:
* let's get rid of the `filter_value` field
* if `event` contains `numpy.nan` those values should be excluded from conditioning

This will be consistent with the `pandas.DataFrame.groupby` and may also lead to a more concise code; e.g.:
```
>>> import pandas as pd
>>> import numpy as np
>>> df = pd.DataFrame({"event": ['a', 'a', np.nan, 'b'], "val": [4, 3, 2, 1]})
>>> df.groupby("event").size()
event
a    2
b    1
dtype: int64
```

fairlearn/reductions/_moments/conditional_selection_rate.py
MiroDudik: See my comment above. Simply enter `numpy.nan` into `event` where y=0 and keep `_LABEL + "=" + str(y)` where y=1.

notebooks/Exp Gradient for DP(adultl).ipynb
```diff
@@ -0,0 +1,192 @@
+{
```
MiroDudik: Make this a separate PR. The notebook should have more than just the code--it needs to provide the motivation, e.g., what does it try to accomplish? And it needs to explain the steps in it more carefully.

notebooks/Exp Gradient for DP(law school).ipynb
```diff
@@ -0,0 +1,249 @@
+{
```
MiroDudik: See the comment for the other notebook.

notebooks/Exp Gradient for DP.ipynb
```diff
@@ -0,0 +1,248 @@
+{
```
MiroDudik: see the comment for the other notebook.

fairlearn/reductions/_moments/conditional_selection_rate.py
romanlutz: So where do you set it to np.nan?

romanlutz:
Closing since #343 replaces this.

Issue #255, opened 2020-01-22
Reductions - Unification and Simplification


riedgar-ms:
Currently `ExponentiatedGradient` and `GridSearch` implement their own 'result' types. However, there is a lot of commonality, which should be separated into an `OracleCall` object, which stores a trained model, and the associated `lambda_vec`. The goal would be to be able to take an `OracleCall` from the ensemble and expand a grid around it for `GridSearch`.

Also, it does not make sense to have an `ExponentiatedGradientResult` property on the `ExponentiatedGradient` object. There is only one result for that algorithm, so the necessary bits and pieces should be directly on the `ExponentiatedGradient` class.

MiroDudik:
In particular, the fitted properties should be stored in the estimator object and follow the naming convention of
https://scikit-learn.org/stable/developers/develop.html

adrinjalali:
Also note that whenever the result is a complicated object, we usually return a dict or a `Bunch` object in sklearn.

romanlutz:
expgrad result frames everything in terms of final result:
- best classifier
- best gap
- classifiers (from each iteration)
- weights
- last t
- best t
- number of oracle calls
- oracle calls execution time (list containing the times for all calls)

expgrad keeps the result object around through `self._expgrad_result`

grid search frames everything in terms of results per "grid point" since it's not sequential:
- predictor
- lambda vec
- objective
- gamma
- oracle call execution time (for the individual call)

All the grid search result objects go into `self._all_results` which is a list.


Design considerations:
- interoperability: if both use the same way to store results then we can take the results of a expgrad run over an iteration and plug it into a grid search run.
- scaling grid search out might mean we don't have all the individual "results" locally


As Miro pointed out a couple of days ago result objects as such may not be suitable. Maybe it's better to just have everything as a property on the resulting predictor? In this case we'd have many instances of the various properties, such as lambda vec, so we'd probably need a list or dictionary. For interoperability it would probably easier to have everything grouped together and "plug" it into the other method that way, as opposed to having to find and use each individual property. The same thought somewhat more concrete:
```
expgrad = ExponentiatedGradient(...)
expgrad.fit(....)
grid_search = GridSearch(expgrad.result, ...)
grid_search.fit(....)
```
vs.
```
expgrad = ExponentiatedGradient(...)
expgrad.fit(....)
grid_search = GridSearch(expgrad.property1, expgrad.property2, expgrad.property3, expgrad.property4, ...)
grid_search.fit(....)
```


[...]

PR #327, opened 2020-03-13
Change reductions inheritance to use sklearn BaseEstimator and mixins

romanlutz:
This is a first step in addressing #255 .


CHANGES.md
```diff
@@ -10,6 +10,11 @@
     an estimator through the `estimator` parameter. To use a pre-fitted
     estimator, pass `prefit=True`.
 * Rename arguments of `create_group_metric_set()` to match the dashboard
+* Remove `Reduction` base class for reductions methods and replace it with
+  `sklearn.base.BaseEstimator`, `sklearn.base.ClassifierMixin`, and
```
riedgar-ms: What functionality do these add, and should there be a smoke test to verify that such functionality really is present?

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -22,7 +22,7 @@ def _mean_pred(X, hs, weights):
     return pred[weights.index].dot(weights)
 
 
-class ExponentiatedGradient(Reduction):
+class ExponentiatedGradient(BaseEstimator, ClassifierMixin):
     """An Estimator which implements the exponentiated gradient approach to reductions.
```
MiroDudik: This should be `BaseEstimator` and `MetaEstimatorMixin`. I don't mind having `ClassifierMixin` for now, but it will go away once we support `ExponentiatedGradient` with regression.

CHANGES.md
adrinjalali: You can eventually pass the sklearn tests as whone [here](https://scikit-learn.org/stable/auto_examples/release_highlights/plot_release_highlights_0_22_0.html#checking-scikit-learn-compatibility-of-an-estimator).

``` python
from sklearn.linear_model import LogisticRegression
from sklearn.tree import DecisionTreeRegressor
from sklearn.utils.estimator_checks import parametrize_with_checks


@parametrize_with_checks([LogisticRegression, DecisionTreeRegressor])
def test_sklearn_compatible_estimator(estimator, check):
    check(estimator)
```

fairlearn/reductions/_grid_search/grid_search.py
```diff
@@ -79,7 +79,7 @@ def accumulate_integer_grid(self, index, max_val):
                 self.accumulate_integer_grid(index + 1, max_val - abs(current_value))   # noqa: E501
 
 
-class GridSearch(Reduction):
+class GridSearch(BaseEstimator, ClassifierMixin, RegressorMixin):
     """Estimator to perform a grid search given a blackbox estimator algorithm.
```
MiroDudik: This should be `BaseEstimator` and `MetaEstimatorMixin`.

As with `ExponentiatedGradient`, it would be to set `_estimator_type` and `score` functionality according to our base estimator... but happy to leave that for a separate PR.

CHANGES.md
romanlutz: That would require a few more changes to work, right?

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
romanlutz: I don't currently see `_estimator_type` and `score`, so I will leave them for another PR. Addressed your other comments, though!

CHANGES.md
adrinjalali: Yes, I'm working on adding the test for ThresholdOptimizer, and once that's done, I guess we'd all have a better idea of what needs to be changed for it to work. In the mean time, this may help understanding what needs to be done:

https://fosdem.org/2020/schedule/event/python2020_scikit_learn_estimator/

PR #332, opened 2020-03-16
WIP Error ratio Implementation in conditional selection rate

parul100495:
**What's there in this PR?**

Implementation of **error ratio** constraint which included the following:

- **Ratio** - Added ratio parameter in initialisation for the relaxation fraction allowed in the constraint. Defaults to 1 for all constraints (compatible with current moments).

- **Utility** - Added utility matrix which corresponds to `g(X,A,Y,h(x)=0)` and `g(X,A,Y,h(x)=1)` from the paper by Agarwal et al ([A reduction approach to fair classification](https://arxiv.org/pdf/1803.02453.pdf) ). Defaults to h(x) for all constraints (compatible with current moments).

**Test**

- Passed all unit tests in the repo, specifically, passed the smoke test `test_smoke` in `test_exponentiatedgradient_smoke.py`.

**TODO**

- Added unit tests for Error Rate Ratio, Error Rate Difference and ratio aspect of EqualizedOdds and DemographicParity.


fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -105,9 +108,13 @@ def signed_weights(self, lambda_vec):
         :param lambda_vec: The vector of Lagrange multipliers indexed by `index`
         :type lambda_vec: :class:`pandas:pandas.Series`
         """
-        lambda_signed = lambda_vec["+"] - lambda_vec["-"]
-        adjust = lambda_signed.sum(level=_EVENT) / self.prob_event \
-            - lambda_signed / self.prob_group_event
+        adjust = self.multiplier[_MULTIPLIER].mul(((lambda_vec["+"] -
```
riedgar-ms: Is there any way this can be broken apart a bit? The previous version was already painfully long.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -173,8 +184,52 @@ class EqualizedOdds(ConditionalSelectionRate):
...
+    The `prob_event` :class:`pandas:pandas.DataFrame` will record the
+    fraction of the samples corresponding to each unique value in
+    the `Y` array.
+
+    The `index` MultiIndex will have a number of entries equal to
+    the number of unique values for the sensitive feature, multiplied by
+    the number of unique values of the `Y` array, multiplied by two (for
+    the Lagrange multipliers for positive and negative constraints).
+
+    With these definitions, the :math:`signed_weights` method
```
riedgar-ms: Might want to expand on this section a bit, since the ratio approach isn't in the paper.

fairlearn/reductions/_moments/conditional_selection_rate.py
parul100495: Sure, I'll format it to make it more readable.

fairlearn/reductions/_moments/conditional_selection_rate.py
parul100495: I'll add some documentation which explains the ratio approach.

romanlutz:
@parul100495 I've marked the PR as WIP which stands for "work in progress". As long as there are TODOs that indicates to us that we can review and provide suggestions, but not complete the PR or expect everything to be perfect. When you're done with the TODOs you should remove the WIP to indicate to us that it's ready for a full review and potentially completion. Thanks for your contribution!

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -48,7 +51,7 @@ def load_data(self, X, y, event=None, **kwargs):
                            names=[_SIGN, _EVENT, _GROUP_ID])
         self.index = signed.index
         self.default_objective_lambda_vec = None
-
+        self.multiplier = self.tags.groupby([_EVENT, _GROUP_ID]).mean()
```
romanlutz: I'm struggling to understand what the multiplier is for, so I assume it's unclear to others as well (especially people who aren't as involved). Can we add a comment explaining what multiplier's purpose is?

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -173,8 +182,57 @@ class EqualizedOdds(ConditionalSelectionRate):
...
+        self.ratio = ratio
+
+    def load_data(self, X, y, **kwargs):
+        """Load the specified data into the object."""
+        super().load_data(X, y,
+                          event=pd.Series(y).apply(lambda y: _LABEL + "=" + str(y)),
+                          **kwargs)
+
+
+class ErrorRatio(ConditionalSelectionRate):
```
romanlutz: @MiroDudik theoretically this is the ratio of error rates, right? Should it be ErrorRateRatio, and the other one ErrorRateDifference?

fairlearn/reductions/_moments/conditional_selection_rate.py
romanlutz: Looks much better than before actually. We usually aim at single lines for math to make it as readable as possible, even if that means that we have to assign more variables.

fairlearn/reductions/_moments/conditional_selection_rate.py
romanlutz: I believe you wanted to revise this once more to remove the privileged/unprivileged formulation in favor of the "single sensitive feature vs. all" approach, so I'll hold off on reviewing the documentation. Thanks for putting in the work to write the doc, though! This will come in handy.

parul100495: I'll update the documentation and get back.

fairlearn/reductions/_moments/conditional_selection_rate.py
parul100495: Sure. I think this is confusing. I'll write the documentation to explain this.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -105,9 +116,11 @@ def signed_weights(self, lambda_vec):
         :param lambda_vec: The vector of Lagrange multipliers indexed by `index`
         :type lambda_vec: :class:`pandas:pandas.Series`
         """
-        lambda_signed = lambda_vec["+"] - lambda_vec["-"]
-        adjust = lambda_signed.sum(level=_EVENT) / self.prob_event \
-            - lambda_signed / self.prob_group_event
+        lambda_event = (lambda_vec["+"] - self.ratio * lambda_vec["-"]) \
```
romanlutz: These already look correct / up-to-date with the updates to the derivations I discussed with @MiroDudik today. Looks like you're way ahead of me :-)

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -173,8 +190,59 @@ class EqualizedOdds(ConditionalSelectionRate):
...
+        super().load_data(X, y,
+                          event=pd.Series(y).apply(lambda y: _LABEL + "=" + str(y)),
+                          **kwargs)
+
+
+class ErrorRatio(ConditionalSelectionRate):
+    r"""Implementation of Error Ratio as a moment.
+
+    Measures the ratio in errors per attribute by overall error.
+    The 2-sided version of error ratio can be written as -
```
romanlutz: nit: please remove the dash at the end of the line. It may confuse people into thinking that it's part of the math that follows

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -173,8 +190,59 @@ class EqualizedOdds(ConditionalSelectionRate):
...
+    the number of unique values for the sensitive feature, multiplied by
+    the number of unique values of the `Y` array, multiplied by two (for
+    the Lagrange multipliers for positive and negative constraints).
+
+    With these definitions, the :math:`signed_weights` method
+    will calculate the costs according to Example 4 of
+    `Agarwal et al. (2018) <https://arxiv.org/abs/1803.02453>`_.
+
+    The `ratio` defines the amount of relaxation that is allowed for the
+    constraint. The value varies between 0-1. The ratio of 1 means, the
```
romanlutz: nit: varies between 0 and 1.

This should probably have some validation as well, i.e. raise exception if ratio <= 0 or ratio > 1

notebooks/Test-DP-ratio-on-adult-dataset.ipynb
```diff
@@ -0,0 +1,187 @@
+{
```
romanlutz: let's keep notebooks separate from the PR for now

fairlearn/reductions/_moments/conditional_selection_rate.py
romanlutz: Still struggling to understand what it looks like. Is it the mean of y-values (labels) of each group (as defined through event + group ID)? That would be 0/1. Overriding this for error rate would mean -1/1. Sorry, I feel like I'm missing something. I'll read your derivations again, but it needs additional description (at least for error rate) saying why we want -1/1 and what that does. It refers to Example 4 below, but example 4 is for equalized odds.

fairlearn/reductions/_moments/conditional_selection_rate.py
MiroDudik: Agreed. Let's rename `ErrorRatio` -> `ErrorRateRatio`.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -33,13 +34,24 @@ def default_objective(self):
         """Return the default objective for moments of this kind."""
         return ErrorRate()
 
-    def load_data(self, X, y, event=None, **kwargs):
+    def load_data(self, X, y, event=None, multiplier=None, **kwargs):
         """Load the specified data into this object.
```
MiroDudik: I'm thinking we should rename `multiplier` -> `<something>_difference` since it corresponds to `g(.,1)-g(.,0)`. Some options:
* `g_difference`
* `target_difference` (since `g` is a target with respect we define the disparity)
* `utility_difference` (I like this the best, but the approach also applies if `g` is negative utility, e.g., error)
@romanlutz , @parul100495 , do you have thoughts on this?

romanlutz: You mean negative utility is a weird thing to specify? I think regardless we'll need to specify that it's the difference between `g(...,1)` and `g(...,0)`. Utility seems intuitive to me as a word for this.

What I don't fully understand right now is whether this kind of formulation will prevent us from doing multi-class classification. @MiroDudik  you may have a better idea about that. Shouldn't be a blocker for this, though.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -173,8 +190,59 @@ class EqualizedOdds(ConditionalSelectionRate):
...
+    def __init__(self, ratio=1.0):
+        """Intialise with the ratio value."""
+        super(ErrorRatio, self).__init__()
+        self.ratio = ratio
+
     def load_data(self, X, y, **kwargs):
         """Load the specified data into the object."""
         super().load_data(X, y,
                           event=pd.Series(y).apply(lambda y: _LABEL + "=" + str(y)),
+                          multiplier=pd.Series(y).apply(lambda y: 1 - 2*y),
```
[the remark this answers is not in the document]
romanlutz: @MiroDudik this is definitely an obstacle for this. I'll try to prioritize this bug in the existing logic. Not entirely sure how to handle it in this PR until then, other than holding off.

fairlearn/reductions/_moments/conditional_selection_rate.py
parul100495: I get where the confusion came from - it was kind of a hack the way I implemented it before. I have simplified the code now. Should be more readable and easy to understand. Also, I'm fixing the documentation to make it more transparent.

fairlearn/reductions/_moments/conditional_selection_rate.py
MiroDudik: Actually, I think that we need to change the way we do things. Instead of providing just the difference `g(.,1)-g(.,0)`, the user should provide full columns `g(.,0)` and `g(.,1)` maybe in the form of n-by-2 matrix, which could be called `g` or  `utilities` or even `costs`. This would clearly generalize well to multi-class settings.

A more important issue is that if we allow a generic `event`, I think we need to rework the derivation of the importance weights, because C^0_i and C^1_i will not line up so nicely.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -105,12 +116,15 @@ def signed_weights(self, lambda_vec):
...
-            - lambda_signed / self.prob_group_event
+        lambda_event = (lambda_vec["+"] - self.ratio * lambda_vec["-"]) \
+            .sum(level=_EVENT) / self.prob_event
+        lambda_group_event = (self.ratio * lambda_vec["+"] - lambda_vec["-"]) / \
+            self.prob_group_event
+        adjust = lambda_event - lambda_group_event
         signed_weights = self.tags.apply(
             lambda row: adjust[row[_EVENT], row[_GROUP_ID]], axis=1
         )
+        signed_weights = self.multiplier.mul(signed_weights)
```
MiroDudik: This is incorrect for generic `event`--I think it's only correct if `g` is independent of `event`. I think we should rework the derivation, so we can handle arbitrary `g` and `event`.

fairlearn/reductions/_moments/conditional_selection_rate.py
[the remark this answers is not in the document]
MiroDudik: so far we are in many ways restricting to y=0/1. i think that it's okay if we do for the time being.

fairlearn/reductions/_moments/conditional_selection_rate.py
romanlutz: Sounds good. I'll resolve this comment and will take a look at the current state :-)

fairlearn/reductions/_moments/conditional_selection_rate.py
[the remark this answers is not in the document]
romanlutz: Okay, then I'll put a note into my bug #339  to make sure that we don't fixing this part.

fairlearn/reductions/_moments/conditional_selection_rate.py
romanlutz: Do you think this is all in scope for this PR?

fairlearn/reductions/_moments/conditional_selection_rate.py
romanlutz: So for existing moments we always have `g` independent of `event`, and that's why this never came up?

Do you think this is in scope for this PR?

fairlearn/reductions/_moments/conditional_selection_rate.py
parul100495: This is done.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -141,6 +156,12 @@ class DemographicParity(ConditionalSelectionRate):
 
     short_name = "DemographicParity"
 
+    def __init__(self, ratio=1.0):
+        super(DemographicParity, self).__init__()
```
romanlutz: Is this common to all subclasses of ConditionalSelectionRate? If so, just move it to the parent class

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -173,8 +194,63 @@ class EqualizedOdds(ConditionalSelectionRate):
...
+    constraint. The value varies between 0 and 1. The ratio of 1 means,
+    the constraint is given no relaxation and thus, the constraint tries to
+    evaluate for
+    error(A=a) / total_error = 1
+    """
+
+    short_name = "ErrorRateRatio"
+
+    def __init__(self, ratio=1.0):
+        """Intialise with the ratio value."""
```
romanlutz: nit: initialize ;-)

test/unit/reductions/exponentiated_gradient/test_exponentiatedgradient_smoke.py
```diff
@@ -69,17 +69,46 @@ def setup_method(self, method):
                         "best_gap": 0.000000, "last_t": 5,
                         "best_t": 5, "disp": 0.005000,
                         "error": 0.442883, "n_oracle_calls": 19,
-                        "n_classifiers": 6}]
+                        "n_classifiers": 6},
+                       {"cons_class": ErrorRateRatio, "eps": 0.1,
+                        "best_gap": 1.1102230246251565e-16, "last_t": 5,
```
romanlutz: shouldn't the gap reach 0? I see that it's super close, just wondering

fairlearn/reductions/_moments/conditional_selection_rate.py
MiroDudik: I think so. This should only add a couple of lines in `signed_weights` and in `gamma`. I just noticed that `gamma` is currently incorrect anyway, as it is still calculating the values of constraint violation for when `g(.,0)=0` and `g(.,1)`.

fairlearn/reductions/_moments/conditional_selection_rate.py
MiroDudik: This is not quite right--we have a hidden dependence on `self.ratio` which is being initialized in subclasses of `ConditionalSelectionRate`. Move it to the constructor of `ConditionalSelectionRate` as @romanlutz suggests below.

test/unit/reductions/exponentiated_gradient/test_exponentiatedgradient_smoke.py
MiroDudik: let's get back to this after `gamma` is fixed.

fairlearn/reductions/_moments/conditional_selection_rate.py
parul100495: Yes, it is common. Will move this to the parent class.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -182,3 +219,45 @@ def load_data(self, X, y, **kwargs):
...
+    error(A=a) / total_error = 1
+    """
+
+    short_name = "ErrorRateRatio"
+
+    def load_data(self, X, y, **kwargs):
+        """Load the specified data into the object."""
+        super().load_data(X, y,
+                          event=_ALL,
+                          utility=pd.Series({_POSITIVE_UTILITY: (1-y), _NEGATIVE_UTILITY: y}),
```
romanlutz, line 262: Series is a one-dimensional datastructure, so perhaps DataFrame is more appropriate?

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -29,17 +34,39 @@ class ConditionalSelectionRate(ClassificationMoment):
...
 
-    def load_data(self, X, y, event=None, **kwargs):
+    def load_data(self, X, y, event=None, utility=None, **kwargs):
         """Load the specified data into this object.
 
         This adds a column `event` to the `tags` field.
+
+        The `utility` is a matrix which correspond to g(X,A,Y,h(X)) as mentioned
+        in the paper `Agarwal et al. (2018) <https://arxiv.org/abs/1803.02453>`
+        The `utility` defaults to h(X), ie. [1 in R^n, 0 in R^n]
```
romanlutz, line 55: I'm unsure what `1 in R^n` means. Are you trying to say that the utility is [1, 0] for every sample?

Also, nit: i.e. instead of ie.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -29,17 +34,39 @@ class ConditionalSelectionRate(ClassificationMoment):
...
+        The `_POSITIVE_UTILITY` is G^1 and the `_NEGATIVE_UTILITY` is G^0.
+        Assumes that binary class of 0/1.
```
romanlutz, line 57: nit: "Assumes binary classification with labels 0/1"
Would that be better?

fairlearn/reductions/_moments/conditional_selection_rate.py
romanlutz: same earlier

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -72,24 +98,31 @@ def load_data(self, X, y, event=None, **kwargs):
...
+        utility_diff = self.utility[_POSITIVE_UTILITY] - self.utility[_NEGATIVE_UTILITY]
+        pred = utility_diff * predictor(self.X) + self.utility[_NEGATIVE_UTILITY]
         self.tags[_PREDICTION] = pred
         expect_event = self.tags.groupby(_EVENT).mean()
         expect_group_event = self.tags.groupby(
             [_EVENT, _GROUP_ID]).mean()
-        expect_group_event[_DIFF] = expect_group_event[_PREDICTION] - expect_event[_PREDICTION]
-        g_unsigned = expect_group_event[_DIFF]
-        g_signed = pd.concat([g_unsigned, -g_unsigned],
+        expect_group_event[_UPPER_BOUND_DIFF] = self.ratio * \
```
romanlutz, line 107: Can we get these on fewer lines? I think you have 99 characters available.

parul100495: I didn't realise. I'll refactor this.

fairlearn/reductions/_moments/conditional_selection_rate.py
parul100495: Yes. I wanted to say that it has n rows (for every sample).

fairlearn/reductions/_moments/conditional_selection_rate.py
parul100495: I'll change it to a DataFrame.

fairlearn/reductions/_moments/conditional_selection_rate.py
romanlutz: If you run `flake8` from the repo root directory you should get a list of all issues like lines that are too long. I suggest ignoring such things by default and running flake8 before committing :-)

test/unit/reductions/exponentiated_gradient/test_exponentiatedgradient_smoke.py
parul100495: It checks for precision till 6 decimal places in tests. Fixed this to 0.000000
Fixed `gamma` too.

parul100495:
Closing this in favour of #364

PR #333, opened 2020-03-16
Rework creation of group metric sets

riedgar-ms:
Make `_create_group_metric_set()` private. Also, change the arguments and innards to require dictionaries for the predictions and sensitive features. This will require users to name each, which will nudge them towards not confusing themselves.

This is a breaking change, but the routine is unlikely to have users right now.


CHANGES.md
```diff
@@ -9,7 +9,8 @@
   * `ThresholdOptmizer` is now a scikit-learn meta-estimator, and accepts
     an estimator through the `estimator` parameter. To use a pre-fitted
     estimator, pass `prefit=True`.
-* Rename arguments of `create_group_metric_set()` to match the dashboard
+* Add `create_dashboard_dictionary()` for creating the Dashboard cache layer
+* Make `create_group_metric_set()` private and change arguments
```
romanlutz: Does that break anyone who's been using it before?

riedgar-ms: Yes, although I somehow doubt that the userbase is larger than "me."

test/unit/metrics/test_dashboard_dictionary.py
```diff
@@ -0,0 +1,67 @@
...
+    assert actual['schemaVersion'] == 0
+
+
+def test_create_dashboard_dictionary_multiple_models_multiple_sensitive_features():
+    y_true = [0, 1, 0, 1, 1, 1, 1, 1, 0, 1, 0]
+
+    y_p1 = [0, 1, 1, 1, 0, 0, 0, 0, 1, 1, 1]
+    y_p2 = [1, 1, 1, 0, 0, 0, 1, 1, 1, 1, 1]
+    y_p3 = [0, 0, 0, 0, 0, 0, 1, 1, 1, 1, 0]
+    m_1 = "M1"
```
romanlutz: What if my keys are of a type other than string? Does that still work? [in python, probably, but I know you're looking at a e2e flow]

riedgar-ms: Which keys do you mean? The `y_pred` or the model name?

CHANGES.md
romanlutz: Fair. If we make such changes in the future we should ideally keep it around for a release longer and issue a warning message indicating the deprecation and impending removal. Maybe that's a little over the top for this one, but just throwing this out there.

test/unit/metrics/test_dashboard_dictionary.py
romanlutz: model name
I essentially tried to confirm whether you considered that keys in a dictionary could be something other than string, and whether that works with UI code when transformed into JSON etc.

riedgar-ms: I don't do any checking. I could check that it's a string... or force it to one. But if users wanted type checking, would they be using Python?

CHANGES.md
riedgar-ms: Agree in general - hence our elaborate migration guide from 0.2.

In this specific case, this functionality is only useful for generating a JSON of the Dashboard cache layer, and right now, you can't even load it back into the dashboard. `_create_group_metric_set()` is only of use for AzureML, and we're still in exploratory stages for that.

fairlearn/metrics/_dashboard_dictionary.py
```diff
@@ -0,0 +1,40 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+from ._group_metric_set import create_group_metric_set
+
+
+def create_dashboard_dictionary(
+        model_type,
+        y_true,
+        models,
```
riedgar-ms: Will be changing to `prediction_type` and `predictions` - @MiroDudik pointed out that 'model' tends to have a rather specific meaning in the ML community, and this usage didn't match that.

test/unit/metrics/test_dashboard_dictionary.py
romanlutz: Sorry, I'm not expressing myself well here. I just wanted to ask whether there's any way a user could do something unintentionally wrong and pass a non-string type that would break the visualizations because it expects a string. I'm probably being paranoid :-)

riedgar-ms: That would ultimately depend on how JS deserialised the JSON file. Since I believe the JS approach to type checking is even more relaxed than Python's, I wouldn't want to hazard a guess at what would happen.

fairlearn/metrics/_group_metric_set.py
```diff
@@ -32,15 +27,12 @@
...
+_DASHBOARD_DICTIONARY = 'dashboardDictionary'
 _VERSION = 'schemaVersion'
 
-_UNSUPPORTED_MODEL_TYPE = "The specified model_type of '{0}' is not supported"
-_DICT_TOO_MANY_Y_PRED = 'Too many y_pred values in dictionary'
-
 BINARY_CLASSIFICATION = 'binary_classification'
 REGRESSION = 'regression'
-_allowed_model_types = frozenset([BINARY_CLASSIFICATION, REGRESSION])
+_allowed_prediction_types = frozenset([BINARY_CLASSIFICATION, REGRESSION])
```
MiroDudik: We should also have `PROBABILISTIC_BINARY_CLASSIFICATION`.

fairlearn/metrics/_group_metric_set.py
```diff
@@ -93,72 +85,82 @@
...
+    preds = []
+    for model_name in sorted(predictions):
+        names.append(model_name)
+        y_p = _convert_to_ndarray_and_squeeze(predictions[model_name])
+        preds.append(y_p.tolist())
+    return names, preds
 
+
+def create_group_metric_set(y_true,
+                            predictions,
```
MiroDudik: let's rename to `_create_group_metric_set` until we stabilize the API.

riedgar-ms: Done

fairlearn/metrics/_group_metric_set.py
riedgar-ms: Yes, but if we're keeping this private pending a review of the API and generated JSON, adding that now is just more stuff which will potentially need a substantial update later.

CHANGES.md
```diff
@@ -9,7 +9,9 @@
   * `ThresholdOptmizer` is now a scikit-learn meta-estimator, and accepts
     an estimator through the `estimator` parameter. To use a pre-fitted
     estimator, pass `prefit=True`.
-* Rename arguments of `create_group_metric_set()` to match the dashboard
+* Change the arguments to `create_group_metric_set()`, so that it requires
+  dictionaries for the predictions and sensitive features. This is a
```
MiroDudik: Please flag that we are changing the name to `_create_group_metric_set` until the API is a bit more stable.

riedgar-ms: Ooops. Thanks for the reminder. Fixed.

fairlearn/metrics/_group_metric_set.py
```diff
@@ -93,72 +86,85 @@
...
-                            y_true,
-                            y_preds,
-                            sensitive_features,
-                            model_titles=None,
-                            sensitive_feature_names=None,
-                            extra_metrics=None):
-    """Create a dictionary matching the Dashboard's cache."""
-    if extra_metrics is not None:
-        raise NotImplementedError("No support for extra_metrics yet")
+def _process_feature_to_integers(feature):
```
romanlutz: This reminds me of #339 a lot. I don't know if you'd consider the examples @adrinjalali  posted there as simpler or better, but we can always consolidate such logic later on if need be.

fairlearn/metrics/_group_metric_set.py
```diff
@@ -93,72 +86,85 @@
...
+    names = []
+    preds = []
+    for model_name in sorted(predictions):
+        names.append(model_name)
+        y_p = _convert_to_ndarray_and_squeeze(predictions[model_name])
+        preds.append(y_p.tolist())
+    return names, preds
 
+
+def _create_group_metric_set(y_true,
```
romanlutz: I think we need some sort of description of what the "dashboard's cache" expects. I suppose the doc here would be a great place for this.

fairlearn/metrics/_group_metric_set.py
```diff
@@ -12,12 +12,7 @@
 from . import group_miss_rate, group_precision_score, group_r2_score
 from . import group_recall_score, group_roc_auc_score, group_root_mean_squared_error
 from . import group_selection_rate, group_specificity_score, group_zero_one_loss
-
-_GROUP_NAMES_MSG = "The sensitive_feature_names property must be a list of strings"
-_METRICS_KEYS_MSG = "Keys for metrics dictionary must be strings"
-_METRICS_VALUES_MSG = "Values for metrics dictionary must be of type GroupMetricResult"
-
-_ARRAYS_NOT_SAME_LENGTH = "Lengths of y_true, y_pred and sensitive_features must match"
+from ._input_manipulations import _convert_to_ndarray_and_squeeze
```
romanlutz: aren't the other methods except for the last one also input manipulations? Should they live there?

fairlearn/metrics/_group_metric_set.py
```diff
@@ -93,72 +86,85 @@
...
+        names.append(model_name)
+        y_p = _convert_to_ndarray_and_squeeze(predictions[model_name])
+        preds.append(y_p.tolist())
+    return names, preds
 
+
+def _create_group_metric_set(y_true,
+                             predictions,
+                             sensitive_features,
+                             prediction_type):
```
romanlutz: Does this function work for all data types? DataFrame, ndarray, Series, list ?

fairlearn/metrics/_group_metric_set.py
riedgar-ms: Good point about `LabelEncoder` - have revised this to match

fairlearn/metrics/_group_metric_set.py
riedgar-ms: That's the doc @MiroDudik  mentions above. @rihorn2 also has a sample. But that's all we've really got for now, beyond this routine and its tests.

fairlearn/metrics/_group_metric_set.py
riedgar-ms: The tests are parameterised for that....

fairlearn/metrics/_group_metric_set.py
riedgar-ms: You mean `_process_predictions()` etc.? I think they're all more specialised.

fairlearn/metrics/_group_metric_set.py
romanlutz: Oh I see the tests already exist, that's why I didn't see them here :-)

PR #334, opened 2020-03-16
capitalize Fairlearn in dashboard

rihorn2:
Change capitalization of fairlearn to Fairlearn. Set starting page for view with precomputed metrics to the feature selection page.


fairlearn/widget/js/src/fairlearnDashboard.tsx
```diff
@@ -60,6 +60,7 @@ export class FairlearnView extends DOMWidgetView {
             predictedY={data.predicted_ys}
             trueY={data.true_y}
             modelNames={data.model_names}
+            customMetrics={[]}
```
riedgar-ms: I believe we need to chat about this at some point. However it's not MVP for the AzureML integration

visualization/FairnessDashboard/src/FairnessWizard.tsx
```diff
@@ -252,7 +252,7 @@ export class FairnessWizard extends React.PureComponent<IFairnessProps, IWizardS
                 parityMetrics: accuracyMetrics,
                 selectedParityKey: accuracyMetrics[0].key,
                 dashboardContext: FairnessWizard.buildPrecomputedFairnessContext(props),
-                activeTabKey: accuracyTabKey,
+                activeTabKey: featureBinTabKey,
```
riedgar-ms: Is this the reason that the dashboard in Mark's part of the PoC always opens to the 'second' page?

riedgar-ms:
Could we have a description, please? I think this is doing more than the PR title imples.

visualization/FairnessDashboard/src/FairnessWizard.tsx
rihorn2: yes. This should fix it to open to the feature page

fairlearn/widget/js/yarn.lock
```diff
@@ -228,70 +239,70 @@
...
+  resolved "https://registry.yarnpkg.com/@uifabric/foundation/-/foundation-7.5.19.tgz#8f84c5c249b006d7b2bf820d6e828d08d7648195"
+  integrity sha512-zBOzpZy5H895r8lmmbeRTrJ8r1MpRTavcbnbGJK1rYGB4/1hcxNmdpH1/Dt96JPCMGHH0wRzCT/w5bdAN99k4Q==
+  dependencies:
+    "@uifabric/merge-styles" "^7.8.7"
+    "@uifabric/set-version" "^7.0.6"
+    "@uifabric/styling" "^7.10.19"
+    "@uifabric/utilities" "^7.14.1"
+    tslib "^1.10.0"
+
+"@uifabric/icons@^7.3.18":
```
riedgar-ms: These are all just general updates?

rihorn2: we don't pin these dependencies to an upper version, so as we publish our package again, it pulls in the latest ones

Issue #256, opened 2020-01-22
Naming Conventions


riedgar-ms:
We should review our property naming conventions. For example, `sklearn` puts trailing underscores on attributes which have been calculated by operation of the class. See e.g.:
https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html

MiroDudik:
I think we should aim to follow the conventions in
https://scikit-learn.org/stable/developers/develop.html

adrinjalali:
in summary:

- `attr` is a public attribute, and is set via the `__init__` and `set_props`
- `_attr` is private, and is used internally by methods of the object, and are set usually in `fit`
- `attr_` is public, and is set in `fit`, and not in `__init__`

romanlutz:
#381 is fixing this for postprocessing, after which we should be conforming. Issue #222 will then track the remaining pieces required to get check_estimator to pass.

PR #338, opened 2020-03-20
Remove ExponentiatedGradientResult and flatten contents within ExponentiatedGradient object

romanlutz:
This is another step related to https://github.com/fairlearn/fairlearn-proposals/issues/4

Removing the result object is something we've wanted to do for a while. The result object is a remnant from fairlearn<v0.3. It existed since ExponentiatedGradient was not a class, but simply a function. When we transitioned to actual classes we simply added the result object as a member with the goal to revisit this decision at a later point which we've now reached.


fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -195,30 +200,22 @@ def _pmf_predict(self, X):
     def _format_results(self, gaps, Qs, lagrangian, B, eta_min):
```
riedgar-ms: Can we give this a better name than _format_ results? 'Populate' might be a better verb

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -195,30 +200,22 @@ def _pmf_predict(self, X):
...
-            best_gap,
-            lagrangian,
-            weights,
-            last_t,
-            best_t)
+        self._last_t = len(Qs) - 1
+        self._best_classifier = lambda X: _mean_pred(X, hs, self._weights)
+        self._best_gap = gaps[self._best_t]
+        self._classifiers = lagrangian.classifiers
+        self._n_oracle_calls = lagrangian.n_oracle_calls
```
riedgar-ms: If these are supposed to be public, shouldn't they have a trailing underscore and not a leading one, to match sklearn conventions?

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
romanlutz: Definitely! From staring at this for a while I feel like it may not be needed at all since we were just formatting results for putting them into a result object which we're not doing anymore. Stay tuned...

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
romanlutz: It's not 100% clear which ones are meant to be public right now. There's another work item to match naming conventions. I stuck to what's been done so far. All of the former result object contents could be public in my opinion, but maybe @MiroDudik  has an opinion either way.

romanlutz: Thought about this a little more. Since most of these are subject to be renamed (see my proposal for an idea about what I mean https://github.com/fairlearn/fairlearn-proposals/issues/4) I think it's wise to keep them private until that's happened. We don't want people to think that those attributes are there to stay (if we do a release), just to rename them in the immediately following release. 
I am working through all the refactoring items suggested by @MiroDudik so I will get to #256 as well eventually. This directly targets such things, just in case I miss out on it as part of the reductions refactoring. Does that make sense?

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -61,8 +60,15 @@ def __init__(self, estimator, constraints, eps=0.01, T=50, nu=None, eta_mul=2.0)
...
         self._eta_mul = eta_mul
+
         self._best_classifier = None
+        self._best_gap = None
         self._classifiers = None
+        self._weights = None
+        self._last_t = None
+        self._best_t = None
+        self._n_oracle_calls = 0
+        self._oracle_calls_execution_time = None
```
MiroDudik: This __init__ is not [sklearn-compliant,](https://scikit-learn.org/stable/developers/develop.html). Are we leaving that for a future PR?

romanlutz: It doesn't strike me as a violation, since there's no logic. But @adrinjalali will have to change a few things with reductions methods anyway, so if this is not compliant we can fix that later (since we aren't compliant yet).

PR #340, opened 2020-03-23
Add tests for _GridGenerator and documentation for the basis

romanlutz:
The existing code wasn't as straightforward as I would have liked it to be. This doesn't entirely fix the issue, but it's a good first step in my opinion. Please check that the descriptions are correct since this is based on my observations from the code.


test/unit/reductions/grid_search/test_grid_generator.py
```diff
@@ -0,0 +1,35 @@
...
+
+from fairlearn.reductions._grid_search import _GridGenerator
+from fairlearn.reductions import DemographicParity
+
+from test.unit.reductions.grid_search import _utilities
+
+
+@pytest.mark.parametrize("grid_size", [5, 6, 10, 11, 100, 101, 102])
+@pytest.mark.parametrize("grid_limit", [1.0, 2.0, 4.0, 10.0])
+def test_grid_generator(grid_size, grid_limit):
```
riedgar-ms: '_1d' since I believe that the grid generator can do multi-d as written.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -58,7 +61,7 @@ def load_data(self, X, y, event=None, **kwargs):
         zero_vec = pd.Series(0.0, self.index)
         i = 0
         for event_val in event_vals:
-            for group in group_vals[:-1]:
+            for group in group_vals[:-1]:  # Leave out last group to avoid linear dependence.
```
riedgar-ms: I have a vague idea about what this means, but it's not completely clear still.

fairlearn/reductions/_grid_search/__init__.py
```diff
@@ -4,9 +4,10 @@
 """Package for Grid Search."""
 
 from .grid_search_result import GridSearchResult  # noqa: F401
-from .grid_search import GridSearch  # noqa: F401
+from .grid_search import GridSearch, _GridGenerator  # noqa: F401
 
 __all__ = [
+    "_GridGenerator",
```
riedgar-ms: Why?

test/unit/reductions/grid_search/test_grid_generator.py
romanlutz: Correct! I should probably add a similar test case for that, and maybe other constraints

fairlearn/reductions/_moments/conditional_selection_rate.py
romanlutz: It's a continuation of the longer comment above that says we need a linearly independent basis Adding this last group into the basis makes it linearly dependent.

romanlutz: Any suggestion on how to rewrite that?

fairlearn/reductions/_grid_search/__init__.py
romanlutz: Couldn't use it in test otherwise for some reason :-| I'll see how I can get rid of it.

romanlutz: Fixed that now. Not sure what happened the previous time.

fairlearn/reductions/_moments/conditional_selection_rate.py
romanlutz: slightly adjusted it, but not sure that helps

riedgar-ms: Well, I _think_ that what you mean is that between two sensitive features there's one tradeoff to be made, between three there are two tradeoffs etc. But I'm not sure. The alternative is for determining the 'size' of the grid in multi-D, which comes down to "if you know the sum of n numbers, then the nth is determined by the other n-1"

romanlutz: That is how I understand it. Maybe @MiroDudik  has an idea on how to rewrite this comment?

fairlearn/reductions/_grid_search/grid_search.py
```diff
@@ -52,6 +52,7 @@ def __init__(self, grid_size, grid_limit, pos_basis, neg_basis, neg_allowed, for
                 # convert the grid of basis coefficients into a grid of lambda vectors
                 self.grid = pos_basis.dot(pos_coefs) + neg_basis.dot(neg_coefs)
                 break
+            # if the grid size is not reached yet increase the maximum value
```
MiroDudik: maximum value -> scaling parameter

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -49,7 +49,10 @@ def load_data(self, X, y, event=None, **kwargs):
         self.index = signed.index
         self.default_objective_lambda_vec = None
 
-        # fill in the information about the basis
+        # Certain mitigation techniques require basis vectors representing the constraint space in
+        # order to generate a grid viable linear combinations. The basis vectors are stored in
+        # pos_basis and neg_basis. Including all constraints would make the vectors linearly
+        # dependent.
```
MiroDudik: Let's move this comment closer to where we define the basis (e.g., right above the line 58).

Also, let's use this description:
```
# The matrices pos_basis and neg_basis contain a lower-dimensional description of constraints,
# which is achieved by removing some redundant constraints.
```

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -58,6 +61,7 @@ def load_data(self, X, y, event=None, **kwargs):
         zero_vec = pd.Series(0.0, self.index)
         i = 0
         for event_val in event_vals:
+            # Leave out last group to avoid linear dependence of the basis.
```
MiroDudik:
```
# Constraints on the final group are redundant, so they are not included in the basis.
```

fairlearn/reductions/_moments/conditional_selection_rate.py
romanlutz: I moved it and used your description.

fairlearn/reductions/_moments/conditional_selection_rate.py
romanlutz: I suppose Miro addressed this with his own comment :-)

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -49,15 +49,18 @@ def load_data(self, X, y, event=None, **kwargs):
         self.index = signed.index
         self.default_objective_lambda_vec = None
 
-        # fill in the information about the basis
         event_vals = self.tags[_EVENT].unique()
         group_vals = self.tags[_GROUP_ID].unique()
+        # The matrices pos_basis and neg_basis contain a lower-dimensional description of
```
riedgar-ms: Did you mean to include this file?

riedgar-ms: OK, seeing that the tests make use of the moments, I can see why you might be commenting here.

test/unit/reductions/grid_search/test_grid_generator.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def calculate_grid(grid_limit, grid_size, disparity_moment, events):
+    X, y, A = _utilities._quick_data()
+
+    disparity_moment.load_data(X, y, sensitive_features=A)
+
+    return _GridGenerator(grid_size, grid_limit,
+                          disparity_moment.pos_basis, disparity_moment.neg_basis,
+                          disparity_moment.neg_basis_present, False).grid
```
riedgar-ms: Do we have tests with whatever that last argument is set to True?

test/unit/reductions/grid_search/test_grid_generator.py
```diff
@@ -0,0 +1,102 @@
...
+    disparity_moment = DemographicParity()
+    events = [_ALL]
+
+    grid = calculate_grid(grid_limit, grid_size, disparity_moment, events)
+
+    expected_index = pd.MultiIndex.from_product([['+', '-'], events, [0, 1]],
+                                                names=[_SIGN, _EVENT, _GROUP_ID])
+    assert (expected_index == grid.index).all()
+    expected_grid = pd.DataFrame()
+    grid_size_or_next_smaller_even_number = int(grid_size / 2) * 2
```
riedgar-ms: As I recall, this was one of the peculiarities of the GridGenerator I found when I was prodding at it - a grid size of 2 would give points {-1,0} and a size of 3 would give {-1,0,1}. Is that what this is about?

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -49,15 +49,18 @@ def load_data(self, X, y, event=None, **kwargs):
...
+        # constraints, which is achieved by removing some redundant constraints.
+        # Including all constraints would make the vectors linearly dependent.
```
MiroDudik: Let's remove line 56--the relationship that we are going after is a bit more complicated than just linear dependence (that's why I'm just calling it `redundance` here). Also, we could have kept all the constraints to begin with--things would be correct--only less efficient. So if you want to give some additional clarification, maybe say:
```
# Considering fewer constraints is not required for correctness, but it can dramatically speed up GridSearch.
```

test/unit/reductions/grid_search/test_grid_generator.py
romanlutz: No, and I'm not going to add them now since I don't quite understand what scenario it's for. The test gap wasn't my main intention here. It should still be addressed at some point, though.

test/unit/reductions/grid_search/test_grid_generator.py
romanlutz: Yes!

fairlearn/reductions/_moments/conditional_selection_rate.py
romanlutz: Done! Thanks for elaborating

PR #344, opened 2020-03-24
Store lambda vectors on the ExponentiatedGradient object

romanlutz:
This is another step towards enabling us to plug them into Grid Search. 
This is related to the reductions refactoring outlined in https://github.com/fairlearn/fairlearn-proposals/issues/4

### DIFF
--- a/fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
+++ b/fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
@@ -69,6 +69,8 @@ class ExponentiatedGradient(BaseEstimator, MetaEstimatorMixin):
         self._best_t = None
         self._n_oracle_calls = 0
         self._oracle_calls_execution_time = None
+        self._lambdas = pd.DataFrame()
+        self._lambdas_LP = pd.DataFrame()
 
     def fit(self, X, y, **kwargs):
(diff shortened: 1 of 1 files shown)


fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -69,6 +69,8 @@ def __init__(self, estimator, constraints, eps=0.01, T=50, nu=None, eta_mul=2.0)
         self._best_t = None
         self._n_oracle_calls = 0
         self._oracle_calls_execution_time = None
+        self._lambdas = pd.DataFrame()
```
romanlutz: do we want it to be lambdas or lambda_vecs?

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -69,6 +69,8 @@ def __init__(self, estimator, constraints, eps=0.01, T=50, nu=None, eta_mul=2.0)
         self._best_t = None
         self._n_oracle_calls = 0
         self._oracle_calls_execution_time = None
+        self._lambda_vecs = pd.DataFrame()
```
MiroDudik: I think we will need to change the name `_lambda_vecs` to something else, but let's do that in a future PR.

PR #349, opened 2020-03-27
Unify inner state/results of ExponentiatedGradient and GridSearch

romanlutz:
The long awaited unification as detailed in https://github.com/fairlearn/fairlearn-proposals/issues/4

The current version does not have `best_predictor`, `best_lambda_vec`, or such for GridSearch. For `ExponentiatedGradient` I've transformed `best_classifier` into `best_predictor`. I'm tending towards making all the "best" attributes available instead of requiring users to run
```
best_lambda_vec = grid_search_obj._lambda_vecs[grid_search_obj._best_grid_index]
```
With this change we also get rid of `best_*` attributes on both EG and GS, with the exception of `best_t` (EG) and `best_grid_index` (GS).


fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -61,9 +61,9 @@ def __init__(self, estimator, constraints, eps=0.01, T=50, nu=None, eta_mul=2.0)
         self._nu = nu
         self._eta_mul = eta_mul
 
-        self._best_classifier = None
+        self._best_predictor = None
```
MiroDudik: I think that calling this `predictor` is confusing, because it is not a predictor in the `sklearn` sense. I don't think that we actually need to return this object. Let's get rid of `self._best_predictor` altogether and merge its functionality directly into `_pmf_predict` (we also don't need the `_mean_pred` wrapper).

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -203,7 +203,7 @@ def predict(self, X):
             the result will be a scalar. Otherwise the result will be a vector
         :rtype: Scalar or vector
         """
-        positive_probs = self._best_classifier(X)
+        positive_probs = self._best_predictor(X)
         return (positive_probs >= np.random.rand(len(positive_probs))) * 1
```
MiroDudik: use `_pmf_predict` here (similarly as we do in `ThresholdOptimizer`)

fairlearn/reductions/_grid_search/_grid_generation.py
```diff
@@ -0,0 +1,65 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
```
MiroDudik: should we name this file `_grid_generator.py` similarly to `_threshold_optimizer.py` we do elsewhere?

fairlearn/reductions/_grid_search/grid_search.py
```diff
@@ -142,23 +82,12 @@ def __init__(self,
...
-        :class:`GridSearchResult` is used in calls to
-        :code:`predict` and :code:`predict_proba`.
-        """
-        return self._best_result
+        self._best_grid_index = None
+        self._predictors = []
+        self._lambda_vecs = pd.DataFrame()
+        self._objectives = []
+        self._gammas = pd.DataFrame()
+        self._oracle_calls_execution_time = []
```
MiroDudik: i think i prefer `oracle_execution_times

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
romanlutz: You're right, of course! One more of the remnants we've wanted to address in a while. I'll fix that...

fairlearn/reductions/_grid_search/grid_search.py
romanlutz: works for me!

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
romanlutz: I just realized that `lagrangian.classifiers` only gets set when it's a new best value. So `len(lagrangian.classifiers)` can be less than `_last_t`. So I guess what I have right now with best_predictor being located at `_best_t` within `_predictors` (which is equivalent to `lagrangian.classifiers`) is not always correct. Is it simply the last entry in `lagrangian.classifiers`?

romanlutz: ... or it may be `expgrad._hs[expgrad._best_t]`. `_hs` has an entry for every iteration

notebooks/Grid Search for Binary Classification.ipynb
```diff
@@ -174,9 +174,11 @@
...
+    "The grid search acts like a normal `sklearn` estimator, implementing `fit()` and `predict()` methods. The `fit()` method performs the grid search, and the best model found (according to a specified selection rule) is used in `predict()` calls. However, after `fit()` is called, there are several extra properties on the estimator. \n",
+    "- `_best_grid_index` indicates the index of the grid point where grid search found the best solution. The `predict()` method uses the corresponding predictor.\n",
+    "- for each grid point we have an entry in the following four collections: `_lambda_vecs`, `_objectives`, `_gammas`, `_predictors`.\n",
     "\n",
```
MiroDudik, line 182: just switch the order of bullets: first say what we store for each grid point and then which is the best.

PR #351, opened 2020-03-27
add notebook test run on last released version of fairlearn

romanlutz:
If this fails it will alert us that we need to release a new version of fairlearn to be compatible.


romanlutz:
> 
> 
> If this fails it will alert us that we need to release a new version of fairlearn to be compatible.
>
Shouldn't these be failing right now?

.github/workflows/run-notebook-tests-on-last-released-version.yml
```diff
@@ -0,0 +1,40 @@
...
+    
+    - name: Upgrade pip, setuptools, and wheel before installing other dependencies
+      run: python -m pip install --upgrade pip setuptools wheel
+
+    - run: pip install -r requirements.txt
+      name: 'Install required packages'
+      shell: bash
+    
+    - run: pip install fairlearn
+      name: 'Install last released version of Fairlearn'
```
riedgar-ms: Since I'm paranoid, I'd suggest deleting the fairlearn directory which was just checked out

.github/workflows/run-notebook-tests-on-last-released-version.yml
```diff
@@ -0,0 +1,40 @@
+name: Notebook-Tests--Against-Last-Release
+
+on:
+  pull_request:
```
riedgar-ms: Can you template this, so that in future we can define our supported versions?

.github/workflows/run-notebook-tests-on-last-released-version.yml
adrinjalali: one other thing you can do is to run the notebooks and the tests from `/tmp`

riedgar-ms: Yes. It occurred to me that this could be built on the stage of the release pipeline which tests against the just-deployed pypi package.

romanlutz: @riedgar-ms I'm much less worried about the release pipeline because it would reflect the state from master at that point, and that's something we have covered through the other notebook tests.

Deleting the fairlearn directory is certainly not a bad idea.

.github/workflows/run-notebook-tests-on-last-released-version.yml
romanlutz: I cannot. There are no templates in GitHub Actions.

romanlutz: They have Actions which you can reuse, but the workflows themselves don't appear to be reusable. Or at least after about 10h of working with them I haven't managed to discover how...

.github/workflows/run-notebook-tests-on-last-released-version.yml
romanlutz: Done

romanlutz:
> 
> 
> > If this fails it will alert us that we need to release a new version of fairlearn to be compatible.
> 
> Shouldn't these be failing right now?

There we go! https://github.com/fairlearn/fairlearn/pull/351/checks?check_run_id=547129918

.github/workflows/run-notebook-tests-on-last-released-version.yml
riedgar-ms: I had missed that this wasn't an ADO yaml. I'm not keen on splitting our builds and releases between two systems. What advantage does GitHub actions offer over ADO?

romanlutz: It's right in GitHub, i.e. we don't have to go to a different system. GitHub Actions seems to be the "future" (don't quote me on that) in that it's brand new and people have a marketplace to share Actions. I've written about this before, but the bottom line was that new things are probably better to write in this, and when we update old pipelines we can move them over if it's not hard. I wouldn't waste a lot of time on that right now because the templating doesn't seem to exist yet.

riedgar-ms: I'm not averse to moving over, but I think we should use one system or the other. Having a mixture is asking for more headaches.

.github/workflows/run-notebook-tests-on-last-released-version.yml
```diff
@@ -0,0 +1,44 @@
+name: Notebook-Tests--Against-Last-Release
```
riedgar-ms: What's the alerting around this, if the build starts failing?

romanlutz: Correct. You can still complete it but it'll make sure whoever completes it knows that it's not compatible anymore.

romanlutz: Like in this PR for example!

.github/workflows/run-notebook-tests-on-last-released-version.yml
riedgar-ms: Also, while I'm happy to see this as a PoC, I don't think we've really had a discussion as to whether we should stay with ADO or move to GitHub Actions. Maybe I've missed something, but the argument for the latter appears to be that it has "GitHub" in the name. Using GitHub Actions for things like automatically adding reviewers to PRs makes a lot of sense. Changing the builds is less clear cut to me.

romanlutz: The discussion exists, although it was less of a discussion and more of me sharing thoughts (i.e. nobody responded). It's not publicly accessible so I'll post it here for the sake of transparency (this was pre-Gitter):
"Depending on whether you're as excited as I am about GitHub actions you may want to check out this little PR https://github.com/fairlearn/fairlearn/pull/323
I wanted to try out GitHub actions with a minimal example, and adding PR reviewers automatically fit the bill. You can also restrict with filters (see documentation link), but I wanted it to be simple to start with.
Eventually all pipelines can run through GitHub actions, but the fundamental limitation I discovered yesterday is that they don't allow for templates (yet). Richard and I have been using templates quite heavily, so for us it would be quite painful to duplicate all the YAML content.... I do plan to gradually move functionality over as they add more features."

This isn't "changing the builds" but merely not adding more of them in ADO.
I have heard from various people that they don't know Azure DevOps well and it's confusing to have to switch to a different kind of page with a different layout. This integrates much better since it's right in GitHub. For that reason I want to avoid adding more complexity to existing Azure pipeline configurations and rather add new ones in GitHub. Whenever they support templates we can be more aggressive about moving everything over, but it's not a pressing issue. As you can see below, the list of checks includes all of them next to each other. I find the one-click experience to get to the build with GitHub Actions vastly superior compared to Azure DevOps.

riedgar-ms: Having builds run on two different platforms is most definitely 'changing the builds.' We now have two places to keep track of things (the PR results might all be listed together, but what about the build histories, where we have the dashboard?).

Are we committing to doing everything in GitHub Actions in future? I don't remember that discussion.

The functionality in this PR feels like it should be a slight modification of the validation stages of the release pipeline template (which pip installs from a given PyPI URL, and runs tests and notebooks). Why not use that?

romanlutz: I'm not changing any existing pipelines.

Like I wrote above nobody seemed keen on the discussion earlier. I'm happy to have that discussion anytime. At the end of the day you want to be on the platform of the future, and the way that GitHub has been building on this and expanding makes me think that it's GitHub Actions, not Azure Pipelines.

With that in mind one can either choose to ignore that and double down on the existing things or move over. Moving over can also happen gradually or at once with lots of effort. My proposal is to just move things as it makes sense with the least resistance. That means
- adding something new? --> GitHub Actions
- changing something old? --> small changes stay in Azure Pipelines, big structural changes can be moved over to GitHub Actions

As you may notice when reading the definition they're not exactly equal so there is extra effort in writing it for one and then rewriting it for the other. For that reason alone I wouldn't want to rewrite it for ADO now.

By dashboard you probably mean this https://dev.azure.com/responsibleai/fairlearn/_dashboards/dashboard/7b669a6f-4535-4777-9349-e38659875c40 ? Do you find that useful?

image

History is obviously interesting for debugging purposes, but we don't lose that either since you have that in GitHub, too. Just click on "Actions" at the top.

Eventually not having to go to ADO at all anymore is something that appeals very much to me, and for every build we have in GitHub that already applies today. Like I said above I'm not advocating for putting in an effort to move pipelines over. I just think it should be done whenever it's just as hard to change/extend existing pipelines or add new ones like in this PR.

riedgar-ms: I'm not objecting to planning a move to GitHub Actions. I'm objecting to supporting two build systems without the prospect of an end date.

riedgar-ms: Specifically, I think that:
https://github.com/fairlearn/fairlearn/blob/master/devops/templates/notebook-pypi-job-template.yml
Is doing almost the same steps (discovering the desired pip version from a build Artifact would be the main wrinkle).

romanlutz: [sidenote: You can have it on your hallway screen! It should be publicly accessible. But I fear you won't see that screen for a while...]

I'm not sure I see a cost of having both. I don't see this discussion going anywhere either, though, so we'll probably have to agree to disagree. :-) 
How do you feel about me updating the ADO pipeline with this AND also having the GitHub Action workflow? At least we'd make a step in the future direction as well, instead of deleting the already written work...

riedgar-ms: OK; I do wish we'd had a better discussion about this, rather than suddenly having the move 'announced' via PR.

I do still think there's a cost to having builds in two places. Especially since GitHub Actions lacks templating, so as you add more pipelines, there won't be reusuability.

romanlutz: I wish the same, but nobody responded at the time. It was communicated via Teams.
I'm not proposing migrating existing templated pipelines to Actions until they have templates.

An interesting compromise might be to run everything through GitHub using this Action that can call Azure Pipelines: https://github.com/marketplace/actions/azure-pipelines-action

riedgar-ms: Did I miss it in one of our Thursday meetings?

I don't see the benefit of having an Action trigger an ADO pipeline. I'm OK with having this in as a proof-of-concept for GitHub Actions. It is, after all, quite small and focussed.

.github/workflows/run-notebook-tests-on-last-released-version.yml
romanlutz: so if I rerun this now it should hopefully succeed... due to the release

romanlutz: It failed, perhaps needs latest master to update the notebooks first. Rerunning after git pull origin master

romanlutz: Oh this changed already since the release. Miro's metrics changes are required for two notebooks. So the failure is actually correct!

PR #354, opened 2020-03-31
Remove group_metric_result and create group_summary transformers

MiroDudik:
Getting started with implementing the [Metrics API proposal](https://github.com/fairlearn/fairlearn-proposals/blob/master/api/METRICS.md).


fairlearn/metrics/_metrics_engine.py
```diff
@@ -3,104 +3,124 @@
...
 import numpy as np
 
-from ._group_metric_result import GroupMetricResult
 from ._input_manipulations import _convert_to_ndarray_and_squeeze
+from sklearn.utils import Bunch
 
 _MESSAGE_SIZE_MISMATCH = "Array {0} is not the same size as {1}"
 
+# Parameters to metrics that should be split according to sensitive features
+_SPLITTABLE_METRIC_PARAMS = {"sample_weight"}
```
riedgar-ms: Isn't this something which should be specified when the metric is invoked?

fairlearn/metrics/_metrics_engine.py
```diff
@@ -3,104 +3,124 @@
...
 
-def metric_by_group(metric_function,
-                    y_true, y_pred, group_membership,
-                    sample_weight=None,
-                    **kwargs):
+
+def group_summary(metric_function, y_true, y_pred,
+                  sensitive_features, **metric_params):
```
riedgar-ms: I thought `**kwargs` was conventional? I can see the sense in renaming it, but I wouldn't want to confuse future readers.

fairlearn/metrics/_metrics_engine.py
```diff
@@ -3,104 +3,124 @@
...
+    :param metric_function: Function ``(y_true, y_pred, \*\*metric_params)``
 
     :param y_true: Array of ground-truth values
 
     :param y_pred: Array of predicted values
 
-    :param group_membership: Array Indicating the group to which each input value belongs
-
-    :param sample_weight: Optional weights to apply to each input value
+    :param sensitive_features: Array indicating the group to which each input value belongs
```
riedgar-ms: It's got to be a single column? If so, shouldn't `sensitive_features` be singular and not plural?

fairlearn/metrics/_metrics_engine.py
```diff
@@ -3,104 +3,124 @@
...
 
-    :param \*\*kwargs: Optional arguments to be passed to the `metric_function`
+    :param \*\*metric_params: Optional arguments to be passed to the `metric_function`
 
     :return: Object containing the result of applying ``metric_function`` to the entire dataset
-        and to each group identified in ``group_membership``.
-        If the ``metric_function`` returns a scalar, then additional fields are populated
-    :rtype: :class:`GroupMetricResult`
+        and to each group identified in ``sensitive_features``
+    :rtype: sklearn.utils.Bunch with the fields ``overall`` and ``by_group``
```
riedgar-ms: Nit: I believe that there's a bit of ReST magic which can be inserted here to generate the link to `sklearn`. Will have to dig it out.

test/unit/metrics/test_group_metric_result.py
```diff
@@ -38,7 +39,7 @@ def test_complex_equality(self):
         assert not(a != b)
         assert not(b != a)
 
-    def test_complex_inequality(self):
+    def dont_test_complex_inequality(self):
```
riedgar-ms: Why are these tests being disabled?

fairlearn/metrics/_metrics_engine.py
```diff
@@ -3,104 +3,124 @@
...
+        result_by_group[group] = metric_function(
+            y_t[group_indices], y_p[group_indices],
+            **_check_metric_params(y_t, metric_params, group_indices))
+
+    return Bunch(overall=result_overall, by_group=result_by_group)
+
+
+# This loosely follows the pattern of _check_fit_params in
+# sklearn/utils/validation.py
+def _check_metric_params(y_true, metric_params, indices=None):
```
riedgar-ms: I'd like to see some tests of this routine.

fairlearn/metrics/_metrics_engine.py
MiroDudik: Do you mean when the metric is invoked or when the group metric function is created?

What I have currently functionally matches our previous implementation. I think a better solution would be to allow providing alternative lists when calling `make_group_metric`. But I'd leave that for a future PR.

fairlearn/metrics/_metrics_engine.py
MiroDudik: it is being partly tested by the old `sample_weights` tests.

test/unit/metrics/test_group_metric_result.py
MiroDudik: A dictionary of numpy arrays throws an exception when called with `==` or `!=` because there are different ways of comparing arrays. These were all tests for `__eq__` and `__ne__` methods that `group_metric_result` was implementing, so I'm not sure that we still need them.

MiroDudik: Okay to delete these tests?

fairlearn/metrics/_metrics_engine.py
MiroDudik: I don't think that `**kwargs` is a major dogma. In this specific instance, calling it with what it contains seems clearer to me, but I don't have a strong opinion on this.

fairlearn/metrics/_metrics_engine.py
MiroDudik: i think that our current convention is to call it `sensitive_features` everywhere... for the sake of future compatibility.

fairlearn/metrics/_metrics_engine.py
MiroDudik: a small issue here is that i don't think that `sklearn.utils.Bunch` is in sklearn docs

fairlearn/metrics/_metrics_engine.py
MiroDudik: okay if we hold this off for a future PR?

test/unit/metrics/test_group_metric_result.py
riedgar-ms: If they're no longer relevant, then delete them. If tests have to be disabled because of a bug, use `pytest.skip`.

fairlearn/metrics/_metrics_engine.py
riedgar-ms: Yes, I believe that @adrinjalali filed an issue with `sklearn` over that. Looking to the future, I believe that what you'll need is:
":py:class:`sklearn.utils.Bunch`"

fairlearn/metrics/_metrics_engine.py
riedgar-ms: Perhaps `sensitive_features_column` which is what we settled on for the AzureML integration?

fairlearn/metrics/_metrics_engine.py
adrinjalali: The bunch object is now documented in master. It'll be on the stable version of the website in the next release.

fairlearn/metrics/_metrics_engine.py
riedgar-ms: Perhaps `metric_function_params` ? I was just surprised, but I'm still on a Python learning curve.

fairlearn/metrics/_metrics_engine.py
riedgar-ms: Would 'future' be 'immediately after this?' I can see the sense in doing this in two stages (less chance for disaster), but I don't think that it should be delayed for long.

I think that this would be specified when the group metric function is created, now you point out the distinction. After all, at that point the underlying metric is bound, and its arguments aren't going to change.

fairlearn/metrics/_metrics_engine.py
riedgar-ms: Given that it's a private routine, OK. However, I do think it's doing some 'magic' so direct testing, even though it's private does make sense.

fairlearn/metrics/_metrics_engine.py
romanlutz: I've never seen it called anything but `kwargs` but I don't see why not...

fairlearn/metrics/_metrics_engine.py
MiroDudik: yes... that's what i was thinking. i think it's easy enough to implement, but i might leave proper testing to a future PR.

fairlearn/metrics/_metrics_engine.py
MiroDudik: I'll rename it to `**metric_function_kwargs` to make this clearer?

fairlearn/metrics/_metrics_engine.py
MiroDudik: I'll hold off. There's more magic that needs to happen in this routine to enable crossvalidation with sklearn. I'll add testing then.

fairlearn/metrics/_metrics_engine.py
adrinjalali: FWIW we call them `**fit_params` in sklearn.

test/unit/metrics/test_group_metric_result.py
adrinjalali: you can also mark them as `xfail` to _allow them to fail_.

fairlearn/metrics/_metrics_engine.py
adrinjalali: Yeah sorry about that. I'm actively working on this one on the sklearn side these days.

fairlearn/metrics/_metrics_engine.py
riedgar-ms: I'm OK with `**metric_func_params` then

fairlearn/metrics/_metrics_engine.py
MiroDudik: I'd rather follow the same convention as we do in `ExponentiatedGradient` and `GridSearch`, where we just use `sensitive_features`.

fairlearn/metrics/_metrics_engine.py
MiroDudik: @adrinjalali awesome! keep us posted.

test/unit/metrics/test_group_metric_result.py
MiroDudik: i'm removing the test file since we have replaced `group_metric_result` by `Bunch`, so it does not need testing.

fairlearn/metrics/_metrics_engine.py
MiroDudik: done.

fairlearn/metrics/_metrics_engine.py
MiroDudik: done.

PR #358, opened 2020-04-01
Bump matplotlib to 3.2.1 and remove TkAgg workaround

### LINKED ISSUE #320: Remove matplotlib TKAgg special case with matplotlib>3.2.0
related to #318 

With matplotlib release 3.2.0 the TKAgg backend stopped working on Linux causing our tests to fail. #319 fixes that by using TKAgg only on MacOS. We should check for future matplotlib releases whether they fixed that and revert the special case, i.e. use TKAgg everywhere like it used to be.

romanlutz:
Closes #320 
Without using TkAgg our plotting tests would fail on MacOS. With 3.2.1 this isn't the case anymore, so we can happily remove the special case for MacOS in general, and apply a skip for MacOS with python 3.5 only. This means all future python versions as well as >=3.6 are tested without TkAgg specifically enabled.

### DIFF
--- a/test/unit/postprocessing/test_plots.py
+++ b/test/unit/postprocessing/test_plots.py
@@ -1,18 +1,13 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License.
 
-# On MacOS we need to use TKAgg before importing matplotlib.pyplot.
-# This used to work on all platforms until matplotlib 3.2.0 broke it on Linux.
-# Consider removing the OS-based if after a future release of matplotlib (issue #320).
-import matplotlib
-matplotlib.use('TkAgg')
-import matplotlib.pyplot as plt  # noqa: E402
-import pkg_resources  # noqa: E402
-import pytest  # noqa: E402
-from fairlearn.postprocessing import ThresholdOptimizer, plot_threshold_optimizer  # noqa: E402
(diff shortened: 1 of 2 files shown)


test/unit/postprocessing/test_plots.py
```diff
@@ -31,6 +25,8 @@
 
 
 def _fit_and_plot(constraints, plotting_data):
+    import matplotlib.pyplot as plt
```
riedgar-ms, line 28: Doesn't flake8 scream about an import which isn't at the top of the file?

riedgar-ms:
Does this mean we no longer need requirements-3.5.txt and all that machinery?

test/unit/postprocessing/test_plots.py
romanlutz: Interestingly, it doesn't!

romanlutz:
> 
> 
> Does this mean we no longer need requirements-3.5.txt and all that machinery?

No that's there for a different reason. This was just about the special case in that test file which is now removed. 
requirements-3.5.txt is there so that we can specify package versions for py3.5 that work. Certain packages (including matplotlib) don't have 3.5 support anymore in the latest versions, so we can't keep bumping. Eventually this will get too tedious and we'll drop py3.5 support ourselves.

PR #362, opened 2020-04-02
Equal Opportunity implementation

hannanabdul55:
## What's there in this PR?
- Implement Equal Opportunity and incorporate comments in #325 
- Add smoke test to check the constraint.


fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -72,7 +72,7 @@ def fit(self, X, y, **kwargs):
         :param y: The label vector
         :type y: numpy.ndarray, pandas.DataFrame, pandas.Series, or list
         """
-        _, y_train, A = _validate_and_reformat_input(X, y, **kwargs)
+        X, y_train, A = _validate_and_reformat_input(X, y, **kwargs)
```
MiroDudik, line 76: we don't want this. why is this change here?

fairlearn/reductions/_grid_search/grid_search.py
```diff
@@ -113,7 +113,7 @@ def fit(self, X, y, **kwargs):
             logger.debug("Regression problem detected")
             is_classification_reduction = False
 
-        _, y_train, sensitive_features_train = _validate_and_reformat_input(
+        X, y_train, sensitive_features_train = _validate_and_reformat_input(
             X, y, enforce_binary_sensitive_feature=True,
```
MiroDudik, line 117: @romanlutz , please also comment here.

fairlearn/reductions/_moments/__init__.py
```diff
@@ -7,7 +7,7 @@
 from .moment import ClassificationMoment, LossMoment  # noq: F401
 
 from .conditional_selection_rate import ConditionalSelectionRate  # noqa: F401
-from .conditional_selection_rate import DemographicParity, EqualizedOdds  # noqa: F401
+from .conditional_selection_rate import DemographicParity, EqualizedOdds, EqualOpportunity  # noqa: F401, E501
 from .error_rate import ErrorRate   # noqa: F401
```
MiroDudik, line 11: We want a different name. Please review and incorporate my comments from the previous PR #325

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -49,19 +52,15 @@ def load_data(self, X, y, event=None, **kwargs):
...
-        # Considering fewer constraints is not required for correctness, but it can dramatically
-        # speed up GridSearch.
         self.pos_basis = pd.DataFrame()
         self.neg_basis = pd.DataFrame()
         self.neg_basis_present = pd.Series()
         zero_vec = pd.Series(0.0, self.index)
         i = 0
         for event_val in event_vals:
-            # Constraints on the final group are redundant, so they are not included in the basis.
             for group in group_vals[:-1]:
```
MiroDudik, line 64: this looks like a bad conflict resolution. the above comments should still be there.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -150,6 +153,45 @@ def load_data(self, X, y, **kwargs):
...
+
+    short_name = "EqualOpportunity"
+
+    def load_data(self, X, y, **kwargs):
+        """Load the specified data into the object."""
+        super().load_data(X, y,
+                          # the where clause is used to put pd.nan on all values
+                          # where y!=1.
+                          event=pd.Series(y).apply(lambda y: _LABEL + "=" + str(y)).where(y == 1),
+                          filter_value=0,
```
MiroDudik, line 191: this needs to be removed

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -40,6 +40,9 @@ def load_data(self, X, y, event=None, **kwargs):
         """
         super().load_data(X, y, **kwargs)
         self.tags[_EVENT] = event
+        self.tags = self.tags.dropna()
+        self.dropped_X = self.X.drop(self.tags[_LABEL].index)
+        self.X = self.X.drop(self.dropped_X.index)
         self.prob_event = self.tags.groupby(_EVENT).size() / self.total_samples
```
MiroDudik, line 46: Do we need to do all this dropping above? I believe that `groupby` automatically drops `np.nan`. also, it would be great if we could avoid transforming `X`.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -115,6 +114,10 @@ def signed_weights(self, lambda_vec):
         signed_weights = self.tags.apply(
             lambda row: adjust[row[_EVENT], row[_GROUP_ID]], axis=1
         )
+        # this flag is set if a `filter_value` was sent to :meth:`load_data()`
+        signed_weights = signed_weights.reindex(
+            list(range(self.dropped_X.shape[0] + self.X.shape[0])), fill_value=0
+        )
```
MiroDudik, line 120: Let's work with `tags` that don't have anything dropped. In that case you can maybe just do
```python
apply(lambda row: 0 if np.isnan(row[_EVENT]) else adjust[row[_EVENT], row[_GROUP_ID]])
```
and no need to reindex anything.

test/unit/input_convertors.py
```diff
@@ -98,4 +98,4 @@ def _map_into_single_column(matrix):
     if len(np.array(matrix).shape) == 1:
         return np.array(matrix)
 
-    return _compress_multiple_sensitive_features_into_single_column(matrix)
+    return _compress_multiple_sensitive_features_into_single_column(np.array(matrix))
```
MiroDudik, line 101: @romanlutz ?

test/unit/reductions/exponentiated_gradient/test_exponentiatedgradient_arguments.py
```diff
@@ -35,7 +35,7 @@ def _get_data(A_two_dim=False):
     if A_two_dim:
         # Stacking the same column a few times will result in the identical groups
         # compared to using a single column, therefore results should be the same.
-        A = np.stack((sensitive_features, sensitive_features), -1)
+        A = pd.DataFrame(np.stack((sensitive_features, sensitive_features), -1))
     else:
```
MiroDudik, line 39: @romanlutz ?

test/unit/input_convertors.py
hannanabdul55: This was something that I am proposing(as of now) to fix the tests with these changes, specifically, I wanted some feedback on this because `drop` is not present in `numpy`, as a result, we would need to ensure the input `X` is a `pandas` object.

fairlearn/reductions/_moments/conditional_selection_rate.py
hannanabdul55: Oh yes, I think I understood what you mean by this. I will check this way and let you know here if it works the same way! Thank you.

fairlearn/reductions/_moments/conditional_selection_rate.py
hannanabdul55: Done.

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
hannanabdul55: Removed this.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -150,6 +151,44 @@ def load_data(self, X, y, **kwargs):
...
 
 
+class TruePositiveRateDifference(ConditionalSelectionRate):
+    r"""Implementation of Equal Opportunity as a moment.
+
+    Adds conditioning on `y=1` compared to Demographic parity, i.e.
+
+    .. math::
+       P[h(X) = 1 | A = a, Y = 1] = P[h(X) = 1 | Y = 1] \; \forall a, y,
+
```
MiroDudik, line 161: This is not quite correct, we have only one event corresponding to `y = 1` now. ALso fix `index` and review the rest of the description to make sure it matches the current implementation.

test/unit/reductions/exponentiated_gradient/test_exponentiatedgradient_smoke.py
```diff
@@ -69,7 +69,13 @@ def setup_method(self, method):
                         "best_gap": 0.000000, "last_t": 5,
                         "best_t": 5, "disp": 0.005000,
                         "error": 0.442883, "n_oracle_calls": 19,
-                        "n_predictors": 6}]
+                        "n_predictors": 6},
+                       {"cons_class": TruePositiveRateDifference, "eps": 0.005,
+                        "best_gap": 0.0, "last_t": 5,
```
MiroDudik, line 74: This test example looks too easy to solve for the algorithm (the fairness constraints are satisfied automatically). Please try to run it on a version of the data where the label 'y' is flipped... in that case you should obtain the error that matches the one obtained for `EqualizedOdds` with the same `eps`.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -49,7 +49,8 @@ def load_data(self, X, y, event=None, **kwargs):
         self.index = signed.index
         self.default_objective_lambda_vec = None
 
-        event_vals = self.tags[_EVENT].unique()
+        # fill in the information about the basis
+        event_vals = self.tags[_EVENT].dropna().unique()
```
romanlutz, line 53: I thought when we chatted earlier today you said `dropna()` isn't needed anymore? Perhaps I misunderstood :-)

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -150,6 +151,45 @@ def load_data(self, X, y, **kwargs):
...
+    Adds conditioning on label `y=1` compared to Demographic parity, i.e.
+
+    .. math::
+       P[h(X) = 1 | A = a, Y = 1] = P[h(X) = 1 | Y = 1] \; \forall a
+
+    This implementation of :class:`ConditionalSelectionRate` defines
+    the event corresponding to `y=1`.
+
+    The `prob_event` :class:`pandas:pandas.DataFrame` will record the
+    fraction of the samples corresponding to each unique value of `y = 1` in
```
romanlutz, line 166: Not sure this sentence makes sense with y = 1.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -150,6 +151,45 @@ def load_data(self, X, y, **kwargs):
...
+    the `Y` array.
+
+    The `index` MultiIndex will have a number of entries equal to
+    the number of unique values for the sensitive feature, multiplied by
+    the number of unique values of the `Y` array, which is just `label=1`
+    (1 unique value), multiplied by two (for the Lagrange multipliers
+    for positive and negative constraints).
+
+    With these definitions, the :meth:`signed_weights` method
```
romanlutz, line 175: Funny typo there: meth -> math

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -150,6 +151,45 @@ def load_data(self, X, y, **kwargs):
...
+    `y=1' samples and subsequently, add a weight of `0` for
+    all other samples (for all `y=0`).
+    """
+
+    short_name = "TruePositiveRateDifference"
+
+    def load_data(self, X, y, **kwargs):
+        """Load the specified data into the object."""
+        super().load_data(X, y,
+                          # the where clause is used to put pd.nan on all values
```
romanlutz, line 187: nit: can this comment go on one line? Can even be right above the expression
```
# the where clause is used to put pd.nan on all values where y!=1
super().load_data(X, y,
                  event=pd.Series(y).apply(lambda y: _LABEL + "=" + str(y)).where(y == 1),
                  **kwargs)
```

fairlearn/reductions/_moments/conditional_selection_rate.py
hannanabdul55: I think this is existing already and this means `method`. I cross-checked this the documentation of `EqualizedOdds` and it seems to be the same there too.

fairlearn/reductions/_moments/conditional_selection_rate.py
hannanabdul55: The `pd.dropna` would not be required in the case of `pd.groupby`. But over here, if the `pd.dropna` is not present, it will take `nan` also as a unique value.

romanlutz: Got it.

fairlearn/reductions/_moments/conditional_selection_rate.py
romanlutz: :O OK the joke's on me! Also, what kind of funny abbreviation for method...

test/unit/reductions/exponentiated_gradient/test_exponentiatedgradient_smoke.py
```diff
@@ -100,7 +106,7 @@ def test_smoke(self, testdata):
 
     def test_simple_fit_predict(self):
         estimator = LeastSquaresBinaryClassifierLearner()
-        constraints = DemographicParity()
+        constraints = TruePositiveRateDifference()
```
romanlutz, line 109: You can parametrize this test to use all the constraints, but don't replace the existing one with DP, please.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -150,6 +151,44 @@ def load_data(self, X, y, **kwargs):
...
+
+    The `prob_event` :class:`pandas:pandas.DataFrame` will record the
+    fraction of the samples corresponding to `y = 1` in
+    the `Y` array.
+
+    The `index` MultiIndex will have a number of entries equal to
+    the number of unique values for the sensitive feature, multiplied by
+    the number of unique values of the `Y` array, which is just `label=1`
+    (1 unique value), multiplied by two (for the Lagrange multipliers
+    for positive and negative constraints).
```
MiroDudik, line 173: The above sounds still a bit confusing. How about the following re-phrase of lines 170 and 171 to clarify things a bit:

```
    ...
    the number of unique values of the sensitive feature, multiplied by
    the number of unique non-NaN values of the constructed `event` array, which contains only `label=1`
    ...
```

test/unit/reductions/exponentiated_gradient/test_exponentiatedgradient_smoke.py
hannanabdul55: Sorry about this, again. Yeah, I will just parameterize it.

fairlearn/reductions/_moments/conditional_selection_rate.py
hannanabdul55: Is this the correction that was expected @MiroDudik ? Since there is only 1 value of y, it is not included in the `\forall` clause. so now it will just be `\forall a` instead of `\forall a,y`.

test/unit/reductions/exponentiated_gradient/test_exponentiatedgradient_smoke.py
hannanabdul55: So does this mean that for `y=1`, there is already no disparity in the test data? So what we should check for is for `y=0`?

hannanabdul55: @MiroDudik  I ran a basic test as in [this gist](https://gist.github.com/hannanabdul55/536a1a9f25a6c39fd60f860bbd0938b5#file-test_exponentiatedgradient_smoke-py-L118). But I did not get the same error value. This is what I got after running the test: 
```python
1. Error: 0.25 Disparity: 0.0050000000000000044
Flipped y
2. Error: 0.4271333333333335 Disparity: 0.004999999999999893
```
- The first set of values is for the original generated data. 
- The second set is for the data with the `y`'s flipped. `list(map(lambda i: int(not i), labels))`. 
I did not understand how the error could be better for either case.  
PS: The error for the Equalized Odds case is `0.442883`(as per the smoke test result for eps=`0.05`.)

test/unit/reductions/exponentiated_gradient/test_exponentiatedgradient_smoke.py
hannanabdul55: Done 👍

fairlearn/reductions/_moments/conditional_selection_rate.py
MiroDudik: looks good

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -150,6 +151,44 @@ def load_data(self, X, y, **kwargs):
         super().load_data(X, y, event=_ALL, **kwargs)
 
 
+class TruePositiveRateDifference(ConditionalSelectionRate):
+    r"""Implementation of Equal Opportunity as a moment.
```
MiroDudik, line 155: "Equal Opportunity" -> "True Positive Rate Difference (also called Equal Opportunity Difference)"

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -150,6 +151,44 @@ def load_data(self, X, y, **kwargs):
...
+
+    Adds conditioning on label `y=1` compared to Demographic parity, i.e.
+
+    .. math::
+       P[h(X) = 1 | A = a, Y = 1] = P[h(X) = 1 | Y = 1] \; \forall a
+
+    This implementation of :class:`ConditionalSelectionRate` defines
+    the event corresponding to `y=1`.
+
+    The `prob_event` :class:`pandas:pandas.DataFrame` will record the
```
MiroDudik, line 165: The class of `prob_event` is `pandas.Series` not `pandas.DataFrame`. I also think that we can just drop the class designation here.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -150,6 +151,44 @@ def load_data(self, X, y, **kwargs):
...
+    the `Y` array.
+
+    The `index` MultiIndex will have a number of entries equal to
+    the number of unique values of the sensitive feature, multiplied by
+    the number of unique non-NaN values of the constructed `event` array,
+    whose entries are either NaN or `label=1` (so only one unique non-NaN value), multiplied
+    by two (for the Lagrange multipliers for positive and negative constraints).
+
+    With these definitions, the :meth:`signed_weights` method
+    will calculate the costs like how it is calculated in Example 4 of
```
MiroDudik, line 176: "like how it is calculated in" -> "similar to"

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -150,6 +151,44 @@ def load_data(self, X, y, **kwargs):
...
+    `Agarwal et al. (2018) <https://arxiv.org/abs/1803.02453>`_ but for all
+    `y=1' samples and subsequently, add a weight of `0` for
+    all other samples (for all `y=0`).
```
MiroDudik, line 179: Replace the part of the comment that starts with "but for all" with:
```
   , but will only use the weights equal to zero for `y=0`.
```

test/unit/reductions/exponentiated_gradient/test_exponentiatedgradient_smoke.py
MiroDudik: This might still be okay. Can you try to run `ExponentiatedGradient` on the flipped labels as well?

Also, try to evaluate both the **equalized odds** based disparity as well as the TPR Difference based disparity in both cases (in other words, you can just cycle through the two `disparity_moment` settings).

fairlearn/reductions/_moments/conditional_selection_rate.py
hannanabdul55:
```
    With these definitions, the :meth:`signed_weights` method
    will calculate the costs like how it is calculated in Example 4 of
    `Agarwal et al. (2018) <https://arxiv.org/abs/1803.02453>`_  for all
    `y=1' samples, but will use the weights equal to zero for `y=0`.
```
Is this better?

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -150,6 +151,44 @@ def load_data(self, X, y, **kwargs):
...
+    Adds conditioning on label `y=1` compared to Demographic parity, i.e.
+
+    .. math::
+       P[h(X) = 1 | A = a, Y = 1] = P[h(X) = 1 | Y = 1] \; \forall a
+
+    This implementation of :class:`ConditionalSelectionRate` defines
+    the event corresponding to `y=1`.
+
+    The `prob_event` :class:`pandas:pandas.DataFrame` will record the
+    fraction of the samples corresponding to `y = 1` in
```
romanlutz, line 166: nit: can you use the full line length? The way lines end seems a bit arbitrary here. We've had some people prefer lengths of 79 chars, others 99 (which is the max as defined in our repo), and I don't really care which one, but try to stick to one if possible :-)

test/unit/reductions/exponentiated_gradient/test_exponentiatedgradient_smoke.py
```diff
@@ -98,9 +105,9 @@ def Q(X): return expgrad._pmf_predict(X)[:, 1]
     def test_smoke(self, testdata):
         self.run_smoke_test(testdata)
 
-    def test_simple_fit_predict(self):
+    @pytest.mark.parametrize("constraints", constraints)
```
romanlutz, line 108: nit: I'd prefer using the list of class names
```
@pytest.mark.parametrize("Constraints", [DemographicParity, EqualizedOdds, ...])
```
and then instantiate it inside the test using
```
constraints = Constraints()
```
because we may add more complex logic to those constraint objects at some point. Better to have a fresh one for each test as opposed to reusing them which is what you're doing with the `constraints` list here.

fairlearn/reductions/_moments/conditional_selection_rate.py
MiroDudik: how about this:
```
    With these definitions, the :meth:`signed_weights` method
    will calculate the costs for `y=1` as they are calculated in Example 4 of
    `Agarwal et al. (2018) <https://arxiv.org/abs/1803.02453>`_,
    but will use the weights equal to zero for `y=0`.
```

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -150,6 +151,44 @@ def load_data(self, X, y, **kwargs):
...
+    the number of unique non-NaN values of the constructed `event` array,
+    whose entries are either NaN or `label=1` (so only one unique non-NaN value), multiplied
+    by two (for the Lagrange multipliers for positive and negative constraints).
+
+    With these definitions, the :meth:`signed_weights` method
+    `Agarwal et al. (2018) <https://arxiv.org/abs/1803.02453>`_ but for all
+    will calculate the costs for `y=1` similar to how it is calculated in Example 4 of
+    `Agarwal et al. (2018) <https://arxiv.org/abs/1803.02453>`_  for all
+    `y=1' samples, but will use the weights equal to zero for `y=0`.
+    """
```
MiroDudik, line 180: I think this got mangled. Please check: https://github.com/fairlearn/fairlearn/pull/362#discussion_r408162648

fairlearn/reductions/_moments/conditional_selection_rate.py
hannanabdul55: I will remove it here. But the `prob_event` is a `DataFrame`  when I checked. I think it would be better to just remove the class designation as you suggested. Will do it 👍

fairlearn/reductions/_moments/conditional_selection_rate.py
hannanabdul55: Sure! Sorry, I think I will also set my editor with these values so that I do not miss it. 👍

fairlearn/reductions/_moments/conditional_selection_rate.py
hannanabdul55: Yes, I think this makes the comment more concise. Thank you!

test/unit/reductions/exponentiated_gradient/test_exponentiatedgradient_smoke.py
hannanabdul55: These are the values I get for the flipped case,
```python
{"cons_class": TruePositiveRateDifference, "eps": 0.005,
         "best_gap": 0.0, "last_t": 5,
         "best_t": 5, "disp": 0.005000,
         "error": 0.427133, "n_oracle_calls": 17,
         "n_predictors": 3},
{"cons_class": EqualizedOdds, "eps": 0.005,
         "best_t": 5, "disp": 0.005000,
         "error": 0.442883, "n_oracle_calls": 19,
         "n_predictors": 6}
```
I think from this, it makes sense because for the same value of `eps`, the error introduced would be more for `EqualizedOdds` than TPR as it tried to account for the fairness in both the values of `y`,rather than just `y=1`. In this process, it injects more error? Could this be a reasonable explanation for this?

fairlearn/reductions/_moments/conditional_selection_rate.py
hannanabdul55: Done 👍

test/unit/reductions/exponentiated_gradient/test_exponentiatedgradient_smoke.py
hannanabdul55: Done 👍

PR #364, opened 2020-04-02
Addition of constraints

parul100495:
### What's there in this PR?
- Implementation of ratio version of DemographicParity, EqualizedOdds, ErrorRate.
- Generalisation of utility function in the conditional selection rate.

### Implementation Details
Implementation of error ratio constraint which included the following:
- Ratio - Added ratio parameter in initialisation for the relaxation fraction allowed in the constraint. Defaults to 1 for all constraints (compatible with current moments).
- Utility - Added utility matrix which corresponds to g(X,A,Y,h(x)=0) and g(X,A,Y,h(x)=1) from the paper by Agarwal et al (A reduction approach to fair classification ). Defaults to h(x) for all constraints (compatible with current moments).

### Test
- Passed all unit tests in the repo, specifically, passed the smoke test test_smoke in test_exponentiatedgradient_smoke.py.
- Added new unit test cases for new constraints.

Coming from #332


fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -2,11 +2,15 @@
 # Licensed under the MIT License.
 
 import pandas as pd
+import numpy as np
 from .moment import ClassificationMoment
-from .moment import _GROUP_ID, _LABEL, _PREDICTION, _ALL, _EVENT, _SIGN
+from .moment import _GROUP_ID, _LABEL, _PREDICTION, _ALL, _EVENT, _SIGN, _POSITIVE_UTILITY, \
+    _NEGATIVE_UTILITY
+from fairlearn._input_validation import _MESSAGE_RATIO_NOT_IN_RANGE, _UTILITY_NOT_SPECIFIED
```
MiroDudik, line 9: Let's rename `_POSITIVE_UTILITY` -> `_UTILITY_FOR_1` and `_NEGATIVE_UTILITY` -> `_UTILITY_FOR_0` or something like that. Using positive and negative in this context is a bit confusing.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -29,17 +33,38 @@ class ConditionalSelectionRate(ClassificationMoment):
 
     """
 
+    def __init__(self, ratio=1.0):
+        """Intialise with the ratio value."""
+        super(ConditionalSelectionRate, self).__init__()
```
MiroDudik, line 38: ise -> ize [for the consistency with the rest of our documentation that uses American spelling]

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -29,17 +33,38 @@ class ConditionalSelectionRate(ClassificationMoment):
...
+    def load_data(self, X, y, event=None, utilities=None, **kwargs):
         """Load the specified data into this object.
 
         This adds a column `event` to the `tags` field.
+
+        The `utilities` is a matrix which correspond to g(X,A,Y,h(X)) as mentioned
+        in the paper `Agarwal et al. (2018) <https://arxiv.org/abs/1803.02453>`
+        The `utilities` defaults to h(X), i.e. [1, 0] for each X_i
+        The `_POSITIVE_UTILITY` is G^1 and the `_NEGATIVE_UTILITY` is G^0.
+        Assumes binary classification with labels 0/1.
```
MiroDudik, line 56: We should change the semantics, so the utility for 0 is in the first column and the utility for 1 is in the second column. This will make it consistent with the output of `predict_proba` for example.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -72,24 +97,29 @@ def load_data(self, X, y, event=None, **kwargs):
...
+                                                    _LOWER_BOUND_DIFF]])
         return g_signed
 
     # TODO: this can be further improved using the overcompleteness in group membership
     def project_lambda(self, lambda_vec):
         """Return the projected lambda values."""
-        lambda_pos = lambda_vec["+"] - lambda_vec["-"]
-        lambda_neg = -lambda_pos
+        lambda_pos = lambda_vec["+"] - self.ratio * lambda_vec["-"]
+        lambda_neg = - self.ratio * lambda_vec["+"] + lambda_vec["-"]
```
MiroDudik, line 122: This is not correct. I guess I never told you what's expected of `project_lambda`... it needs to return lambda which is guaranteed to lead to the same or higher value of the Lagrangian compared with `lambda_vec` for *all* possible choices of the classifier _h_.

So let's do the following. If `self.ratio != 1.0` then just return `lambda_vec` as given, and when `self.ratio == 1.0` then use the previous code and return `lambda_projected`.

MiroDudik: In the meantime, I think I figured what the right code should be. Please try this instead of lines 121--122:
```python
lambda_pos = lambda_vec["+"]
lambda_neg = lambda_vec["-"]
lambda_min = np.minimum(lambda_pos, lambda_neg)
lambda_max = np.maximum(lambda_pos, lambda_neg)
lambda_subtract = np.minimum(lambda_max, lambda_min/self.ratio)
lambda_pos = lambda_pos - lambda_subtract
lambda_neg = lambda_neg - lambda_subtract
```
See if things work with the above.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -29,17 +33,38 @@ class ConditionalSelectionRate(ClassificationMoment):
 
     """
 
+    def __init__(self, ratio=1.0):
+        """Intialise with the ratio value."""
+        super(ConditionalSelectionRate, self).__init__()
+        if ratio <= 0 or ratio > 1:
```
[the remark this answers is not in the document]
parul100495: Yeah. Ratio should reside in (0,1].

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -182,3 +216,45 @@ def load_data(self, X, y, **kwargs):
         super().load_data(X, y,
                           event=pd.Series(y).apply(lambda y: _LABEL + "=" + str(y)),
                           **kwargs)
+
+
+class ErrorRateRatio(ConditionalSelectionRate):
+    r"""Implementation of Error Rate Ratio as a moment.
+
```
MiroDudik, line 223: This description does not look correct currently... for example, the definition of `event` looks wrong.

romanlutz: This is what `prob_event` says right now
```
The `prob_event` :class:`pandas:pandas.DataFrame` will record the fraction of the samples corresponding to each unique value in the `Y` array.
```
but the event is `_ALL` so it should actually be 1.

fairlearn/reductions/_moments/conditional_selection_rate.py
parul100495: This is nice. I'll change `self.utilities` to have 2d-array-like format (similar to `predict_proba` format) as it definitely feels more intuitive. I'll do away with specific strings `_POSITIVE_UTILITY` and `NEGATIVE_UTILITY`.

fairlearn/reductions/_moments/conditional_selection_rate.py
parul100495: I will try this out and get back.
I'll also document the `project_lambda` for ease.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -29,17 +33,38 @@ class ConditionalSelectionRate(ClassificationMoment):
 
     """
 
+    def __init__(self, ratio=1.0):
```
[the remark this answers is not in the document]
MiroDudik: On one hand, I think it would be nice to just do `DemographicParity`, which would default to difference unless 'ratio' is specified.

On the other hand, we have examples such as `AccuracyScoreDifference` and `AccuracyScoreRatio` which would be confusing to handle by a class called `AccuracyScore` (and similarly for `ErrorRate`), because that strongly evokes the overall performance metric... so I think I'm in favor of `<Metric>Ratio` and `<Metric>Difference` approach.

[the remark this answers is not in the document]
MiroDudik: that said... let's hold off on the renaming for a brief moment and iron everything else first.

fairlearn/reductions/_moments/conditional_selection_rate.py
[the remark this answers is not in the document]
MiroDudik: agreed. this works for me. resolving.

fairlearn/reductions/_moments/conditional_selection_rate.py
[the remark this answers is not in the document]
romanlutz: That's what I thought. We can deprecate the existing "DemographicParity", etc. in favor of <metric>Difference, <metric>Ratio...

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -29,17 +32,35 @@ class ConditionalSelectionRate(ClassificationMoment):
...
         """Return the default objective for moments of this kind."""
         return ErrorRate()
 
-    def load_data(self, X, y, event=None, **kwargs):
+    def load_data(self, X, y, event=None, utilities=None, **kwargs):
         """Load the specified data into this object.
 
         This adds a column `event` to the `tags` field.
+
+        The `utilities` is a 2-d array which correspond to g(X,A,Y,h(X)) as mentioned in the paper
```
MiroDudik, line 51: This is not quite right. The array `utilities` should have the utilities in columns not in rows--similarly to how `X` is sharped and how multi-class labels `y` and `predict_proba` are treated in `sklearn`.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -72,26 +93,38 @@ def load_data(self, X, y, event=None, **kwargs):
...
+        Lagrangian compared with lambda_vec for all possible choices of the classifier, h.
+        """
+        lambda_pos = lambda_vec["+"]
+        lambda_neg = lambda_vec["-"]
+        lambda_min = np.minimum(lambda_pos, lambda_neg)
+        lambda_max = np.maximum(lambda_pos, lambda_neg)
+        lambda_subtract = np.minimum(lambda_max, lambda_min / self.ratio)
+        lambda_pos = lambda_pos - lambda_subtract
+        lambda_neg = lambda_neg - lambda_subtract
         lambda_projected = pd.concat([lambda_pos, lambda_neg],
```
MiroDudik, line 128: Please put the deleted lines 93 & 94 after the current line 127, i.e.:
```
        lambda_pos[lambda_pos < 0.0] = 0.0
        lambda_neg[lambda_neg < 0.0] = 0.0
```

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -182,3 +219,35 @@ def load_data(self, X, y, **kwargs):
...
+    r"""Implementation of Error Rate Ratio as a moment.
+
+    Measures the ratio in errors per attribute by overall error.
+    The 2-sided version of error ratio can be written as
+    ratio <= error(A=a) / total_error <= 1/ratio
+    .. math::
+    ratio <= E[abs(h(x) - y)| A = a] / E[abs(h(x) - y)] <= 1/ratio\; \forall a
+
+    This implementation of :class:`ConditionalSelectionRate` defines a single event, `all`.
+    Consequently, the `prob_event` :class:`pandas:pandas.DataFrame` will only have a single
```
MiroDudik, line 234: Drop the class designation or fix to `pandas.Series`

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -182,3 +219,35 @@ def load_data(self, X, y, **kwargs):
...
+    entry, which will be equal to 1.
+
+    The `index` property will have twice as many entries (corresponding to the Lagrange multipliers
+    for positive and negative constraints) as there are unique values for the sensitive feature.
+
+    The :meth:`signed_weights` method will compute the costs according to Example 3 of
+    `Agarwal et al. (2018) <https://arxiv.org/abs/1803.02453>`_.
```
MiroDudik, line 241: "similar to Example 3" and at the end add something like "but with g=abs(h(x)-y) rather than g=h(x)".

fairlearn/reductions/_moments/conditional_selection_rate.py
parul100495: Fixed to `pandas.Series`

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -182,3 +219,35 @@ def load_data(self, X, y, **kwargs):
...
+
+    The `ratio` defines the amount of relaxation that is allowed for the constraint.
+    The value varies in the range (0,1]. The ratio of 1 means, the constraint is given
+    no relaxation and thus, the constraint tries to evaluate for
+    error(A = a) / total_error = 1
+    """
```
MiroDudik, line 247: This is not quite right. How about something like this:
```
    The `ratio` specifies the multiple at which error(A = a) should be compared with total_error and vice
    versa. The value of `ratio` has to be in the range (0,1] with smaller values corresponding to weaker
    constrain. The `ratio` equal to 1 corresponds to the constraint
    error(A = a) = total_error
```

fairlearn/reductions/_moments/conditional_selection_rate.py
parul100495: This doesn't work right. Example: When I try to run test data with this code for `ErrorRateRatio` with `eps=0.05`, I get `error=0.75` and `disparity=-0.042857` which is incorrect. Ideally, for negative disparity, it should give minimum error.

I have reverted the change to original code for now.

romanlutz:
From my perspective this looks good to go. We should think about adding unit tests for the moments themselves (in addition to the smoke tests for expgrad with those moments), but I'm happy to defer that to its own PR if @MiroDudik agrees.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -72,22 +97,31 @@ def load_data(self, X, y, event=None, **kwargs):
...
 
     # TODO: this can be further improved using the overcompleteness in group membership
     def project_lambda(self, lambda_vec):
-        """Return the projected lambda values."""
+        """Return the projected lambda values.
+
+        i.e., returns lambda which is guaranteed to lead to the same or higher value of the
+        Lagrangian compared with lambda_vec for all possible choices of the classifier, h.
+        """
         lambda_pos = lambda_vec["+"] - lambda_vec["-"]
```
MiroDudik, line 125: I think my feedback was confusing here. The current version is not correct for the ratio, please go with the following:
```python
        lambda_pos = lambda_vec["+"]
        lambda_neg = lambda_vec["-"]
        lambda_min = np.minimum(lambda_pos, lambda_neg)
        lambda_max = np.maximum(lambda_pos, lambda_neg)
        lambda_subtract = np.minimum(lambda_max, lambda_min / self.ratio)
        lambda_pos = lambda_pos - lambda_subtract
        lambda_neg = lambda_neg - lambda_subtract
        lambda_pos[lambda_pos < 0.0] = 0.0
        lambda_neg[lambda_neg < 0.0] = 0.0
```

parul100495: As discussed offline, kept the code as is for `self.ratio == 1` and returning `lambda_vec` otherwise.

fairlearn/reductions/_moments/conditional_selection_rate.py
romanlutz: This was addressed after today's discussion. Thanks!

romanlutz:
For some reason the PR-Gate shows as "waiting" here, but I double-checked with the "Checks" page and they ran successfully. Completing since all the concerns were addressed and Miro signed off.

Issue #316, opened 2020-03-04
Improve ExponentiatedGradient Testing


riedgar-ms:
The smoke tests for `ExponentiatedGradient` are overly self-validating. In particular, when the [expected disparity and error are calculated by the test](https://github.com/fairlearn/fairlearn/blob/23d535a628856ba4dec37a09096355977b44a694/test/unit/reductions/exponentiated_gradient/test_exponentiatedgradient_smoke.py#L87), the `Moment` objects are used. Since the `Moment` objects drive `ExponentiatedGradient` itself, these checks only verify consistency, not actual correctness.

The tests should be augmented to compute disparity and errors using the `metrics` portion of the package, and then to compare with the result from `ExponentiatedGradient`

riedgar-ms:
There are more issues with the testing of `ExponentiatedGradient` - basically, we have smoke tests, but we do not test enough. For example, if some portion of the algorithm were to be tweaked, we do not have tests (or benchmarking) to assess whether the tweak has improved the performance (where performance encompasses accuracy and correctness, as well as speed).

TamaraAtanasoska:
cc @StefanieSenger (just FIY that this was discussed multiple times), these kinds of tests would be lovely for the whole library and especially for the reductions module.

Issue #61, opened 2019-10-02
ExponentiatedGradient needs comprehensive tests


romanlutz:
The current set of tests is too limited. We need to extend that asap.

UPDATE: argument tests are now there thanks to @riedgar-ms but we still need functional tests

riedgar-ms:
Subtask - #316

romanlutz:
@riedgar-ms I've committed a PR for this, but won't get to more for the next few weeks. If you want to wait for this with a release, your release may not happen for a long time...

romanlutz:
This is represented through #316 so closing this issue

PR #372, opened 2020-04-13
Add unit tests for exponentiated gradient

romanlutz:
A first set of unit tests to complement the existing smoke and argument tests. The unit tests are only for two of the methods so far. In the process of creating these I also 
- simplified the code in _lagrangian.py by merging `_eval` and `_eval_from_error_gamma` after observing that the `callable(h)` case in `_eval` was not reachable (probably a remnant of the earlier implementation).
- added the `_call_oracle` method since it made my life easier and it represented a logical unit within the code.
- moved `_get_data` into a test_utilities file so that I can use it across all exponentiated gradient test files.
- added `pytest-mock` as a requirement since I'm using it in the tests.

TODO:
- clarify the open questions with @MiroDudik (see TODO in the code)

More tests will be added in the future. I would prefer keeping these PRs short.

Relevant for #61 and #316


test/unit/reductions/exponentiated_gradient/test_lagrangian.py
```diff
@@ -0,0 +1,107 @@
...
+        'X2': 0.457627,
+        'X3': 0.021186})
+    assert (np.isclose(fitted_estimator.weights, expected_estimator_weights, atol=1.e-6)).all()
+
+
+@pytest.mark.parametrize("Constraints", [DemographicParity, EqualizedOdds])
+@pytest.mark.parametrize("eps", [0.001, 0.01, 0.1])
+def test_call_oracle(Constraints, eps, mocker):
+    X, y, A = _get_data(A_two_dim=False)
+    estimator = mocker.MagicMock()
```
riedgar-ms, line 73: Why do we need a mock? I don't actually see any behaviour being set for the mock, beyond the pickle (and that might be something we should just change to a `copy.deepcopy()` internally anyway).

test/unit/reductions/exponentiated_gradient/test_lagrangian.py
```diff
@@ -0,0 +1,107 @@
...
+    constraints = Constraints()
+
+    # ExponentiatedGradient pickles and unpickles the estimator, which isn't possible for the mock
+    # object, so we mock that process as well.
+    import pickle
```
riedgar-ms, line 78: Shouldn't this be at the top of the file?

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -56,39 +56,39 @@ def __init__(self, X, sensitive_features, y, estimator, constraints, eps, B, opt
         self.last_linprog_n_hs = 0
         self.last_linprog_result = None
 
-    def _eval_from_error_gamma(self, error, gamma, lambda_vec):
+    def _eval(self, h, lambda_vec):
         """Return the value of the Lagrangian.
```
riedgar-ms, line 60: This is a docstring which could use some expansion :-)

test/unit/reductions/exponentiated_gradient/test_lagrangian.py
romanlutz: I don't actually want to do anything with pickle here, so it's there to stress that it's just for this. I guess we can move it anywhere.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
romanlutz: Sure, but that's not the point of this PR.

test/unit/reductions/exponentiated_gradient/test_lagrangian.py
romanlutz: The point of this test is that we don't care what estimator you pass in. It just has to be a thing with a fit method that is called exactly once. This is the minimal way to test this. The point is exactly that I don't need any behavior. I just want that fit method to be called and have that be verifiable at the end with my asserts.

test/unit/reductions/exponentiated_gradient/test_lagrangian.py
riedgar-ms: I'm sure I've seen linters scream at imports anywhere than the top of the file.... although `flake8` does not appear to be complaining for now.

test/unit/reductions/exponentiated_gradient/test_lagrangian.py
riedgar-ms: OK, I think I get it now (and yes, because you only call `fit()` there's no return value to be specified), and I do see the asserts later.

In general, mocks make me nervous (although this one is benign), since specifying their behaviour is another source of bugs and incorrect tests. Could you add a comment explaining exactly what the mock is for, and the behaviour which is checked (namely that the mock has had its `fit()` method invoked with the correct arguments; there is no return value from that call).

test/unit/reductions/exponentiated_gradient/test_lagrangian.py
romanlutz: Actually I wonder why it didn't complain... anyway it's changed now

test/unit/reductions/exponentiated_gradient/test_lagrangian.py
romanlutz: I documented that further as suggested.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -56,39 +56,39 @@ def __init__(self, X, sensitive_features, y, estimator, constraints, eps, B, opt
...
+        :type h: pandas.Series
+        :param lambda_vec: lambda vector
+        :type h: pandas.Series
+
+        :return: tuple `(L, L_high, gamma, error)` where `L` is the value of the Lagrangian,
+            `L_high` is the value of the Lagrangian under the best response of the lambda player,
+            `gamma` is the vector of constraint violations, and `error` is the empirical error
         """
-        lambda_projected = self.constraints.project_lambda(lambda_vec)
+        error = self.errors[h.index].dot(h)
```
MiroDudik, line 72: Let's revert this piece of code to what it was previously, meaning:
```python
        if callable(h):
            error = self.obj.gamma(h)[0]
            gamma = self.constraints.gamma(h)
        else:
            error = self.errors[h.index].dot(h)
            gamma = self.gammas[h.index].dot(h)
```
I think that the evaluation of Lagrangian should work for both h and Q.

MiroDudik: And you can of course update the doc string at the top appropriately--that h can be either a callable or a Series representing Q.

test/unit/reductions/exponentiated_gradient/test_exponentiatedgradient_arguments.py
```diff
@@ -75,3 +63,8 @@ def Q(X): return expgrad._pmf_predict(X)[:, 1]
         assert error == pytest.approx(0.25, abs=_PRECISION)
         assert expgrad._n_oracle_calls == 32
         assert n_predictors == 3
+
+        # ensure that the input data wasn't changed by our mitigator
+        assert isinstance(transformed_X, type(transformX(X)))
+        assert isinstance(transformed_y, type(transformY(y)))
+        assert isinstance(transformed_A, type(transformA(A)))
```
MiroDudik, line 70: I don't think that these three added asserts will ever throw an exception. Is the intention to ensure that we don't apply some validation routine that will turn a list into a Series or a DataFrame? I don't think we would detect that here. That's because python passes around either immutable types or pointers, but there's no calling by reference.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
romanlutz: Are you aware that it's never called with callable(h)? It's a dead code path.

test/unit/reductions/exponentiated_gradient/test_exponentiatedgradient_arguments.py
romanlutz: Yes, not sure how I thought that would work. There's a different way to test that, though. Let me rewrite this.

romanlutz: I changed this. Instead I'm now checking for the data type of the args to estimator.fit(...)
LMK what you think!

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
MiroDudik: I know it's a dead code right now, but let's keep it--we could even use it in tests with a callable rather than Q.

riedgar-ms: If we want to keep it around, it would be best that there's a test which touches it somehow. Otherwise, there's a tendency for the code to 'drift' and you eventually find out that the code which you've been keeping around 'in case it's useful' doesn't actually work any more.

test/unit/reductions/exponentiated_gradient/test_lagrangian.py
```diff
@@ -0,0 +1,110 @@
...
+    mocker.patch('pickle.dumps')
+    pickle.loads = mocker.MagicMock(return_value=estimator)
+
+    lagrangian = _Lagrangian(X, A, y, estimator, constraints, eps, 1/eps)
+
+    # set up initial lambda vector based on a 0-initialized theta
+    theta = pd.Series(0, lagrangian.constraints.index)
+    lambda_vec = np.exp(theta) / (1 + np.exp(theta).sum())
+
+    signed_weights = lagrangian.obj.signed_weights() + \
```
MiroDudik, line 91: Maybe this is okay, but I think it would be better if `theta` and `signed_weights` did not have a dependence on `lagrangian`. This only tests that the oracle call is "internally" consistent within the lagrangian logic, but it will not catch that `signed_weights` are incorrect. A better approach would be to get `signed_weights` from a separate copy of Constraints() and the default objective (this is one of the fields of Moment).

test/unit/reductions/exponentiated_gradient/test_lagrangian.py
```diff
@@ -0,0 +1,110 @@
...
+    theta = pd.Series(0, lagrangian.constraints.index)
+    lambda_vec = np.exp(theta) / (1 + np.exp(theta).sum())
+
+    # call oracle to determine error and gamma and calculate exp
+    fitted_estimator = lagrangian._call_oracle(lambda_vec)
+    def h(X): return fitted_estimator.predict(X)
+    best_h_error = lagrangian.obj.gamma(h)[0]
+    best_h_gamma = lagrangian.constraints.gamma(h)
+
+    # opt_lambda affects only the calculation of L
```
MiroDudik, line 40: The current code tests just the internal logic of the lagrangian. Maybe that's what's intended? However two points:
* when `opt_lambda=False`, then there is a correct value of the lagrangian, so you should maybe just check that value
* when `opt_lambda=True`, then this is probably the best you can do, since the only requirement is that `projected_lambda` is an improvement (but it's left unspecified which of many possible improvements to choose)

test/unit/reductions/exponentiated_gradient/test_lagrangian.py
```diff
@@ -0,0 +1,110 @@
...
+
+    # call _eval to get the desired results L, L_high, gamma, error
+    L, L_high, gamma, error = lagrangian._eval(h_vec, lambda_vec)
+
+    assert L == L_expected
+    assert L_high == L_high_expected
+    assert error == 0.25
+    assert (gamma == best_h_gamma).all()
+
+    # in this particular example the estimator is always the same
```
MiroDudik, line 62: this check should go first... before all other checks.

test/unit/reductions/exponentiated_gradient/test_lagrangian.py
romanlutz: I'd like to explain what `opt_lambda` is for through a comment.
It seems to only decide whether or not we use `project_lambda` or just the `lambda_vec`. Does `opt_lambda` stand for "optimal lambda", or "optimize lambda", or perhaps something else? I'd like to either rename it or add a comment to explain what it is.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -56,39 +56,39 @@ def __init__(self, X, sensitive_features, y, estimator, constraints, eps, B, opt
...
         self.last_linprog_result = None
 
-    def _eval_from_error_gamma(self, error, gamma, lambda_vec):
+    def _eval(self, h, lambda_vec):
         """Return the value of the Lagrangian.
 
-        :return: tuple `(L, L_high)` where `L` is the value of the Lagrangian, and
-            `L_high` is the value of the Lagrangian under the best response of the lambda player
-        :rtype: tuple of two floats
+        :param h: TODO: should this be Q instead of h? It seems to indicate how much weight each h
```
romanlutz, line 62: @MiroDudik see the question in the code:
:param h: TODO: should this be Q instead of h? It seems to indicate how much weight each h gets in the randomized classifier, summing up to 1

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -56,39 +56,39 @@ def __init__(self, X, sensitive_features, y, estimator, constraints, eps, B, opt
...
+
+        :return: tuple `(L, L_high, gamma, error)` where `L` is the value of the Lagrangian,
+            `L_high` is the value of the Lagrangian under the best response of the lambda player,
+            `gamma` is the vector of constraint violations, and `error` is the empirical error
         """
-        lambda_projected = self.constraints.project_lambda(lambda_vec)
+        error = self.errors[h.index].dot(h)
+        gamma = self.gammas[h.index].dot(h)
+
+        # TODO: some entries in gamma can be negative, presumably because for certain sensitive
```
romanlutz, line 75: @MiroDudik  see the question in the code:
some entries in gamma can be negative, presumably because for certain sensitive feature values the (one-sided) constraint is satisfied and then some. However, below we multiply the lambda vector with it to calculate the Lagrangian value L, so negative values should reduce that. Is that intended? Just trying to make sure that's not unintended.

test/unit/reductions/exponentiated_gradient/test_lagrangian.py
romanlutz: We need some sort of test for the Lagrangian, otherwise we'll go into that code later, make tweaks and it'll behave differently. I'm not too happy with the tests either, but I couldn't think of anything better.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
romanlutz: added back + tests for the callable(h) case

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
MiroDudik: Correct. The name of the argument should be Q, and the type should be either: Series with weights that sum to 1, or a callable (corresponding to a *deterministic* predict).

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
MiroDudik: Yes--it is totally fine for gamma to be negative. The formulas below are correct.

test/unit/reductions/exponentiated_gradient/test_lagrangian.py
MiroDudik: Re. `opt_lambda`--put in a comment along the lines "`opt_lambda` indicates whether to optimize lambda during the calculation of Lagrangian". The part of the code with `opt_lambda` is going to change shortly, so don't invest too much time on that part of the code.

MiroDudik: Re. testing of the Lagrangian: sure--leave as is for now. I might tweak at some indefinite point in future.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
romanlutz: Great, adjusted that

Issue #371, opened 2020-04-10
Simplify Release Pipeline


riedgar-ms:
Our release pipeline currently release first to Test-PyPI and then to PyPI itself. In order to avoid 'burning' version numbers in Test-PyPI (because PyPI does not separate 'upload package' from 'publish package' and once a version is published, it can never be replaced), there is some elaborate machinery using a `DEV_VERSION` pipeline variable to append `.dev[n]` to the package uploaded to Test-PyPI.

This is confusing, and actually doesn't get us much benefit.

@adrinjalali  has pointed out that once the wheel is created, with can just `pip install` that file, and run our validation tests against it. If we do this, we can dump all the machinery associated with `DEV_VERSION` which requires some supporting Python scripts for the build and also infects `setup.py`. We would skip the upload to Test-PyPI entirely (or perhaps make it available, if we wanted to keep versions on Test-PyPI and PyPI in sync... although we might be unique in that).

A final (minor) benefit is that we would also then recover the `.dev[n]` option for mainline versioning.

riedgar-ms:
The new pipeline is in place, and tested against Test-PyPI. Once we have a successful PROD release, I'll clean out the old pipeline and close this issue.

PR #375, opened 2020-04-16
Improve pipeline job generation

riedgar-ms:
Refactor the job generation portions of the pipelines, so that more than the Python version is automatically expanded. The new system generates a job matrix which is a cartesian product over platforms, test runs and Python versions.

This is related to #371, but does not yet close that issue.

### DIFF
--- a/devops/pypi-release.yml
+++ b/devops/pypi-release.yml
@@ -31,19 +31,9 @@ stages:
     
   jobs:
   - template: templates/all-tests-job-template.yml
-      parameters:
-      platforms:  { Linux: ubuntu-latest, Windows: vs2017-win2016 }
-      installationType: 'PipLocal'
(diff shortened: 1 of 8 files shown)


devops/pypi-release.yml
```diff
@@ -32,53 +32,20 @@ stages:
...
       pyVersions: [3.5, 3.6, 3.7, 3.8]
       freezeArtifactStem: "PreDeployment"
       freezeFileStem: $(FreezeFileStem)
       
   - template: templates/all-tests-job-template.yml
     parameters:
-      platform: MacOS
-      vmImage:  'macos-latest'
-      testRunType: 'Unit'
+      platforms: { MacOS: macos-latest }
```
romanlutz, line 43: Why is this separate?

riedgar-ms: Because it's only doing the Unit tests

PR #377, opened 2020-04-17
Enable GridSearch with more than 2 sensitive feature values

### LINKED ISSUE #376: Enable GridSearch with more than 2 sensitive feature values
GridSearch can't be run with more than 2 sensitive feature values due to a check in the validation stage.

Enable GridSearch for >2 sensitive feature values.

romanlutz:
Closes #376 by removing the check for the number of sensitive features during input validation. To ensure that users still know that they shouldn't use GridSearch with many groups to avoid high-dimensional grids (which would be extremely costly to search thoroughly) we add a warning in the grid generation if `true_dim > 4` as an indication that the grid may get large. At the end of the day it should be up to the user to decide whether they want to explore it anyway, and this solution allows for that.

### DIFF
--- a/fairlearn/_input_validation.py
+++ b/fairlearn/_input_validation.py
@@ -20,8 +20,6 @@ _MESSAGE_RATIO_NOT_IN_RANGE = "ratio must lie between (0,1]"
 _INPUT_DATA_FORMAT_ERROR_MESSAGE = "The only allowed input data formats for {} are: {}. " \
                                      "Your provided data was of type {}."
 _EMPTY_INPUT_ERROR_MESSAGE = "At least one of sensitive_features, labels, or scores are empty."
-_SENSITIVE_FEATURES_THRESHOLD_ERROR_TEMPLATE = "Sensitive features contain more than {} unique" \
-                                               " values"
 _LABELS_NOT_0_1_ERROR_MESSAGE = "Supplied y labels are not 0 or 1"
 _MORE_THAN_ONE_COLUMN_ERROR_MESSAGE = "{} is a {} with more than one column"
 _NOT_ALLOWED_TYPE_ERROR_MESSAGE = "{} is not an ndarray, Series or DataFrame"
@@ -35,8 +33,7 @@ _ALLOWED_INPUT_TYPES_Y = [np.ndarray, pd.DataFrame, pd.Series, list]
 _SENSITIVE_FEATURE_COMPRESSION_SEPARATOR = ","
 
 
-def _validate_and_reformat_input(X, y=None, expect_y=True, enforce_binary_labels=False, 
-                                 group_warn_threshold=None, **kwargs):
+def _validate_and_reformat_input(X, y=None, expect_y=True, enforce_binary_labels=False, **kwargs):
     """Validate input data and return the data in an appropriate format.
 
     :param X: The feature matrix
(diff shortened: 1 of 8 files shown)


fairlearn/_input_validation.py
```diff
@@ -42,12 +45,12 @@ def _validate_and_reformat_input(X, y=None, expect_y=True, enforce_binary_sensit
...
     :param expect_y: if True y needs to be provided, otherwise ignores the argument; default True
     :type expect_y: bool
-    :param enforce_binary_sensitive_feature: if True raise exception if there are more than two
-        distinct values in the `sensitive_features` data from `kwargs`; default False
-    :type enforce_binary_sensitive_feature: bool
     :param enforce_binary_labels: if True raise exception if there are more than two distinct
         values in the `y` data; default False
     :type enforce_binary_labels: bool
+    :param group_warn_threshold: threshold on the number of groups as defined by sensitive
+        features above which a warning is displayed; default None
```
riedgar-ms, line 52: The comment should say what happens for None

fairlearn/_input_validation.py
```diff
@@ -82,9 +85,10 @@ def _validate_and_reformat_input(X, y=None, expect_y=True, enforce_binary_sensit
         sensitive_features = \
             _compress_multiple_sensitive_features_into_single_column(sensitive_features)
 
-    if enforce_binary_sensitive_feature:
-        if len(np.unique(sensitive_features)) > 2:
-            raise ValueError(_SENSITIVE_FEATURES_NON_BINARY_ERROR_MESSAGE)
+    if group_warn_threshold is not None:
```
riedgar-ms, line 88: I think it should default to 2, so we keep the previous behaviour as a default.

riedgar-ms:
Could you make the description a bit more descriptive? Just pointing at an issue isn't really ideal.

riedgar-ms:
Could you also add a smoke test with a ternary sensitive feature?

fairlearn/_input_validation.py
MiroDudik: I don't think we should check for the number of sensitive feature values; instead, we should add a warning if `true_dim>4` on:
https://github.com/fairlearn/fairlearn/blob/78d18451afadc1365af47e8b5a502c28c2a51db4/fairlearn/reductions/_grid_search/_grid_generator.py#L22
(we could also consider a more sophisticated condition that would be based on both the grid size and the grid dimension, but I'm not exactly sure what it should be).

fairlearn/_input_validation.py
romanlutz: Removed this after feedback from Miro. The warning now comes from grid generator

fairlearn/_input_validation.py
romanlutz: OK, I adjusted it according to your specification @MiroDudik . Resolving this since it's based on an old version of the code.

romanlutz:
> 
> 
> Could you also add a smoke test with a ternary sensitive feature?

Sounds good. I've extended the existing smoke tests to cover various scenarios with up to 10 sensitive feature values.

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -194,6 +194,7 @@ def predict(self, X):
         :rtype: Scalar or vector
         """
         positive_probs = self._pmf_predict(X)[:, 1]
+        print(positive_probs)
```
MiroDudik, line 197: you probably meant to remove those print statements here.

fairlearn/reductions/_grid_search/_grid_generator.py
```diff
@@ -1,10 +1,22 @@
...
 import numpy as np
 import pandas as pd
 
 
+logger = logging.getLogger(__name__)
+
+GRID_DIMENSION_WARN_THRESHOLD = 4
+GRID_DIMENSION_WARN_TEMPLATE = "The grid has {} dimensions. It is not recommended to use more " \
+                               "than {}, otherwise a prohibitively large grid size is required " \
+                               "to explore the grid thoroughly. For such cases consider using " \
```
MiroDudik, line 14: "to explore the grid" -> "to explore the space"

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
romanlutz: Yes! Thank you

PR #379, opened 2020-04-17
Continue implementing the metrics API proposal

MiroDudik:


fairlearn/metrics/__init__.py
```diff
@@ -18,81 +18,71 @@
...
-from ._skm_wrappers import group_median_absolute_error  # noqa: F401
-from ._skm_wrappers import group_root_mean_squared_error  # noqa: F401
-from ._skm_wrappers import group_r2_score  # noqa: F401
-
-from ._metrics_engine import make_group_metric, group_summary  # noqa: F401
-from ._metrics_engine import group_min_from_summary, group_max_from_summary  # noqa: F401
-from ._metrics_engine import difference_from_summary, ratio_from_summary  # noqa: F401
-
-# -------------------------------------------
+from ._extra_metrics import (  # noqa: F401
```
riedgar-ms, line 21: I didn't know about this syntax... :-)

fairlearn/metrics/__init__.py
```diff
@@ -18,81 +18,71 @@
...
+)
+
 
 _extra_metrics = [
+    "true_positive_rate",
+    "true_negative_rate",
+    "false_positive_rate",
+    "false_negative_rate",
     "balanced_root_mean_squared_error",
-    "fallout_rate",
```
riedgar-ms, line 54: We do need to get a bit better about removing things from our API. I know that we're still <1.0.0, but I think we do it a bit too often.

fairlearn/metrics/__init__.py
romanlutz: @riedgar-ms  there are at least 3 different ways you can do imports ;-)

fairlearn/metrics/__init__.py
```diff
@@ -18,81 +18,71 @@
...
+    "make_derived_metric",
     "group_min_from_summary",
     "group_max_from_summary",
     "difference_from_summary",
     "ratio_from_summary"
 ]
 
 
-__all__ = _engine + _extra_metrics + _group_metrics
+globals().update(_metric_group_summary_dict)
```
romanlutz, line 71: why is this required?

fairlearn/metrics/_disparities.py
```diff
@@ -0,0 +1,42 @@
...
+from ._metrics_engine import (
+    selection_rate_difference,
+    selection_rate_ratio,
+    true_positive_rate_difference,
+    true_positive_rate_ratio,
+    false_positive_rate_difference,
+    false_positive_rate_ratio)
+
+
+def demographic_parity_difference(y_true, y_pred, *, sensitive_features, sample_weight=None):
```
riedgar-ms, line 15: Do we want to keep these `*` in the API? I'm sure I had a good reason for putting them there in the first place, but I can't quite remember it right now.

fairlearn/metrics/_disparities.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def demographic_parity_ratio(y_true, y_pred, *, sensitive_features, sample_weight=None):
+    r"""Calculate the demographic parity ratio."""
+    return selection_rate_ratio(
+        y_true, y_pred, sensitive_features=sensitive_features, sample_weight=sample_weight)
+
+
+def equalized_odds_difference(y_true, y_pred, *, sensitive_features, sample_weight=None):
+    r"""Calculate the demographic parity difference."""
```
riedgar-ms, line 28: This docstring appears incorrect.

fairlearn/metrics/_disparities.py
```diff
@@ -0,0 +1,42 @@
...
+    return max(
+        true_positive_rate_difference(
+            y_true, y_pred, sensitive_features=sensitive_features, sample_weight=sample_weight),
+        false_positive_rate_difference(
```
riedgar-ms, line 32: Should the docstring be explicit in how this method reduces the two comparisons (TP and FP) to a single number?

fairlearn/metrics/_metrics_engine.py
```diff
@@ -167,3 +258,66 @@ def group_max_from_summary(summary):
...
+                metric_group_summary,
+                name=derived_metric_name)
+            derived_metric_dict[derived_metric_name] = derived_metric
+
+    return metric_group_summary_dict, derived_metric_dict
+
+
+_metric_group_summary_dict, _derived_metric_dict = _derive_metrics(METRICS_SPEC)
+
+globals().update(_metric_group_summary_dict)
```
riedgar-ms, line 322: I'm not familiar with this bit of Python at all, but I'm guessing that the above generates a large set of 'functions' which can then be called?

test/unit/metrics/sample_dashboards/bc-2p-3f.json
```diff
@@ -450,19 +450,19 @@
...
                 },
                 "miss_rate": {
                     "bins": [
-                        0.6470588235294117,
-                        0.5833333333333333,
+                        0.6470588235294118,
+                        0.5833333333333334,
                         0.25,
-                        0.4545454545454546
+                        0.45454545454545453
```
riedgar-ms, line 465: It would appear that I unwisely assumed I didn't need `pytest.approx()` in a few tests....

fairlearn/metrics/__init__.py
romanlutz: and each of those changes requires an update in CHANGES.md

riedgar-ms:
This also needs more extensive docstrings on the new methods

fairlearn/metrics/_metrics_engine.py
riedgar-ms: If so, how do we document them?

fairlearn/metrics/_disparities.py
romanlutz: It means keyword-name is required, so you can't write
`func(a, b, c)`
but instead you have to write
`func(a, b, c=c)`
We all agreed that's a good idea months ago when the project started, at least for sensitive_features since that's the part that's really different from other libraries like sklearn

riedgar-ms: I knew I'd had a reason ;-)

fairlearn/metrics/__init__.py
MiroDudik: I know--I'm really hoping to get all of the metrics changes into the current version. In addition to the above, there are two metrics whose API I'm currently hesitating about: `mean_{over,under}prediction`, but we need them for the dashboard, so I'm going to prepend their names with an underscore to give us some flexibility.

fairlearn/metrics/__init__.py
```diff
@@ -18,81 +18,71 @@
...
+
+from ._metrics_engine import (  # noqa: F401
+    make_metric_group_summary, group_summary,
+    make_derived_metric,
+    group_min_from_summary, group_max_from_summary,
+    difference_from_summary, ratio_from_summary,
+    _metric_group_summary_dict, _derived_metric_dict)
+
+from ._disparities import (  # noqa: F401
+    demographic_parity_difference,
```
riedgar-ms, line 40: Something (slightly off-topic) which came out of an AzureML-related review.... is 'demographic parity' a good name, given that different people might understand different things by 'demographic' (much the same way as we steer away from 'bias' in our writing?

We did explain that it's a term used in the literature, but it's something to consider. A more neutral term would be 'selection rate parity' I believe.

fairlearn/metrics/__init__.py
riedgar-ms: Related to this, I've proposed on Gitter that because of these changes (and others) we bump ourselves to v0.5.0. Not to be done in this PR (and I don't have a release timeframe in mind), but do you have thoughts on that?

fairlearn/metrics/__init__.py
MiroDudik: This adds all the metrics of the form `<metric>_group_summary` to the package module and similarly the next line adds `<metric>_{difference,ratio,group_min,group_max}`. I'll add a comment to explain. Let me know if you are aware of a more idiomatic way to do this. (But that could be done in a future PR too.)

fairlearn/metrics/_metrics_engine.py
MiroDudik: I am not sure that we want to generate the documentation of the derived metrics automatically (at least not yet). For now, I was thinking that our documentation should just include a table similar to the one at the end of the [API metrics proposal](https://github.com/fairlearn/fairlearn-proposals/blob/master/api/METRICS.md), but with links to the docs for the base metrics.

test/unit/metrics/sample_dashboards/bc-2p-3f.json
MiroDudik: I looked into whether I can use pytest.approx(), but it doesn't seem work with deep data structures, so I just changed the numbers. Maybe there's a simple tweak--that involves flattening json or invoking an approximate comparison on deep structures--but I'm going to leave that for a future PR.

fairlearn/metrics/__init__.py
MiroDudik: It makes sense to me. I had the same thought.

fairlearn/widget/_fairlearn_dashboard.py
```diff
@@ -49,91 +63,83 @@ def __init__(
         self._metric_methods = {
             "accuracy_score": {
                 "model_type": ["classification"],
-                "function": group_accuracy_score
+                "function": accuracy_score_group_summary
```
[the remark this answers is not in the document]
MiroDudik: They shouldn't. I have made no changes with the dashboard key strings (just with how they are calculated). I have also built a new wheel with the dashboard and made sure it works for all metrics across our three task types.

MiroDudik:
Agreed that we should bump to v0.5.0 with the next release.

fairlearn/widget/_fairlearn_dashboard.py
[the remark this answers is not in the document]
riedgar-ms: I'm not certain that building a new wheel is sufficient (we've been burned by that before), however, I do agree that `accuracy_score` is probably the important string in the highlighted section.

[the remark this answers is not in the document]
MiroDudik: Let me know if you can think of another way how to test things locally. if we run into any new issues, we will update pre-release notes.

fairlearn/metrics/_disparities.py
MiroDudik: sounds good.

CHANGES.md
```diff
@@ -1,9 +1,25 @@
 # Changes
 
 ### v0.4.6
-* Refactoring of metrics:
+* Refactor metrics:
   * Remove `GroupMetricResult` type in favour of a `Bunch`.
```
romanlutz, line 5: nit: if you make other changes as well can we make this favor instead of favour?

romanlutz: otherwise ignore

fairlearn/metrics/__init__.py
romanlutz: So this is to avoid listing all of them individually?
I've never seen this before, and I wonder whether there are any negative consequences of doing this.
Perhaps @adrinjalali has thoughts?

test/unit/metrics/sample_dashboards/bc-2p-3f.json
romanlutz: Something like
```
np.isclose(A, B).all()
```
may work

fairlearn/metrics/__init__.py
riedgar-ms: Discussed separately. This has been talked about extensively, and 'demographic parity' found to be an appropriate term.

fairlearn/metrics/_disparities.py
```diff
@@ -0,0 +1,86 @@
...
+
+
+def equalized_odds_difference(y_true, y_pred, *, sensitive_features, sample_weight=None):
+    """Calculate the equalized odds difference.
+
+    :param 1D-array y_true: Ground truth (correct) labels :math:`Y`.
+    :param 1D-array y_pred: Predicted labels :math:`h(X)` returned by the classifier.
+    :param 1D-array sensitive_features: Sensitive features.
+    :param 1D-array sample_weight: Sample weights.
+    :return: The greater of two metrics: `true_positive_rate_difference` and
```
riedgar-ms, line 52: I'd tend to put larger blocks of text like this up in the main description, and keep the parameters/return text shorter. Given the amoutn of reworking which is about to hit our API documentation, this isn't an issue for this checkin.

CHANGES.md
```diff
@@ -1,9 +1,25 @@
...
+  * base metrics `{true,false}_{positive,negative}_rate`
+  * group summary metrics `<metric>_group_summary`
+  * derived metrics `<metric>_{difference,ratio,group_min,group_max}`
+  * disparity metrics `{demographic_parity,equalized_odds}_{difference,ratio}`
+* Remove metrics:
+  * `fallout_rate` in favor of `false_positive_rate`
+  * `miss_rate` in favor of `false_negative_rate`
+  * `specificity_score` in favor of `true_negative_rate`
+* Change from public to private:
+  * `mean_{over,under}prediction` changed to `_mean_{over,under}prediction`
```
riedgar-ms, line 22: Can I ask why, given that (I believe) these are used by the Dashboard?

fairlearn/metrics/_metrics_engine.py
```diff
@@ -82,33 +95,114 @@ def _check_metric_params(y_true, metric_params,
...
 
-    # Improve the name of the returned function
-    wrapper.__name__ = "group_{0}".format(metric_function.__name__)
 
-    return wrapper
+class _DerivedMetricCallable:
+    """Callable that calculates a derived metric.
+
+    :param transformation_function: A transformation function with the signature
+        ``transformation_function(summary)``
```
riedgar-ms, line 144: … where 'summary' is a `Bunch` with keys `overall` and `by_group` ?

fairlearn/metrics/_metrics_engine.py
```diff
@@ -82,33 +95,114 @@ def _check_metric_params(y_true, metric_params,
...
+def make_metric_group_summary(metric_function, indexed_params=None, name=None):
+    """Make a callable that calculates the group summary of a metric.
+
+    :param metric_function: A metric function with the signature
+        ``metric_function(y_true, y_pred, **metric_params)``
+    :type metric_function: func
+
+    :param indexed_params: The names of parameters of ``metric_function`` that
+        should be split according to ``sensitive_features`` in addition to ``y_true``
+        and ``y_pred``. Defaults to ``None`` corresponding to ``['sample_weight']``.
```
riedgar-ms, line 179: I've just spotted this, and.... should this parameter be `extra_indexed_params`? What happens if the user specifies a list which doesn't include `sample_weight` ?

fairlearn/metrics/_metrics_engine.py
```diff
@@ -167,3 +261,66 @@ def group_max_from_summary(summary):
...
+    "ratio": ratio_from_summary,
+    "group_min": group_min_from_summary,
+    "group_max": group_max_from_summary,
+}
+
+# Base metrics and the variants that are implemented by the metrics engine
+METRICS_SPEC = [
+    # base metrics from _extra_metrics
+    (true_positive_rate, ["difference", "ratio"]),
+    (true_negative_rate, []),
```
riedgar-ms, line 277: Why no difference and ratio?

fairlearn/metrics/_metrics_engine.py
MiroDudik: My sense is that the user would only supply this after they inspect the signature of the base metric in question, and if `sample_weight` is not part of the signature, I think it would be weird to *always* include it... An alternative would be for the default value to be `['sample_weight']` rather than `None`. But given that the behavior of that is equivalent, I would leave that to a future PR.

fairlearn/metrics/_metrics_engine.py
MiroDudik: This follows the API proposal. The only reason I ended up expanding `true_positive_rate` and `false_positive_rate` is because they are used to define `equalized_odds_{difference,ratio}`. There is also a separate rationale for `true_positive_rate` since it correponds to the "equal opportunity" metric.

Beyond that, I tried to keep the number of automatically expanded metrics to the minimum. It will be easy to add new derived metrics in future, but we won't be able to remove metnics easily. So I'm trying to start with the "minimal sensible set" and expand based on the usage pattern.

CHANGES.md
MiroDudik: Re. dashboard--this specific plot was confusing to users, so we might rework it.

Re. why make it private--I looked around online and couldn't find a standard name for this metric. But I found something called `mean_overprediction_error` and `mean_underprediction_error` and their meaning is similar to ours, but very slightly different (e.g., overprediction is positive and underprediction is negative, and they differ in how they are normalized). So I would like to do a bit of asking around to figure what the right API should be. Since we're reworking so much about metrics, I think it makes sense to introduce this change now, so we have more freedom to make the API change in future.

fairlearn/widget/_fairlearn_dashboard.py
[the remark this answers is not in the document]
riedgar-ms: I think this should be OK, since we just change the function in the value, not the key. We should check on release, though.

fairlearn/metrics/_metrics_engine.py
riedgar-ms: I'm not a huge fan, but I'm not sure which would be considered more Pythonic.

fairlearn/metrics/_metrics_engine.py
riedgar-ms: OK

CHANGES.md
MiroDudik: done

PR #342, opened 2020-03-24
[WIP] FIX make `ThresholdOptimizer` sklearn compatible

adrinjalali:
Closes #282 
This PR runs the `check_estomator` tests from sklearn to ensure compatibility.

There are issues which may require some workarounds, and I'll be working on them.

Issue #314, opened 2020-03-04
Question: should we move to numpydoc format?


adrinjalali:
Many libraries use the [`numpydoc`](https://numpydoc.readthedocs.io/en/latest/format.html) format, which is understood by many tools in the ecosystem as well.

Would there be any interest in moving to it instead of what we have?

romanlutz:
We did quite a bit of work to make it somewhat consistent in the current format. @riedgar-ms spent many hours getting this right. When you say "understood" what do you actually mean? Do tools do further processing using docs?

riedgar-ms:
I believe I did try the `numpydoc` format for the ReST strings for a while. It's basically a different set of markup tags to the existing ones we have.

Changing would be nontrivial (we have to edit all of our existing documentation to conform to the new ReST variant, and possibly persuade ReadTheDocs to process it). As @romanlutz asks, what are the specific benefits from this?

adrinjalali:
It's just that the part of the ecosystem I know follows that format and we're all used to it, and people write tools to parse the format and extract information from it. We even have tests in our test suit which make sure all the parameters of a function or a method are documented in the docstring and we can check for the presence of the default values etc. I'm not sure how you could do that for the one we have here right now.

riedgar-ms:
We'd have to investigate, but since this is just the default `autodoc` format (as I understand it), I'd guess that there are similar tools for it. Those are definitely significant benefits which you list, but I think we should make sure that we can't get them with the existing docstrings before we commit to rewriting all of them.

riedgar-ms:
Bump.... this should be part of the wider discussion we're having about documentation. Tagging @MiroDudik and @mesameki for that reason.

romanlutz:
In general, such changes are best addressed as soon as possible, since the amount of work for switching only increases over time. I don't know whether we see enough of a need to allocate time for this (unless there are volunteers, of course). @vingu  FYI

adrinjalali:
I'm really not sure if `autodoc` gives you the options I mention, but for a better understanding of what I mean, you can check the [`sklearn/tests/test_docstring_parameters.py`](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/tests/test_docstring_parameters.py) file as an example.

hildeweerts:
In addition to the previous points, in my opinion, numpy docstrings are also a much more human-readable format than the one that is currently used.

MiroDudik:
[...]

PR #381, opened 2020-04-20
Add new constraints to ThresholdOptimizer and add InterpolatedThresholder

MiroDudik:


fairlearn/postprocessing/_constants.py
```diff
@@ -11,6 +13,45 @@
 
 DEMOGRAPHIC_PARITY = "demographic_parity"
 EQUALIZED_ODDS = "equalized_odds"
+ACCURACY_SCORE = "accuracy_score"
+
+def confusion_matrix_summary(*, true_positives, false_positives, true_negatives, false_negatives):
```
romanlutz, line 18: That's not really a "constant" :-)

test/unit/postprocessing/test_threshold_optimization.py
```diff
@@ -150,24 +150,19 @@ def test_threshold_optimization_different_input_lengths(data_X_y_sf, constraints
...
                                                    sensitive_features_transform):
     y = y_transform(labels_ex)
     sensitive_features = sensitive_features_transform(sensitive_features_ex1)
-    scores = score_transform(scores_ex)
-    estimator = ThresholdOptimizer(estimator=ExamplePredictor(scores),
+    estimator = ThresholdOptimizer(estimator=PassThroughPredictor(),
                                    constraints=DEMOGRAPHIC_PARITY)
-    estimator.fit(X_ex, y, sensitive_features=sensitive_features)
-
+    estimator.fit(pd.DataFrame(scores_ex), y, sensitive_features=sensitive_features)
```
romanlutz, line 161: Why was this required? Did we break something about supported data types?

test/unit/postprocessing/test_threshold_optimization.py
```diff
@@ -150,24 +150,19 @@ def test_threshold_optimization_different_input_lengths(data_X_y_sf, constraints
...
 
-@pytest.mark.parametrize("score_transform", candidate_Y_transforms)
 @pytest.mark.parametrize("y_transform", candidate_Y_transforms)
 @pytest.mark.parametrize("sensitive_features_transform", candidate_A_transforms)
-def test_threshold_optimization_demographic_parity(score_transform, y_transform,
+def test_threshold_optimization_demographic_parity(y_transform,
                                                    sensitive_features_transform):
     y = y_transform(labels_ex)
     sensitive_features = sensitive_features_transform(sensitive_features_ex1)
-    scores = score_transform(scores_ex)
```
romanlutz, line 160: I'm fine with the other changes, but score_transform should still be used, otherwise we don't have test coverage for other cases.

test/unit/postprocessing/test_curve_utilities.py
```diff
@@ -89,24 +89,37 @@ def test_calculate_roc_points():
...
 
-    roc_points = _calculate_roc_points(grouped_data, "A")
+    roc_points = _calculate_roc_points(grouped_data, "A", flip=True)
     expected_roc_points = pd.DataFrame({
-        "x": [0, 0.25, 0.5, 0.5, 1],
-        "y": [0, 1/3,  2/3, 1,   1],
+        "x": [0, 1, 0.75, 0.25, 0.5, 0.5, 0.5, 0.5, 1, 0],
+        "y": [0, 1, 2/3,  1/3,  1/3, 2/3, 0,   1,   1, 0],
         "operation": [ThresholdOperation('>', np.inf),
+                      ThresholdOperation('<', np.inf),
```
romanlutz, line 97: Why change the original example?

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -93,9 +214,16 @@ def fit(self, X, y, *, sensitive_features, **kwargs):
         if self.estimator is None:
             raise ValueError(ESTIMATOR_ERROR_MESSAGE)
 
-        if self.constraints not in _SUPPORTED_CONSTRAINTS:
+        if self.constraints not in SUPPORTED_CONSTRAINTS:
             raise ValueError(NOT_SUPPORTED_CONSTRAINTS_ERROR_MESSAGE)
 
+        if self.constraints == EQUALIZED_ODDS:
+            if self.objective not in SUPPORTED_OBJECTIVES_FOR_EQUALIZED_ODDS:
+                raise ValueError(NOT_SUPPORTED_OBJECTIVES_FOR_EQUALIZED_ODDS_ERROR_MESSAGE)
```
romanlutz, line 222: There should be a test case for this (and any error message really)

fairlearn/postprocessing/_constants.py
romanlutz, line 18: Any chance we could reuse something that already exists for this? Perhaps https://scikit-learn.org/stable/modules/generated/sklearn.metrics.confusion_matrix.html is of use?

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -38,12 +39,131 @@
...
 ESTIMATOR_ERROR_MESSAGE = "'estimator' cannot be `None`."
 
-_SUPPORTED_CONSTRAINTS = [DEMOGRAPHIC_PARITY, EQUALIZED_ODDS]
-
 logger = logging.getLogger(__name__)
 
+# All allowed utilities must have values between 0 and 1 and attain both extremes as
+# the threshold goes from -Inf to Inf.
+SUPPORTED_UTILITIES = {
+    'selection_rate',
```
romanlutz, line 47: This is equivalent to DP. Why add it?

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -38,12 +39,131 @@
...
+    "Currently only the following constraints are supported: {}.".format(
+        ", ".join(SUPPORTED_CONSTRAINTS)))
+NOT_SUPPORTED_OBJECTIVES_ERROR_MESSAGE = (
+    "Currently only the following objectives are supported: {}.".format(
+    ", ".join(METRIC_DICT.keys())))
+NOT_SUPPORTED_OBJECTIVES_FOR_EQUALIZED_ODDS_ERROR_MESSAGE = (
+    "Currently only the following objectives are supported: {}.".format(
+    ", ".join(SUPPORTED_OBJECTIVES_FOR_EQUALIZED_ODDS)))
+
+class InterpolatedThresholder(BaseEstimator, MetaEstimatorMixin):
```
romanlutz, line 72: I'm not sure I like that it's in the same file. Makes the entire file more crowded with things. :-(
So far, if I search for `fit` in this file I know where I end up. Now I have to scroll back up and search for the class name.

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -38,12 +39,131 @@
...
+    """Binary predictor that thresholds continuous predictions of a base estimator.
+
+    At prediction time, the predictor takes as input both standard and sensitive features.
+    Based on the values of sensitive features, it then applies a randomized thresholding
+    transformation according to the provided `interpolation_dict`.
+    
+    :param estimator: base estimator
+    :param dict interpolation_dict: maps sensitive feature values to `Bunch` that describes the
```
romanlutz, line 80: Is this a new way of writing the doc, or did you mean
```
:param interpolation_dict: …
:type interpolation_dict: dict
```
?

fairlearn/postprocessing/_constants.py
MiroDudik: shall we rename `_constants` -> `_base`?

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -38,12 +39,131 @@
...
+        :rtype: numpy.ndarray
+        """
+        check_is_fitted(self)
+        _, _, sensitive_feature_vector = _validate_and_reformat_input(
+            X, y=None, sensitive_features=sensitive_features, expect_y=False,
+            enforce_binary_labels=True)
+        base_predictions = np.array(self.estimator_.predict(X))
+        
+        positive_probs = 0.0*base_predictions
+        for a, interpolation in self.interpolation_dict.items():
```
romanlutz, line 133: Can we use `sensitive_feature_value` or at least `sf` instead of `a`?

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -38,12 +39,131 @@
...
+        
+    def predict(self, X, *, sensitive_features, random_state=None):
+        """Provide a prediction for the given input data.
+
+        Note that this is non-deterministic, due to the nature of the
+        interpolation.
 
-class ThresholdOptimizer(BaseEstimator, ClassifierMixin, MetaEstimatorMixin):
+        :param X: Feature data
+        :type X: numpy.ndarray or pandas.DataFrame
```
romanlutz, line 152: Can we also add the other params in the documentation?

test/unit/postprocessing/test_curve_utilities.py
romanlutz: Keeping the existing tests means we didn't regress anywhere so that would be very important for a PR of this size. You can add as many additional examples as you can write :-)

fairlearn/postprocessing/_constants.py
MiroDudik: This is doing something very different. It is taking an already calculated confusion matrix and filling in the names of implied fields. Would it help to rename this utility function?

test/unit/postprocessing/test_curve_utilities.py
MiroDudik: The logic of `_calculate_roc_points` has changed--that's why the change here. New logic is more generic, so I think I prefer it.

test/unit/postprocessing/test_threshold_optimization.py
MiroDudik: I don't think we need to vary score transform in this instance. The base estimator is supposed to be implementing an sklearn estimator and those are always supposed to return 1D-arrays (not lists and not DataFrames). I don't think we need to try to be more general than sklearn here.

test/unit/postprocessing/test_threshold_optimization.py
MiroDudik: This is because I'm now working with `PassThroughPredictor` instead of `ExamplePredictor`.

fairlearn/postprocessing/_threshold_optimizer.py
MiroDudik: I don't mind pulling out into a separate file.

fairlearn/postprocessing/_threshold_optimizer.py
MiroDudik: this is supported by sphinx and looks more natural to me... i'm even thinking we should consider going numpy style like sklear, because it reads nicer when you call `help` in interactive python.

fairlearn/postprocessing/_constants.py
romanlutz: Base kind of makes me think "base class". So far everything like that is named utilities. The constants just don't belong in the same file, though. Can you put the functions/logic in a utilities (or other name) file?

MiroDudik: i see your point. let me see what i can do. i resist creating a lot of small files...

test/unit/postprocessing/test_curve_utilities.py
romanlutz: I'm actually not entirely sure whether I understand it anymore. These used to be ordered in an understandable way, but now it's hard to grasp how this all came to be.

You mean the old example doesn't work anymore? Because it's hard to tell if we break something if we change the test logic itself.

test/unit/postprocessing/test_threshold_optimization.py
MiroDudik: On a second thought, let me try to add this back. I don't trust people to implement sklearn estimators 100% faithfully to the spec, and it would be nice to cover these simple cases.

romanlutz: Then we need to adjust the expected data types to exclude those. At every other point we've parametrized over the different data types.

test/unit/postprocessing/test_threshold_optimization.py
romanlutz: This goes back to the same point as in the other comment. If we allow more data types we should test the entire matrix of data types. If we don't allow it we need to adjust the places where we've said we do.

MiroDudik: This is a different issue. The feature matrix should always be a pass-through, we don't touch it, we just pass it on to the base estimator. We cannot guarantee that the base estimator will work well with one or another kind of feature encoding (this is what burnt us with LightGBM). Maybe best is to remove all of the validation that we do around X. But that should be a separate PR.

test/unit/postprocessing/test_curve_utilities.py
MiroDudik: [exactly... let me re-order them in an understandable way again]

fairlearn/postprocessing/_threshold_optimizer.py
romanlutz: @adrinjalali recently pointed that out in an issue, I think #314

fairlearn/postprocessing/_threshold_optimizer.py
romanlutz: Or rather: why add this one and not others like equality of opportunity and other duplicate names?

fairlearn/postprocessing/_constants.py
romanlutz: I kind of understood why it's there better after reading the entire PR. You're mapping the various objectives to these metrics inside postprocessing, so this is required. I think it's fine the way it is, but it would be awesome to have a short description saying why this is needed (i.e. the second sentence of this very comment, but perhaps more nicely written?)

test/unit/postprocessing/test_curve_utilities.py
romanlutz: OK, thanks for elaborating!
Would it be possible to keep the old case (probably with adjusted result) and add this new case additionally?

test/unit/postprocessing/test_threshold_optimization.py
romanlutz: Wow I completely confused myself... this is passed as the feature matrix, of course, not as scores. 

What burnt us with LightGBM was that we changed X (and that was in reductions, not postprocessing). That's not the case here, though. You should be able to pass in any kind of format that we allow for X, and therefore we should also be testing any kind of format as input.

fairlearn/postprocessing/_constants.py
romanlutz: If you find a more suitable name than constants or base I don't mind that either... :-)

fairlearn/postprocessing/_threshold_optimizer.py
MiroDudik: I'd like to limit the duplication unless there is a strong reason against it. The "systematic" terminology is "selection_rate", but "demographic_parity" is extremely common, so it's added as a synonym (and for the backward compatibility).

test/unit/postprocessing/test_curve_utilities.py
MiroDudik: Technically, the behavior of `_calculate_roc_points` is now different (it produces lists twice as long as before), so this is a breaking change. If I left the previous test unmodified it would fail. What I have here is an equivalent test (on the exactly same input data). Do you want to separate out the old functionality of the previous  `_calculate_roc_points`? (I'm not sure I see why.)

romanlutz: I feel like we already agree, but I'm expressing myself badly :-)
What I meant to say is that when you change logic like in `_calculate_roc_points`, then it's natural that the test case expectations may need to be adjusted. For example, the expected result may be twice as long as you mention. However, in this particular case the input data itself was also changed (or at least it looks like that to me), which means it's not actually possible for us to understand how it affects the test case without calculating everything by hand. From the diff I expected to see the identical input data and then an expected result list that's twice as long.
Perhaps git/GitHub is playing tricks on me and you didn't change the test input data?

fairlearn/postprocessing/_threshold_optimizer.py
romanlutz: Sure, if you feel strongly then go with that. My question was rather out of curiosity.

fairlearn/postprocessing/_constants.py
adrinjalali: `extended_confusion_matrix`?

fairlearn/postprocessing/_roc_curve_utilities.py
```diff
@@ -4,13 +4,17 @@
...
+    LABEL_KEY, SCORE_KEY, P0_KEY, P1_KEY,
+    METRIC_DICT,
+    confusion_matrix_summary)
 from ._threshold_operation import ThresholdOperation
 
 DEGENERATE_LABELS_ERROR_MESSAGE = "Degenerate labels for sensitive feature value {}"
 
 
-def _get_roc(data, sensitive_feature_value, flip=True):
+def _get_roc(data, sensitive_feature_value, flip=False,
```
adrinjalali, line 16: is this very different from https://scikit-learn.org/dev/modules/generated/sklearn.metrics.roc_curve.html#sklearn.metrics.roc_curve?

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -38,12 +39,131 @@
...
+        If `prefit` is set to `True` then the base estimator is kept as is.
+        Otherwise it is fitted from the provided arguments.
+        """
+        if self.estimator is None:
+            raise ValueError(ESTIMATOR_ERROR_MESSAGE)
+        
+        if not self.prefit:
+            self.estimator_ = clone(self.estimator).fit(X, y, **kwargs)
+        else:
+            self.estimator_ = self.estimator
```
adrinjalali, line 110: this should also check if the estimator is already fitted. Otherwise it should refit anyway.

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -38,12 +39,131 @@
...
+            of the two numbers in each tuple needs to add up to 1.
+        :rtype: numpy.ndarray
+        """
+        check_is_fitted(self)
+        _, _, sensitive_feature_vector = _validate_and_reformat_input(
+            X, y=None, sensitive_features=sensitive_features, expect_y=False,
+            enforce_binary_labels=True)
+        base_predictions = np.array(self.estimator_.predict(X))
+        
+        positive_probs = 0.0*base_predictions
```
adrinjalali, line 132: maybe [`np.zeros_like`](https://numpy.org/doc/stable/reference/generated/numpy.zeros_like.html?highlight=zeros_like#numpy.zeros_like)?

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -65,12 +185,13 @@ class ThresholdOptimizer(BaseEstimator, ClassifierMixin, MetaEstimatorMixin):
     """
 
     def __init__(self, *, estimator=None,
-                 constraints=DEMOGRAPHIC_PARITY, grid_size=1000, flip=True,
+                 constraints=DEMOGRAPHIC_PARITY, objective=ACCURACY_SCORE, grid_size=1000, flip=True,
```
adrinjalali, line 188: the issue with having these constants such as `ACCURACY_SCORE` is that when user sees the autocomplete in their IDE, it would look less confusing if they saw `"accuracy_score"` I think.

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -170,16 +282,11 @@ def _pmf_predict(self, X, *, sensitive_features):
...
-            self._post_processed_predictor_by_sensitive_feature, sensitive_feature_vector,
-            self.estimator_.predict(X))
-        return np.array([[1.0 - p, p] for p in positive_probs])
-
-    def _threshold_optimization_demographic_parity(self, sensitive_features, labels, scores,
-                                                   grid_size=1000, flip=True):
+        return self.interpolated_thresholder_._pmf_predict(
+            X, sensitive_features=sensitive_features)
+        
+    def _threshold_optimization_for_utility(
```
adrinjalali, line 288: a better name here maybe

test/unit/postprocessing/conftest.py
```diff
@@ -113,6 +113,14 @@ def predict(self, X):
...
 
 
+class PassThroughPredictor(BaseEstimator, ClassifierMixin):
+    def fit(self, X, y=None, **kwargs):
+        return self
+
+    def predict(self, X):
+        return X[0]
+
+
```
adrinjalali, line 123: I think all these would much better fit in the test file themselves instead of in the `conftest.py`. I find it odd to have these here.

fairlearn/postprocessing/_constants.py
```diff
@@ -11,6 +13,45 @@
 
 DEMOGRAPHIC_PARITY = "demographic_parity"
 EQUALIZED_ODDS = "equalized_odds"
+ACCURACY_SCORE = "accuracy_score"
```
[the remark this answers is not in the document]
MiroDudik: I think that our names of metrics are now stable enough that we don't need these three constants as long as we have tests that exercise all valid strings.

fairlearn/postprocessing/_threshold_optimizer.py
MiroDudik: This actually points out to a deeper issue.

I tried to do it the way you suggested first, but running with `LGBMClassifier` was triggering refitting even though it was already fitted. Obviously they are not implementing sklearn spec correctly, but that might end up being quite common when people create simple wrappers around their non-sklearn classifiers to pull into our library. I am considering two options here:
- Don't make `InterpolatedThresholder` a sklearn estimator; we would not bring in the mixins and would only implement `predict`, but not `fit`. (Alas, I would use the pretty printing of `BaseEstimator.__repr__`
- Throw a warning or even an exception when there's a mismatch between `prefit` and `check_fitted` (and required a manual override for an exception, e.g., by setting an additional argument? or is that too much?)
Thoughts?

test/unit/postprocessing/conftest.py
romanlutz: Perhaps this continues an odd pattern we've accidentally established...
I agree with @adrinjalali, though, and will make a mental note to clean other similar cases up.

fairlearn/postprocessing/_threshold_optimizer.py
romanlutz: I love this comment, because the name makes perfect sense to me, but it shows that it's not quite as clear as I thought! 
It does perform threshold optimization for a provided utility, though, so I wonder how we could find a better name.

fairlearn/postprocessing/_threshold_optimizer.py
romanlutz: Fair point! Does sklearn repeat all the strings all over the place, or is there a better pattern that we just haven't come across?

fairlearn/postprocessing/_threshold_optimizer.py
romanlutz: I dislike not conforming to sklearn because other packages don't follow conventions properly.
I also dislike adding yet another argument, because that would just completely confuse me as the user.

Let me see whether I can write up expected behavior (IMO)
- `prefit=True` and `check_fitted` returns True: no additional fitting
- `prefit=True` and `check_fitted` returns False: user tells us explicitly that it's fitted, so don't fit again, but log warning and if it fails on `predict` provide an explanation. This seems like a typical scenario of "user error" to me. If a user provides `prefit=True` we should never call `fit`.
- `prefit=False` and `check_fitted` returns True: call `fit` anyway because the user seems to want us to refit it.
- `prefit=False` and `check_fitted` returns False: call `fit`

What do you think?

fairlearn/postprocessing/_constants.py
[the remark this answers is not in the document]
romanlutz: I rather like the verbosity here (and in general), but I'm not too invested in this.

fairlearn/postprocessing/_constants.py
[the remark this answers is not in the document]
romanlutz: and perhaps we should put them into the doc for ThresholdOptimizer to clearly say what the options are

fairlearn/postprocessing/_threshold_optimizer.py
adrinjalali: The issue with `prefit=True` and warning or raising is that `clone` will keep `prefit=True`, but the underlying estimator will be a fresh unfitted one. To me it's a straightforward issue, `LGBMClassifier` has an issue which can very easily be fixed, so I'd open an issue on https://github.com/microsoft/lightgbm and most probably have it fixed. It'll probably be better received if the issue report comes from withing Microsoft (gut feeling, not sure about your politics ;) )

fairlearn/postprocessing/_constants.py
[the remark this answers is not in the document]
adrinjalali: ideally the docstring should include all the options, and a link to a "user guide" is beneficial for further explanation ;)

fairlearn/postprocessing/_threshold_optimizer.py
adrinjalali: I guess it's just a norm that different communities follow. And in a dynamically typed lang like python, having the strings makes sense IMO. Alternatively, you could go for an `Enum`, but I haven't seen that often in python, and it can complicate things. What makes sense to me, is ['Literal`](https://www.python.org/dev/peps/pep-0586/) which we probably will see more of once people start adopting Python 3.8 standards.

fairlearn/postprocessing/_threshold_optimizer.py
romanlutz: That helps with LightGBM, but that won't fix the same issue for other packages.

This is what I filed with them: https://github.com/microsoft/LightGBM/issues/3014

adrinjalali: Well, if you go with a nice sklearn compat, then it makes sense for the user to work in that ecosystem I'd say.

fairlearn/postprocessing/_threshold_optimizer.py
MiroDudik: I will be renaming `utility` to `simple_metrics` as opposed to `equalized_odds` (which is the one "not simple" metric that requires a separate treatment).

fairlearn/postprocessing/_threshold_optimizer.py
MiroDudik: I kinda want to have a cake and eat it too. @romanlutz : your proposal exactly matches mine.

A slight elaboration:
- if `prefit=True` and `check_fitted=True`: no additional fitting
- if `prefit=True` and `check_fitted=True`: no additional fitting, but give a warning--the warning should say that if the user is running crossvalidation etc., they should do so with `prefit=False` and if the estimator is already fitted, they should file the issue to make it sklearn compliant. 
- if `prefit=False`: always call `fit`

@adrinjalali :  I'm also a bit worried about a silent failure (where the user is setting `prefit=True`, but is unaware that the base estimator is being refitted). This proposal addresses that issue too.

adrinjalali: I was checking `SelectFromModel` in sklearn, and apparently there we even raise if the user calls `fit` and `prefit=True`, which means users can't give an estimator with `prefit=True` to a `GridSearchCV` for instance. So a warning, or even raising, would be consistent with that.

fairlearn/_input_validation.py
```diff
@@ -54,9 +54,9 @@ def _validate_and_reformat_input(X, y=None, expect_y=True, enforce_binary_labels
     if y is not None:
         # calling check_X_y with a 2-dimensional y causes a warning, so ensure it is 1-dimensional
         if isinstance(y, np.ndarray) and len(y.shape) == 2 and y.shape[1] == 1:
-            y = y.squeeze()
+            y = y.reshape(-1)
```
romanlutz, line 57: why not squeeze anymore?

fairlearn/postprocessing/_constants.py
MiroDudik: I'm moving it elsewhere and renaming to `_extend_confusion_matrix`.

fairlearn/_input_validation.py
MiroDudik: it reduced the number of dimensions too much, down to a scalar, instead down to a one-dimensional array.

fairlearn/postprocessing/_roc_curve_utilities.py
MiroDudik: This function is in some ways more general than the sklearn's (but in some ways less general). We allow picking  `x_metric` and `y_metric` that are different from FPR and TPR, and also we allow the ability to flip the decision if that improves `y_metric`. I think we should rename this from `roc_curve` to a `tradeoff_curve`. I'll update.

test/unit/postprocessing/test_threshold_optimization.py
MiroDudik: `score_ex` is now again included and I've also added a comment about the `PassThroughPredictor`

test/unit/postprocessing/test_threshold_optimization.py
MiroDudik: The previous version was only trying different transformed versions of `y`, `scores_ex` and `sensitive_features`, so I've retained that functionality. Let me know if you think I should still be changing anything here.

test/unit/postprocessing/test_curve_utilities.py
MiroDudik: Please take a look again. To make it clear how I changed things, the additional returned points are written in the form `.0` and `.5` rather than `0` and `0.5` etc.

fairlearn/postprocessing/_constants.py
MiroDudik: Adding a more detailed docstring.

fairlearn/postprocessing/_constants.py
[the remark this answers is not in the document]
MiroDudik: I'm implementing all the docstring suggestions, but not pointing to the User Guide yet--until we write that content in a future PR.

fairlearn/postprocessing/_constants.py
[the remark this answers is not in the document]
MiroDudik: i'll stick to the verbose form now, but making this method private, so we can revisit a more concise API if we want to make it public.

fairlearn/postprocessing/_threshold_optimizer.py
MiroDudik: Great, I'll then implement my last suggestion.

fairlearn/postprocessing/_threshold_optimizer.py
MiroDudik: Moving to explicit strings since metric names are now stable.

test/unit/postprocessing/conftest.py
MiroDudik: Done.

fairlearn/postprocessing/_threshold_optimizer.py
MiroDudik: I ended up keeping as is, because `base_predictions` is `int`, but `positive_probs` needs to be `float`.

romanlutz:
This PR looks great. Is this still WIP?

MiroDudik:
Actually--you are right. This is definitely no longer WIP. The "only" missing piece I believe are tests for the new functionality.

fairlearn/postprocessing/_interpolated_thresholder.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+class InterpolatedThresholder(BaseEstimator, MetaEstimatorMixin):
+    """Binary predictor that thresholds continuous predictions of a base estimator.
+
+    At prediction time, the predictor takes as input both standard and sensitive features.
+    Based on the values of sensitive features, it then applies a randomized thresholding
+    transformation according to the provided `interpolation_dict`.
+
+    :param estimator: base estimator
```
romanlutz, line 26: Did you want to use numpydoc all over this PR?

MiroDudik:
@adrinjalali : This PR overlaps with #282 and #342 . It fixes some of the issues noted in #282 (such as `prefit`), but not all of them. Is it okay if I go ahead and merge this one? I can help you incorporate the additional changes from #282.

adrinjalali:
Yes of course. My PR requires some changes in scikit-learn and that's why it's on hold. Please don't wait for that one. I'll probably need to open a new one once certain issues are fixed on the sklearn side.

Issue #380, opened 2020-04-20
ENH - ExponentiatedGradient: use only classifiers with non-zero weights for predictions


romanlutz:
#### Is your feature request related to a problem? Please describe.
Some classifiers within ExponentiatedGradient have weights of 0. That means they don't contribute to the prediction, yet we calculate their predictions for every sample. This feature request is about optimizing ExponentiatedGradient to not waste time on this. The more iterations EG runs, the more potential there is for improvement due to the larger number of resulting classifiers.
Example based on EG tests in the repo:
```>>> eg = ExponentiatedGradient(LogisticRegression(), DemographicParity())
>>> eg.fit(X, y, sensitive_features=A)
>>> eg._last_t
5
// i.e. 6 iterations
>>> eg.predict(X)
predictions per classifier (column) per sample (row)
    0  1  2  3  4  5
0   0  0  1  1  1  0
1   1  1  0  1  1  0
2   1  1  0  1  1  0
3   0  0  1  1  1  0
4   1  1  1  0  1  1
5   0  0  1  1  1  0
6   1  1  0  1  1  0
7   0  0  1  1  1  0
8   1  1  0  1  1  0
9   1  1  0  1  1  0
10  1  1  0  1  1  0
11  1  1  0  1  1  0
12  1  0  1  0  0  1
13  1  1  1  0  1  1
14  1  0  1  0  0  1
15  1  0  1  0  0  1
16  1  1  1  0  1  1
17  1  0  1  0  0  1
18  1  1  1  0  1  1
19  1  1  1  0  1  1


aggregated product of predictions multiplied times the weight of the classifiers 
           0
0   0.430000
1   0.713846
2   0.713846
3   0.430000
4   1.000000
5   0.430000
6   0.713846
7   0.430000
8   0.713846
9   0.713846
10  0.713846
11  0.713846
12  0.286154
13  1.000000
14  0.286154
15  0.286154
16  1.000000
17  0.286154
18  1.000000
19  1.000000
output from predict(X):
array([0, 0, 1, 1, 1, 0, 1, 0, 0, 1, 1, 1, 0, 1, 0, 0, 1, 0, 1, 1])

>>> eg._weights
0    0.000000
1    0.283846
2    0.000000
3    0.000000
4    0.430000
5    0.286154
```
The last output shows that half the weights are actually zero, so we could have avoided using them for predictions.

#### Describe the solution you'd like
Ignore classifiers with weights of zero for predictions.

[...]

Issue #65, opened 2019-10-02
Expectations for predict, predict_proba, etc.


romanlutz:
Since fairlearn uses randomized classifiers we need to set expectations for what `predict`, `predict_proba` etc. return.

romanlutz:
API documentation exists, but it doesn't reflect randomization well yet.

harvineet:
For ExponentiatedGradient, what is the preferred way of getting predicted **probabilities**?
Can we use `_best_classifier`?

The notebook example [link](https://github.com/fairlearn/fairlearn/blob/master/notebooks/Mitigating%20Disparities%20in%20Ranking%20from%20Binary%20Data.ipynb) expects `predict` to return scores for `LogisticRegression`:

`scores_expgrad_X = pd.Series(expgrad_X.predict(X_test), name="scores_expgrad_X")`

However, the result seems to be the predicted class.

romanlutz:
Right. For probabilities we currently provide `_pmf_predict` from https://github.com/fairlearn/fairlearn/blob/master/fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py but as you can tell from the underscore there's no guarantee it'll stay that way.

harvineet:
I looked at `_pmf_predict` in Exponentiated Gradient. I have a question on how the predictions from the returned classifiers are combined to get predicted probabilities.

**Current**
In `_mean_pred`[function](https://github.com/fairlearn/fairlearn/blob/master/fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py#L17), the approach taken is to  call `predict` and then weight the result. This does not pass on the 'full uncertainty information' from the classifiers. As `predict` only returns 0 or 1 instead of probabilities. Encountered this in an example where the `_classifiers` had only one classifier. Then, the result from `_pmf_predict` had only 0s and 1s which is what `predict` returns. This is an issue.

**Alternative**
Instead, one can call `predict_proba` for each classifier and then weight. In expectation, the results should be same i.e. expectation of the two variables `weight*Bernoulli(prob)` and `Bernoulli(weight*prob)` is `weight*prob`.

Is this a better way for getting predicted probabilities? Trying it on the [notebook example](https://github.com/fairlearn/fairlearn/blob/master/notebooks/Mitigating%20Disparities%20in%20Ranking%20from%20Binary%20Data.ipynb) seem to give better AUC and disparity values. Did not do a thorough test.

```
def weighted_score(dataset, algorithm):
[...]

PR #382, opened 2020-04-21
Document probabilistic classifier behavior in ExponentiatedGradient & calculate predictions only for non-zero weighted classifiers

romanlutz:
There had been a gap between the description of probabilistic classifiers in the paper vs. what we do in practice, even though the procedures are equivalent. This added documentation makes that clear.

Additionally, the iterative calculation of predictions for each classifier is replaced with a vectorized implementation that omits predicting with zero-weighted classifiers as they can't contribute to the result anyway.

Related to #5 #65 #380 


fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -183,8 +183,13 @@ def fit(self, X, y, **kwargs):
     def predict(self, X):
         """Provide a prediction for the given input data.
 
-        Note that this is non-deterministic, due to the nature of the
-        exponentiated gradient algorithm.
+        Chooses the prediction label randomly based on the probabilities from `_pmf_predict`,
+        i.e., based on the aggregate of the weighted predictions of all underlying classifiers.
+        This is equivalent to randomly choosing an individual classifier based on the weights and
```
riedgar-ms, line 188: An equation might be helpful.

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -199,13 +204,21 @@ def predict(self, X):
...
         :rtype: pandas.DataFrame
         """
-        pred = pd.DataFrame()
-        for t in range(len(self._hs)):
-            pred[t] = self._hs[t](X)
+        n_samples = len(X)
+
+        # ignore classifiers with weight 0 since they can only contribute 0 to the result
+        pred = pd.DataFrame(0, index=range(n_samples), columns=range(len(self._hs))) \
+            .apply(lambda row: np.zeros(n_samples) if self._weights[int(row.name)] == 0
```
riedgar-ms, line 220: Is this saving us a call to the predictors, or have those already happened? This statement is a bit of a monstrosity to set things to zero which would be zero anyway. Unless it saves an expensive call.

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
romanlutz: I think that's more appropriate for the user guides, don't you think?
In fact, if I had known we'd have user guides soon I probably wouldn't have written this here at all, so I consider this much more of a "let's document it and see where it lands eventually" sort of PR

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
romanlutz: The first line doesn't do anything other than create a data structure with zeros.
I would call it expensive if we have a billion rows and a thousand classifiers of which only 5 have non-zero weights :-)

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
riedgar-ms: OK.

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
riedgar-ms: Fair enough, but I still think this is a monstrosity of a statement. Rather than a lambda, could you put an inline `def evaluate_for_nonzero_weight()` (on the line immediately preceding this one) and have that in the `apply()` method?

I grant that at least part of this is due to my lack of familiarity with pandas, but I do think that when a lambda is covering multiple lines, it should probably get a proper function definition. I also while I have used it in list comprehensions, I'm not a huge fan of the "stmt if cond else stmt" construct.

romanlutz: Sure we can do that. The lambda is not the important part, the `apply` is, so we can certainly pre-define it.

riedgar-ms: Are you going to?

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -198,13 +203,26 @@ def predict(self, X):
...
         :type X: numpy.ndarray or pandas.DataFrame
         :return: Array of tuples with the probabilities of predicting 0 and 1.
         :rtype: pandas.DataFrame
         """
-        pred = pd.DataFrame()
-        for t in range(len(self._hs)):
-            pred[t] = self._hs[t](X)
+        n_samples = len(X)
+
+        def get_weighted_predictions(row):
```
romanlutz, line 217: rename: it's not a weighted prediction, weighting happens later

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -182,8 +182,13 @@ def fit(self, X, y, **kwargs):
     def predict(self, X):
         """Provide a prediction for the given input data.
```
MiroDudik, line 183: "predictions" instead of "a prediction"

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -182,8 +182,13 @@ def fit(self, X, y, **kwargs):
...
-        exponentiated gradient algorithm.
+        Chooses the prediction label randomly based on the probabilities from `_pmf_predict`,
+        i.e., based on the aggregate of the weighted predictions of all underlying classifiers.
+        This is equivalent to randomly choosing an individual classifier based on the weights and
+        returning its prediction.
+        Repeatedly calling `predict` with the same feature data may yield different output. This
+        non-deterministic behavior is intended and stems from the nature of exponentiated gradient
+        algorithm, which produces probabilistic classifiers.
 
         :param X: Feature data
```
MiroDudik, line 193: move to numpy format

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -182,8 +182,13 @@ def fit(self, X, y, **kwargs):
     def predict(self, X):
         """Provide a prediction for the given input data.
 
-        Note that this is non-deterministic, due to the nature of the
-        exponentiated gradient algorithm.
+        Chooses the prediction label randomly based on the probabilities from `_pmf_predict`,
```
MiroDudik, line 185: Here I would only say that the predictions are randomized, so the output might be different on the same data and that this behavior is intended. We could provide more detail either in the user guide or, for now, in the `Notes` section (see [numpy standard](https://numpydoc.readthedocs.io/en/latest/format.html#sections)), e.g., there we can say that a fitted `ExponentiatedGradient` has `predictors_` and `weights_` and on each data point we pick a predictor randomly and apply it. I don't think it makes sense to mention anything about `_pmf_predict` here--it's such a low level detail, that it just belongs into a comment in the code.

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
MiroDudik: I think that this was implemented in #442 (slightly more transparently)--but the docstring for `_pmf_predict` needs to be updated there appropriately.

romanlutz: removed this as agreed, and will update #442 with more detail

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
romanlutz: Done for the entire file!

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
romanlutz: Thanks! Addressed this.

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -191,17 +191,30 @@ def fit(self, X, y, **kwargs):
...
+        the same feature data may yield different output. This
+        non-deterministic behavior is intended and stems from the nature of
+        the exponentiated gradient algorithm.
+
+        Notes
+        -----
+        A fitted `ExponentiatedGradient` has the members `predictors_` and
+        `weights_`. On each data point in `X` it picks one of its underlying
+        predictors randomly and applies it. `weights_` determines the
+        probabilities to choose the predictors.
```
MiroDudik, line 206: A minor rewrite for improved clarity and also, a small thing, I think that standard terminology is that Python class instances have `attributes`:

> A fitted `ExponentiatedGradient` has an attribute `predictors_`, an
> array of predictors, and an attribute `weights_`, an array of non-negative floats of the same length.
> The prediction on each data point in `X` is obtained by first picking a
> random predictor according to the probabilities in `weights_` and then
> applying it. Different predictors can be chosen on different data points.

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -211,10 +224,19 @@ def predict(self, X):
     def _pmf_predict(self, X):
         """Probability mass function for the given input data.
 
-        :param X: Feature data
-        :type X: numpy.ndarray or pandas.DataFrame
-        :return: Array of tuples with the probabilities of predicting 0 and 1.
-        :rtype: pandas.DataFrame
+        Calculates predictions for each of the underlying classifiers. The predictions are
```
MiroDudik, line 227: I think we only need to describe "what" and not "how"? And if you want, you could put "how" into Notes--but I don't think that we need to say "how". How about just:
> For each data point, provide the probabilities with which 0 and 1 is returned as a prediction.

I really don't think that you need to provide Notes, but if you do, you can do something like:
> Probabilities are obtained by applying the predictors in `predictors_` and
> then calculating the probability of returning 0 or 1 when one of them is
> chosen according to `weights_`.

romanlutz: I agree. If we ever do a more detailed description of Exponentiated Gradient in the user guide or contributor guide we can write this with more detail.

PR #384, opened 2020-04-21
Add Lagrangian lambdas to exponentiated gradient object

parul100495:
### What's in the PR?

Addition of Lagrangian lambda in exponentiated gradient object. This helps us access the values of Lagragian lambdas from exponentiated gradient object directly.


fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -62,6 +62,7 @@ def __init__(self, estimator, constraints, eps=0.01, T=50, nu=None, eta_mul=2.0)
         self._oracle_execution_times = None
         self._lambda_vecs = pd.DataFrame()
         self._lambda_vecs_LP = pd.DataFrame()
+        self._lagrangian_lambdas = pd.DataFrame()
```
romanlutz, line 65: so far we've named all of them `_lambda_vecs_*`, can we keep that convention? That would make it `_lambda_vecs_lagrangian`

PR #386, opened 2020-04-23
Conform to naming conventions for attributes

romanlutz:
Related to #256  (but doesn't fully fix it since Postprocessing needs to be addressed as well)

Adjust the attributes to naming conventions, specifically for an attribute named `attr`:
- everything set during `__init__` becomes `attr`
- everything set during `fit` becomes `attr_` unless we want to keep it hidden (then `_attr`)

Since we have example notebooks where we use some of these, e.g. `_predictors`, we should expose them properly.

Additionally, this cleanup includes:
- providing `run_lp_step` as an argument in ExponentiatedGradient. Right now it's a hard-coded constant which isn't very useful for people who want to actually use it.
- renaming two arguments for ExponentiatedGradient, namely 
  - `T` to `max_iterations` to make it more obvious
  - `eta_mul` to `learning_rate` to imply the purpose


fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -48,17 +49,16 @@ def __init__(self, X, sensitive_features, y, estimator, constraints, eps, B, opt
...
-        self.errors = pd.Series(dtype="float64")
-        self.gammas = pd.DataFrame()
-        self.lambdas = pd.DataFrame()
-        self.n = self.X.shape[0]
-        self.n_oracle_calls = 0
-        self.n_oracle_calls_dummy_returned = 0
-        self.oracle_execution_times = []
-        self.last_linprog_n_hs = 0
-        self.last_linprog_result = None
+        self.hs_ = pd.Series(dtype="float64")
```
MiroDudik, line 52: I don't think that we need to be using this convention in classes that are not estimators. I'd rather revert these.

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -35,37 +35,43 @@ class ExponentiatedGradient(BaseEstimator, MetaEstimatorMixin):
         violation is at most :code:`2*(eps+best_gap)`
     :type eps: float
 
-    :param T: Maximum number of iterations
-    :type T: int
+    :param max_iterations: Maximum number of iterations
+    :type max_iterations: int
```
MiroDudik, line 40: if we are renaming this, let's follow [sklearn](https://scikit-learn.org/stable/glossary.html#parameters): `max_iter`

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -35,37 +35,43 @@ class ExponentiatedGradient(BaseEstimator, MetaEstimatorMixin):
...
     :type nu: float
 
-    :param eta_mul: Initial setting of the learning rate
-    :type eta_mul: float
+    :param learning_rate: Initial setting of the learning rate
+    :type learning_rate: float
+
+    :param run_lp_step: if True each step of exponentiated gradient is followed by the saddle
+        point optimization over the convex hull of classifiers returned so far; default True
+    :type run_lp_step: bool
```
MiroDudik, line 51: maybe `run_lp_step` -> `linprog_step`?

romanlutz: `linprog` seems easier to read.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
romanlutz: Sounds good!

CHANGES.md
```diff
@@ -25,6 +29,8 @@
 * Fix warning due to changing default `dtype` when creating an empty
   `pandas.Series`.
 * Enable `GridSearch` for more than two sensitive features values.
+* Change `ExponentiatedGradient` signature by renaming argument `T` to
+  `max_iterations`, `eta_mul` to `learning_rate`, and by adding `run_lp_step`.
 * Add new disparity constraints for reductions methods as moments in
```
MiroDudik, line 34: remember to update the names here according to corrections

romanlutz: ... and it's also in the wrong release because it didn't make 0.4.6 :-)

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -35,37 +35,43 @@ class ExponentiatedGradient(BaseEstimator, MetaEstimatorMixin):
...
+        self.max_iterations = max_iterations
+        self.nu = nu
+        self.learning_rate = learning_rate
+        self.run_lp_step = run_lp_step
+
+        self.best_gap_ = None
+        self.predictors_ = None
+        self.weights_ = None
+        self.last_t_ = None
+        self.best_t_ = None
```
MiroDudik, line 68: `last_iter_` and `best_iter_` to match `max_iter`?

romanlutz: Good point! I like that

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -35,37 +35,43 @@ class ExponentiatedGradient(BaseEstimator, MetaEstimatorMixin):
...
+        self.n_oracle_calls_ = 0
+        self.n_oracle_calls_dummy_returned_ = 0
+        self.oracle_execution_times_ = None
+        self.lambda_vecs_ = pd.DataFrame()
+        self.lambda_vecs_LP_ = pd.DataFrame()
+        self.lambda_vecs_lagrangian_ = pd.DataFrame()
```
MiroDudik, line 75: Actually, I think that we will need to tweak these names:
* `lambda_vecs_` -> `lambda_vecs_EG_` [the dimensionality of EG and LP variants is `n_iter_`]
* `lambda_vecs_lagrangian_` ->`lambda_vecs_` [this is because it matches the dimension of `predictors_` and `weights_`]

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
MiroDudik: also... i wonder whether we don't want to replace `last_iter_` by `n_iter_` which would be equal to `last_iter_+1`

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -35,37 +35,43 @@ class ExponentiatedGradient(BaseEstimator, MetaEstimatorMixin):
...
 
     :param nu: Convergence threshold for the duality gap, corresponding to a
         conservative automatic setting based on the statistical uncertainty in measuring
         classification error
     :type nu: float
 
-    :param eta_mul: Initial setting of the learning rate
-    :type eta_mul: float
+    :param learning_rate: Initial setting of the learning rate
+    :type learning_rate: float
```
MiroDudik, line 47: i'm not sure how i feel about renaming this one... I know that this better matches the current description, but the actual initial learning rate is `eta_mul / B`. Plus, `sklearn` is also using `eta0` in this context: https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.SGDClassifier.html

romanlutz: okay, if this is commonly used we can do `eta0`. I just didn't like it to be `mul` because that kind of conflicts with the "initial setting" description. I expected it to be multiplied with `eta` (learning rate). `eta0` makes intuitively sense.

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
romanlutz: Done!

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -35,37 +35,43 @@ class ExponentiatedGradient(BaseEstimator, MetaEstimatorMixin):
...
 
-    :param eta_mul: Initial setting of the learning rate
-    :type eta_mul: float
+    :param eta0: Initial setting of the learning rate
+    :type eta0: float
+
+    :param run_linprog_step: if True each step of exponentiated gradient is followed by the saddle
+        point optimization over the convex hull of classifiers returned so far; default True
+    :type run_linprog_step: bool
     """
```
MiroDudik, line 52: i'm being a bit nitpicky. but can we just do `linprog_step` rather than `run_linprog_step`?

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -35,37 +35,43 @@ class ExponentiatedGradient(BaseEstimator, MetaEstimatorMixin):
...
+                 eta0=2.0, run_linprog_step=True):  # noqa: D103
+        self.estimator = estimator
+        self.constraints = constraints
+        self.eps = eps
+        self.max_iter = max_iter
+        self.nu = nu
+        self.eta0 = eta0
+        self.run_linprog_step = run_linprog_step
+
+        self.best_gap_ = None
```
MiroDudik, line 64: These initializations should not be here, but at the very beginning of fit (only if we need them). See: https://github.com/scikit-learn/scikit-learn/blob/fd237278e/sklearn/utils/validation.py#L956

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
romanlutz: That doesn't soudn like a bool, though.

fairlearn/reductions/_grid_search/grid_search.py
```diff
@@ -90,12 +90,12 @@ def __init__(self,
...
-        self._lambda_vecs = pd.DataFrame()
-        self._objectives = []
-        self._gammas = pd.DataFrame()
-        self._oracle_execution_times = []
+        self.best_grid_index_ = None
+        self.predictors_ = []
+        self.lambda_vecs_ = pd.DataFrame()
+        self.objectives_ = []
+        self.gammas_ = pd.DataFrame()
+        self.oracle_execution_times_ = []
```
MiroDudik, line 98: There's a lot of stuff in this `__init__` that should go into `fit`. But we can hold off for a future PR?

romanlutz: Just did that locally. :-) Thanks!

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
romanlutz: Most of them are not even needed. We just didn't like the practice of suddenly defining things outside of `__init__` so far but that's simply how Python folks do that...

MiroDudik: We can possibly hold off with this for a future PR--your call.

fairlearn/reductions/_grid_search/grid_search.py
```diff
@@ -90,12 +90,12 @@ def __init__(self,
         self.grid_offset = grid_offset
         self.grid = grid
 
-        self._best_grid_index = None
-        self._predictors = []
-        self._lambda_vecs = pd.DataFrame()
-        self._objectives = []
-        self._gammas = pd.DataFrame()
-        self._oracle_execution_times = []
+        self.best_grid_index_ = None
```
MiroDudik: since we're in business of renaming things. i think i might prefer `best_index_` or even `best_idx_` over `best_grid_index` (is it an index of the best grid?)

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
MiroDudik: this is fine...

fairlearn/reductions/_grid_search/grid_search.py
MiroDudik: this is reasonable. there's also the validation code and a conversion using `float`, but we can tackle that in a future PR.

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
MiroDudik: sure

notebooks/Binary Classification with the UCI Credit-card Default Dataset.ipynb
```diff
@@ -227,7 +227,7 @@
     "lgb_params = {\n",
     "    'objective' : 'binary',\n",
     "    'metric' : 'auc',\n",
-    "    'learning_rate': 0.03,\n",
+    "    'eta0': 0.03,\n",
```
MiroDudik, line 230: this is a bug. revert back to `learning_rate`.

romanlutz: Yes, found that last week :-) Already changed locally but I ran into some more issues that I'm fixing now.

fairlearn/reductions/_grid_search/grid_search.py
romanlutz: renamed to `best_idx_`

fairlearn/reductions/_grid_search/grid_search.py
romanlutz: The validation code needs to be addressed here because it requires initialized members which we don't have anymore now :-)

PR #393, opened 2020-04-27
Simplified Release Pipeline

riedgar-ms:
Create a new, simplified release pipeline. This is based on the suggestion of @adrinjalali , who pointed out that uploading and then installing from Test-PyPI didn't offer much advantage over telling `pip` to install from the build wheel file. This pipeline:
1. Performs a redeployment test run
1. Creates the wheel and stores it in an Artifact
1. Installs from that wheel file, and runs tests
1. Uploads to PyPI
1. Runs tests against the version uploaded to PyPI

Whether to use Test or Prod PyPI is set by the user at queue time.

Once this has been tested with an actual production release, we will be able to remove the old release pipeline, and remove the logic surrounding the `DEV_VERSION` queue variable and `.dev[n]` version suffix in `__init__.py` (and the release scripts).

Related to #371  but will not close until the aforementioned deletions have taken place.


devops/templates/pypi-variables.yml
```diff
@@ -0,0 +1,21 @@
+parameters:
```
riedgar-ms, line 1: This file is actually unused right now... I can't figure out how to use a template for variables (and the ADO documentation doesn't have a full example).

devops/pypi-release-new.yml
```diff
@@ -0,0 +1,102 @@
+# Simplified PyPI release pipeline
+
+# At queue time, the user selects a Test or Production deployment. The following stages
+# then Run
```
romanlutz, line 4: The sentence "The following stages then Run" seems odd to me. Did you mean to write "run" with lowercase R or something entirely different?

riedgar-ms: Yes, I meant lower case 'r' - I've fixed it.

devops/pypi-release-new.yml
```diff
@@ -0,0 +1,102 @@
...
+  - Test
+  - Production
+
+variables:
+  poolImage: "ubuntu-latest"
+  poolPythonVersion: 3.6
+  # Following is actually the name of the service connection from the Project Settings page
+  kvSubscription: "Fairness - Automation (cecafb73-04ae-4432-9f96-d96925d28058)"
+  kvVaultName: fairlearndeploy
+  ${{ if eq(parameters.releaseType, 'Test')}}:
```
romanlutz, line 26: I thought the point was to get rid of pypi Test deployments?

devops/templates/post-upload-validation.yml
```diff
@@ -0,0 +1,63 @@
+# Stage template to validate uploaded packag from PyPI
```
romanlutz, line 1: nit: typo packag

devops/templates/post-upload-validation.yml
```diff
@@ -0,0 +1,63 @@
...
+
+# Downloads fairlearn from the specified PyPI (Test or Prod) at the version
```
romanlutz, line 3: Do we want to write Fairlearn with uppercase F? I'm not sure...

devops/templates/post-upload-validation.yml
```diff
@@ -0,0 +1,63 @@
...
+# specified in the Version Artifact. Then runs tests against that install
+
+parameters:
```
romanlutz, line 6: I wonder whether it's simpler not to have this as a template. Briefly touched on this with @eedeleon today and we've created such a high barrier to entry with all the templating that I'd actually take a step back and perhaps remove all templates that aren't used multiple times. We have 30 lines of input parameter spec here for 30 lines of actual logic.

devops/templates/pypi-deployment-stages-template.yml
```diff
@@ -75,6 +75,12 @@ stages:
                 keyVaultName:  ${{parameters.kvVaultName}}
                 secretsFilter: '${{parameters.kvUsername}},${{parameters.kvPassword}}'
 
+            - script: echo $(usernametest)
```
romanlutz, line 78: Did you mean to remove this?

devops/templates/pypi-variables.yml
romanlutz: If it's unused at the very least document that at the top of the file. Ideally let's not check in things that aren't used. As you may have noticed I'm not a huge fan of templating everything because it becomes unreadable.

devops/pypi-release-new.yml
riedgar-ms: This new pipeline can deploy to either independently. The old pipeline required a Test-PyPI deployment before going to Prod.

devops/templates/post-upload-validation.yml
riedgar-ms: The idea was to make the top level file as clean as possible, so the individual stages could be seen clearly. I agree that mustache templating is not the nicest, though.

devops/templates/post-upload-validation.yml
riedgar-ms: Good point

riedgar-ms:
I'm going to experiment with removing a layer of templating. Closing this to avoid a tsunami of builds

PR #396, opened 2020-04-29
Extend CircleCI workflow for GitHub pages deployment

romanlutz:
This is another step to make our documentation show up on our website according to https://github.com/fairlearn/fairlearn-proposals/pull/8

I've validated this using my own forks and the entire setup there. Just have to make sure I generate the key the right way again for our repos.  See this commit as the example: https://github.com/romanlutz/fairlearn.github.io/commit/ae959f7f76937cd0c1ca99505c07624ee91b54a7


romanlutz:
> 
> 
> This is another step to make our documentation show up on our website according to [fairlearn/fairlearn-proposals#8](https://github.com/fairlearn/fairlearn-proposals/pull/8)
> 
> I've elaborated on the reasons for using a GitHub Action as opposed to CircleCI in the YAML file itself. If you have a simple solution to add it to CircleCI I'm all ears. Since this is a rather simple definition I'd prefer getting this in and perhaps exchanging it (or the CircleCI one) later in favor of a unified solution.
> 
> Unfortunately, we can't validate this since it's triggered only by pushes to the master branch. I have previously tried it on my fork and that worked (although with a different repo). I'll try to replicate that once more.
> 
> Signed-off-by: Roman Lutz [rolutz@microsoft.com](mailto:rolutz@microsoft.com)

OK, works with token, although I'm unsure what I'd use for the token in this repo. In my own I can use my personal token,  but here I'm more reluctant to do that...

This is the result btw: https://github.com/romanlutz/fairlearn.github.io

PR #398, opened 2020-04-30
Simplify Release Pipeline

riedgar-ms:
Create a new, simplified release pipeline. This is based on the suggestion of @adrinjalali , who pointed out that uploading and then installing from Test-PyPI didn't offer much advantage over telling `pip` to install from the build wheel file. This pipeline:
1. Performs a redeployment test run
1. Creates the wheel and stores it in an Artifact
1. Installs from that wheel file, and runs tests
1. Uploads to PyPI
1. Runs tests against the version uploaded to PyPI

Whether to use Test or Prod PyPI is set by the user at queue time.

Once this has been tested with an actual production release, we will be able to remove the old release pipeline, and remove the logic surrounding the `DEV_VERSION` queue variable and `.dev[n]` version suffix in `__init__.py` (and the release scripts).

Related to #371  but will not close until the aforementioned deletions have taken place.


devops/pypi-release-new.yml
```diff
@@ -0,0 +1,257 @@
...
+      freezeFileStem: $(freezeFileStem)
+      installationType: 'PyPI'
+      targetType: ${{parameters.releaseType}}
+      versionArtifactName: $(versionArtifactName)
+      versionArtifactFile: $(versionFileName)
+
+  - template: templates/all-tests-job-template.yml
+    parameters:
+      platforms: { MacOS: macos-latest }
+      testRunTypes: ['Unit']
```
romanlutz, line 251: This one has Unit only, while Windows/Linux above doesn't have that. Why is that?

riedgar-ms: We've never run notebook tests of MacOS. We should probably look into that again, but that's not for this PR.

PR #404, opened 2020-05-01
Fix for Nightly-Fixed

riedgar-ms:
Get the Nightly-Fixed build green again by dropping Python 3.5. Since several of our dependencies are dropping Python 3.5 support, it does not make sense to keep it in our builds if giving trouble.


riedgar-ms:
Green build:
https://dev.azure.com/responsibleai/fairlearn/_build/results?buildId=5824&view=results

PR #407, opened 2020-05-04
Handle degenerate y values in reductions

### LINKED ISSUE #395: Handle exceptions when relabeling results in single class
#### Describe the bug
Our reductions methods relabel samples occasionally, so it's possible that a class (label) is entirely erased, resulting in a single class as opposed to two classes (for binary classification). Certain estimators are unable to handle this and throw and exception. Since this isn't unintended from a Fairlearn perspective we should catch this error and handle it appropriately.

#### Steps/Code to Reproduce
Example with `LogisticRegression` as a representative of classifiers that don't work with a single class.
```python
>>> from sklearn.linear_model import LogisticRegression
>>> LogisticRegression().fit([[1],[2],[3]], [0,0,0])
  File "<stdin>", line 1, in <module>
  File "~\AppData\Local\Continuum\anaconda3\lib\site-packages\sklearn\linear_model\_logistic.py", line 1558, in fit
    " class: %r" % classes_[0])
ValueError: This solver needs samples of at least 2 classes in the data, but the data contains only one class: 0
```

Another example with different error message
```python
>>> from sklearn.ensemble import GradientBoostingClassifier
>>> GradientBoostingClassifier().fit([[1],[2],[3]], [0,0,0])
  File "<stdin>", line 1, in <module>
  File "~\AppData\Local\Continuum\anaconda3\lib\site-packages\sklearn\ensemble\_gb.py", line 1455, in fit
    y = self._validate_y(y, sample_weight)
  File "~\AppData\Local\Continuum\anaconda3\lib\site-packages\sklearn\ensemble\_gb.py", line 2097, in _validate_y
    % n_trim_classes)
ValueError: y contains 1 class after sample_weight trimmed classes with zero weights, while a minimum of 2 classes are required.
```

#### Expected vs. Actual Results
Fairlearn should just skip fitting in this case and use a classifier that always returns the one remaining class.
We received a `ValueError` in the case above. However, even within `sklearn` the returned error messaged isn't consistent as shown above. For that reason we can't assume that we'll be able to check for those error messages. `ValueError` is too generic to simply catch, either. It may be best to first check for the number of unique classes in `y` before calling `fit`.

#### Versions

System:
    python: 3.7.4 (default, Aug  9 2019, 18:34:13) [MSC v.1915 64 bit (AMD64)]
executable: ~\AppData\Local\Continuum\anaconda3\python.exe
   machine: Windows-10-10.0.18362-SP0

Python dependencies:
       pip: 19.2.3
setuptools: 41.4.0
   sklearn: 0.22.2.post1
     numpy: 1.18.1
     scipy: 1.4.1
    Cython: 0.29.13
    pandas: 0.25.1
matplotlib: 3.1.3
    tempeh: 0.1.12

riedgar-ms:
During the course of our `reductions` algorithms, it is possible for a relabelled dataset with only a single label to be present. Some underlying Estimators (such as `LogisticRegression` in `sklearn`) throw an exception for this. In these cases, substitute the `DummyClassifier` from `sklearn`, set to return a constant.

Closes #395


test/unit/reductions/exponentiated_gradient/test_exponentiatedgradient_smoke.py
```diff
@@ -238,3 +241,36 @@ def test_simple_fit_predict(self, Constraints):
...
+        A = A + 11
+        A[0] = 10
+
+        estimator = LogisticRegression(solver='liblinear',
+                                       fit_intercept=True,
+                                       random_state=97)
+        expgrad = ExponentiatedGradient(estimator, DemographicParity())
+
+        # Following line should not throw an exception
+        expgrad.fit(X, y, sensitive_features=A)
```
riedgar-ms, line 267: This `fit()` call is not triggering the single-class case. @MiroDudik , is there an easy way of forcing the algorithm to produce a single label? The approach above is borrowed from the equivalent `GridSearch` test, but expgrad is avoiding the problem.

romanlutz: Could you just provide only one class?

test/unit/reductions/grid_search/test_grid_search_demographicparity.py
```diff
@@ -222,3 +223,36 @@ def test_can_specify_and_generate_lambda_vecs(A_two_dim):
...
+    # Setup with data designed to result in "all single class"
+    # at some point in the grid
+    X_dict = {
+        "c": [0, 1, 4, 1, 5, 1, 6, 0, 2, 4],
+        "d": [1, 5, 1, 6, 2, 3, 5, 1, 5, 2]
+    }
+    X = pd.DataFrame(X_dict)
+
+    y = [0, 0, 0, 0, 0, 0, 0, 0, 0, 1]
+    A = [3, 3, 3, 3, 3, 3, 3, 3, 3, 11]
```
romanlutz, line 238: Any specific reason for 3 and 11? Might confuse people. Perhaps best to make it "A" and "B" or at least something simpler than random numbers. Or comment that these are just group identifiers

test/unit/reductions/exponentiated_gradient/test_exponentiatedgradient_smoke.py
riedgar-ms: Actually, now I look, yes. I thought we had an upfront check that required a binary input, but it actually only requires that the unique values of Y are a _subset_ of {0, 1}.

test/unit/reductions/grid_search/test_grid_search_demographicparity.py
riedgar-ms: Will do. Quotes are more tedious to type.

fairlearn/reductions/_constant_predictor.py
```diff
@@ -0,0 +1,9 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+class ConstantPredictor:
+    def __init__(self, constant_value):
+        self._constant_value = constant_value
+
+    def predict(self, *args, **kwargs):
```
riedgar-ms, line 8: Actually, this is bad. The predictor needs to respond to array-valued argument correctly. I'll have to fix this.

test/unit/reductions/exponentiated_gradient/test_exponentiatedgradient_smoke.py
romanlutz: We could restrict it further, but I don't see why...

riedgar-ms: I'm happy with not adding more restrictions :-)

notebooks/Mitigating Disparities in Ranking from Binary Data.ipynb
```diff
@@ -77,9 +77,140 @@
   },
   {
    "cell_type": "code",
-   "execution_count": null,
+   "execution_count": 1,
```
MiroDudik, line 80: I don't think you meant to check in the changes here. Please double check this one and all other notebook changes.

riedgar-ms: Ouch. No I didn't. I hadn't spotted the number of changed files exploding :-(

test/unit/reductions/exponentiated_gradient/test_lagrangian.py
```diff
@@ -119,3 +120,40 @@ def test_call_oracle(Constraints, eps, mocker):
     assert (kwargs['sample_weight'] == redW).all()
     assert lagrangian.n_oracle_calls == 1
     assert len(lagrangian.oracle_execution_times) == 1
+
+
+@pytest.mark.parametrize("Constraints", [DemographicParity, EqualizedOdds])
```
MiroDudik, line 125: This is okay, but I'm a bit worried about the growing copy-paste overlap among all the tests of the Lagrangian. I wonder if there's a good way to refactor. But maybe better to leave that for a future PR.

test/unit/reductions/grid_search/test_grid_search_demographicparity.py
```diff
@@ -222,3 +223,36 @@ def test_can_specify_and_generate_lambda_vecs(A_two_dim):
...
+                             grid_limit=0.1,
+                             grid_offset=5)
+
+    # We want to avoid an exception on the following line
+    grid_search.fit(X, y, sensitive_features=A)
+
+    # Check all predictors are DummyClassifiers
+    for p in grid_search._predictors:
+        assert isinstance(p, DummyClassifier)
+        assert np.array_equal(p.predict([1, 2]), [y_val, y_val])
```
MiroDudik, line 258: The test is a bit weird since the predictor was trained on the `X` array with two features 'c' and 'd', but I think you are evaluating it on a list which is interpreted as a single column rather than a single row. Maybe better to do `np.array_equal(p.predict(X), y)` or at least some `X_test` whose type matches `X`?

test/unit/reductions/exponentiated_gradient/test_lagrangian.py
```diff
@@ -119,3 +120,40 @@ def test_call_oracle(Constraints, eps, mocker):
...
+    # object for it to avoid the dependence on the lagrangian object.
+    constraints.load_data(X, y, sensitive_features=A)
+    objective = constraints.default_objective()
+    objective.load_data(X, y, sensitive_features=A)
+    theta = pd.Series(0, constraints.index)
+    lambda_vec = np.exp(theta) / (1 + np.exp(theta).sum())
+
+    result_estimator = lagrangian._call_oracle(lambda_vec)
+    assert isinstance(result_estimator, DummyClassifier)
+    assert result_estimator.predict([1]) == 1
```
MiroDudik, line 155: This test is a bit weird, see my comment in `test_grid_search`.

test/unit/reductions/exponentiated_gradient/test_lagrangian.py
riedgar-ms: I would agree that test refactoring is something which needs doing. Although the thing which triggered that realisation was needing to identify which of the twenty-or-so dictionaries in the smoke-test data required changing. I do think it's better to leave that for a future PR, when we're _only_ changing the tests, and not the underlying functionality.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -53,6 +55,7 @@ def __init__(self, X, sensitive_features, y, estimator, constraints, eps, B, opt
         self.lambdas = pd.DataFrame()
         self.n = self.X.shape[0]
         self.n_oracle_calls = 0
+        self.n_dummy_classifiers = 0
```
MiroDudik, line 58: Can we rename to something like: `n_oracle_calls_dummy_returned`. Below I think we should include the dummy calls among oracle calls (since they are technically the oracle calls).

test/unit/reductions/grid_search/test_grid_search_demographicparity.py
riedgar-ms: Yes, that was a curiosity which I had not been paying too much attention to. I had been hoping that the `DummyClassifier` would use the supplied `X` to generate a schema check, but it appears that it doesn't. I'll update the test.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -137,11 +140,19 @@ def _call_oracle(self, lambda_vec):
...
-        oracle_call_start_time = time()
-        classifier.fit(self.X, redY, sample_weight=redW)
-        self.oracle_execution_times.append(time() - oracle_call_start_time)
-        self.n_oracle_calls += 1
+        redY_unique = np.unique(redY)
+        if len(redY_unique) == 1:
+            logger.debug("redY had single value. Using DummyClassifier")
+            classifier = DummyClassifier(strategy='constant',
+                                         constant=redY_unique[0])
+            classifier.fit(self.X, redY, sample_weight=redW)
```
MiroDudik, line 148: Part of me wants to pull `fit` out of the if statement and add the `fit` time to oracle execution times. If the goal is to track performance, we shouldn't assume that `DummyClassifier` is a no-op.

MiroDudik: Also--I think that we should always increment `n_oracle_calls`.

fairlearn/reductions/_grid_search/grid_search.py
```diff
@@ -165,12 +167,20 @@ def fit(self, X, y, **kwargs):
...
-            oracle_call_start_time = time()
-            current_estimator.fit(X, y_reduction, sample_weight=weights)
-            oracle_call_execution_time = time() - oracle_call_start_time
-            logger.debug("Call to underlying estimator complete")
+            y_reduction_unique = np.unique(y_reduction)
+            if len(y_reduction_unique) == 1:
+                logger.debug("y_reduction had single value. Using DummyClassifier")
+                current_estimator = DummyClassifier(strategy='constant',
+                                                    constant=y_reduction_unique[0])
+                current_estimator.fit(X, y_reduction, sample_weight=weights)
```
MiroDudik, line 175: Similar to `ExponentiatedGradient`, can we pull out `fit` and time tracking out of `if`?

test/unit/reductions/exponentiated_gradient/test_lagrangian.py
MiroDudik: absolutely!

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
riedgar-ms: I see your point, but.... to my mind, the use of `n_oracle_calls` is for finding out the average time spent in the underlying classifier. That would suggest renaming `n_oracle_calls` to something like `n_oracle_calls_supplied_estimator`  but that's a bit unwieldly.

Is `n_oracle_calls` interesting from an algorithmic standpoint as well? I could perhaps add a get-only property calculated from `n_oracle_calls` and `n_oracle_calls_dummy_returned` to provide a way of getting at that mean.

MiroDudik: How about this:
* let's make `n_oracle_calls` count the number of times `_call_oracle` is called, so it's inclusive of both kinds of calls; this is an important characteristic of the optimization algorithm and a more fundamental property than how often the actual classifier gets called (the fact that some standard classifiers don't handle the trivial case is just a fluke that put us down this path)
* I think it's also useful to count `n_oracle_calls_dummy_returned`
* if you want to get an idea about the average time you're talking about, you can either do:
  * `sum(oracle_execution_times) / (n_oracle_calls - n_oracle_calls_dummy_returned)`; it's a slight overestimation, but for now I think it will do just fine
  * at some point in future, we can improve tracking of oracle calls and in addition to the times, we can also track which estimator was invoked, what was the incoming lambda vector, whether that oracle call was "saved in the cache" (as part of `best_h`) etc.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
riedgar-ms: Modified as discussed

test/unit/reductions/grid_search/test_grid_search_demographicparity.py
riedgar-ms: Fixed.

test/unit/reductions/exponentiated_gradient/test_lagrangian.py
riedgar-ms: Fixed

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
riedgar-ms: Fixed as discussed.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
MiroDudik: is there a reason why not pull out `fit` from the if/then statement?

PR #409, opened 2020-05-04
Convert terminology & contribution guides into generated docs on the webpage

romanlutz:
To see the generated output click on the CircleCI check, build_doc step, and select the artifacts, where you need to select the one ending in `index.html`. Example from first push https://105-133444044-gh.circle-artifacts.com/0/docs/_build/html/index.html

There's no new content here, it already existed in CONTRIBUTING.d and TERMINOLOGY.md, and has been moved to the website.


README.md
```diff
@@ -32,7 +32,8 @@ An AI system can behave unfairly for a variety of reasons. In Fairlearn, we defi
 We follow the approach known as **group fairness**, which asks: _Which groups of individuals are at risk for experiencing harms?_ The relevant groups need to be specified by the data scientist and are application specific.
 
 Group fairness is formalized by a set of constraints, which require that some aspect (or aspects) of the AI system's behavior be comparable across the groups. The Fairlearn package enables assessment and mitigation of unfairness under several common definitions.
-To learn more about our definitions of fairness, please visit our [terminology page](./TERMINOLOGY.md#fairness-of-ai-systems).
+To learn more about our definitions of fairness, please visit our
+[terminology page](https://fairlearn.github.io/user_guide/fairness_in_machine_learning.html#fairness-of-ai-systems).
```
riedgar-ms, line 36: It appears that the word 'terminology' doesn't actually appear on the linked page.... so I wouldn't call it a 'terminology page.'

docs/conf.py
```diff
@@ -77,3 +96,15 @@
      'examples_dirs': '../examples',
      'gallery_dirs': 'auto_examples',
 }
+
+# -- LaTeX macros ------------------------------------------------------------
+
+mathjax_config = {                  
+    "TeX": {                        
+        "Macros": {                 
+            "E": '{\\mathbb{E}}',
```
riedgar-ms, line 105: We're going to need to tell users where to find these macros.

riedgar-ms: (not necessarily in this PR)

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -1,4 +1,144 @@
...
+For example, in Fairlearn, we consider the following types of parity constraints.
+
+*Binary classification*:
+
+* *Demographic parity* (also known as *statistical parity*): A classifier
+  :math:`h` satisfies demographic parity under a distribution over
+  :math:`(X, A, Y)` if its prediction :math:`h(X)` is statistically
+  independent of the sensitive feature :math:`A`. This is equivalent to
+  :math:`\E[h(X) \given A=a] = \E[h(X)] \quad \forall a`.
+  `[Agarwal, Beygelzimer, Dudik, Langford, Wallach] <https://arxiv.org/pdf/1803.02453.pdf>`__
```
riedgar-ms, line 94: Can I suggest adding the year to all paper citations? Unless that's not the done thing in CS?

docs/conf.py
romanlutz: Not sure I understand. Why would users have to know about logic to generate a webpage that they're reading?

docs/user_guide/fairness_in_machine_learning.rst
romanlutz: There are as many citation styles as there are definitions of what constitutes a sandwich :-) I think it mainly depends on what people like. I've certainly seen all kinds of variations. 
[I personally always prefer seeing all authors since it gives me a better intuition on what paper it was. We had two different versions of Agarwal et al. in the terminology page, for example, which isn't terribly helpful until you click on the link.]
What we should be doing is define a standard format that every citation should abide by. Let's just do whatever scikit-learn does.

docs/conf.py
riedgar-ms: s/users/contributors/

docs/user_guide/fairness_in_machine_learning.rst
riedgar-ms: Personally, I always let BibTeX do its thing :-)

I'd be in favour of including the year, since sometimes the same set of collaborators do a series of papers.

Actually, do we want a bibliography page instead?

romanlutz: BibTex also just takes a configuration, so it depends on what document type you use... you probably just use the default of your document type without realizing it.

Scikit-learn does the following:
- inline it's just `Wu et al. (2004).`
- and it's linked in the "references" section per subsection https://scikit-learn.org/stable/modules/svm.html#scores-and-probabilities 
image

adrinjalali: As you can see in https://raw.githubusercontent.com/scikit-learn/scikit-learn/master/doc/modules/svm.rst

we use two styles of referencing:

```
.. _`liblinear`: https://www.csie.ntu.edu.tw/~cjlin/liblinear/

.. topic:: References:

   .. [#1] Platt `"Probabilistic outputs for SVMs and comparisons to
      regularized likelihood methods"
      <https://www.cs.colorado.edu/~mozer/Teaching/syllabi/6622/papers/Platt1999.pdf>`_.
```
and in the `.rst` code, we either have `[#1]_` or:

    `liblinear`_

rendered version: https://scikit-learn.org/dev/modules/svm.html

docs/conf.py
romanlutz: I see! It says "LaTeX macros" right above. That's what I would search for. I do agree that we need some sort of "readme" for the doc generation, though. I'll start with that, and we can keep adding details over the next few PRs.

riedgar-ms: Yes... having something you have to find using `grep -ilr` is not ideal :-)

docs/user_guide/fairness_in_machine_learning.rst
romanlutz: I've copied that pattern and got a very surprising result! Still trying to determine why/how
Output inline:
image

references section: [note that 4 and 6 were cited twice, so this seems to be an odd way of handling the formatting... I don't like it.]
image

riedgar-ms: Wandering off-topic, I do remember a few `\bibliographystyle{}` commands too. As you say, most of the time, I was using that set by the `\documentclass{}` though :-)

Actually, just searching a little, perhaps:
https://sphinxcontrib-bibtex.readthedocs.io/en/latest/quickstart.html
would be helpful?

romanlutz: Looks like this is the intended pattern. We can always adapt it in the future, of course.

docs/conf.py
romanlutz: I've added a README. I still think people either know how sphinx works in which case they'll go to conf.py, or they don't, in which case they'll search for LaTeX and will find it. Anyway, it's documented now.

README.md
```diff
@@ -57,8 +58,7 @@ Fairlearn contains the following algorithms for mitigating unfairness in binary
...
 
 > _Note_:
-> DP refers to demographic parity, EO to equalized odds, and BGL to bounded group loss. For more information on these and other terms we use in this repository please refer to the [terminology page](./TERMINOLOGY.md). To request additional algorithms or fairness definitions, please open a [new issue](https://github.com/fairlearn/fairlearn/issues).
-
+> DP refers to demographic parity, EO to equalized odds, and BGL to bounded group loss. For more information on these and other terms we use in this repository please refer to the [terminology page](https://fairlearn.github.io/user_guide/fairness_in_machine_learning.html). To request additional algorithms or fairness definitions, please open a [new issue](https://github.com/fairlearn/fairlearn/issues).
```
riedgar-ms, line 61: This 'terminology' should change too.

romanlutz: I know. I have that covered with lots of other changes since, but didn’t want to club actual changes with just migrations too much.

Issue #339, opened 2020-03-21
Bug: class labels not preserved


romanlutz:
```
>>> eg = ExponentiatedGradient(constraints=DemographicParity(), estimator=LogisticRegression())
>>> eg.fit(pd.DataFrame([[1], [2], [3]]), [2, 0, 0], sensitive_features=[3,2,4])
>>> eg.predict([[1], [2], [3]])
0    0
1    0
2    1
dtype: int32
```

romanlutz:
@MiroDudik should we restrict to just 0/1? We definitely want to consider multi-class classification in the not so distant future, so I don't think it makes sense.

adrinjalali:
I've been wanting to mention that there's an issue with handling the outputs in the current codebase. As examples, here's how we handle the output in the `GradientBoostingClsasifier`:

https://github.com/scikit-learn/scikit-learn/blob/dcfb3df9a3df5aa2a608248316d537cd6b3643ee/sklearn/ensemble/_gb.py#L1095-L1105

and in `HistGradientBoostingClassifier`:

https://github.com/scikit-learn/scikit-learn/blob/dcfb3df9a3df5aa2a608248316d537cd6b3643ee/sklearn/ensemble/_hist_gradient_boosting/gradient_boosting.py#L1129-L1142

romanlutz:
If this is addressed after the completion of #332 then we need to adjust the parts where we expect labels 0/1.

adrinjalali:
generally, you just need to limit the case to a binary classification, not necessarily a `0/1` output, and where not clear, you can accept a `pos_label` to set the positive output.

romanlutz:
I agree. At least, that applies unless we have logic that relies on 0/1 like in PR #332 
The other alternative is to store the mapping from original labels to 0...m
I'm very much thinking of the longer term here because we won't always want to restrict Exponentiated Gradient to binary classification.

riedgar-ms:
Should this be closed? @moprescu  signed off on the fix.

romanlutz:
Which fix?

MiroDudik:
Our reductions for fair classification strongly rely on y being 0/1. Going to multi-class would require non-trivial algorithmic extensions. 

Handling arbitrarily encoded **binary** classification should be doable, but it is more involved than meets the eye, because:
* Fairness criteria / fairness metrics rely on a specific `pos_label` (which is right now a constant)
[...]

PR #410, opened 2020-05-06
Check for 0-1 labels in ExponentiatedGradient classification tasks

riedgar-ms:
This is a quick fix for #339  pending a more uniform fix which will enable users to specify a positive label. It makes the input validation logic match for `ExponentiatedGradient` match that of `GridSearch`.


fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -73,7 +75,16 @@ def fit(self, X, y, **kwargs):
         :param y: The label vector
         :type y: numpy.ndarray, pandas.DataFrame, pandas.Series, or list
         """
-        _, y_train, sensitive_features = _validate_and_reformat_input(X, y, **kwargs)
+
+        if isinstance(self._constraints, ClassificationMoment):
+            logger.debug("Classification problem detected")
+            is_classification_reduction = True
+        else:
+            logger.debug("Regression problem detected")
```
romanlutz, line 83: We don't do regression with expgrad yet, right? Won't this err out somewhere?

riedgar-ms: No more than the code would right now surely?

MiroDudik: Eh. It appears that we had a bug here. If somebody tried calling ExponentiatedGradient with GroupLossMoment, there would be no complaint, but possibly a mysterious error / strange behavior down the line. I'm okay leaving this as Richard has right now. Or we could possibly raise an error right after "is_classification_reduction=False", but keep the rest of the PR including `enforce_binary_labels=is_classification_reduction`.

riedgar-ms: I believe you're right - passing in a moment for regression would result in 'interesting' behaviour.

I can raise an exception in place of logging "Regression problem detected" but since the Stanford students should be about to start adding something to cope, unless one of you insists.....

PR #414, opened 2020-05-07
Update the set of supported derived metrics

MiroDudik:
Updating the set of supported derived metrics to be more systematic. Also make the metrics for root_mean_squared_error private.


MiroDudik:
The changes re. creating derived metrics were already mentioned in CHANGES.md (so I didn't add any new information on that), but I updated the info about the functions with new names.

CHANGES.md
```diff
@@ -20,7 +20,8 @@
   * `miss_rate` in favor of `false_negative_rate`
   * `specificity_score` in favor of `true_negative_rate`
 * Change from public to private:
-  * `mean_{over,under}prediction` changed to `_mean_{over,under}prediction`
+  * `mean_{over,under}prediction` and `{balanced_,}root_mean_squared_error`
```
romanlutz, line 23: were these public pre-0.4.6? Because if these were just introduced and already made private all within the work leading up to 0.4.6 then there's no need to mention that.

riedgar-ms: Yes, they were public before, and they're used in the dashboard.

PR #416, opened 2020-05-11
Usecase

srayagarwal:


romanlutz:
Still quite a few binary files in there. Could you remove them please? You can use .gitignore if it helps.

notebooks/__init__.py
```diff
@@ -0,0 +1,91 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+"""Functionality for computing metrics, with a particular focus on group metrics.
```
romanlutz, line 4: I'm not sure I understand the purpose of this file. This is all functionality from fairlearn.metrics, isn't it? Why not directly import into your notebook?

requirements.txt
```diff
@@ -37,5 +47,16 @@ shap
...
 sphinx
+
+
+
+
+
+
+
 sphinx-gallery
+<<<<<<< HEAD
```
romanlutz, line 58: looks like a merge conflict

srayagarwal:
I used this as there were some bugs in my path. Removed it from my branch now

srayagarwal:
Removed the files. Let me know if there are still few files. Also please point me to the directory where you see binary files

srayagarwal:
> Still quite a few binary files in there. Could you remove them please? You can use .gitignore if it helps.

Done

romanlutz:
Is this data file from somewhere else? We usually prefer not having required files outside of the notebook, because that means the notebook can't run independently. Perhaps this can be downloaded in the notebook itself? If not, I'm happy to add it to `tempeh`, our library for such things. Then that dependency is externalized and we can fully focus on the notebook.

srayagarwal:
I downloaded this data from an open-source platform and then conducted massive data transformation. What I am using is a transformed/processed data. Thus merely downloading it wont work. Can we keep the data somewhere in the repository from where user can access and download it.

romanlutz:
> I downloaded this data from an open-source platform and then conducted massive data transformation. What I am using is a transformed/processed data. Thus merely downloading it wont work. Can we keep the data somewhere in the repository from where user can access and download it.


Depends on various factors:

- where is it from?
- what’s the license on the dataset?
- do you have the steps to transform the original dataset?

I always find it rather odd to add in datasets to tool packages, especially if we don’t own the data. IMO it’s simpler to move such logic to a place like https://github.com/microsoft/tempeh or similar. That way people can use your notebook without having to download additional files. At some point this will be part of the project website with a download button, but without additionally finding the data it won’t be usable.

srayagarwal:
1.  where is it from? https://www.bondora.com/en/public-reports [manually downloadable]
  2.  Do you have the steps to transform the original dataset? Yes
  3.  Can we upload the transformed dataset to https://github.com/microsoft/tempeh or similar so that user wont have to go through multiple steps for data transformation. I have uploaded the transformed data on my branch.

romanlutz:
> 
> 1. where is it from? https://www.bondora.com/en/public-reports [manually downloadable] 2. Do you have the steps to transform the original dataset? Yes 3. Can we upload the transformed dataset to https://github.com/microsoft/tempeh or similar so that user wont have to go through multiple steps for data transformation. I have uploaded the transformed data on my branch. 

I'd certainly prefer adding it to `tempeh`. I'll add it there and will update you on how to use it. The terms from that website seem to indicate that we can't just store it in this repository (or any other): "Portal, Portal’s website and the copyright of the contents thereof belong to the Company. The User does not have the right to save, copy, change, transfer, forward or disclose the pages of the Portal for a purpose other than personal use."

srayagarwal:
Sure, please add it to temph and then let me know the next steps

requirements.txt
```diff
@@ -3,8 +3,18 @@ matplotlib>=3.2.1
 numpy>=1.17.2
 pandas>=0.25.1
 scikit-learn>=0.22.1
+
+
+math
```
romanlutz, line 8: These seems to be part of the standard library
https://docs.python.org/3/library/
except for seaborn (the only truly new one here) and scipy.stats (which is part of scipy)

srayagarwal:
Shall I edit the requirements.txt and recommit?

srayagarwal:
Closing for duplication

PR #417, opened 2020-05-11
Cleanup of release pipelines

### LINKED ISSUE #371: Simplify Release Pipeline

riedgar-ms:
Cleaning up the old release pipeline, and removing the logic around `DEV_VERSION` in `__init__.py`. This leaves things substantially simpler.

Closes #371


devops/pypi-release-new.yml
```diff
@@ -208,7 +207,7 @@ stages:
           - script: 'twine upload --verbose --repository-url $(pypiUrl) -u $(${{variables.kvUsername}}) -p "$(${{variables.kvPassword}})" $(Pipeline.Workspace)/$(packageArtifactName)/*'
             displayName: Upload to ${{parameters.releaseType}} PyPI
 
-        # TODO: Add GitHub Release task, so links in PyPI ReadMe will work without manual intervention
+        # TODO: Add GitHub Release task, so links in PyPI ReadMe will work without manual intervention (Prod only)
```
romanlutz, line 210: Moving away from "Prod" is a separate change I presume?

docs/contribution_guide.rst
```diff
@@ -168,37 +168,20 @@ file.
 1. Check that the dashboard loads in the notebooks
 
 We have a
-`Azure DevOps Pipeline <https://dev.azure.com/responsibleai/fairlearn/_build?definitionId=48&_a=summary>`_
+`Azure DevOps Pipeline <https://dev.azure.com/responsibleai/fairlearn/_build?definitionId=60&_a=summary>`_
```
romanlutz, line 171: This file has changed considerably in my PR. FYI

fairlearn/__init__.py
```diff
@@ -3,42 +3,19 @@
...
-            print(msg.format(_FAIRLEARN_DEV_VERSION_ENV_VAR), file=sys.stderr)
-    except ValueError:
-        msg = "Value of {0} in {1} did not parse to integer. Ignoring"
-        print(msg.format(dev_version_string, _FAIRLEARN_DEV_VERSION_ENV_VAR), file=sys.stderr)
-
-
 __name__ = "fairlearn"
-__version__ = "{0}{1}".format(_base_version, _dev_version)
-
+__version__ = "0.4.6.post0.dev0"
```
romanlutz, line 12: post0 AND dev0?

devops/pypi-release-new.yml
riedgar-ms: Yes it would be.

docs/contribution_guide.rst
riedgar-ms: OK.

fairlearn/__init__.py
riedgar-ms: Yes. `post` because it's coming after `v0.4.6` and `dev` to stop it automatically superseding `v0.4.6` if this should get released by accident. I could do `v0.4.7.dev0` instead.

romanlutz: I somehow expected to see 0.4.7.dev0
That's also consistent with sklearn https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/__init__.py

docs/contribution_guide.rst
romanlutz: This file itself doesn't exist anymore in my PR so I've copied the changes from this file into my branch. Just FYI

fairlearn/__init__.py
adrinjalali: I'd say the master branch is not a post release of the past main release, it's more of a dev release of the next release. If the last release is 0.4.6, then master should be 0.4.7.devN, since it's a development happening to release 0.4.7.

riedgar-ms: Changed to 0.4.7.dev0 :-)

PR #420, opened 2020-05-12
Build fixes

riedgar-ms:
Addressing broken builds
- Fix new `flake8` issue
- Add checks for the UCI Credit Card notebook
- Workaround `sklearn` and `LightGBM` issue in the UCI Credit Card notebook
- Support changing exception message for bad array shapes in `sklearn`

The fix for the changed exception message is ugly. We will need a better solution if this keeps happening.

### DIFF
--- a/test/notebooks/test_notebooks.py
+++ b/test/notebooks/test_notebooks.py
@@ -139,4 +139,12 @@ def test_mitigating_disparities_in_ranking_from_binary_data():
 def test_binary_classification_with_the_uci_credit_card_default_dataset():
     nb_name = "Binary Classification with the UCI Credit-card Default Dataset"
     test_values = {}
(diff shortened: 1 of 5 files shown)


test/notebooks/test_notebooks.py
```diff
@@ -139,4 +139,12 @@ def test_mitigating_disparities_in_ranking_from_binary_data():
 def test_binary_classification_with_the_uci_credit_card_default_dataset():
     nb_name = "Binary Classification with the UCI Credit-card Default Dataset"
     test_values = {}
+    test_values["GridSearch_17"] = ScrapSpec(
```
romanlutz, line 142: You could directly initialize the dict with these values... not a big deal I guess

PR #422, opened 2020-05-15
notes from first developer call

romanlutz:
Adding the writeup - please comment if you want anything to be added/changed!

If there are no comments I'll complete the PR by Friday 5/22

### DIFF
--- a/docs/community/developer_call_notes/2020_05_07.rst
+++ b/docs/community/developer_call_notes/2020_05_07.rst
@@ -5,19 +5,19 @@ Scribe: `Roman Lutz <https://github.com/romanlutz>`_
 
 Attendees:
 
-- Andrew Anderson
-- Matthijs Brouns
-- Miro Dudik
-- Richard Edgar
-- Parul Gupta
-- Ken Holstein
-- Adrin Jalali
-- Abdul Hannan Kanji
+- `Andrew Anderson <https://www.linkedin.com/in/andrewanderson05/>`_
+- `Matthijs Brouns <https://github.com/mbrouns>`_
+- `Miro Dudik <https://github.com/MiroDudik>`_
+- `Richard Edgar <https://github.com/riedgar-ms>`_
+- `Parul Gupta <https://github.com/parul100495>`_
+- `Ken Holstein <https://kenholstein.myportfolio.com/>`_
+- `Adrin Jalali <https://github.com/adrinjalali>`_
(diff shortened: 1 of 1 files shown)


docs/community/developer_call_notes/2020_05_07.rst
```diff
@@ -1,2 +1,101 @@
...
+- `Hanna Wallach <https://www.microsoft.com/en-us/research/people/wallach/>`_
+
+Organizational updates
+^^^^^^^^^^^^^^^^^^^^^^
+
+#. Microsoft & Fairlearn:
+
+    - goal is to be an open source, community driven project
+    - sociotechnical framing is central to the project, careful and precise
+      wording matter
```
MiroDudik, line 29: A couple small suggestions:
* maybe split this off the first two sub bullets separately under `#. Fairlearn project`
* before the last two sub-bullets add one more explanatory sub-bullet "currently many core members are from Microsoft, but the project is an independent open source project" (a bit repetitive, but doesn't hurt)
* it seems that the sub-bullets jumped an extra level (just me being nitpicky)

romanlutz: which ones jumped? 
Adding all the changes you suggested except for the last one which I don't understand yet.

romanlutz: should look like this now 
image

romanlutz:
> This is pretty cool, and I really like the attention you've put into this. Just a note that I rather have the notes even if not perfect, in case it becomes too much work as we go on. But I really like this one.

Absolutely! I don't mind even making adjustments later on if people want something reworded or fix typos or what not. 
In fact, I'm unhappy with how long it took me to get this PR out. That was due to the webpage/documentation generation taking a couple days longer than I thought, so at least it won't happen next time.

adrinjalali:
one thing I tend to do in meetings is to take the minutes as the meeting goes, on something shared (like hackmd.io) where people can see and edit together, and by the time the meeting is over, the minutes are ready to be published.

romanlutz:
> one thing I tend to do in meetings is to take the minutes as the meeting goes, on something shared (like hackmd.io) where people can see and edit together, and by the time the meeting is over, the minutes are ready to be published.

I love that. Let's see what makes sense! I don't want people to have to sign up for too many new tools, but this seems promising

Issue #431, opened 2020-05-21
Proposal: Datasets Module


koaning:
Just like scikit-learn offers datasets that make it easy for people to play around, it would be nice if this package also offers datasets where fairness is a theme. Over at scikit-fairness we've got a [few of these implemented already](https://scikit-fairness.netlify.app/api/datasets.html) and we could port them over here. 

There's one feature about these `load_`/`fetch_` methods that is worth pointing out; they all raise a `FairnessWarning`. Note that this is a feature, not a bug! It makes people aware that they should not play with this data naively. 

Another benefit of having these datasets around is that it potentially makes writing guides a lot easier as well.

romanlutz:
I'm always a little bit reluctant on that front because those datasets are probably owned by someone, or distributed with some sort of license. If you have experience with these sorts of things I'm happy to learn more :-)

To sort of work around that I've been dumping all such code into a different package called `tempeh` (https://github.com/microsoft/tempeh) which I'm not super eager to maintain because I simply don't have the time to do this properly, but it provides access to all kinds of datasets (among other thigns).

Curious to hear what you think!

koaning:
Sofar I've been able to find enough datasets that (at least I think) do not have ownership issues. Thusfar it seemed fine as long as you properly attribute but this is something that we can check on a dataset by dataset basis. You do probably want to be critical here because you don't want to become a package of datasets. 

In any case; one thing to be aware of is that scikit-learn is dropping support for `load_boston` because of the controversy. I agree that the boston dataset should be removed there, but there are lessons in that dataset in terms of fairness that should not be lost. It may be valid to host here, it will certainly help with education.

romanlutz:
I haven't read up on the details yet, but if there's something to be learned from it I think it would be fabulous to have that illustrated / explained. We've started converting some of the notebooks we currently have into these example notebooks on the webpage https://fairlearn.github.io/auto_examples/notebooks/index.html. The one that's there exists because @adrinjalali put in the effort to get this started. I'll try and do something similar for the other ones we want there.

Do you have a rough idea what this could look like/contain? Any links to read up more on the topic are welcome, of course. Starting to feel like I should open another issue for this specifically :-)

To address the broader question: why isn't there an ML package that provides *all* the datasets?
I know that lots of packages have a few datasets sort of attached to them, but I can't say I like this pattern. Not that that should prevent us from going down that path if it's what everyone else does...

koaning:
[...]

Issue #430, opened 2020-05-21
Proposal: InformationFilter


koaning:
I'm making this issue so that we can disucss how to move the `InformationFilter` from [scikit-fairness](https://scikit-fairness.netlify.app/api/preprocessing.html#skfair.preprocessing.InformationFilter) to this project. While moving it to this project I could also add a [guide](https://scikit-fairness.netlify.app/fairness_boston_housing.html#How-it-Works) to the docs on the boston housing dataset (the same we have on scikit-lego side). The details of the approach can also be seen on youtube [here](https://youtu.be/Z8MEFI7ZJlA?t=948). 

The goal of this issue is to first gather feedback on the implementation and, assuming people agree this tool is useful, to discuss the steps to make it compatible for this project. Once agreed on the path forward I'll open up a PR.

romanlutz:
Thanks for sharing the video!

Adding @MiroDudik and @hannawallach as well. I think it's interesting, but we need to be very careful in the description / documentation. De-biasing data is usually a tricky topic since there are so many different kinds, and they can affect each other etc., so the way we've been looking at it so far is described in the [documentation](https://fairlearn.github.io/user_guide/fairness_in_machine_learning.html) (quoted below). That's much more focused on "harms", that is looking at the outcome or result.
That said, applying this could reduce the unfairness in the outcome at the end of the pipeline. We've discussed adding preprocessing techniques a few times, but never quite got down to it. I think it would be great to add it. 


> AI systems can behave unfairly for a variety of reasons. Sometimes it is because of societal biases reflected in the training data and in the decisions made during the development and deployment of these systems. In other cases, AI systems behave unfairly not because of societal biases, but because of characteristics of the data (e.g., too few data points about some group of people) or characteristics of the systems themselves. It can be hard to distinguish between these reasons, especially since they are not mutually exclusive and often exacerbate one another. Therefore, we define whether an AI system is behaving unfairly in terms of its impact on people — i.e., in terms of harms — and not in terms of specific causes, such as societal biases, or in terms of intent, such as prejudice.
Usage of the word bias. Since we define fairness in terms of harms rather than specific causes (such as societal biases), we avoid the usage of the words bias or debiasing in describing the functionality of Fairlearn.

koaning:
[...]

PR #434, opened 2020-05-22
fetch_boston (#431)

koaning:
This is a work in progress PR and it is discussed in https://github.com/fairlearn/fairlearn/issues/431. 

The goal here is to discuss the pro/cons of the current implementation such that we may potentially set an example of how other datasets can be added. 

In particular; 

- Do we want such functions to depend on `sklearn.datasets.fetch_openml`?  This function is currently experimental but inventing our own standard feels like a bad choice. 
- We currently rely on [this](https://www.openml.org/d/531) uploaded dataset. We don't own this dataset and if the owner makes a change then this will cause the behavior of this function to change. I've set a parameter for the base version, which should protect us. Do we prefer to upload our own version of the dataset so that we can maintain it? 
- I've only implemented the function but we may appreciate a guide in the docs as well. That could be a seperate issue though. With that in mind ... what do we want to have in the docstring and what don't we want to have there? 
- How might we want to unit test this? I could check for shapes but then we would have unit tests that depend on the availability of an external resource.


fairlearn/datasets/_common.py
```diff
@@ -0,0 +1,4 @@
+class FairnessWarning(UserWarning):
```
adrinjalali, line 1: I think `fairlearn/utils/_common.py` would be a better place for this one?

fairlearn/datasets/_fetch_data.py
```diff
@@ -0,0 +1,91 @@
...
+    ==============   ==============
+
+    This dataset has known fairness issues. There's a "lower status of
+    population" (LSTAT) parameter that you need to look out for and a column
+    that is a derived from the proportion of people with a black skin color
+    that live in a neighborhood (B). This dataset should be used to explore
+    and benchmark fairness. Predicting a house-price while ignoring these
+    sensitive attributes is naive.
+
+    Parameters
```
adrinjalali, line 27: Just a note that this is `numpydoc` which I very much like, but it's not the same as the rest of the repo (also cc @romanlutz)

fairlearn/datasets/_fetch_data.py
```diff
@@ -0,0 +1,91 @@
...
+    ----------
+    version : integer or 'active', default='active'
+        Version of the dataset. Can only be provided if also ``name`` is given.
```
adrinjalali, line 30: AFAIK we either do `(name, version)` or `data_id`. This function calls `fetch_openml` with `data_id` which ignores the `version` (if I'm not mistaken).

fairlearn/datasets/_fetch_data.py
```diff
@@ -0,0 +1,91 @@
...
+        there may be more than one active version of a dataset, and those
+        versions may fundamentally be different from one another, setting an
+        exact version is highly recommended.
+
+    cache : boolean, default=True
+        Whether to cache downloaded datasets using joblib.
+
+    data_home : optional, default: None
+        Specify another download and cache folder for the datasets. By default
+        all scikit-learn data is stored in '~/scikit_learn_data' subfolders.
```
adrinjalali, line 41: maybe use a `~/.fairlearn` folder or something instead?

fairlearn/datasets/_fetch_data.py
```diff
@@ -0,0 +1,91 @@
...
+        a pandas DataFrame or Series depending on the number of target_columns.
+        The Bunch will contain a ``frame`` attribute with the target and the
+        data. If ``return_X_y`` is True, then ``(data, target)`` will be pandas
+        DataFrames or Series as describe above.
+
+    return_X_y : boolean, default=False.
+        If True, returns ``(data.data, data.target)`` instead of a Bunch
+        object.
+
+    warn : boolean, default=False.
```
adrinjalali, line 55: I think the default value should be `True`, and the user can disable it explicitly.

fairlearn/datasets/_fetch_data.py
```diff
@@ -0,0 +1,91 @@
...
+            If ``as_frame`` is True, ``data`` is a pandas object.
+        target : numpy array of shape (20640,)
+            Each value corresponds to the average
+            house value in units of 100,000.
+            If ``as_frame`` is True, ``target`` is a pandas object.
+        feature_names : list of length 8
+            Array of ordered feature names used in the dataset.
+        DESCR : string
+            Description of the California housing dataset.
+
```
adrinjalali, line 76: should also mention `(data, target)` if `return_X_y=True`.

fairlearn/datasets/_fetch_data.py
```diff
@@ -0,0 +1,91 @@
...
+    Notes
+    -----
+    This dataset consists of 506 samples and 14 features.
+    """
```
adrinjalali, line 80: ideally we should have a reference explaining the data, and another one explaining the fairness issues in it.

fairlearn/datasets/_common.py
koaning: Agreed.

fairlearn/datasets/_fetch_data.py
koaning: I copied scikit-learn here, if this project prefers a different standard I'll gladly hear it. Sphinx seemed to render it fine tho.

fairlearn/datasets/_fetch_data.py
koaning: Will check.

fairlearn/datasets/_fetch_data.py
koaning: My bad! I agree.

fairlearn/datasets/_fetch_data.py
koaning: Do we want to explain that here in the docstring or have a reference to the docs? To properly explain the issues with *that* dataset, you're gonna need an article.

fairlearn/datasets/_fetch_data.py
koaning: There's something to be said for that. Curious to hear if anybody has an objection to this.

fairlearn/datasets/_fetch_data.py
adrinjalali: I would just put the two references in the `Refenreces` section and refer to them from the docstring, with not much explanation.

fairlearn/datasets/_common.py
MiroDudik: How about `fairlearn.exceptions` (to mirror `sklearn.exceptions`)? Also--how specific do we want to scope this warning? If we only expect to use it for data sets, we could use a more specific name.

fairlearn/datasets/_fetch_data.py
MiroDudik: I agree with @adrinjalali . It makes more sense to store data in a folder with "fairlearn" in its name. I don't have a strong opinion if it should be named ".fairlearn" or "fairlearn-data" or something else of the sort.

fairlearn/datasets/_fetch_data.py
MiroDudik: Agreed with just providing links to the data explanation and fairness issues, but I'm not sure that we will be able to always find suitable existing external references for the fairness issues. Also, when it comes to fairness issues, we want to be able to incorporate any new issues as they become known, so I think we will probably want to point to a section of our "User Guide" where we will host these?

fairlearn/datasets/_fetch_data.py
MiroDudik: The docstring standard is currently being debated (#314) and we might end up switching to `numpydoc`. I'm not sure it makes sense to rewrite the docstring if it renders correctly on the webpage--unless #314 is resolved in favor of status quo by the time this PR gets in.

fairlearn/datasets/_fetch_data.py
```diff
@@ -0,0 +1,91 @@
...
+    if warn:
+        msg = "You are about to use with an unfair dataset. Mind the `B` and `LSTAT` columns."
```
MiroDudik, line 82: Let's tweak this a bit for consistency with the language we use elsewhere:
  "You are about to use a dataset with fairness issues. Mind the `B` and `LSTAT` columns."
Or possibly something more generic:
  "You are about to use a dataset with fairness issues. Please review the documentation."
The benefit is that if more issues are discovered, there are fewer places to update.

adrinjalali: I think people hardly even read the docstrings. I like that this message is very specific, and if we find other issues, then we can change this message as well.

fairlearn/datasets/_fetch_data.py
romanlutz: I’m happy with either one. Just make sure this is a directory that one usually has permission. I remember running into trouble with shap’s dataset utility on CircleCi because it downloaded the data into the directory where the package is stored. The CI didn’t have permissions for that so it failed. There’s a special case for it under examples/quickstart right now. Would love to never have to do this again.

fairlearn/datasets/_fetch_data.py
romanlutz: I like @adrinjalali 's proposal of just switching gradually. That would mean you can leave this in numpydoc format. FYI this is what it looks like in the generated documentation: https://250-133444044-gh.circle-artifacts.com/0/docs/_build/html/api_reference/fairlearn.datasets.html 
Is that what you would have expected? Would be nice if the feature/column names showed up.

fairlearn/datasets/_fetch_data.py
```diff
@@ -0,0 +1,91 @@
+import warnings
```
romanlutz, line 1: Perhaps we should have individual files per dataset? A single _fetch_data file for all datasets may get super big soon and then hard to read. What do you think?

fairlearn/datasets/_fetch_data.py
MiroDudik: Sounds good to me.

fairlearn/datasets/_fetch_data.py
```diff
@@ -0,0 +1,91 @@
...
+    Features                   real
+    Target           real 5. - 50.
+    ==============   ==============
+
+    This dataset has known fairness issues. There's a "lower status of
+    population" (LSTAT) parameter that you need to look out for and a column
+    that is a derived from the proportion of people with a black skin color
+    that live in a neighborhood (B). This dataset should be used to explore
+    and benchmark fairness. Predicting a house-price while ignoring these
+    sensitive attributes is naive.
```
MiroDudik, line 25: This paragraph needs a bit of tweaking. Mostly, the user needs to know where to go for details about the data description and the fairness issues.

Two clarification questions:
* By explore/benchmark fairness do you mean "explore/benchmark unfairness mitigation algorithms"?
* In the last sentence, by "naive" do you mean the fact that many other features can serve as proxies for sensitive features, so the strategy of simply removing sensitive features is often an ineffective mitigation strategy?

fairlearn/datasets/_fetch_data.py
```diff
@@ -0,0 +1,91 @@
...
+        If True, it raises an extra warning to make users aware of the unfairness
+        aspect of this dataset.
+
+
+    Returns
+    -------
+    dataset : :class:`~sklearn.utils.Bunch`
+        Dictionary-like object, with the following attributes.
+
+        data : ndarray, shape (506, 14)
```
romanlutz, line 65: below it says 8 feature values, but the shape is (506,14)... I feel like I'm missing something

fairlearn/datasets/_fetch_data.py
```diff
@@ -0,0 +1,91 @@
+import warnings
+from ._common import FairnessWarning
+
+from sklearn.datasets import fetch_openml
+
+
+def fetch_boston(*, version='1', cache=True, data_home=None,
```
romanlutz, line 7: I played around with this a little because I'm interested in adding more datasets once you've established the pattern. However, the version argument seems to only be relevant when passing "name" as opposed to the `data_id` that's used below.

fairlearn/datasets/_common.py
koaning: Working on an iteration now. I've renamed it to `UnfairDataWarning` and moved the warning to `fairlearn.exceptions`.

fairlearn/datasets/_fetch_data.py
koaning: @romanlutz that's a nice point, datasets have an opportunity to render the description in the documentation. 

Here's what it looks like using base sphinx. 

image

Would this suffice?

fairlearn/datasets/_fetch_data.py
koaning: I've removed the version. Am using just `data_id` now.

fairlearn/datasets/_fetch_data.py
koaning: Fixed.

fairlearn/datasets/_fetch_data.py
koaning: The current iteration links to a medium article discussing the problems with the dataset as well to the github issue where the removal from scikit-learn is discussed. I feel the latter link is relevant because it tells the story of how the community responded.

fairlearn/datasets/_fetch_data.py
koaning: Keeping it as is.

fairlearn/datasets/_fetch_data.py
koaning: I agree seperating is best. Made the improvement locally.

fairlearn/datasets/_fetch_data.py
koaning: I've though about it and came to the conclusion that the issues with the dataset deserve an article. For now I've changed the information here to include references that folks can read. 

Benchmarking fairness deserves a seperate page on the docs as well, explaining that is something that should not go here.

fairlearn/datasets/_fetch_data.py
koaning: Yep. Discovered that. Should be fixed in next commit.

fairlearn/datasets/_fetch_data.py
koaning: Well spotted. Fixing now.

fairlearn/datasets/_fetch_data.py
koaning: I don't know about windows systems actually. But for now I've implemented it such that we save it in;

```python
pathlib.Path().home() / ".fairlearn-data"
```

I think the home folder is allocated to the user, so this should suffice.

koaning:
@romanlutz I've made another iteration on this PR this weekend but the notebook CI step seems to break. I've not touched anything there that I am aware of so just to check ... is this something in this PR or something outside of this PR that is causing the CI to fail there?

riedgar-ms:
Sorry if I've missed a previous conversation, but isn't this the sort of thing which @romanlutz  was putting in `tempeh` rather than Fairlearn? Again, apologies if I've missed a discussion....

fairlearn/datasets/_fetch_boston.py
```diff
@@ -0,0 +1,121 @@
...
+    PTRATIO  pupil-teacher ratio by town
+    B        1000(Bk - 0.63)^2 where Bk is the proportion of blacks by town
+    LSTAT    % lower status of the population
+    MEDV     Median value of owner-occupied homes in $1000's
+    =======  ======================================================================
+
+
+    Parameters
+    ----------
+    version : integer or 'active', default='active'
```
MiroDudik, line 50: I think this parameter was dropped?

fairlearn/exceptions.py
```diff
@@ -0,0 +1,4 @@
+class UnfairDataWarning(UserWarning):
+    """Custom warning meant to be raised when loading unfair datasets."""
```
MiroDudik, line 2: A small language tweak: a dataset is not unfair, it has "fairness issues". Suggestions:
* `UnfairDataWarning` -> `DataFairnessWarning`
* docstring : "... when loading datasets with fairness issues."

MiroDudik:
@koaning : re:

> @romanlutz I've made another iteration on this PR this weekend but the notebook CI step seems to break. I've not touched anything there that I am aware of so just to check ... is this something in this PR or something outside of this PR that is causing the CI to fail there?

This is because we checked in some breaking API changes into master. Notebooks are all compatible with the master (Linux/Windows notebooks are passing), but the notebooks are now not compatible with the most recent version on pypi--hence the "against last release" check is failing, but that's okay--"merging can be performed automatically". Is there a way to make the FYI nature of that "against last release" check clearer?

@riedgar-ms : see #431 -- @romanlutz was very amenable to stop maintaining `tempeh`.

fairlearn/datasets/_fetch_boston.py
koaning: d0h

fairlearn/datasets/_fetch_data.py
romanlutz: This is great!

fairlearn/datasets/_fetch_boston.py
```diff
@@ -0,0 +1,121 @@
...
+    -----
+    This dataset consists of 506 samples and 13 features. It is notorious for the fairness
+    issues related to the `B` column. There's more information in the references.
+
+    References
+    ----------
+    https://medium.com/@docintangible/racist-data-destruction-113e3eff54a8
+    https://github.com/scikit-learn/scikit-learn/issues/16155
+    """
+    if warn:
```
kevinrobinson, line 110: fwiw, I'll put in my final vote against this approach, as previously raised in https://github.com/fairlearn/fairlearn/issues/431#issuecomment-632682693.  I've shared elsewhere (https://github.com/fairlearn/fairlearn/issues/430#issuecomment-632690069, https://github.com/fairlearn/fairlearn/issues/466#issuecomment-640852091, https://github.com/koaning/scikit-fairness/issues/31) a critique of using this dataset in isolation to illustrate the value of algorithmic approaches to assessment and mitigation on a prediction task.  My perception is that I am probably alone in this view :)

I'd love to see more discussion on the use cases for different algorithmic approaches, efforts to connect technical methods to their impact on real world harms, examples and benchmarks set in specific deployment contexts, and to generally see that the work in the project is approaching fairness as a sociotechnical issues.

I've also heard folks on the project talk about the importance of careful language when talking about fairness issues, and that an aspiration for this project is to advance the perspective that fairness is a sociotechnical challenge.  So I'd love to see discussion of the linked articles in that lens, or to see `fairlearn` taking advantage of the opportunity to engage in fairness questions in a deeper way than say `scikit-learn`, which doesn't have the same focus or purpose.

Thanks for reading! 😄 I'll make this my last comment on the topic and step back.  @koaning I hope you and others will consider these perspectives before moving forward in whatever way you all decide is best.

romanlutz: @kevinrobinson I've read your comments, but there's a lot of material and various points you're making, so I'm not entirely sure what specifically you're voting against. I suspect it's adding the dataset(?).

It's not entirely clear to me what's objectionable about including the dataset here. From my perspective the presentation (through user guide or examples) is crucial, and that's something @koaning said we should focus on next.
If there's a specific way in which you'd like to present this dataset please feel empowered to propose an approach or submit a PR directly. I think we all value your thorough analysis and I think that could be a great way for anyone interacting with the educational material to learn about this.

In general, I think it's as useful to show failures as it is to show successes (perhaps more!), as long as it's clear that it is a failure and we don't want folks to repeat it.

What do you think? I think this broader discussion will certainly be part of the call on Thursday as well since a lot of us would like to focus on fairness as a sociotechnical issue.

koaning: @kevinrobinson I also see a lot of material and various points so for me it's also unclear what you're voting against.

The goal here is not to only add one dataset, but many. This is but the beginning. Having a few of these datasets readily available will making it *much* easier to write guides on how to use techniques as well as give context. I'd even go as far as say without examples it'll be *really* hard to explain the merits of using fair techniques. Certainly, there's a sociotechnical context that needs to be adressed (and having datasets means that it's easier to write about those) but without examples of datasets it's going to be really hard to demonstrate to people how (and if) algorithmic fairness techniques work. 

Also note that this PR is merely to add a dataset, there's future PRs which would include blogposts/articles that can give more context. I just think it's much easier to write a blogpost about a dataset if the dataset is readily available.

koaning:
@MiroDudik @romanlutz the merge conflict is a mystery to me but other than that I'd say this one may be good to go.

romanlutz:
> @MiroDudik @romanlutz the merge conflict is a mystery to me but other than that I'd say this one may be good to go.

Bad rebase or something like that? There are 180 changed files, most of which you probably didn't touch, so that indicates to me that something with git went wrong. Can you try pulling in the latest `master` branch from the fairlearn repo?

koaning:
@romanlutz I couldn't find a nice way to deal with all the merge conflicts that made an appearance. I do not know what went wrong, but I decided on making a new PR with a clean branch from master. Here: https://github.com/fairlearn/fairlearn/pull/494. Once that is merged I propose that we delete this PR.

PR #438, opened 2020-05-28
remove python 3.5 support

romanlutz:
Dropping python 3.5 since a lot of our dependencies don't support it anymore. I'm leaving the special case for requirements.txt in place in case we ever want to resort to something like that again.

@adrinjalali you probably support this change :-)


adrinjalali:
Yep, and welcome to the world of f-strings, good type annotations, and more!

riedgar-ms:
> Yep, and welcome to the world of f-strings, good type annotations, and more!

Type annotations.... look desirable....

PR #440, opened 2020-05-28
Establish notebook editing process

romanlutz:
UPDATE: This PR will not migrate the UCI Credit Card Default notebook into .py format. The notebook is currently subject of our deep dive sessions (reach out if you want to join!) and will probably not exist in its current state in the future.

Instead, this PR focuses only on establishing the notebook editing process (or at least on providing suggestions).


docs/contributor_guide/contributing_example_notebooks.rst
```diff
@@ -13,4 +15,39 @@ include:
...
+notebook and then make the same change in the `.py` file. To make the
+development process smoother we recommend using 
+`Visual Studio Code <https://code.visualstudio.com/docs/python/jupyter-support>`_.
+VS Code recognizes the lines starting with :code:`$ %%` as new cells.
+Each cell can be executed individually by clicking on *Run Cell*, and VS Code
+opens a *Python Interactive* tab to show the output.
+
+.. image:: ../_static/images/vscode-jupyter.png
+
+Note that the text portion of these notebooks should be written in
```
riedgar-ms, line 47: This paragraph says "use ReST but don't use ReST." I'm guessing that the tool which produces the notebooks from the `.py` files doesn't know how to translate ReST into the variant of markdown used by the notebooks?

examples/notebooks/plot_binary_classification_COMPAS.py
```diff
@@ -18,6 +16,8 @@
 #
 # To start, let's download the dataset using `tempeh`
 
+print(__doc__)
```
riedgar-ms, line 19: Why is this being moved?

test/notebooks/test_notebooks.py
```diff
@@ -133,18 +120,3 @@ def test_mitigating_disparities_in_ranking_from_binary_data():
         "sel_expgrad_X_alt.loc[ 'disparity', :][0]",
         pytest.approx(0.35, abs=0.08))
     assay_one_notebook(nb_name, test_values)
-
-
-@pytest.mark.notebooks
-def test_binary_classification_with_the_uci_credit_card_default_dataset():
-    nb_name = "Binary Classification with the UCI Credit-card Default Dataset"
-    test_values = {}
-    test_values["GridSearch_17"] = ScrapSpec(
```
riedgar-ms, line 142: I'm not seeing this test in the new `.py` file. Am I missing something? It's not a great test, but it does ensure that the notebook has run and produced results as expected.

examples/notebooks/plot_binary_classification_COMPAS.py
romanlutz: Because there's no evident need to have it at the top AFAIK.

docs/contributor_guide/contributing_example_notebooks.rst
romanlutz: It does, but only for certain things. I've asked the sphinx-gallery folks about this and they're working on it, but I wouldn't expect this anytime soon. That's why I'd rather have the disclaimer there.

test/notebooks/test_notebooks.py
romanlutz: With notebooks moved out of the notebooks directory we don't have these tests anymore.

riedgar-ms: We're losing that functionality? I assume that the notebooks at least get run (although I can't remember where that happens now).... can we have some function like this back?

romanlutz: They get run because that's how the webpage is generated.

riedgar-ms: Thanks. I'm not hugely keen on the loss of the assertions - without them, there's a bit of a risk of silent failures. However, I don't see an easy way to add them back in - you can't just run the Python files because they pop up things from `matplotlib`.

docs/contributor_guide/contributing_example_notebooks.rst
riedgar-ms: It might be more helpful to give a list of the ReST markup which works.

romanlutz: That's fair, although I don't think it's worth spending much time on that since sphinx-gallery folks already fixed this problem and we're just waiting for their next release. Once that's out I'll try a few things (links, more complicated directives, etc.) and if any of them don't work yet I'll be happy to update. Does that make sense?

docs/contributor_guide/contributing_example_notebooks.rst
```diff
@@ -14,3 +16,38 @@ Good examples of existing notebooks that abide by these requirements are:
 
 * `Mitigating Disparities in Ranking from Binary Data <https://github.com/fairlearn/fairlearn/blob/master/notebooks/Mitigating%20Disparities%20in%20Ranking%20from%20Binary%20Data.ipynb>`_
 * `Binary Classification with the UCI Credit-card Default Dataset <https://github.com/fairlearn/fairlearn/blob/master/notebooks/Binary%20Classification%20with%20the%20UCI%20Credit-card%20Default%20Dataset.ipynb>`_
+
+Working with example notebooks
```
MiroDudik, line 20: I have some high-level structuring issues with this section. This is how I would structure the section:
* Maybe we can rename the section to something like: **Formatting of example notebooks**
* Start by saying that the notebooks are provided as `.py` files in the [percent format](https://jupytext.readthedocs.io/en/latest/formats.html#the-percent-format) (is there a better link to the format description?)
* Then describe what's allowed and what's not allowed in the comments
  * I am personally in favor of allowing a limited markdown in the comments (something that can be translated into ReST using simple regex), and disallowing ReST (because of .ipynb export issues)... but maybe that's exactly what the next sphinx version will accomplish?
* Then say how to easily create and edit such notebooks in Visual Studio
* Describe how to transition if you have a notebook in .ipynb (do we have a strategy here? would VS Code work? if we don't have a strategy, it's also okay to say we're not aware of a good strategy)
* Maybe also mention that if you want to continue developing an existing notebook using Jupyter, you can do so with the [Jupytext](https://jupytext.readthedocs.io/en/latest/index.html) extension (which allows you to open, run and edit `.py` files in the percent format)

romanlutz: - done
- done (although I left the order as is, because the first sentence tells people why there's any kind of special formatting in the first place)
- The next sphinx-gallery version will convert ReST to markdown. I'm not sure I understand what you mean by "allowing a limited markdown" because sphinx-gallery works off of ReST and can't parse markdown (I tried).
- The strategy with jupyter is to just open the .py file. Everything will work fine. I've updated the file accordingly.
- I've found jupytext to be more trouble than it's worth. I don't use it myself because VS Code knows how to work with .py files as notebooks.

docs/contributor_guide/contributing_example_notebooks.rst
romanlutz: I've upgraded to `sphinx-gallery==0.8.1` and activated `pypandoc` but that doesn't really seem to be helping. Perhaps it's restricted to certain types of rest-to-md conversions...

riedgar-ms (approved):

PR #441, opened 2020-05-29
[WIP] Add segment parity constraint

romanlutz:
Please provide feedback on this new moment. The name isn't perfect, but the best I could come up with so far.

The idea is to provide a constraint that can enforce parity between subgroups within a group, e.g.
- parity between groups defined by sensitive features within group A
- parity between groups defined by sensitive features within group B
- parity between groups defined by sensitive features within group C
- but not necessarily parity across (sub-)groups (within) A, B, and C
Notably, A, B, and C are feature values of a non-sensitive feature column. 

The feature itself is something people have asked for and it's easy enough to implement, so I'd like to get the ball rolling at least.

I'm holding off with tests until we know whether this is what we want the class to look like.


romanlutz:
> I think that this PR has the following prerequisites:
> 
> * harmonize Moment names with metric names following [metrics API proposal](https://github.com/fairlearn/fairlearn-proposals/blob/master/api/METRICS.md#harmonizing-metrics-across-modules)
> * augment the definitions of various metrics in `fairlearn.metrics` to allow such "subgroups".
> 
> By the way, I think that we should call these subgroups _segments_?
> 
> When it comes to updating metrics in `fairlearn.metrics` my suggestion is to update metrics `<metric>_{difference, ratio}` so that they take an additional optional argument `segments`. Similarly, for Moments, I would use existing moment names from the proposal and just add an optional argument `segments` which would need to be provided at the fit time (similarly to `sensitive_features` and `sample_weight`).
> 
> We also need to update `<metric>_group_summary` with an optional argument `segments`. With this argument, the output bundle could have fields `overall`, `by_segment`, `by_group`, `by_segment_group` (there are other choices we could make here, but you get the idea).

If it's called segment, then I don't like Segment Parity as a name. In fact I know now what's wrong with Subgroup Parity as a name! It just says parity without saying what kind of parity. If we call them segment we're essentially enforcing Demographic Parity on segments. Maybe that should be Segmented Demographic Parity? Do you think other constraints will be useful in that setting as well? I'm thinking of "Segmented <insert any other parity>"

romanlutz:
@riedgar-ms is taking another stab at this in #631 which is based on this PR, so I'm closing this one.

PR #442, opened 2020-05-29
Exponentiated gradient implementation for bounded group loss

### LINKED ISSUE #380: ENH - ExponentiatedGradient: use only classifiers with non-zero weights for predictions

dgiova:
Summary of contributions:
* added signed_weights method for default objective.
* implemented logic to compute the best response estimator for bounded group loss.
* developed method to produce randomized predictions, given the set of estimators produced by the exponentiated gradient optimization.

closes #380


fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -83,15 +85,16 @@ def _eval(self, Q, lambda_vec):
...
-            L = error + np.sum(lambda_projected * gamma) - self.eps * np.sum(lambda_projected)
+            L = error + np.sum(lambda_projected * (gamma - self.constraints.bound()))
         else:
-            L = error + np.sum(lambda_vec * gamma) - self.eps * np.sum(lambda_vec)
+            L = error + np.sum(lambda_vec * (gamma - self.constraints.bound()))
 
         max_gamma = gamma.max()
-        if max_gamma < self.eps:
+        max_bound = self.constraints.bound().max()
+        if max_gamma < max_bound:
```
MiroDudik, line 94: we need
```
max_constraint = (gamma - self.constnraint.bound).max()
```
followed by `if max_constraint < 0`

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -114,7 +117,7 @@ def solve_linprog(self, nu):
         if self.last_linprog_n_hs == n_hs:
             return self.last_linprog_result
         c = np.concatenate((self.errors, [self.B]))
-        A_ub = np.concatenate((self.gammas - self.eps, -np.ones((n_constraints, 1))), axis=1)
+        A_ub = np.concatenate((self.gammas - self.constraints.bound().max(), -np.ones((n_constraints, 1))), axis=1)
```
MiroDudik, line 120: just `self.gammas - self.constaints.bound()`

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -218,8 +225,19 @@ def _pmf_predict(self, X):
...
+
+        def get_predictions(clf):
+            """ignore classifiers with weight 0 since they can only contribute 0 to the result"""
+            if self.weights_[int(clf.name)] == 0:
+                return np.zeros(n_samples)
+            else:
+                return self._hs[int(clf.name)](X)
+
+        pred = pd.DataFrame(0, index=range(n_samples),
+                            columns=range(len(self._hs))).apply(get_predictions)
```
MiroDudik, line 238: Let's just insert a simple if statement in the previous version?

dgiova:
I agree to the developer certificate of origin https://developercertificate.org/ - signed by Davide Giovanardi, email address: dgiova27@gmail.com

adrinjalali:
I don't see any documentation for the changes though, and it also needs to add an entry to the change log.

romanlutz:
> I don't see any documentation for the changes though, and it also needs to add an entry to the change log.

Absolutely! I'm still following up on these things from the PR that this builds on with PR #443 and will get to this right after. Thanks for your patience!

romanlutz:
CHANGES.md has this already from the previous PR, documentation (user guide) exists now for the moment, although we still need to write the expgrad user guide (separate PR), eps removed from lagrangian constructor, but tests for BGL still missing.

fairlearn/reductions/_moments/utility_parity.py
```diff
@@ -13,6 +13,10 @@
 _MESSAGE_INVALID_BOUNDS = "Only one of difference_bound and ratio_bound can be used."
 _DEFAULT_DIFFERENCE_BOUND = 0.01
 
+_POSITIVE_RATE_PARITY_IMPOSSIBLE_ERROR_MESSAGE_TEMPLATE = "The dataset " \
```
riedgar-ms, line 16: This doesn't quite seem part of this PR

romanlutz: Needed to make a few adjustments to the tests to avoid too much duplication for BGL, and this was one of a few small things that surfaced.

romanlutz: I guess the question is whether this should raise or just not do anything. It is at the moment (pre this PR) a bug since we end up with an empty list due to 0 rows that correspond to label=1 (or 0 for the other one), and then we try to put this in a DataFrame and do math with it.

test/unit/reductions/exponentiated_gradient/test_exponentiatedgradient_smoke.py
```diff
@@ -10,268 +11,335 @@
...
          "error": 0.442883, "n_oracle_calls": 19, "n_oracle_calls_dummy_returned": 0,
          "n_predictors": 6}
     ]
 
-    def run_smoke_test(self, data, flipped=False):
-        if flipped:
-            y = self.flipped_y
-        else:
-            y = self.y
+    smoke_test_data_regression = [
```
MiroDudik, line 214: Two issues with the current tests:
* you also need to check `expgrad.weights_` not just `expgrad.predictors_`
* it would be good to try setting `upper_bound` to a value that's different from `eps` and something that would make the problem feasible--currently all your examples have infeasible constraints... so you could maybe try `upper_bound=0.05`, which would make some of the examples feasible and some infeasible; i'd also suggest setting `nu=1e-6` or `1e-3` and `max_iter=20` in these runs (since the automatic setting of `nu` needs to be double-checked and might change in the future)

test/unit/reductions/exponentiated_gradient/test_utilities.py
```diff
@@ -5,15 +5,22 @@
...
 sensitive_features = [str(x) for x in 'AAAAAAA' 'BBBBBBB' 'CCCCCC']
-labels =     [int(x) for x in '0110100' '0010111' '001111']  # noqa: E222
-X1 =         [int(x) for x in '0110101' '0111101' '001011']  # noqa: E222
-X2 =         [int(x) for x in '0000100' '0000011' '111111']  # noqa: E222
-X3 =         [int(x) for x in '1111111' '1111111' '111111']  # noqa: E222
+labels =             [int(x) for x in '0110100' '0010111' '001111']  # noqa: E222
+X1 =                 [int(x) for x in '0110101' '0111101' '001011']  # noqa: E222
+X2 =                 [int(x) for x in '0000100' '0000011' '111111']  # noqa: E222
+X3 =                 [int(x) for x in '1111111' '1111111' '111111']  # noqa: E222
+scores =          [int(x)/10 for x in '0001124' '0223346' '134447']  # noqa: E222
```
MiroDudik, line 12: this looks buggy, you probably mean `float(x)/10` or `int(x)/10.0`?

test/unit/reductions/exponentiated_gradient/test_utilities.py
```diff
@@ -5,15 +5,22 @@
...
 
 
-def _get_data(A_two_dim=False):
+def _get_data(A_two_dim=False, y_as_scores=False, flip_y=False):
```
MiroDudik, line 15: I don't love this calling convention. The issue is that if you sat `y_as_scores` to `True`, it's not just a matter of returning `y` as real values, you are actually returning a completely different `y`. But... I don't necessarily want to be overly pedantic here... so okay to leave as is.

fairlearn/reductions/_moments/utility_parity.py
MiroDudik: This is not the right way to do it. Can we get this out of the current PR and file an issue?

The fix needs to happen in `UtilityParity`, because this bug occurs whenever we have zero events. In that case, we should just have zero constraints (vacuous constraints).

test/unit/reductions/exponentiated_gradient/test_utilities.py
romanlutz: Yeah, just a test utility, not an API that we expect anyone to use.

test/unit/reductions/exponentiated_gradient/test_utilities.py
romanlutz: Nope. 
```
>>> int(7)/10
0.7
```

romanlutz: It's all identical.

fairlearn/reductions/_moments/utility_parity.py
romanlutz: OK, sure.

romanlutz: opened #543 and removed this from the PR

test/unit/reductions/exponentiated_gradient/test_exponentiatedgradient_smoke.py
romanlutz: - yes! Updated.
- Good point. I thought that's an issue since the gap doesn't go to 0. Will add a few more tests, although I'll leave the existing ones there as well. In practice that can happen, too, and we should have tests covering that. I'm adding the tests for `upper_bound=0.05` with `best_gap=0.0`. Regarding `nu` and `max_iter` restrictions: The number of iterations is never larger than 5 or 6 so that wouldn't do anything, and that applies even when I set `nu` to 1e-6. I've added it anyway for now, but just so you're aware.

PR #443, opened 2020-05-30
Documentation & naming updates after constraints generalization to use bounds

romanlutz:
This is a follow-up to #424 as well as the remaining renaming change from the metrics harmonization proposal https://github.com/fairlearn/fairlearn-proposals/blob/master/api/METRICS.md

Overall it provides:
- renaming of moments/constraints to abide by the new conventions from the proposal
- CHANGES.md updates that were missing from PR #424
- widespread documentation updates for the webpage for reductions & moments in particular


fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -17,25 +17,43 @@
...
-    This serves as the base class for both :class:`DemographicParity`
-    and :class:`EqualizedOdds`. The two are distinguished by
-    the events they define, which in turn affect the
-    `index` field created by :meth:`load_data()`.
+    This serves as the base class for :class:`DemographicParity` :class:`EqualizedOdds`, and
+    others. All subclasses can be used as difference-based constraints or ratio-based constraints.
+    
+    Difference-based constraints are written with a difference bound :math:`\\epsilon`.
+    Ratio-based constraints have a similar :math:`\\epsilon`, referred to as ratio bound slack
+    below. Additionally, they require a ratio bound. Constraints are set up for each group
```
riedgar-ms, line 25: "ratio bound slack below" ?

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -17,25 +17,43 @@
...
+    compared to the overall population (unless further events are specified, e.g., in Equalized
+    Odds). Constraint violation for difference-based constraints starts if the difference between
+    a group and the overall population with regard to a utility exceeds the difference bound. For
+    ratio-based constraints we consider the difference between the overall population's utility
```
riedgar-ms, line 29: Putting in the two equations here might be useful. If you'd rather not do that, can you link to the relevant section of the user guide?

romanlutz: I would, if it existed. That one will take some significant time.

riedgar-ms: Could you at least put in the two equations, rather than describing them in the text?

romanlutz: Miro said in the past that we shouldn’t add that level of detail in docstrings. I don’t think I can properly explain formulas in here either. I’ll write the doc pages later this week and can link them then.

adrinjalali: I personally find it useful when the PR adds a method and the user guide at the same time, and the docstring refers to the user guide for further explanation.

riedgar-ms: I agree that that would be the best way to do these things in future :-)

romanlutz: @adrinjalali agreed! I guess I hadn't thought the organization with student groups through. They don't necessarily have enough time to do all the tasks required for the open source repo due to how long semesters are and other commitments. For the future we should do this differently and have all changes at the same time as you suggest. Perhaps we can just add to their PRs until they're ready, as opposed to completing them and then adding follow-ups.

adrinjalali: > For the future we should do this differently and have all changes at the same time as you suggest. Perhaps we can just add to their PRs until they're ready, as opposed to completing them and then adding follow-ups.

Either way is valid IMO. I'd say the best way is the way you're comfortable with. I'm used to adding to people's PRs usually rather than follow up PRs, but if you're happy with openning an issue/submitting a followup PR, then that's all good.

fairlearn/reductions/_moments/conditional_selection_rate.py
romanlutz: What do you mean by this?

riedgar-ms: I was asking what was meant by "ratio bound slack below" because I didn't understand it.

romanlutz: I'm unsure whether this would be fixed from your perspective by saying
`below referred to as *ratio bound slack*`
OR
is it unclear what ratio bound slack is and you'd like me to elaborate on that?

riedgar-ms: I've now reparsed, and the 'below' is referring to the _text_ below. Now it makes sense, but you might want to rephrase slightly.

CHANGES.md
```diff
@@ -5,8 +5,24 @@
...
+    have new arguments on the constructor:
+    * `difference_bound` - for difference-based constraints such as demographic
+      parity difference
+    * `ratio_bound_slack` - for ratio-based constraints such as demographic
+      parity ratio
+    * Additionally, there's a `ratio_bound` argument which represents the
+      argument previously called `ratio`.
+  * Regression moments: `ConditionalLossMoment` and its subclasses have a new
+    argument `upper_bound` with the same purpose for newly enabled regression
+    scenarios on `ExponentiatedGradient`.
```
kevinrobinson, line 22: @romanlutz I think this in intentionally developer-facing, but I'd vote to make this more user-facing.  If someone doesn't already know how the guts of the library works, there's nothing in here that communicates "the way you set bounds on DemographicParity has changed."  Relatedly, has the behavior of `DemographicParity()` changed?

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -15,27 +15,45 @@
...
+    :param: ratio_bound: the constraints' ratio bound for constraints that are expressed as
+        ratios. The specified value needs to abide by :math:`0 < \\text{ratio_bound} \\leq 1`.
+        If `difference_bound` is used then `ratio_bound` needs to be None.
+        Default None.
+    :type ratio_bound: float
+    :param ratio_bound_slack: the constraints' ratio bound slack for constraints that are
+        expressed as ratios, usually referred to as :math:`\\epsilon`.
+        `ratio_bound_slack` is ignored if `ratio_bound` is not specified.
+        Default 0.0
+    :type ratio_bound_slack: float
```
kevinrobinson, line 56: It seems like the decision to use a superclass here is obscuring how to use user-facing classes.  By using a common API across all subclasses, it means this documentation is all generic and describes a combination of options.  Further, it means there is no place in the code comments that directly express what the parameters to `DemographicParity` are, to pick one example.  Focusing on just the comments, what do you think about removing the comments here in the superclass, and instead updating the comments on the user-facing classes to describe to users how they can use the classes that the library provides, and how they can configure them?

Alternately, focusing on the code, a more direct code change might break the inheritance chain so that `DemographicParity` can stand alone, define its interface, and then internally use `ConditionalSelectionRate` to do whatever it needs, including parameterizing it properly and providing the common interface that's needed to the algorithms.  That would avoid the explosion of parameters in the superclass (and thus each subclass), and let user-facing classes focus on describing to users how they can use them.  I realize that's probably impossible since there are things in-flight right now, but otherwise it seems like anytime the team wants to change the parameters to any of the subclasses, it will require changing the superclass and also changing the parameters that all the other subclasses take.

CHANGES.md
romanlutz: This came up on another PR already.

CHANGES.md is only shown in the repo and intentionally intended for developers or folks who are familiar with the code. We should have a section on the webpage (that we link from the README perhaps) to say how to upgrade from version X to Y (for all versions). 

I'm not going to address this in this PR, though, because it'll already get pretty huge with the user guide updates I'm working on right now.

fairlearn/reductions/_moments/conditional_selection_rate.py
romanlutz: I'm writing a user guide that explains how to use these classes (the user facing ones). Please reevaluate the PR when that's there. Right now it's work in progress (WIP).

From a developer perspective these are perfectly reasonable things to do (using inheritance), and for people familiar with the code it's very useful to know what changed in this format. Your other comment indicates that you're looking for a user facing change description, and as I outlined there we should have that. That's not the purpose of this PR, though.

I don't think breaking inheritance makes sense since it's explicitly designed for this purpose. If you read Miro's proposal #9 in the proposals repo (it's actually accepted and checked in now) you may get the higher level perspective. All constraints have in common that we can express them as differences or ratios, that's why all of this can live together in the parent class. That means the API can stay like this. Again, the user guide section I'm writing will make this a lot clearer, so I'll have to ask for patience.

kevinrobinson: @romanlutz oop!  I saw "widespread documentation updates for the webpage for reductions & moments in particular" in the pull request description at the time but missed the "WIP" in the title, sorry about that :)

@riedgar-ms It seems like the `mitigation` file in this pull request is what you were referring to in https://github.com/fairlearn/fairlearn/pull/480#discussion_r441066021, and the intention is for this PR to land first?

Either way I'm happy to help bring a helpful lack of experience and knowledge to reading and reviewing whenever that would actually be helpful :)

riedgar-ms: @kevinrobinson Correct. I also bring a certain lack of experience to this area (I've had other priorities since the beginning of the year), which had been my motivation for starting to write the other document.

test/unit/reductions/exponentiated_gradient/test_exponentiatedgradient_smoke.py
```diff
@@ -249,7 +254,7 @@ def test_smoke(self, testdata):
     def test_smoke_flipped(self, testdata):
         self.run_smoke_test(testdata, flipped=True)
 
-    @pytest.mark.parametrize("Constraints", [TruePositiveRateDifference, DemographicParity])
+    @pytest.mark.parametrize("Constraints", [TruePositiveRateParity, DemographicParity])
```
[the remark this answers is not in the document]
romanlutz: Added all constraints here.

setup.cfg
```diff
@@ -14,11 +14,12 @@ exclude =
 # D107 requires docstrings on __init__ but these are ignored in our current sphinx configuration
 # RST304 is about "unknown text roles" in sphinx. Unfortunately, things like :class: appear to be unknown
 # RST902 is from trouble parsing the __all__ entry. This only happens in a single file
+# D413 requires an empty line at the end of every docstring.
```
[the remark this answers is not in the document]
romanlutz: Why exclude it? Because it adds another 15 files that I didn't mean to touch.

```
.\fairlearn\datasets\_fetch_boston.py:14:1: D413 Missing blank line after last section
.\fairlearn\datasets\_fetch_boston.py:14:1: D413 Missing blank line after last section
.\fairlearn\metrics\_extra_metrics.py:22:1: D413 Missing blank line after last section
.\fairlearn\metrics\_extra_metrics.py:22:1: D413 Missing blank line after last section
.\fairlearn\metrics\_extra_metrics.py:80:1: D413 Missing blank line after last section
.\fairlearn\metrics\_extra_metrics.py:80:1: D413 Missing blank line after last section
.\fairlearn\metrics\_extra_metrics.py:111:1: D413 Missing blank line after last section
.\fairlearn\metrics\_extra_metrics.py:111:1: D413 Missing blank line after last section
.\fairlearn\metrics\_extra_metrics.py:142:1: D413 Missing blank line after last section
.\fairlearn\metrics\_extra_metrics.py:142:1: D413 Missing blank line after last section
.\fairlearn\metrics\_extra_metrics.py:173:1: D413 Missing blank line after last section
.\fairlearn\metrics\_extra_metrics.py:173:1: D413 Missing blank line after last section
.\fairlearn\postprocessing\_threshold_optimizer.py:94:1: D413 Missing blank line after last section
.\fairlearn\postprocessing\_threshold_optimizer.py:94:1: D413 Missing blank line after last section
.\fairlearn\postprocessing\_tradeoff_curve_utilities.py:38:1: D413 Missing blank line after last section
.\fairlearn\postprocessing\_tradeoff_curve_utilities.py:38:1: D413 Missing blank line after last section
.\fairlearn\postprocessing\_tradeoff_curve_utilities.py:71:1: D413 Missing blank line after last section
.\fairlearn\postprocessing\_tradeoff_curve_utilities.py:71:1: D413 Missing blank line after last section
.\fairlearn\reductions\_exponentiated_gradient\exponentiated_gradient.py:20:1: D413 Missing blank line after last section
.\fairlearn\reductions\_exponentiated_gradient\exponentiated_gradient.py:20:1: D413 Missing blank line after last section
.\fairlearn\reductions\_exponentiated_gradient\exponentiated_gradient.py:65:1: D413 Missing blank line after last section
.\fairlearn\reductions\_exponentiated_gradient\exponentiated_gradient.py:65:1: D413 Missing blank line after last section
.\fairlearn\reductions\_exponentiated_gradient\exponentiated_gradient.py:194:1: D413 Missing blank line after last section
.\fairlearn\reductions\_exponentiated_gradient\exponentiated_gradient.py:194:1: D413 Missing blank line after last section
.\fairlearn\reductions\_exponentiated_gradient\exponentiated_gradient.py:228:1: D413 Missing blank line after last section
.\fairlearn\reductions\_exponentiated_gradient\exponentiated_gradient.py:228:1: D413 Missing blank line after last section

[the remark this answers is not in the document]
Comment (romanlutz) [reply setup.cfg:17]: Must be some kind of new flake8 rule?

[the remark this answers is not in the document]
Comment (riedgar-ms) [reply setup.cfg:17]: Odd that it hasn't blown up our other builds.

File: docs/user_guide/mitigation.rst [hunk 7]
```diff
@@ -42,8 +42,9 @@ Fairlearn contains the following algorithms for mitigating unfairness:
 
 .. [*] coming soon!
 
-DP refers to demographic parity, EO to equalized odds, TPRD to true positive
-rate difference, ERR to error rate ratio, and BGL to bounded group loss. For
+DP refers to Demographic Parity, EO to Equalized Odds, TPRP to True Positive
```
riedgar-ms, line 45: This is hard to read. Perhaps a separate table of acronym and definition?

docs/user_guide/mitigation.rst
```diff
@@ -42,8 +42,9 @@ Fairlearn contains the following algorithms for mitigating unfairness:
...
+Rate Parity, FPRP to False Positive Rate Parity, ERP to Rrror Rate Parity, and
```
riedgar-ms, line 46: Typo: 'Error Rate Parity'

docs/user_guide/mitigation.rst
```diff
@@ -80,11 +81,408 @@ Reductions
...
+groups affected by a decision. For a broader discussion of fairness as a
+sociotechnical challenge and how to view Fairlearn in this context refer to
+:ref:`fairness_in_machine_learning`.
+
+The reductions approach is based on the idea in [#2]_ to reduce binary
+classification subject to fairness constraints to a sequence of cost-sensitive
+classification problems. As a result, the reductions methods in Fairlearn
+require only black-box access to a cost-sensitive classification algorithm,
+which does not need to have any knowledge of the desired definition of
+fairness or sensitive features. [#1]_ extends this to regression scenarios.
```
riedgar-ms, line 101: Do not start a sentence with an `[n]` reference. Author(s) & year would be OK, but we're not using those.

docs/user_guide/mitigation.rst
```diff
@@ -80,11 +81,408 @@ Reductions
...
+
+From an API perspective this looks as follows in all situations
+
+>>> reduction = Reduction(estimator, constraints, **kwargs)  # doctest: +SKIP
+>>> reduction.fit(X_train, y_train, sensitive_features=sensitive_features)  # doctest: +SKIP
+>>> reduction.predict(X_test)  # doctest: +SKIP
+
+Fairlearn doesn't impose restrictions on the referenced :code:`estimator`
+other than the existence of :code:`fit` and :code:`predict` methods.
```
riedgar-ms, line 110: Suggest putting `()` on all methods, to make it obvious what they are.

docs/user_guide/mitigation.rst
```diff
@@ -80,11 +81,408 @@ Reductions
...
+At the moment, the :code:`estimator`'s :code:`fit` method also needs to
+provide a :code:`sample_weight` argument which the reductions techniques use
+to reweight samples.
+In the future Fairlearn will provide functionality to handle this even
+without a :code:`sample_weight` argument.
+
+Before looking more into the reductions approaches themselves this section
+examines the different constraints that are compatible with them. All of them
+are expressed as moments inheriting from :code:`ClassificationMoment` or
+:code:`LossMoment`, which in turn inherit from the base class :code:`Moment`.
```
riedgar-ms, line 120: Not a fan of "...base class Moment. Moment's main purpose...."

docs/user_guide/mitigation.rst
```diff
@@ -80,11 +81,408 @@ Reductions
...
+constraints.
+
+Difference-based constraints consider the difference with respect
+to a metric between a group of samples and the overall population or
+cohorts thereof.
+Parity represents the state where
+
+.. math::
+
+    \text{metric}_a - \text{metric}_* = 0 \; \forall a
```
riedgar-ms, line 141: Suggest a larger hspace than `\;` (also below)

docs/user_guide/mitigation.rst
```diff
@@ -80,11 +81,408 @@ Reductions
...
+but rather violations that go beyond their :code:`difference_bound`. Such
+a difference-based constraint could be instantiated as
+
+    >>> ClassificationMoment(difference_bound=0.01)  # doctest: +SKIP
+
+Note that achieving parity with the :code:`difference_bound` does not mean
+that the difference between the group with the highest metric value and the
+group with the lowest metric value is below the :code:`difference_bound`, but
+rather the difference between each individual group and the metric value for
+the overall population (or cohort thereof) is below :code:`difference_bound`.
```
riedgar-ms, line 161: 'or a cohort thereof' ? What's going on with that?

docs/user_guide/mitigation.rst
```diff
@@ -80,11 +81,408 @@ Reductions
...
+In constrast, ratio-based constraints consider the ratio as opposed to the
+difference. Parity is represented as
+
+.. math::
+
+    r \leq \dfrac{\text{metric}_a}{\text{metric}_*} \leq \dfrac{1}{r} \; \forall a
+
+with a :code:`ratio_bound` :math:`r`. Similar to the :code:`difference_bound`
+for difference-based constraints, ratio-based constraints have a
+:code:`ratio_bound_slack`. The two inequalities can also be considered
```
riedgar-ms, line 172: I think it would be helpful to show the slack for the ratios. I can think of a couple of possible ways it might appear, and I'm not sure which is correct.

docs/user_guide/mitigation.rst
```diff
@@ -80,11 +81,408 @@ Reductions
...
+
+    It is not possible to specify both :code:`difference_bound` *and*
+    :code:`ratio_bound` for the same constraint object.
+
+.. _demographic_parity:
+
+Demographic Parity
+^^^^^^^^^^^^^^^^^^
+
+A classifier :math:`h(X)` satisfies *Demographic Parity* if
```
riedgar-ms, line 208: A _binary_ classifier

docs/user_guide/mitigation.rst
```diff
@@ -80,11 +81,408 @@ Reductions
...
+In other words, the selection rate or percentage of samples with label 1
+should be equal between all groups. Implicitly this means the percentage
+with label 0 is equal as well.
+
+In the example below class :code:`"a"` has a selection rate of 60%,
+:code:`"b"` has a selection rate of 20%. The overall selection rate is 40%,
+so :code:`"a"` is `0.2` above the overall selection rate, and :code:`"b"` is
+`0.2` below. The seemingly redundantly stored constraint violation with both
+:code:`"a"` and :code:`"b"` showing up twice stems from the constraint
+definition outlined in :ref:`constraints_binary_classification`. Each equality
```
riedgar-ms, line 223: I think the 'Each equality constraint' sentence belongs above where you first introduce them.

docs/user_guide/mitigation.rst
```diff
@@ -80,11 +81,408 @@ Reductions
...
+constraint is written as two inequalities of which one contains
+:math:`\text{metric}_a` positively (sign :math:`+`) and one negatively
+(sign :math:`-`).
+Note that the specified :code:`difference_bound` is not used
```
riedgar-ms, line 227: I don't think this sentence makes a lot of sense.

docs/user_guide/mitigation.rst
```diff
@@ -80,11 +81,408 @@ Reductions
...
+since the constraint object is not used in a mitigation in this example.
+Similarly, :code:`y_true` is technically irrelevant to the calculations
```
riedgar-ms, line 229: This sentence doesn't quite feel like it belongs here. Not sure of what to do with it, though. Perhaps stick it up where DP is first introducted?

docs/user_guide/mitigation.rst
```diff
@@ -80,11 +81,408 @@ Reductions
...
+because the underlying metric of Demographic Parity, selection rate, does not
+consider performance relative to the true labels, but rather proportions in
+the predicted labels.
+
+.. note::
+
+    In mitigation examples we never need to call :code:`load_data` or
```
riedgar-ms, line 236: You haven't mentioned `load_data()` to this point. This is also skipping into implementation when I thought this section was about the theory behind the moments.

docs/user_guide/mitigation.rst
riedgar-ms: OK... now I get down to the code sample, I see why, but.... I think the fact that DP doesn't need y_true merits mention where it's first defined.

docs/user_guide/mitigation.rst
```diff
@@ -80,11 +81,408 @@ Reductions
...
+    corresponding inequalities in :ref:`constraints_binary_classification`.
+
+.. doctest:: mitigation
+
+    >>> from fairlearn.reductions import DemographicParity
+    >>> from fairlearn.metrics import selection_rate_group_summary
+    >>> import numpy as np
+    >>> dp = DemographicParity(difference_bound=0.01)
+    >>> X                  = np.array([[0], [1], [2], [3], [4], [5], [6], [7], [8], [9]])
+    >>> y_true             = np.array([ 1 ,  1 ,  1 ,  1 ,  0,   0 ,  0 ,  0 ,  0 ,  0 ])
```
riedgar-ms, line 248: Since y_true doesn't matter, I suggest you just put in y_pred, then specify (with a comment about not being needed) y_true = y_pred.

docs/user_guide/mitigation.rst
```diff
@@ -80,11 +81,408 @@ Reductions
...
+
+.. math::
+    
+    \P[h(X) = y' \given A = a] = \P[h(X) = y'] \; \forall a, y'
+ 
+In other words, the selection rate or percentage of samples with label 1
+should be equal between all groups. Implicitly this means the percentage
+with label 0 is equal as well.
+
+In the example below class :code:`"a"` has a selection rate of 60%,
```
riedgar-ms, line 218: I think that some of this belongs after the example. And.... this whole section feels like an uncomfortable mix of theory and code explanations.

docs/user_guide/mitigation.rst
```diff
@@ -80,11 +81,408 @@ Reductions
...
+    >>> dp.load_data(X, y_true, sensitive_features=sensitive_features)
+    >>> dp.gamma(lambda X: y_pred)
+    sign  event  group_id
+    +     all    a           0.2
+                 b          -0.2
+    -     all    a          -0.2
+                 b           0.2
+    dtype: float64
+ 
+Rewriting this into a ratio-based constraint with :code:`ratio_bound`
```
riedgar-ms, line 262: "Alternatively, we can write this as a ratio-based constraint"

I don't believe that it will give the same ultimate answers will it? In which case, it's not really 'rewriting' I think.

docs/user_guide/mitigation.rst
```diff
@@ -80,11 +81,408 @@ Reductions
...
+
+and *False Positive Rate Parity* if
+
+.. math::
+
+    \P[h(X) = 1 \given A = a, Y = 0] = \P[h(X) = 1 \given Y = 0] \; \forall a
+
+Note that this formulation of True Positive Rate Parity ignores what happens
+to samples with :math:`Y=0`, and the False Positive Rate Parity ignores
+:math:`Y=1`.
```
riedgar-ms, line 317: 'False Positive Rate Parity ignores those with'

docs/user_guide/mitigation.rst
```diff
@@ -80,11 +81,408 @@ Reductions
...
+Refer to :ref:`equalized_odds` for a constraint that extends this by
```
riedgar-ms, line 318: Why not introduce equalised odds first, and then point out that TPRP and FPRP are the two halves of it. I feel that referring forward like this is a bit odd.

docs/user_guide/mitigation.rst
```diff
@@ -80,11 +81,408 @@ Reductions
...
+    >>> from fairlearn.reductions import TruePositiveRateParity
+    >>> from fairlearn.metrics import true_positive_rate_group_summary
+    >>> import numpy as np
+    >>> tprp = TruePositiveRateParity(difference_bound=0.01)
+    >>> X                  = np.array([[0], [1], [2], [3], [4], [5], [6], [7], [8], [9]])
+    >>> y_true             = np.array([ 1 ,  1 ,  1 ,  1 ,  1,   1 ,  1 ,  0 ,  0 ,  0 ])
+    >>> y_pred             = np.array([ 1 ,  1 ,  1 ,  1 ,  0,   0 ,  0 ,  1 ,  0 ,  0 ])
+    >>> sensitive_features = np.array(["a", "b", "a", "a", "b", "a", "b", "b", "a", "b"])
+    >>> true_positive_rate_group_summary(y_true, y_pred, sensitive_features=sensitive_features)
+    {'overall': 0.5714285714285714, 'by_group': {'a': 0.75, 'b': 0.3333333333333333}}
```
riedgar-ms, line 335: Can you get these numbers to be rounder?

docs/user_guide/mitigation.rst
```diff
@@ -80,11 +81,408 @@ Reductions
...
+    \P[h(X) = 1 \given A = a, Y = 0] = \P[h(X) = 1 \given Y = 0] \; \forall a
+
+Note that this formulation of True Positive Rate Parity ignores what happens
+to samples with :math:`Y=0`, and the False Positive Rate Parity ignores
+:math:`Y=1`.
+Refer to :ref:`equalized_odds` for a constraint that extends this by
+considering all samples, i.e., the samples with :math:`Y=0` as well as the
+ones with :math:`Y=1`.
+
+In practice this can be used in a difference-based version as follows:
```
riedgar-ms, line 322: I would start off here by saying that you're going to construct a dataset with given TPR values (overall, a and b). Otherwise people have to pick too much out of the code.

docs/user_guide/mitigation.rst
```diff
@@ -80,11 +81,408 @@ Reductions
...
+    dtype: float64
+
+.. note::
+
+    In mitigation examples we never need to call :code:`load_data` or
+    :code:`gamma` since the reductions techniques all automatically take care
+    of it. The example merely helps in illustrating the deviation from the
+    corresponding inequalities in :ref:`constraints_binary_classification`.
+
+Similarly to other constraints a ratio-based version is also available:
```
riedgar-ms, line 352: Why isn't there an example like the others?

docs/user_guide/mitigation.rst
```diff
@@ -80,11 +81,408 @@ Reductions
...
+^^^^^^^^^^^^^^
+
+A classifier :math:`h(X)` satisfies *Equalized Odds* if it satisfies both
+*True Positive Rate Parity* and *False Positive Rate Parity*, i.e.,
+
+.. math::
+
+    \P[h(X) = y' \given A = a, Y = y] = \P[h(X) = y' \given Y = y] \; \forall a, y, y'
+
+From a usage standpoint this works just like :ref:`true_positive_rate_parity`.
```
riedgar-ms, line 378: Just a usage standpoint? Isn't EO _exactly_ TPRP \union FPRP?

docs/user_guide/mitigation.rst
```diff
@@ -80,11 +81,408 @@ Reductions
...
+    >>> import numpy as np
+    >>> dp = DemographicParity(difference_bound=0.01)
+    >>> X                  = np.array([[0], [1], [2], [3], [4], [5], [6], [7], [8], [9]])
+    >>> y_true             = np.array([ 1 ,  1 ,  1 ,  1 ,  0,   0 ,  0 ,  0 ,  0 ,  0 ])
+    >>> y_pred             = np.array([ 1 ,  1 ,  1 ,  1 ,  0,   0 ,  0 ,  0 ,  0 ,  0 ])
+    >>> sensitive_features = np.array(["a", "b", "a", "a", "b", "a", "b", "b", "a", "b"])
+    >>> selection_rate_group_summary(y_true, y_pred, sensitive_features=sensitive_features)
+    {'overall': 0.4, 'by_group': {'a': 0.6, 'b': 0.2}}
+    >>> dp.load_data(X, y_true, sensitive_features=sensitive_features)
+    >>> dp.gamma(lambda X: y_pred)
```
riedgar-ms, line 254: I think this merits a comment. I just noticed that it was a bit bizarre. Both what the actual argumment of gamma really is, and what you're doing here.

docs/user_guide/mitigation.rst
```diff
@@ -80,11 +81,408 @@ Reductions
...
+^^^^^^^^^^^^^^^^^
+
+The basic idea behind *Error Rate Parity* is that the error rate should be
+similar between different groups of people.
+
+A classifier :math:`h(X)` satisfies *Error Rate Parity* if it satisfies
+
+.. math::
+
+   \E[\lvert h(X) - Y \rvert \given A = a] - \E[\lvert h(X) - Y \rvert] \leq 0 \; \forall a
```
riedgar-ms, line 412: The \rvert and \given symbols are indistinguishable in the rendered equation. And the brackets aren't too different from them either. As a result, this equation is hard to parse.

riedgar-ms: Perhaps split into two - first define an err(X) \def | h(X) - Y | and then use err(X) in the parity definition.

docs/user_guide/mitigation.rst
```diff
@@ -80,11 +81,408 @@ Reductions
...
+    dtype: float64
+
+.. note::
+
+    In mitigation examples we never need to call :code:`load_data` or
+    :code:`gamma` since the reductions techniques all automatically take care
+    of it. The example merely helps in illustrating the deviation from the
+    corresponding inequalities in :ref:`constraints_binary_classification`.
+
+Similarly, *Error Rate Parity* can be expressed through a ratio constraint as
```
riedgar-ms, line 440: I don't think that 'Similarly' is quite the right word here. I feel it implies that the ratio and difference constraints will lead to the same place. 'Alternatively' might be a better word - and say that you're following a similar procedure to those above.

docs/user_guide/mitigation.rst
```diff
@@ -80,11 +81,408 @@ Reductions
...
+A classifier :math:`h(X)` satisfies *Equalized Odds* if it satisfies both
+*True Positive Rate Parity* and *False Positive Rate Parity*, i.e.,
+
+.. math::
+
+    \P[h(X) = y' \given A = a, Y = y] = \P[h(X) = y' \given Y = y] \; \forall a, y, y'
+
+From a usage standpoint this works just like :ref:`true_positive_rate_parity`.
+The constraint violations with :code:`label=1` match exactly the ones from
+True Positive Rate Parity above, while the ones with :code:`label=0` represent
```
riedgar-ms, line 380: Need to point out that the reader needs to look at the _difference_ code boxes, not the _ratio_ ones.

docs/user_guide/mitigation.rst
romanlutz: I left it off because 
1. we say `methods` at the end of the sentence
2. you can't actually call them as `fit()`, it needs args.

docs/user_guide/mitigation.rst
romanlutz: Well, sometimes you don't evaluate against the overall population, but a cohort thereof. Example: Equalized Odds evaluates positives (one cohort) and negatives (another cohort) separately.

docs/user_guide/mitigation.rst
romanlutz: We have all the moments in a separate section, but the definition of fairness constraints should probably be in a a different section altogether (the fairness in ML section). I don't quite know what you find hard to read there since everything is listed right below the table.

docs/user_guide/mitigation.rst
romanlutz: done

docs/user_guide/mitigation.rst
romanlutz: replaced second one with "It"

docs/user_guide/mitigation.rst
romanlutz: replaced with \quad

docs/user_guide/mitigation.rst
romanlutz: This is the binary classification section, but sure why not...

docs/user_guide/mitigation.rst
romanlutz: Hmm. So the problem with that is that I actually lied above. In general it's not metric_a, but metric_{a,cohort}, e.g. in Equalized Odds. Probably needs a bit more updating.

docs/user_guide/mitigation.rst
romanlutz: Why not? The difference bound is indeed not used because we're not doing any mitigation here. The fact that *a* difference bound was provided is used to calculate gamma, but not the difference bound itself.

docs/user_guide/mitigation.rst
romanlutz: But this is the section where it's first defined?!?

docs/user_guide/mitigation.rst
romanlutz: Correct. It's because I don't want to actually talk about load_data or gamma, because they are irrelevant to users. This is a user guide and therefore won't have much theory behind moments other than a definition, which is provided above the note. I feel like you're expecting something more?

docs/user_guide/mitigation.rst
romanlutz: I thought about that and decided not to do it, because the point here is to show what it does. Doing that would risk confusing readers for no reason other than avoiding the repetition of a line.

romanlutz: Our API says `y_true` after all

docs/user_guide/mitigation.rst
romanlutz: I think we may have different expectations of the user guide. If you look at sklearn's user guide it's very much a mix, because for theory you wouldn't go to the user guide for a software package.

docs/user_guide/mitigation.rst
romanlutz: Of course the answers are different, that's the entire point of showing the example twice :-)

> "To rewrite" = to make a revision of (something, such as a story)

We are rewriting it by changing the argument.

docs/user_guide/mitigation.rst
romanlutz: Because then I have to refer the other way forward. I did
```
A
B
C (depends on A, B)
```
and you're suggesting

```
C (depends on A, B which at this point we don't even know)
A
B
```
I really don't see how that's better.

docs/user_guide/mitigation.rst
romanlutz: Why?

docs/user_guide/mitigation.rst
romanlutz: You mean write the content of the printed dictionary before the code?
```
{'overall': 0.5714285714285714, 'by_group': {'a': 0.75, 'b': 0.3333333333333333}}
```

docs/user_guide/mitigation.rst
romanlutz: There is, right below this line. Not sure what you're referring to.

docs/user_guide/mitigation.rst
riedgar-ms: Then you could write it as `fit(...)`

docs/user_guide/mitigation.rst
romanlutz: I don't quite see what you mean to say here. The usage ("how to write the code") is indeed the same (same API, args, etc.)

Regarding the second question: it depends. The state of complete EO is exactly TPRP AND FPRP, but in real world scenario you never have full parity, so there's always some disparity, it just depends on how much. From a moments perspective EO is exactly the combination of those two constraints.

docs/user_guide/mitigation.rst
riedgar-ms: I think that the list of "<Acronym> refers to <Definition>" might work better as a table.

docs/user_guide/mitigation.rst
romanlutz: I guess this goes back to "I don't want users to use this, this is just an internal way of representing things". That's why that note says as much.

[Ideally I'd like gamma not to be named gamma, and also not to take predictors this way, but unfortunately that requires more rewriting than I'm willing to do in this already huge PR.]

docs/user_guide/mitigation.rst
riedgar-ms: OK, understood.

docs/user_guide/mitigation.rst
riedgar-ms: If you're not going to show it, I don't see why you'd mention it at this point. In the difference version, doesn't it change the zeros on the right hand side into epsilons?

docs/user_guide/mitigation.rst
riedgar-ms: Doesn't hurt to repeat.

docs/user_guide/mitigation.rst
riedgar-ms: The point is, if it's unused, why talk about it?

docs/user_guide/mitigation.rst
riedgar-ms: Section yes, paragraph no.

docs/user_guide/mitigation.rst
riedgar-ms: It doesn't make sense to say that you're not calling something when you haven't mentioned that it exists yet.

docs/user_guide/mitigation.rst
riedgar-ms: I had meant as a declaration of y_true, not in the actual method call itself. However, I do see your point.

docs/user_guide/mitigation.rst
riedgar-ms: But aren't we going for more educational material than scikit-learn?

docs/user_guide/mitigation.rst
riedgar-ms: To me, 'rewriting' an equation implies that it's the same one, rearranged a bit. I don't believe that's the case for difference vs ratio bounds

docs/user_guide/mitigation.rst
riedgar-ms: No, you'd introduce A, and then note that B and C are subsets of it. In this case, it might just be because I learned about EO first, and so to me it's the foundational one. I suppose it doesn't really matter.

docs/user_guide/mitigation.rst
riedgar-ms: Because it's a nice thing to do for readers?

docs/user_guide/mitigation.rst
riedgar-ms: Effectively - but written as equations and fractions, not just spit out a Python dictionary.

docs/user_guide/mitigation.rst
riedgar-ms: Hmmm... I'm wondering if I looked at the wrong doc build.

docs/user_guide/mitigation.rst
riedgar-ms: If you're going to use them in the user guide, they should be explained.

docs/user_guide/mitigation.rst
riedgar-ms: From what standpoints does it not work just like TPRP? If there aren't any, then "From a usage standpoint" is pointless verbiage. If there are other standpoints, they should at least be mentioned.

CHANGES.md
```diff
@@ -5,10 +5,42 @@
 * Adjust classes to abide by naming conventions for attributes.
 * Change `ExponentiatedGradient` signature by renaming argument `T` to
   `max_iter`, `eta_mul` to `eta0`, and by adding `run_linprog_step`.
+* API refactoring to separate out different uses of `eps` within
+  `ExponentiatedGradient`. It is now solely responsible for setting a bound on
+  the lambda vectors. The other usage of `eps` as a bound for constraints is
```
MiroDudik, line 10: this description of the remaining use of `eps` is not quite correct. i'll do a direct edit after i finish my pass.

fairlearn/reductions/_moments/conditional_selection_rate.py
```diff
@@ -14,33 +14,54 @@
...
-    `index` field created by :meth:`load_data()`.
-
-    The `index` field is a :class:`pandas:pandas.MultiIndex` corresponding to the rows of
-    the DataFrames either required as arguments or returned by several
-    of the methods of the `ConditionalSelectionRate` class. It is the cartesian
-    product of:
-
-    - The unique events defined for the particular object
-    - The unique values for the sensitive feature
+class UtilityParity(ClassificationMoment):
```
MiroDudik, line 17: the file name should change to `utility_parity` as well.

fairlearn/reductions/_moments/conditional_selection_rate.py
MiroDudik: I have made some tweaks in `mitigation.rst` (within this PR), so there is now the description of the constraints in `UtilityParity`, so you can just point to there.

docs/user_guide/mitigation.rst
MiroDudik: this is fixed now.

docs/user_guide/mitigation.rst
MiroDudik: I slightly rearranged, but there is space for improvements (not sure whether they need to happen in this PR).

docs/user_guide/mitigation.rst
MiroDudik: I fixed this.

docs/user_guide/mitigation.rst
MiroDudik: I used a more natural definition that makes it clear that this is an error rate.

fairlearn/reductions/_moments/conditional_selection_rate.py
romanlutz: Done! Good catch

CHANGES.md
romanlutz: Looks like you got to this based on commit messages.

docs/user_guide/mitigation.rst
romanlutz: Sure, I agree 100%. However, so should 100 other (higher priority) things. I don't want to block this PR until every single aspect is perfect, because right now the page is completely empty. There are at least 2 other PRs currently waiting for this one.

docs/user_guide/mitigation.rst
romanlutz: I'm showing it because I'm describing the moment objects. If I don't provide a constructor arg then I can't instantiate it (technically no arg = difference bound, so it would work but then I'm just hiding and confusing). I'm trying to help users understand what these are about, and how they relate to metrics (which are explained in the previous section). I can show mitigations in the mitigation technique section. Why should I show it in both? Doesn't quite make sense to me. 

We need to generally explain what the slack means at some point, but I don't want to block this PR on that. Documentation is more of an iterative process so we can come back to this.

docs/user_guide/mitigation.rst
romanlutz: Agreed, there will be plenty more we can do over time, but *something* is better than nothing to start with.

docs/user_guide/mitigation.rst
romanlutz: I think Miro's revisions removed this.

docs/user_guide/mitigation.rst
romanlutz: Same, and I found it terribly confusing because of that. I would have much preferred hearing about TPRP and FPRP first.

fairlearn/reductions/_moments/conditional_selection_rate.py
romanlutz: linked to user guide

docs/user_guide/mitigation.rst
romanlutz: added ellipsis

docs/user_guide/mitigation.rst
romanlutz: done

docs/user_guide/mitigation.rst
romanlutz: This text doesn't quite exist anymore after Miro's revisions

romanlutz:
The main points I'm not addressing in this PR that we may want to consider revisiting: 
- explain slack through mitigation doc
- explain load_data & gamma
- consider constructing a dataset with given TPR, FPR, etc. values ahead of time and then using it in there

@riedgar-ms FYI

PR #446, opened 2020-06-01
Add F1 score and log loss

romanlutz:
Also attempted to add F1-score to the widget as a performance metric, but I don't think that was sufficient. @rihorn2 is there any particular place where I need to do something other than the ones that I adjusted already? Perhaps I'm doing something wrong.

@MiroDudik if you don't want it part of the widget at this point we can also hold off on that and just do the metrics changes.


fairlearn/metrics/_balanced_root_mean_squared_error.py
```diff
@@ -41,6 +41,6 @@ def _balanced_root_mean_squared_error(y_true, y_pred, sample_weight=None):
         y_ta_s = y_ta[indices]
         y_pa_s = y_pa[indices]
         s_w_s = s_w[indices]
-        errs[i] = math.sqrt(skm.mean_squared_error(y_ta_s, y_pa_s, s_w_s))
+        errs[i] = math.sqrt(skm.mean_squared_error(y_ta_s, y_pa_s, sample_weight=s_w_s))
```
riedgar-ms, line 44: We had a lurking bug?

visualization/FairnessDashboard/src/Localization/en.json
```diff
@@ -137,8 +140,8 @@
         "r2Description": "The fraction of variance in the labels explained by the model.",
         "aucDescription": "The quality of the predictions, viewed as scores, in separating positive examples from negative examples.",
         "balancedRMSEDescription": "Positive and negative examples are reweighted to have equal total weight. Suitable if the underlying data is highly imbalanced.",
-        "balancedAccuracy": "Balanced accuracy",
         "balancedAccuracyDescription": "Positive and negative examples are reweighted to have equal total weight. Suitable if the underlying data is highly imbalanced."
+        "f1ScoreDescription": "F1-Score is the harmonic mean of precision and recall."
```
riedgar-ms, line 144: Any idea why the colon is showing up in red in the GitHub diff?

fairlearn/metrics/_balanced_root_mean_squared_error.py
romanlutz: Not a bug, but sklearn is deprecating certain positional args and making them keyword only. I saw the warnings and fixed it right away. I'm sure you don't mind :-)

visualization/FairnessDashboard/src/Localization/en.json
romanlutz: I'm so glad you spotted this! The comma from the previous line was missing. Not sure why I don't see errors when building the widget, though....

fairlearn/metrics/_balanced_root_mean_squared_error.py
riedgar-ms: Not at all :-)

romanlutz:
> You won't see the TS updates until you rebuild and publish the `npm` package.

Are there steps for this?

fairlearn/widget/_fairlearn_dashboard.py
```diff
@@ -129,6 +131,14 @@ def __init__(
                 "model_type": ["regression"],
                 "function": r2_score_group_summary
             },
+            "f1_score": {
+                "model_type": ["classification"],
+                "function": f1_score_group_summary
+            },
+            "log_loss": {
+                "model_type": ["regression"],
```
MiroDudik, line 139: it should be `["probability"]`

visualization/FairnessDashboard/src/AccuracyMetrics.ts
```diff
@@ -181,5 +181,20 @@ export const AccuracyOptions: {[key: string]: IAccuracyOption} = {
...
+        title: localization.Metrics.f1Score,
+        description: localization.Metrics.f1ScoreDescription,
+        isMinimization: false,
+        isPercentage: false,
+        userVisible: true
+    },
+    "log_loss": {
+        key: "log_loss",
+        title: localization.Metrics.logLoss,
+        description: localization.loremIpsum,
```
rihorn2, line 196: @romanlutz  or @MiroDudik  we definitely don't want placeholder description. Though this is currently set to not be user visible (is that intentional?)

romanlutz: I just copied from other metrics here. I was similarly surprised to see this (hence my message to you yesterday)

fairlearn/widget/_fairlearn_dashboard.py
romanlutz: so in _group_metric_set.py does it go under
BINARY_CLASSIFICATION_METRICS or
REGRESSION_METRICS?

I had it under regression so far.

visualization/FairnessDashboard/src/AccuracyMetrics.ts
romanlutz: This isn't currently visible. If we want the ones currently with loremIpsum to be visible we need to write all the descriptions. Not doing that in this PR

visualization/FairnessDashboard/src/AccuracyMetrics.ts
```diff
@@ -156,15 +157,17 @@ export const AccuracyOptions: {[key: string]: IAccuracyOption} = {
...
     },
     "root_mean_squared_error": {
         key: "root_mean_squared_error",
         title: localization.Metrics.rms_error,
         description: localization.Metrics.rmseDescription,
         isMinimization: true,
         isPercentage: false,
-        userVisible: true
+        userVisible: true,
+        alwaysUpperCase: true
```
rihorn2, line 170: great catch

visualization/FairnessDashboard/src/Controls/ModelComparisonChart.tsx
```diff
@@ -182,9 +182,11 @@ export class ModelComparisonChart extends React.PureComponent<IModelComparisonPr
             return series;
         });
         const accuracyMetricTitle = selectedMetric.title;
```
rihorn2, line 184: Github may be cutting out instances when I search, but it seems there are only two uses of this, both with the same ternary to do lowercase, so you could just set here using the ternary to correctly capitalize, then use the variable below without modification?

romanlutz: done

fairlearn/widget/_fairlearn_dashboard.py
MiroDudik: this seems like a bug to me--we should have three task types, but let's leave that for a future PR. (tagging @riedgar-ms )

Issue #433, opened 2020-05-22
Microsoft Copyright Notice


koaning:
When you look at `setup.py` you'll notice these two lines: 

```
# Copyright (c) Microsoft Corporation. All rights reserved.
# Licensed under the MIT License.
```

Just to check; does the project intend on keeping the copyright line? It appears in every file.

MiroDudik:
It's because the file was contributed while we were on MSFT payroll--MSFT employees can contribute to OSS projects as long as we put the copyright notice in the source files we create. If you modify the file, you can of course add your copyright notice to the top (but don't need to). Similarly, if MSFT folks modify files initially contributed by others, we're not expected (but can) add the additional copyright notice. I hope this is not an issue--but we should explain this in our contributor guide probably?

koaning:
Since there's an MIT license I suppose that technically, you can still argue for openness. But it is a turnoff. 

Contributing to a file that starts with `Copyright (c) Microsoft Coporation. All rights reserved.` feels like I am using my spare time to contribute to something that is owned by a big company rather than a open source project that I am building with peers. 

Also, side note, the link to the contributor guide on the readme is broken; https://fairlearn.github.io/contributor_guide.html.

adrinjalali:
I agree. There's also the `SECURITY.md` file which I still think should be removed.

MiroDudik:
I see your points--let me consult with MSFT lawyers to get to the bottom of the rationale and see what can be relaxed. For example, would it help if we just kept the copyright notice in the LICENSE file? On a basic level, I'm not sure whether we should be preventing any contributor (e.g., folks from Google, Facebook, or any other corporations) for stating their copyright as long as they keep the MIT license.

I think that we're running into a misconception--but a turnoff nevertheless. The way I understand this, in case of the MIT license, the copyright notice is like stating the authorship and so when somebody copies substantial portions of the code, they need to give credit to the author (just by stating their names as given in the copyright notice). Other than that--everybody is free to do as they please with the code. When you update the code, you become a co-author and you can add your name in the form of the copyright notice at the top.

Re. `SECURITY.md`, I'll try to get some clarity around that as well. Last we heard is that until we have an independent entity, e.g., "a Fairlearn foundation", we should keep `SECURITY.md`. @romanlutz : is that right?

romanlutz:
@vingu was involved in those discussions, but I think you’re right. 

[...]

PR #447, opened 2020-06-02
update copyright to include contributors

romanlutz:
This is related to #433 . As a first step we're extending the copyright to contributors.


fairlearn/metrics/_extra_metrics.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) Microsoft Corporation. All rights reserved.
```
riedgar-ms, line 1: Random question: why remove 'all rights reserved' ?

romanlutz: Qouting our lawyers: “we no longer use ‘all rights reserved’”

riedgar-ms: Fair enough :-)

Issue #448, opened 2020-06-02
AttributeError: 'GridSearch' object has no attribute 'predictors_'


jingwora:
I have done pip install -r requirements.txt
And run the "Grid Search with Census Data" example notebook.
Everything go well until below cell.

------------------------------------------
sweep.fit(X_train, Y_train,
          sensitive_features=A_train)
predictors = sweep.predictors_

------------------------------------------

AttributeError                            Traceback (most recent call last)
<ipython-input-24-49c8aaf5d801> in <module>
----> 1 predictors = sweep.predictors_

AttributeError: 'GridSearch' object has no attribute 'predictors_'

Could anyone quide me how to solve this problem?

Environment: Python 3.6

kevinrobinson:
It sounds like you are running this from source on a recent commit?  The issue may be related to  https://github.com/fairlearn/fairlearn/pull/386/files#diff-9a9553e456a6225df4024e931e062bc6R181, where there are changes to this part of the code since the 0.4.6 release.  I'm figuring the example notebook you're trying might not have been updated as part of those code changes.

MiroDudik:
I think that the issue is actually the opposite: `predictors_` were added _after_ 0.4.6 and they will be part of a new release, so the notebooks on the `master` branch in the repository have already been updated. I am guessing that your are running a pip-installed Fairlearn package, which does not have `predictors_`.

You have two options
* run an older version of the notebook with the pip-installed Fairlearn. Call `pip freeze` to check your installed version, and then find the notebook on, e.g., https://github.com/fairlearn/fairlearn/tree/v0.4.6
* run Fairlearn as well as notebooks from a cloned repository: https://fairlearn.github.io/contributor_guide/development_process.html#advanced-installation-instructions

Let me know if these solve your issue!

jingwora:
Thank you for all help!!! 
It is due to mismatch version. 
I have decided to move to v0.4.6.

riedgar-ms:
Good to hear you're unblocked @jingwora . We are working to improve our backwards compatibility story, since we know we've broken a lot of things recently.

PR #449, opened 2020-06-02
Add ReadMe about notebook versions

riedgar-ms:
Add a short ReadMe to the notebooks, explaining the need to keep the version aligned with whatever is `pip install`ed. This may help with things like #448 , but the real solution is to reduce our API breaks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,12 @@ For instructions on how to install Fairlearn check out our [Quickstart guide](ht
 ## Usage
 
 For common usage refer to the [Jupyter notebooks](./notebooks) and our
-[user guide](https://fairlearn.github.io/user_guide/index.html)
+[user guide](https://fairlearn.github.io/user_guide/index.html).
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -63,7 +63,12 @@ For instructions on how to install Fairlearn check out our [Quickstart guide](ht
 ## Usage
 
 For common usage refer to the [Jupyter notebooks](./notebooks) and our
-[user guide](https://fairlearn.github.io/user_guide/index.html)
+[user guide](https://fairlearn.github.io/user_guide/index.html).
+Please note that our APIs are subject to change, so notebooks downloaded
+from `master` may not be compatible with Fairlearn installed with `pip`.
+In this case, please navigate the tags in the repository
+(e.g. [v0.4.5](https://github.com/fairlearn/fairlearn/tree/v0.4.5))
```
kevinrobinson, line 70: nice! 👍

PR #450, opened 2020-06-02
add build_widget.py extension to build local changes

romanlutz:
This currently involves all these steps to be run manually, so this script should make things easier.

### DIFF
--- a/scripts/build_widget.py
+++ b/scripts/build_widget.py
@@ -1,3 +1,9 @@
+"""Fairlearn dashboard widget build scripts.
+
+To build the widget in order to validate local changes to the visualizations
+add the --use-local-changes option and run `pip install .` after completion.
+"""
+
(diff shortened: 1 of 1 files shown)


scripts/build_widget.py
```diff
@@ -52,18 +64,42 @@ def main(argv):
         subprocess.check_call([args.yarn_path, "build:all"],
```
rihorn2, line 64: I think you want to do the install at line 47, then if the use_local_changes is true, do those steps (all look correct), then resume the normal flow of build (line 63) and clean up (line 91, which is currently in an else block). The new automation should be very helpful, thank you for adding this script!

romanlutz: The reason I didn't do that is because build:all didn't work. Only build did. For reasons I don't understand...

Issue #456, opened 2020-06-04
Add links to source code to API reference


hildeweerts:
#### Describe the issue linked to the documentation
It would be nice to be able to quickly navigate from the API reference to the source code on Github.

#### Suggest a potential alternative/fix
I believe scikit-learn uses the sphinx extension linkcode to achieve this.

riedgar-ms:
That would certainly be useful. @adrinjalali , can you comment from the scikit-learn perspective?

adrinjalali:
Yep, we use `linkcode`, and the configuration is pretty easy:

https://github.com/scikit-learn/scikit-learn/blob/master/doc/conf.py
https://github.com/scikit-learn/scikit-learn/blob/master/doc/sphinxext/github_link.py

PR #461, opened 2020-06-04
add links to github from API reference

romanlutz:
Followed the same pattern as pandas, but it doesn't show up yet for some reason that I haven't figured out yet....

Related to #456 @hildeweerts


romanlutz:
@adrinjalali if you have any clue what's missing there lmk. I've spent a couple hours debugging this and comparing it with pandas and sklearn without any progress. The [source] link just doesn't show up. The filenames are generated properly, though, so there must be something wrong in terms of the options we're selecting in conf.py or the template itself (although pandas has the link show up and we use their template...)

romanlutz:
> @adrinjalali if you have any clue what's missing there lmk. I've spent a couple hours debugging this and comparing it with pandas and sklearn without any progress. The [source] link just doesn't show up. The filenames are generated properly, though, so there must be something wrong in terms of the options we're selecting in conf.py or the template itself (although pandas has the link show up and we use their template...)

Hah! Update: it renders properly in CircleCI, just not on my machine :-)

PR #462, opened 2020-06-05
Add FAQ and Further Reading

riedgar-ms:
Add a FAQ based on questions asked at Build. Also add a Further Reading section for generally interesting books and papers.


docs/user_guide/faq.rst
```diff
@@ -0,0 +1,112 @@
...
+Fairlearn questions
+-------------------
+
+Can Fairlearn make my model fair?
+    Naive use of Fairlearn is not going to result in fairer models.
+    Indeed, when used incorrectly, Fairlearn could produce a model which is *less* fair.
+
+    What Fairlearn can do is generate a model which has lower disparity as measured by
+    a user-specified metric.
+    Whether this makes the model fairer is outside the remit of Fairlearn - or indeed of
```
romanlutz, line 51: I re-read this a few times and I don't like how it portrays the project. I think a big part of what we want to achieve is that Fairlearn is *not* just code, but educational material to help stakeholders understand fairness and apply the toolkit in practice. That would mean it's very much not "ouside the remit of Fairlearn" (unless you mean just the code), but we should rather provide the resources to enable this to be done properly

docs/user_guide/faq.rst
```diff
@@ -0,0 +1,112 @@
...
+    machine learning in general.
+    To determine whether the new model is fairer, the whole system of which it is a part
+    needs to be considered in terms of the societal context in which it is used.
+    See, for example, "Fairness and Abstraction in Sociotechnical Systems" by Selbst et al.,
+    listed on our :ref:`further_reading` page. 
+
+What sort of fairness related harms can Fairlearn address?
+    Fairlean concentrates on group fairness.
```
romanlutz, line 59: typo: Fairlearn

docs/user_guide/faq.rst
```diff
@@ -0,0 +1,112 @@
...
+    That is, obtaining the minimum disparity on some metric evaluated on
```
romanlutz, line 60: nit: This isn't a real sentence and should probably be part of the previous one

Fairlearn concentrates on group fairness, i.e., obtaining ....

docs/user_guide/faq.rst
```diff
@@ -0,0 +1,112 @@
...
+    See :ref:`types_of_harms` for more information.
+    Beyond these, there are also fairness concepts which are note amenable to
+    mathematical (and hence algorithmic) expression - justice, due process and
+    righting historic iniquities for example.
+
+Can Fairlearn be used to detect bias in datasets?
+    We do not have concrete plans for this at the present time.
+
+Can Fairlearn recommend ways to make my model fairer?
+    No. Fairlearn is not magic.
```
romanlutz, line 75: I'm not sure we want to be snarky in the responses....

It's actually a good question, and recommendations *could* theoretically be made. They may just not be very good unless we highly contextualize them, and that's a lot of work with very uncertain outcomes as far as I see it. Better to provide thorough examples.

This could link to the fairness checklist http://www.jennwv.com/papers/checklists.pdf

docs/user_guide/further_reading.rst
```diff
@@ -0,0 +1,25 @@
...
+rather than technical details of algorithms.
+
+The following books provide excellent overviews of the various ways in which machine
+learning algorithms can have harmful effects:
+
+- `Weapons of Math Destruction <https://weaponsofmathdestructionbook.com/>`_ by Cathy O'Neill
+- `Artificial Unintelligence <https://mitpress.mit.edu/books/artificial-unintelligence>`_ by Meredith Broussard
+- `Algorithms of Oppression <https://nyupress.org/9781479837243/algorithms-of-oppression/>`_ by Safiya Umoja Noble
+
+There are also papers which focus less on individual algorithms, and more on societal impacts.
```
romanlutz, line 17: Probably should add gender shades there as well ;-)

docs/user_guide/index.rst
```diff
@@ -9,4 +9,6 @@ User Guide
 
    fairness_in_machine_learning
    assessment
-   mitigation
\ No newline at end of file
+   mitigation
+   further_reading
+   faq
```
romanlutz, line 14: if you pull you'll see the "about" section (as a peer to user_guide) and you could add it there. Definitely a better place than the user guide. Additionally, you can try and add it also to contents.rst which would also add it at the top level. Not sure whether both of these things are possible. Otherwise just as a top-level item would also be fine.

romanlutz: Probably still want to remove it from the user guide, right?

riedgar-ms: Oops, yes.

docs/faq.rst
```diff
@@ -42,45 +42,47 @@ Won't making a model fairer reduce its accuracy?
...
     things fair automatically.
-    In order to use Fairlearn successfully, you first need to work out
+    In order to use the Fairlearn library successfully, you first need to work out
     what fairness means for your particular problem.
-    Once that is determined, you can then work out how to use the
-    various algorithms and fairness metrics which Fairlearn supports
-    to develop models which are *fairer by the standard you have chosen*.
+    We aim to provide a rich discussion of the topic of fairness in machine learning
+    on this website, together with links to the literature.
+    Once you have determined what fairness means for your machine pipeline,
```
romanlutz, line 82: machine pipeline? Did you mean "machine learning pipeline"?

Usually the term AI system is preferable.

docs/user_guide/further_reading.rst
```diff
@@ -14,12 +14,21 @@ learning algorithms can have harmful effects:
...
 - `Measures and Mismeasures of Fairness <https://5harad.com/papers/fair-ml.pdf>`_ (Corbett-Davies & Goel, 2018)
 - `Fairness through Awareness <https://arxiv.org/abs/1104.3913>`_ (Dwork et al., 2011)
 - `Algorithmic Fairness from a Non-ideal Perspective <http://zacklipton.com/media/papers/fairness-non-ideal-fazelpour-lipton-2020.pdf>`_ (Fazelpour & Lipton, 2020)
-- `A Survey on Bias and Fairness in Machine Learning <https://arxiv.org/abs/1908.09635>`_ (Mehrabi et al. 2019)
-- `Fairness and Abstraction in Sociotechnical Systems <https://papers.ssrn.com/sol3/papers.cfm?abstract_id=3265913>`_ (Selbst et al., 2018)
\ No newline at end of file
+- `A Survey on Bias and Fairness in Machine Learning <https://arxiv.org/abs/1908.09635>`_ (Mehrabi et al., 2019)
+- `Fairness and Abstraction in Sociotechnical Systems <https://papers.ssrn.com/sol3/papers.cfm?abstract_id=3265913>`_ (Selbst et al., 2018)
+
+A paper discussing a process (using checklists) to ensure that fairness
```
romanlutz, line 32: seems like there's an issue with the sentence, either the "that" in the end should be removed or something else is missing at the end.

docs/faq.rst
riedgar-ms: Switched to AI system

docs/user_guide/further_reading.rst
riedgar-ms: Urghh yes. Obviously got distracted halfway through writing that sentence :-/

docs/faq.rst
```diff
@@ -0,0 +1,114 @@
+.. _faq:
+
+Frequently asked questions
+==========================
+
+General fairness questions
```
MiroDudik, line 6: Can we remove the subsections for the time being?

docs/faq.rst
```diff
@@ -0,0 +1,114 @@
...
+--------------------------
+
+Where can I learn more about fairness in machine learning?
+    For a high level view of the subject, please see the :ref:`further reading <further_reading>`
```
MiroDudik, line 10: I think that we should replace "further reading" with "further resources" and just send the users there. That page should include both general links as well as academic papers, so we don't need to call out academic papers in this answer.

docs/faq.rst
```diff
@@ -0,0 +1,114 @@
...
+    section, where we provide links to a variety of books and papers.
+    In our :ref:`user guide <user_guide>` we also have links to the papers describing
+    Fairlearn's mitigation algorithms.
+
+Why not just ignore the sensitive features?
+    Because ignoring the sensitive feature usually doesn't work.
```
MiroDudik, line 16: We shouldn't say "it usually doesn't work" without calling out the use case scenario where it does not work. In this question, there is a hidden assumption about what notion of fairness the user is interested in--most likely the goal is to achieve statistical independence between the sensitive feature and prediction. The answer needs to start by clarifying this notion of fairness and then pointing out that there will be statistical dependence between the sensitive feature and other features, which may will lead to the statistical dependence between the predictor and sensitive features.

docs/faq.rst
```diff
@@ -0,0 +1,114 @@
...
+    algorithms *will* uncover these links (it is what they are designed to do).
+    For example, in the US, the ZIP code where a person lives is well correlated with their
+    race.
+    Even if the model is not provided with race as a feature, the model will pick up on it
+    implicitly via the ZIP code (and other features).
+    Worse, without having the race available in the dataset, it is not possible to determine
+    whether any use of ZIP code was as a proxy for race, or for some other reason.
+
+The model is unfair because the data are biased. Isn't it better to get better data?
+    If a model is unfair due to biased data (note that this is not the only reason a model
```
MiroDudik, line 28: Again the first step in answering the question is to dissect the notion of fairness. The model can be "unfair" in different ways. I am not sure if the notion of fairness here is related to the harms of allocation or harms of quality of service. The answers will be a bit different. Also, we need to clarify what is meant by "biased data"--this can mean different things. Does this refer to the fact that different groups might not be sufficiently well represented (so data is skewed towards certain groups)? Or does this mean that the labels are generated by some process that needs to be accounted for (otherwise a statistical bias ensues).

docs/faq.rst
```diff
@@ -0,0 +1,114 @@
...
+    may be unfair), then obtaining unbiased data is definitely the best solution.
+    However, doing so might not always be practical.
+    In these cases, the mitigation techniques in Fairlearn and similar packages can redistribute
```
MiroDudik, line 31: I don't like the optics of redistribution here, because it encourages zero-sum thinking about allocation harms. I would instead say something like. "There are often many machine learning models that achieve similar levels of accuracy or other performance metrics, but that dramatically differ in how they affect different subgroups. Mitigation algorithms seek to improve the fairness metrics without strongly affecting the accuracy, or more generally to navigate the trade-offs between performance and fairness metrics." This point is similar to the one we're making in the next question.

docs/faq.rst
```diff
@@ -0,0 +1,114 @@
...
+    harms between subgroups in the data.
+    The extent to which this is 'fair' will depend on how the harms are redistributed, and the
+    broader context of the system of which the model is a part.
+
+Won't making a model fairer reduce its accuracy?
+    Any technique which reduces disparity in some metric is most likely to impact other metrics.
```
MiroDudik, line 37: Actually I would probably just use the answer that I gave [above](https://github.com/fairlearn/fairlearn/pull/462/files#r444376612) .

docs/user_guide/further_reading.rst
```diff
@@ -0,0 +1,35 @@
+.. _further_reading:
+
+Further Reading
```
MiroDudik, line 3: I'm in favor of just creating a file "further_resources.rst" rather than just "further_reading.rst", because we can then include links to videos / toolkits etc. 

For the list of resources, I'd rather start with just a very simple list grouped into different kinds of resources, but I'd rather not provide the annotations (they're going to be harder to scale and they'll be necessarily subjective, so possibly a subject of discussion).

We can use high-level groupings:
* "Books" (the three below)
* "Academic papers" (including the ones below)
* "Popular writings" (including things like blog posts; maybe there's a better name than "Popular writings")
* "Video tutorials"
* "Online textbook", e.g., https://fairmlbook.org/
* "Academic conferences"

...

docs/faq.rst
riedgar-ms: Done

docs/faq.rst
riedgar-ms: I've renamed "Further Reading" to "Further Resources"

docs/faq.rst
riedgar-ms: I was enjoying being blunt :-)

To tell the truth, that hidden assumption hadn't occurred to me. I've made it explicit.

docs/faq.rst
riedgar-ms: Rephrased.

docs/faq.rst
riedgar-ms: I have done so.

docs/faq.rst
riedgar-ms: Updated with an example

docs/user_guide/further_reading.rst
riedgar-ms: Don't have entries for all those headings yet, but have gathered things under them and removed annotations.

docs/faq.rst
```diff
@@ -0,0 +1,112 @@
...
+    Fairness is a social concept, and no technological solution can make
+    things fair automatically.
+    In order to use the Fairlearn library successfully, you first need to work out
+    what fairness means for your particular problem.
+    We aim to provide a rich discussion of the topic of fairness in machine learning
+    on this website, together with links to the literature.
+    Once you have determined what fairness means for your AI system,
+    you can then work out how to use the various algorithms and fairness metrics
+    which Fairlearn supports to develop models which are
+    *fairer by the standard you have chosen*.
```
romanlutz, line 87: Should we add that fairness may not always be quanitifiable.

riedgar-ms: Like on line 70?

romanlutz: Right. It's a different answer. Duplication isn't a big deal here since we're really trying to make the message stick :-)

riedgar-ms: Duplicated. Well, rephrased slightly.

docs/faq.rst
```diff
@@ -0,0 +1,124 @@
...
+Can the mitigation algorithms in Fairlearn make my model fair?
+    There are many ways in which a model can be unfair. Fairlearn mitigation algorithms
+    only address some of them: those that can be quantified by our supported
+    fairness metrics.
+    However, to assess whether the new model is fairer, it is important to consider
+    not only the fairness metrics, but also the societal and technical context in which
+    the model is applied.
+    
+What sort of fairness-related harms can the Fairlearn library address?
+    We focus on two kinds of harms:
```
riedgar-ms, line 70: Should this be 'We currently focus' ?

PR #464, opened 2020-06-06
Information Filter

koaning:
This is a PR for https://github.com/fairlearn/fairlearn/issues/430.


koaning:
This is my initial proposal for the information filter. I wonder about testing; if I am adhering to the correct standard. Very much open to feedback.

fairlearn/preprocessing/__init__.py
```diff
@@ -0,0 +1,10 @@
+# Copyright (c) Microsoft Corporation and contributors.
+# Licensed under the MIT License.
+
+"""This module contains algorithms implementing preprocessing steps that might help
+filter information that correlates with sensitive attributes.
```
romanlutz, line 5: I guess for right now it's fine, but in general preprocessing can do things other than filtering information, right? I remember Sanmi Koyejo's NeurIPS 2019 presentation which summarized ideas in the area of representation learning for fairness, for example, which could also fit in here. http://sanmi.cs.illinois.edu/documents/Representation_Learning_Fairness_NeurIPS19_Tutorial.pdf

fairlearn/preprocessing/__init__.py
```diff
@@ -0,0 +1,10 @@
...
+"""
+
+from ._informationfilter import InformationFilter
```
romanlutz, line 8: so far we've used the convention of separating words in file names with underscores, i.e., _information_filter.py

Any particular reason not to do that?

fairlearn/preprocessing/_informationfilter.py
```diff
@@ -0,0 +1,120 @@
+import numpy as np
```
romanlutz, line 1: nit: can we add the license header?

fairlearn/preprocessing/_informationfilter.py
```diff
@@ -0,0 +1,120 @@
...
+
+def _vector_projection(vec, unto):
+    return _scalar_projection(vec, unto) * unto
+
+
+class InformationFilter(BaseEstimator, TransformerMixin):
+    r"""
+    The `InformationFilter` filters out sensitive correlations in a dataset.
+
+    The `InformationFilter` uses a variant of the gram smidt process
```
romanlutz, line 21: Did you mean Gram-Schmidt?

romanlutz: also occurs in the code itself

test/unit/preprocessing/informationfilter/test_sklearn_compat.py
```diff
@@ -0,0 +1,54 @@
...
+        estimator_checks.check_estimators_pickle,
+    ]
+)
+def test_estimator_checks(test_fn):
+    test_fn(InformationFilter.__name__, InformationFilter(columns=[]))
+    test_fn(InformationFilter.__name__, InformationFilter(columns=[0]))
+
+
+def test_linear_dependence():
+    X = np.array([[0, 0, 1, 1, ],
```
romanlutz, line 48: Does it work just with numpy ndarrays or also with pandas DataFrames? In lots of other tests we use `pytest.mark.parametrize` with pre-defined combinations of input transformations to ensure that all combinations work.

Example: https://github.com/fairlearn/fairlearn/blob/fee71c194e5fbc08d2d17f431d30788752adba5c/test/unit/reductions/exponentiated_gradient/test_exponentiatedgradient_arguments.py#L34

That may be useful here, too!

fairlearn/preprocessing/_informationfilter.py
```diff
@@ -0,0 +1,120 @@
...
+       \\begin{split}
+       v_1 & = x_1 \\\\
+       v_2 & = x_2 - \\frac{x_2 v_1}{v_1 v_1}\\\\
+       v_3 & = x_3 - \\frac{x_k v_1}{v_1 v_1} - \\frac{x_2 v_2}{v_2 v_2}\\\\
+       ... \\\\
+       v_k & = x_k - \\frac{x_k v_1}{v_1 v_1} - \\frac{x_2 v_2}{v_2 v_2}
+       \\end{split}
+
+    Concatenating our vectors (but removing the sensitive ones) gives us
+    a new training matrix :math:`X_{fair} =  [v_3, ..., v_k]`.
```
romanlutz, line 40: I'd suggest not to use "fair" here, because fairness is inherently sociotechnical. Whether it's fair or not to use this technique depends entirely on the context. You could call it `X_{filtered}`, of course. I currently have a PR out with a styleguide on fairness that may be worth taking a look at. If you have any feedback on that please comment as well :-)

fairlearn/preprocessing/_informationfilter.py
```diff
@@ -0,0 +1,120 @@
...
+def _scalar_projection(vec, unto):
+    return vec.dot(unto) / unto.dot(unto)
+
+
+def _vector_projection(vec, unto):
+    return _scalar_projection(vec, unto) * unto
+
+
+class InformationFilter(BaseEstimator, TransformerMixin):
+    r"""
```
romanlutz, line 18: I'm happy with this doc section. @adrinjalali pointed out that it's probably best to always write the documentation along with the code, i.e., in the same PR. That now also applies to the user guide in the docs/user_guide directory. Since we haven't done that so far I'll leave it up to you for now, but if you already have an idea of what that should look like please go ahead and add it.

Some more context: I'm currently making up for what we don't have in terms of existing mitigation technique, metrics, etc., so until that's in a good state I can't really tell anyone else to do things we haven't done.

fairlearn/preprocessing/__init__.py
koaning: Did not realise that. But can change for sure.

fairlearn/preprocessing/_informationfilter.py
koaning: It seems my dyslexia is making a come-back. Will fix.

test/unit/preprocessing/informationfilter/test_sklearn_compat.py
koaning: I'm using the standard scikit-learn estimator checks. These should already test for a lot of these but I wouldn't mind adding more tests just to be safe.

fairlearn/preprocessing/_informationfilter.py
koaning: It's certainly valid to be critical to the language here. I think filtered is more factual and therefore preferable. Will make the fix.

fairlearn/preprocessing/_informationfilter.py
koaning: I'd love to write some documents on `fetch_boston` and to show mitigation techniques on that dataset. Before I can do that, the dataset needs to get accepted as well as the technique. I am totally planning on doing this tho. It's my preference to have a seperate PR for it but I suppose I could add it here too. I suspect the PR will take a lot longer though.

koaning:
@romanlutz just to check, what does DCO stand for?

Also, is it just me, or does `flake8` run with settings in CI that it does not run with locally?

fairlearn/preprocessing/_informationfilter.py
koaning: Sure.

test/unit/preprocessing/informationfilter/test_sklearn_compat.py
koaning: Adding now.

fairlearn/preprocessing/__init__.py
koaning: Will have a look at that video. I'll keep the information as is for now, it feels best to update it once other tools go in there. There are many preprocessing steps possible but since I can't predict all of them now, it's probably better to iteratively add them.

fairlearn/preprocessing/_information_filter.py
```diff
@@ -0,0 +1,123 @@
...
+
+import numpy as np
+import pandas as pd
+from sklearn.base import BaseEstimator, TransformerMixin
+from sklearn.utils import check_array
+from sklearn.utils.validation import check_is_fitted
+from sklearn.preprocessing import StandardScaler
+
+
+def _scalar_projection(vec, unto):
```
riedgar-ms, line 12: Do you mean `onto` rather than `unto` ? In any case, a longer name might be helpful.

fairlearn/preprocessing/_information_filter.py
```diff
@@ -0,0 +1,123 @@
...
+    columns :math:`x_1, ..., x_k`. If we assume columns :math:`x_1` and :math:`x_2`
+    to be the sensitive columns then the information-filter will
+    remove information by applying these transformations;
+
+    .. math::
+
+       \\begin{split}
+       v_1 & = x_1 \\\\
+       v_2 & = x_2 - \\frac{x_2 v_1}{v_1 v_1}\\\\
+       v_3 & = x_3 - \\frac{x_k v_1}{v_1 v_1} - \\frac{x_2 v_2}{v_2 v_2}\\\\
```
riedgar-ms, line 37: Nitpicking docs, but if these are all vectors, should there be things like `v1 \cdot v1` here?

fairlearn/preprocessing/_information_filter.py
```diff
@@ -0,0 +1,123 @@
...
+       ... \\\\
+       v_k & = x_k - \\frac{x_k v_1}{v_1 v_1} - \\frac{x_2 v_2}{v_2 v_2}
+       \\end{split}
+
+    Concatenating our vectors (but removing the sensitive ones) gives us
+    a new training matrix :math:`X_{filtered} =  [v_3, ..., v_k]`.
+
+    :param columns: list of columns to filter out this can be a sequence of either int
+                    (in the case of numpy) or string (in the case of pandas).
+    :param alpha: parameter to control how much to filter, for alpha=1 we filter out
```
riedgar-ms, line 47: How does `\alpha` affect the equations above?

fairlearn/preprocessing/_information_filter.py
```diff
@@ -0,0 +1,123 @@
...
+                  all information while for alpha=0 we don't apply any.
+    :param center: setting to tell if this preprocessing step should center the data for
+                   numerical stability
+    """
+
+    def __init__(self, columns, alpha=1, center=True):
+        self.columns = columns
```
riedgar-ms, line 54: Should the `self` properties have trailing underscores @romanlutz ?

fairlearn/preprocessing/_information_filter.py
```diff
@@ -0,0 +1,123 @@
...
+        X_fair = X.copy()
+        v_vectors = self._make_v_vectors(X, self.col_ids_)
+        # Gram-Schmidt process but only on sensitive attributes
+        for i, col in enumerate(X_fair.T):
+            for v in v_vectors.T:
+                X_fair[:, i] = X_fair[:, i] - _vector_projection(X_fair[:, i], v)
+        # We want to learn matrix P: X P = X_filtered
+        # This means we first need to create X_filtered in order to learn P
+        self.projection_, resid, rank, s = np.linalg.lstsq(X, X_fair, rcond=None)
+        return self
```
riedgar-ms, line 108: I don't believe we've done this on our other `fit()` methods... I'm guessing it allows users to do things like `est.fit(X).predict(X_test)` ?

fairlearn/preprocessing/_information_filter.py
MBrouns: in sklearn, this should only be the case when they are fitted parameters (e.g. when learned in the `train` method)

fairlearn/preprocessing/_information_filter.py
MBrouns: Exactly, this is also what for example sklearns `TransformerMixin` relies on:

```
class TransformerMixin:
    ....
    def fit_transform(self, X, y):
        ....
        return self.fit(X, **fit_params).transform(X)```
```

fairlearn/preprocessing/_information_filter.py
koaning: It smoothes between the original columns and the newly created ones. I could add some maths if that makes it clearer.

riedgar-ms: It might be more of a userguide thing, but I think it would be helpful to have the way it enters the equations shown.

fairlearn/preprocessing/_informationfilter.py
romanlutz: At least until we're in a better state documentation-wise I'm very much on board with your suggested path forward.

romanlutz:
> @romanlutz just to check, what does DCO stand for?
> 
> Also, is it just me, or does `flake8` run with settings in CI that it does not run with locally?

DCO - developer certificate of origin. For more information see https://fairlearn.github.io/contributor_guide/development_process.html#developer-certificate-of-origin and let me know if you have any questions!

Re flake8: I responded to this on Gitter as well in case you saw that already. There's a config file that handles flake8 settings https://github.com/fairlearn/fairlearn/blob/master/setup.cfg
However, I've seen a difference twice in the last 10 months or so, and they only happened on certain operating systems. The tests run on MacOS, Linux, and Windows, so the differences may be due to that. Judging from the specific errors you have in the pipelines I infer that you perhaps didn't install all the packages from `requirements.txt`. There are some flake8 extension in there that can cause extra errors. Mind trying that? If you prefer not to do that, you can always fix those three errors by hand. Seems simple enough:

```
./fairlearn/preprocessing/__init__.py:4:1: D205 1 blank line required between summary line and description
./fairlearn/preprocessing/__init__.py:4:1: D400 First line should end with a period
./fairlearn/preprocessing/_information_filter.py:111:1: D401 First line should be in imperative mood
```

koaning:
@MiroDudik any comments on the implementation?

fairlearn/preprocessing/_information_filter.py
```diff
@@ -0,0 +1,80 @@
+# Copyright (c) Microsoft Corporation and contributors.
+# Licensed under the MIT License.
+
+import numpy as np
+from sklearn.base import BaseEstimator, TransformerMixin
+from sklearn.utils import check_array
+from sklearn.utils.validation import check_is_fitted
+
+
+class InformationFilter(BaseEstimator, TransformerMixin):
```
MiroDudik, line 10: I am worried about our terminology here. _Filter_ is used frequently in signal processing and there might be some risk of confusion, e.g., https://en.wikipedia.org/wiki/Kalman_filter#Information_filter . Can we do something more descriptive? How about `CorrelationRemover` ? Other ideas?

fairlearn/preprocessing/_information_filter.py
```diff
@@ -0,0 +1,80 @@
...
+    :param columns: list of columns to filter out this can be a sequence of either int
+                    (in the case of numpy) or string (in the case of pandas).
+    :param alpha: parameter to control how much to filter, for alpha=1 we filter out
+                  all information while for alpha=0 we don't apply any.
+    :param center: setting to tell if this preprocessing step should center the data for
+                   numerical stability
+    """
+
+    def __init__(self, columns, alpha=1, center=True):
+        self.columns = columns
```
MiroDudik, line 50: A couple of comments here:
* Elsewhere in Fairlearn, we say `sensitive_features` for the sensitive feature data. I wonder if we can use some parallel terminology here. How about `sensitive_feature_ids` ? I don't love the verbosity, but since it's not a forced keyword argument, maybe not a big issue.
* Based on the writeup [linear-information-filter.pdf](https://github.com/fairlearn/fairlearn/files/4761588/linear-information-filter.pdf), I think I convinced myself that the cleanest course of action is to (1) center the sensitive features, but (2) don't center the remaining features. So--I'm now in favor of dropping `center` argument. Thoughts?

fairlearn/preprocessing/_information_filter.py
```diff
@@ -0,0 +1,80 @@
...
+        non_sensitive = [i for i in range(X.shape[1]) if i not in sensitive]
+        return X[:, non_sensitive], X[:, sensitive]
+
+    def fit(self, X, y=None):
+        """Learn the projection required to make the dataset orthogonal to sensitive columns."""
+        X = check_array(X, estimator=self)
+        X_use, X_sensitive = self._split_X(X)
+        X_s_center = X_sensitive - X_sensitive.mean()
+        self.beta_, _, _, _ = np.linalg.lstsq(X_s_center, X_use, rcond=None)
+        self.X_shape_ = X.shape
```
MiroDudik, line 66: You should also save:
```
self.X_s_mean_ = X_sensitive.mean()
```
(or maybe `self.X_sensitive_mean_`). And that's the mean to be used at the evaluation time.

fairlearn/preprocessing/_information_filter.py
```diff
@@ -0,0 +1,80 @@
...
+        return self
+
+    def transform(self, X):
+        """Transform X by applying the information filter."""
+        X = check_array(X, estimator=self)
+        check_is_fitted(self, ["beta_", "X_shape_"])
+        if self.X_shape_[1] != X.shape[1]:
+            raise ValueError(f"The trained data has {self.X_shape_[1]} while this dataset has {X.shape[1]}.")
+        X_use, X_sensitive = self._split_X(X)
+        X_s_center = X_sensitive - X_sensitive.mean()
```
MiroDudik, line 76: For centering, you should subtract the mean measured on the train data, see the [previous comment](https://github.com/fairlearn/fairlearn/pull/464/files#r443773442).

fairlearn/preprocessing/_information_filter.py
```diff
@@ -0,0 +1,80 @@
...
+        self.beta_, _, _, _ = np.linalg.lstsq(X_s_center, X_use, rcond=None)
+        self.X_shape_ = X.shape
+        return self
+
+    def transform(self, X):
+        """Transform X by applying the information filter."""
+        X = check_array(X, estimator=self)
+        check_is_fitted(self, ["beta_", "X_shape_"])
+        if self.X_shape_[1] != X.shape[1]:
+            raise ValueError(f"The trained data has {self.X_shape_[1]} while this dataset has {X.shape[1]}.")
```
MiroDudik, line 74: Add "columns":
```
"... has {self.X_shape_[1]} columns while ... has {X.shape[1]} columns."
```

fairlearn/preprocessing/__init__.py
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) Microsoft Corporation and contributors.
+# Licensed under the MIT License.
+
+"""This module contains preprocessing steps that help make pipelines more fair."""
```
MiroDudik, line 4: A tweak:
```
This module contains preprocessing algorithms for unfairness mitigation.
```
That'll better match the language we use elsewhere.

fairlearn/preprocessing/_information_filter.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def __init__(self, columns, alpha=1, center=True):
+        self.columns = columns
+        self.alpha = alpha
+        self.center = center
+
+    def _split_X(self, X):
+        """Split up X into a sensitive and non-sensitive group."""
+        sensitive = self.columns
+        non_sensitive = [i for i in range(X.shape[1]) if i not in sensitive]
```
MiroDudik, line 57: Have you tested this when `self.columns` is a list of strings? In that case it looks like you retain all of the original columns.

test/unit/preprocessing/informationfilter/test_sklearn_compat.py
```diff
@@ -0,0 +1,63 @@
...
+    X_sensitive = X[:, 0]
+    X = transform(X)
+
+    X_tfm = InformationFilter(columns=[0]).fit(X).transform(X)
+
+    cov_col_0 = np.cov(X_sensitive, X_tfm[:, 0])
+    cov_col_1 = np.cov(X_sensitive, X_tfm[:, 1])
+    assert np.isclose(cov_col_0[0, 1], 0.0)
+    assert np.isclose(cov_col_1[0, 1], 0.0)
+    assert X_tfm.shape[1] == 2
```
MiroDudik, line 62: We should have an analogous test when the columns of `X` are name--only for `pd.DataFrame`.

fairlearn/preprocessing/_information_filter.py
```diff
@@ -0,0 +1,80 @@
...
+
+import numpy as np
+from sklearn.base import BaseEstimator, TransformerMixin
+from sklearn.utils import check_array
+from sklearn.utils.validation import check_is_fitted
+
+
+class InformationFilter(BaseEstimator, TransformerMixin):
+    r"""
+    A component that filters out sensitive correlations in a dataset.
```
MiroDudik, line 12: Suggestions for the docstring:
* we have now moved to `numpydoc`
* In the extended summary, we should just describe what the transformation does, and only provide the implementation details in the `Notes` section (see the recommendation [here](https://numpydoc.readthedocs.io/en/latest/format.html#sections))--but in this case we might not even need the implementation notes.
* For the extended description, I was thinking about something like:
```
CorrelationRemover applies a linear transformation to the non-sensitive feature columns in order 
to remove their correlation with the sensitive feature columns while retaining as much information
as possible (as measured by the least-squares error).
```
* If you think it would be useful to add implementation notes, I think that those should contain:
  * statement of the optimization problem that the transformation is solving (minimize the least-squares error subject to the no-correlation constraint)
  * statement that the solution is found by centering sensitive features, fitting the linear regression model to non-sensitive features, and reporting the residual

fairlearn/preprocessing/_information_filter.py
MBrouns: I do think sklearn deprecated non-keyworded arguments in estimator constructors, something to keep in mind.
That said, I don't mind `sensitive_feature_ids`.

fairlearn/preprocessing/_information_filter.py
romanlutz: Right, we should definitely check that `X` actually had those columns. Otherwise users may think they correctly applied it but they didn't.

fairlearn/preprocessing/_information_filter.py
koaning: That can work.

fairlearn/preprocessing/_information_filter.py
koaning: Resolved!

fairlearn/preprocessing/_information_filter.py
koaning: The current implementation I've made assumes a numpy array which has numeric column names. I didn't know what standard you folks use if it is a pandas dataframe. Do the fairlearn components assume that the input can be either `np.ndarray` or `pd.DataFrame`?

test/unit/preprocessing/informationfilter/test_sklearn_compat.py
koaning: Ah this answers my previous question.

Does fairlearn offer an abstraction to check if `X` is a dataframe or an array? It's something we've been munging over on the scikit-lego side of things as well.

fairlearn/preprocessing/_information_filter.py
koaning: This is no longer relevant.

fairlearn/preprocessing/_information_filter.py
koaning: I've added these changes to the docstring now. I'd say it renders nicely. 

image

romanlutz: I agree, this looks nice.

Elsewhere we either just say sensitive features or use `A`, not `S`. `Z` as `X \ A` is also new to me. Perhaps we need to standardize on these kinds of things at some point to avoid confusing users? @MiroDudik ?

nit: small typo "mathmatically"

test/unit/preprocessing/informationfilter/test_sklearn_compat.py
romanlutz: We have a utility to parametrize the tests to run with all kinds of configurations. You can see that in use here: https://github.com/fairlearn/fairlearn/blob/4dbf97d6fccfbd38be94ac5f7e7f236e4d7ea816/test/unit/reductions/grid_search/test_grid_search_arguments.py#L43

We don't really do anything with X other than passing it to the underlying estimator, so from our perspective we don't even really want to know right now. We do run scikit-learn validation like check_X_y, but we don't take the reformatted output from that.

I'm still not sure I quite understand what you're trying to do. Checking for type is easy with `type(...)` so I must be missing something.

fairlearn/preprocessing/_information_filter.py
romanlutz: X can be ndarray or DataFrame, yes. The API doc (hopefully) says that in every mitigation technique.

We don't do a lot with X itself since we just pass it on to the underlying estimator so far, so we haven't really hit this problem a lot.

It's only two types, so I think it's fine to do this with an `if` and check whether the columns exist. What do you think?

romanlutz:
#605 is completed and @koaning did the same there. Closing this accordingly.

Issue #458, opened 2020-06-04
Roadmap for interpretability of fairness metrics?


hildeweerts:
The correct interpretation of almost any fairness metric relies on which label is positive and which sensitive groups are compared. At the moment, for most metrics, these considerations are implicit in the API. For example, all confusion_matrix related performance metrics will return labels as: “those that appear at least once in y_true or y_pred are used in sorted order” due to the dependency on scikit-learn, but this is not explicitly mentioned in the fairlearn documentation. 

I was wondering whether there are plans to help users (particularly those who do not use the dashboard) interpret the metrics correctly.

I do not have a strong opinion on what it should look like, but I can imagine e.g. consistent usage of ``pos_label`` and an equivalent argument for privileged group membership might help (although the latter is tricky for polytomous sensitive features). Or perhaps some accompanying textual information, e.g. if demographic_parity_difference = 0.10, one could retrieve an 'explanation' along the lines of "selection rate of group=1 (largest) is 0.10 higher than the selection rate of group=2 (smallest)".

riedgar-ms:
Thank you for the feedback. The matter of `pos_label` has certainly come up recently... I'm not sure but I think the reworked metrics API will actually pass that through to the underlying method. Having said that, the fact that you're asking about it means that (even if it would be passed through) it's not documented properly.

The point about a privileged group is an interesting one. Our metrics are always based on max and min, rather than a particular group. But different groups could get the max and min by different metrics. We probably need to support both - i.e. if `privileged_group=None` then do what we do now, and base on max and min. Otherwise, compare to the specified group (and concoct some way of handling the case where the privileged group might not be the max or min - easy for differences, ratios are not so obvious).

hildeweerts:
On a related note, disparities between more than two sensitive groups are very difficult to capture in a single number. To be honest, I cannot really think of a practical scenario in which the min/max difference would be sufficient for me to judge the fairness of a model, let alone compare several models. You would always like to know between *which* groups the disparity is large, what the performance is like for the other groups that are not min/max, etc.

romanlutz:
Thanks for raising the issue @hildeweerts !

There's indeed a lot to be improved in terms of the documentation. Ideally our documentation (API reference) should be clear enough that people understand what they're actually calling. Providing a string "explanation" is a pattern I haven't seen before, but if it's something people find useful we should consider it. As a user I usually don't use things unless I have an idea of what it does, which is why from my perspective the documentation improvements sound like the obvious place to fix this.

[...]

PR #467, opened 2020-06-08
Enable pos_label on metrics

riedgar-ms:
Enable the `pos_label` argument for appropriate metrics. Releated to #458. This involved a few steps:
- Adding tests of the metrics engine, to ensure that extra arguments were passed correctly
- Adding the `pos_label` argument to the functions in `_extra_metrics` which are computed from the confusion matrix. :
  - Ensure the passed in arrays are binary
  - Use the `labels` argument of `confusion_matrix` to select the `pos_label` (the latter needs to be the second & last entry in the list)
  - Add appropriate tests (including tests of the true positive rate, which had been overlooked)
- Creating tests of the `pos_label` argument for the appropriate metrics. Re-use the previous tests, inverting the input arrays, and then set `pos_label=0`


test/unit/metrics/test_metrics_engine_dicts.py
```diff
@@ -88,6 +88,27 @@
     metrics.log_loss_group_min: 7.67546133,
 }
 
+classification_keys = [
+    metrics.true_positive_rate_group_summary,
+    metrics.true_negative_rate_group_summary
+]
+
+metric_group_summary_pos_label_keys = [
+    metrics.true_positive_rate_group_summary,
```
riedgar-ms, line 97: Please let me know if I've overlooked any metrics which should be here.

fairlearn/metrics/_extra_metrics.py
```diff
@@ -6,38 +6,63 @@
...
+
+def _get_labels_for_confusion_matrix(y_true, y_pred, pos_label):
+    """Figure out the labels argument for skm.confusion_matrix.
+
+    This assumes the input arrays are binary, and we need to specify a pos_label.
+    Given the way the calling routines work, this is achieved by ensuring that
+    the pos_label is last in the (two element) list.
+    """
+    my_labels = list(np.unique(np.concatenate((y_true, y_pred), axis=None)))
+    if len(my_labels) != 2:
```
romanlutz, line 28: Right now we require it to be 0 and 1 because some of the moments depend on this. This seems somewhat more general, so I want to be careful and not break that. This code looks like it could allow any two values to pass through, but perhaps I'm missing something

fairlearn/metrics/_extra_metrics.py
```diff
@@ -6,38 +6,63 @@
...
+        raise ValueError(_NEED_TWO_UNIQUE_Y_VALS)
+    if pos_label not in my_labels:
+        raise ValueError(_NEED_TWO_UNIQUE_Y_VALS)
+    if my_labels[1] != pos_label:
+        my_labels = list(reversed(my_labels))
+    return my_labels
+
+
+def true_positive_rate(y_true, y_pred, sample_weight=None, pos_label=1):
```
romanlutz, line 37: I think we could really use the docstrings here. As a user I'd be confused about `pos_label` because TPR has a definition that relies on labels being 0 and 1 and positives being the samples with `y_true=1`.

test/unit/metrics/test_metrics_engine.py
```diff
@@ -43,6 +47,35 @@ def test_smoke(self, transform_y_a, transform_y_p, transform_gid):
         assert metrics.difference_from_summary(result) == 1
         assert metrics.ratio_from_summary(result) == pytest.approx(0.6666666667)
 
+    @pytest.mark.parametrize("transform_gid", conversions_for_1d)
+    @pytest.mark.parametrize("transform_y_p", conversions_for_1d)
+    @pytest.mark.parametrize("transform_y_a", conversions_for_1d)
+    def test_smoke_extra_arg(self, transform_y_a, transform_y_p, transform_gid):
+        y_a = transform_y_a([0, 0, 1, 1, 0, 1, 1, 1])
```
romanlutz, line 54: I'm guessing `y_p` is predicted `y`, but what does `y_a`  stand for?

fairlearn/metrics/_extra_metrics.py
riedgar-ms: Moments call into metrics? I hadn't realised that. But in any case, this doesn't change the existing behaviour.

fairlearn/metrics/_extra_metrics.py
riedgar-ms: Surely for true positive rate, you just need to know which label of the two is considered the 'positive' one?

Fair point about doing the docstrings, though.

test/unit/metrics/test_metrics_engine.py
riedgar-ms: You're asking me to remember something from six or more months ago.... I suspect I was thinking of 'actual'.

fairlearn/metrics/_extra_metrics.py
riedgar-ms: New style docstrings:
https://406-133444044-gh.circle-artifacts.com/0/docs/_build/html/api_reference/fairlearn.metrics.html#fairlearn.metrics.true_positive_rate
Before I go and write them all, let me know what you think

test/unit/metrics/test_metrics_engine.py
romanlutz: That’s a great indication that the abbreviation isn’t working.

fairlearn/metrics/_extra_metrics.py
romanlutz: The current behavior had 0/1 explicitly from what I can tell.

hildeweerts: The current behavior does explicitly state `labels=[0,1]` when calling `confusion_matrix()`, but scikit-learn does not raise an error if you do not adhere to 0/1, it will just ignore other values. I.e. `confusion_matrix(y_true=[0,2,2,0], y_pred=[0,2,0,0], labels=[0,1])` returns `array([[2, 0], [0, 0]])`. So if you want to *require* 0/1, we need to add an additional check.

fairlearn/metrics/_extra_metrics.py
```diff
@@ -6,38 +6,63 @@
...
     r"""Calculate the true positive rate (also called sensitivity, recall, or hit rate)."""
+    my_labels = _get_labels_for_confusion_matrix(y_true, y_pred, pos_label)
```
hildeweerts, line 84: I'd probably just call them 'labels' rather than 'my_labels'.

fairlearn/metrics/_metrics_engine.py
```diff
@@ -159,14 +158,14 @@ def __init__(self, transformation_function, summary_function, name=None):
...
+            _function_name(self._transformation_function),
+            _function_name(self._summary_function))
 
     def __call__(self, y_true, y_pred, *, sensitive_features, **metric_params):
         return self._transformation_function(self._summary_function(
-                y_true, y_pred,
-                sensitive_features=sensitive_features,
-                **metric_params))
+            y_true, y_pred,
+            sensitive_features=sensitive_features,
```
hildeweerts, line 167: This is not unique to this PR, but I feel the argument `sensitive_features` implies a list of feature names (e.g. `['gender', 'race']`) instead of an array indicating group membership. I'm not sure what a better name would be - perhaps `group` or even just `a` or `s` which is often used in fairness lit.

adrinjalali: I'd say [hopefully] most people who end up using this package are not gonna be FAT-ML scholars if we achieve a broad adoption of the library, and hence `A` would be very confusing (it certainly is to me). `sensitive_features` however, does give the kinda feeling to people that we want them to get.

As in passing an array or just the feature names, I think eventually we'd support both, but supporting feature names means a pipeline which passes along feature names which we don't have right now and is in active development in sklearn. So I'd say we can keep the name and add feature name support later?

MiroDudik: Another angle that we were looking at was that `sensitive_features` is a keyword argument that should feel syntactically parallel to `sample_weight`.

fairlearn/metrics/_extra_metrics.py
```diff
@@ -6,38 +6,175 @@
...
+
+    This is an internal method used by the true/false positive/negative
+    rate metrics (and hence are restricted to binary data). We compute
+    these using the confusion matrix.
+    In general, we expect the input data to be :math:`\in {0, 1}`
+    with 1 being the 'positive' label. However, users might want to have
+    a different pair of values, and then specify the positive label
+    for themselves.
+    This method prepares the `labels` argument of
+    :py:func:\sklearn.metrics.confusion_matrix` based on the
```
MiroDudik, line 31: \sklearn -> `sklearn

fairlearn/metrics/_extra_metrics.py
```diff
@@ -6,38 +6,175 @@
...
-def true_positive_rate(y_true, y_pred, sample_weight=None):
-    r"""Calculate the true positive rate (also called sensitivity, recall, or hit rate)."""
+
+def _get_labels_for_confusion_matrix(y_true, y_pred, pos_label):
+    r"""Figure out the labels argument for skm.confusion_matrix.
+
+    This is an internal method used by the true/false positive/negative
+    rate metrics (and hence are restricted to binary data). We compute
+    these using the confusion matrix.
+    In general, we expect the input data to be :math:`\in {0, 1}`
```
MiroDudik, line 26: no need for :math: here; just `to be in {0, 1}`

fairlearn/metrics/_extra_metrics.py
```diff
@@ -6,38 +6,175 @@
...
+    :py:func:\sklearn.metrics.confusion_matrix` based on the
+    user's specifications.
+
+    Parameters
+    ----------
+    y_true : array-like
+        The true values for the metric computation
+
+    y_pred : array-like
+        The predicted alues for the metric computation
```
MiroDudik, line 40: alues -> values

fairlearn/metrics/_extra_metrics.py
```diff
@@ -6,38 +6,175 @@
...
+        The value in the true and predicted arrays to treat as positive
+
+    Returns
+    -------
+    list
+        A two element list, consisting of the unique values of the two arrays
+        with the positive label listed last
+    """
+    my_labels = list(np.unique(np.concatenate((y_true, y_pred), axis=None)))
+    if len(my_labels) != 2:
```
MiroDudik, line 52: we should only be raising if there are more than 2 labels. if there is only 1 label present, there is no ambiguity in defining TPR, FPR, TNR, FNR--simply depending on whether that label equals 'pos_label' or no.

test/unit/metrics/test_metrics_engine_dicts.py
```diff
@@ -88,6 +88,27 @@
...
+    metrics.true_negative_rate_group_summary,
+    metrics.false_positive_rate_group_summary,
+    metrics.false_negative_rate_group_summary,
+    metrics.selection_rate_group_summary,
+    metrics.precision_score_group_summary,
+    metrics.recall_score_group_summary,
+    metrics.f1_score_group_summary
+]
+
+# Subset of metrics which accept a pos_label argument
```
MiroDudik, line 107: I don't think we need this dictionary.

test/unit/metrics/test_metrics_engine_dicts.py
```diff
@@ -109,3 +130,18 @@ def test_metric_group_summary_smoke(func):
...
     assert result == pytest.approx(derived_metric_results[func])
+
+
+@pytest.mark.parametrize("func", metric_group_summary_results_pos_label.keys())
+def test_metric_group_summary_pos_label_0(func):
+    # We're going to set pos_label=0, so for simplicity invert the previous inputs
+    y_true_invert = [1-y for y in y_true]
+    y_pred_invert = [1-y for y in y_pred]
+    result = func(y_true_invert, y_pred_invert, sensitive_features=sf_binary, pos_label=0)
+    assert result.overall == pytest.approx(
```
MiroDudik, line 141: you can just use `metric_group_summary_results[func]` everywhere below, right?

test/unit/metrics/test_metrics_engine_dicts.py
```diff
@@ -88,6 +88,27 @@
     metrics.log_loss_group_min: 7.67546133,
 }
 
+classification_keys = [
```
MiroDudik, line 91: is this used anywhere?

fairlearn/metrics/_extra_metrics.py
```diff
@@ -6,38 +6,175 @@
...
+        raise ValueError(_NEED_TWO_UNIQUE_Y_VALS)
+    if pos_label not in my_labels:
+        raise ValueError(_NEED_POS_LABEL_IN_Y_VALS)
+    if my_labels[1] != pos_label:
+        my_labels = list(reversed(my_labels))
+    return my_labels
+
+
+def true_positive_rate(y_true, y_pred, sample_weight=None, pos_label=1):
```
MiroDudik, line 61: default in all of these should be `pos_label=None`. the meaning of that will be the same as `pos_label=1`, but only if `my_labels` in `_get_labels_for_confusion_matrix` are a subset of `{0, 1}`. otherwise, an exception needs to be raised.

this will prevent a bad corner case, e.g., when the unique labels are `[1, 2]`. in that case defaulting to `pos_label=1` might not be what's intended.

riedgar-ms: Actually, scikit-learn is not consistent:
https://scikit-learn.org/stable/modules/generated/sklearn.metrics.precision_score.html?highlight=precision_score

@adrinjalali do you know why scikit-learn has these different defaults for `pos_label`?

fairlearn/metrics/_extra_metrics.py
riedgar-ms: Thank you.... I should have looked at the doc build

fairlearn/metrics/_extra_metrics.py
riedgar-ms: I think we've gone for the LaTeX rendering of these in other places.

fairlearn/metrics/_extra_metrics.py
riedgar-ms: Thank you :-/

test/unit/metrics/test_metrics_engine_dicts.py
riedgar-ms: Why not? I use it as a quick way of getting a subset of results which can be tested with fiddled `pos_label`

test/unit/metrics/test_metrics_engine_dicts.py
riedgar-ms: I suppose I could, but I thought it would be more natural to use the same dictionary as provides the test data.

test/unit/metrics/test_metrics_engine_dicts.py
riedgar-ms: In a previous attempt at writing the tests :-)

fairlearn/metrics/_extra_metrics.py
riedgar-ms: Yes.... I had been going back and forth on that point. My thinking had been that users might like to know that their dataset was (to coin a phrase) trivial, but I do not feel strongly on the point. I'll update.

fairlearn/metrics/_extra_metrics.py
MiroDudik: The issue is not with the webpage, but with calling `help` in python. Also, as a matter of style, I think that we should only resort to math symbols if they make things clearer. I don't think it's the case here.

test/unit/metrics/test_metrics_engine_dicts.py
MiroDudik: Isn't that goal achieved by `metric_group_summary_pos_label_keys`?

fairlearn/metrics/_extra_metrics.py
MiroDudik: My 2c: on a high level, if the labels appearing in data are `[1,2]`, then it seems wrong to default to 1 as a `pos_label`.

test/unit/metrics/test_metrics_engine_dicts.py
riedgar-ms: Good point, given that the 'data-driven' part is over the keys of the dictionary.

fairlearn/metrics/_extra_metrics.py
riedgar-ms: I don't disagree - however, looking at `recall_score` (which *is* the true positive rate), `pos_label` defaults to 1.

riedgar-ms: Having just experimented with defaulting `pos_label` to `None`, suppose that all the y values are the same....how do we automatically determine the 'positive' value - or do we just raise an error at that point?

riedgar-ms: I'm going with raising an error.

adrinjalali: I prefer the proposal in https://github.com/fairlearn/fairlearn/pull/467#discussion_r439187689 . I'd say the inconsistencies in sklearn come from the fact that they're not written at the same time and not by the same people, and therefore there are quite a bit of those inconsistencies all around the library. But if I wanna chose one, I'd go or the `None` default.

fairlearn/metrics/_extra_metrics.py
```diff
@@ -6,38 +6,194 @@
...
+        raise ValueError(_TOO_MANY_UNIQUE_Y_VALS)
+    if pos_label is None:
+        if len(my_labels) == 1:
+            raise ValueError(_POS_LABEL_AMBIGUOUS)
+    else:
+        if pos_label not in my_labels and len(my_labels) == 2:
+            raise ValueError(_NEED_POS_LABEL_IN_Y_VALS)
+        if len(my_labels) == 2 and my_labels[1] != pos_label:
+            my_labels = list(reversed(my_labels))
+    return my_labels
```
MiroDudik, line 61: this is different from my suggestion in the comment. there are some issues with this default implementation. can you explain why this is preferred?

fairlearn/metrics/_extra_metrics.py
```diff
@@ -6,38 +6,194 @@
...
+        then the largest unique value of the y arrays will be used.
+
+    Returns
+    -------
+    float
+        The false negative rate for the data
+    """
+    unique_labels = _get_labels_for_confusion_matrix(y_true, y_pred, pos_label)
+    if len(unique_labels) == 1:
+        fnr = 0
```
MiroDudik, line 192: `0` -> `0.0` and similarly return `float` elsewhere

fairlearn/metrics/_extra_metrics.py
riedgar-ms: Do you mean:
https://github.com/fairlearn/fairlearn/pull/467#discussion_r439472357
If so, then if the unique values are {1,2} then this does select 2 as the 'positive' case. I've just added some tests to show that.

Or did you mean something else?

fairlearn/metrics/_extra_metrics.py
riedgar-ms: <mutter>type systems</mutter>

fairlearn/metrics/_extra_metrics.py
MiroDudik: Let me explain the proposed logic better:
```
if pos_label is None:
   if sorted unique labels are one of [0], [1], [-1], [0, 1], [-1, 1]:
       create `labels` as if pos_label = 1
   else:
       raise ValueError
else:
   ...       
```
The behavior in the current PR differs when unique values are [0], [-1] or ['a', 'b'] and pos_label=None.

riedgar-ms: Ahhh got it. I missed the bit in your initial comment about only supporting values in {-1,0,1}.

riedgar-ms: I think I've sorted this out now....

fairlearn/metrics/_extra_metrics.py
riedgar-ms: Done.

fairlearn/metrics/_metrics_engine.py
riedgar-ms: I'll resolve this specific thread for now, but this naming issue is certainly one we might wish to revisit.

fairlearn/metrics/_extra_metrics.py
riedgar-ms: I've reworked the handling, I believe in line with what @MiroDudik  suggested.

fairlearn/metrics/_extra_metrics.py
```diff
@@ -6,38 +6,214 @@
...
+        labels11 = frozenset([-1, 1])
+        if labels01.issuperset(unique_labels) or labels11.issuperset(unique_labels):
+            pos_label = 1
+        else:
+            raise ValueError(_RESTRICTED_VALS_IF_POS_LABEL_NONE)
+
+    # Ensure unique_labels has two elements
+    if len(unique_labels) == 1:
+        if unique_labels[0] == pos_label:
+            unique_labels = [_NO_SUCH_ELEMENT, pos_label]
```
MiroDudik, line 65: if pos_label is numeric then i think this will fail:
```
>>> skm.confusion_matrix([1,1,1,1],[1,1,1,1], labels=['not 1',1])
  File "<stdin>", line 1, in <module>
  File "/mnt/c/Users/mdudik/projects/fairlearn/.venv/lib/python3.6/site-packages/sklearn/metrics/_classification.py", line 277, in confusion_matrix
    raise ValueError("At least one label specified must be in y_true")
ValueError: At least one label specified must be in y_true
>>>
```

MiroDudik: oh... i see why this doesn't trigger an error below--because you have unique_labels and cm_labels.

riedgar-ms: Yes, that exception looks rather familiar from when I was developing this.

test/unit/metrics/test_extra_metrics.py
```diff
@@ -1,102 +1,331 @@
...
+        r1 = _get_labels_for_confusion_matrix(['a'], 0)
+        assert np.array_equal(r1, ['a', 0])
+
+    def test_too_many_values(self):
+        expected_msg = "Must have no more than two unique y values"
+        expected_msg_no_pos = \
+            "If pos_label is not specified, values must be from {0, 1} or {-1, 1}"
+        with pytest.raises(ValueError) as e0:
+            _get_labels_for_confusion_matrix([0, 1, 2], None)
+        assert str(e0.value) in {expected_msg, expected_msg_no_pos}
```
riedgar-ms, line 59: The point of checking for the message text is to make sure that we've hit the error condition we're checking for. Having multiple options negates that.

fairlearn/metrics/_extra_metrics.py
```diff
@@ -6,38 +6,197 @@
...
+        labels11 = frozenset([-1, 1])
+        if labels01.issuperset(unique_labels) or labels11.issuperset(unique_labels):
+            pos_label = 1
+        else:
+            raise ValueError(_RESTRICTED_VALS_IF_POS_LABEL_NONE)
+
+    # Ensure unique_labels has two elements
+    if len(unique_labels) == 1:
+        if unique_labels[0] == pos_label:
+            unique_labels = [None, pos_label]
```
riedgar-ms, line 63: It should have occurred to me that `None` was a viable option for the non-existent element.... :-/

test/unit/metrics/test_extra_metrics.py
MiroDudik: there are two error conditions that apply!

riedgar-ms: Certainly, but one or the other should appear deterministically. It's why my version of the checks was very flat - checking for errors early and raising immediately makes it easier to see where you're going to end up.

I can see that if we're looking at the test 'being a user' then this does make sense. I'll just add a note to the test before checkin.

MiroDudik: Thanks for clarifying here. My intention was to make this test a bit future-proof in case we tweak the flow in `_get_labels_for_confusion_matrix`.

Issue #472, opened 2020-06-09
How to proceed with PR Gate that runs notebooks against the last released Fairlearn version?


romanlutz:
This keeps causing confusion, e.g. most recently https://github.com/fairlearn/fairlearn/pull/434#issuecomment-640945245

Any suggestions for how to rename the pipeline? This is complex because in general it should be green/passing. If we make the conscious decision to break something compared to the last version it gets red. From that point onward it stays red until the next release. So we definitely do want to be alerted when it goes red for the first time. That's what makes this tricky. 

@adrinjalali suggested in the past that things like this could perhaps be fixed just before a release. [I hope I'm not rephrasing his ideas in a way that misrepresent them...]

So I'm curious how people feel about this. In a way I feel like solving the documentation versioning in the way scikit-learn does it, i.e. the webpage can be viewed based on whatever version of the package you're using incl. notebook, and therefore you will have no trouble finding the compatible notebook. That would perhaps address the issue sufficiently.

FYI @kevinrobinson helped out some folks over the past week or so who raised this issue here on GitHub, so perhaps you also have thoughts.

FYI @koaning who faced this issue as well.

FYI @riedgar-ms @MiroDudik

riedgar-ms:
I'd say that there's no value to a build which is perpetually red, and hence ignored. Fundamentally, we need to get better about compatibility (which is what my PR over in `fairlearn-proposals` is about). Of course, the build is currently checking _forwards_ compatibility, which is necessarily going to break as features are added (although I believe that the current issues are simply due to naming instability). To keep this build green through such changes, we would need to have tests and notebooks declare their 'minimum supported version.'

Having the documentation versioned is related, and very good to have, but doesn't solve the issue that we have too many breaking changes too often.

Until we get to that v0.5.0 compatibility commitment, it might be worth putting '[Optional]' or something like that on the build name. That will at least reduce the confusion.

adrinjalali:
I don't mean putting anything more than what we already have in master @riedgar-ms . When reviewing a PR, if the PR has to change anything in the tests or the docs to make the CI pass, that means it's probably not backward compatible, and that's enough, isn't it?

riedgar-ms:
@adrinjalali Agreed. But I do like to have builds which double check things :-)

adrinjalali:
[...]

PR #476, opened 2020-06-11
Website: Fix buttons on desktop layout

kevinrobinson:
The main call to action buttons on the website overlap at ~1000px width.

I didn't look at the build failures, as I'm assuming they are being discussed in https://github.com/fairlearn/fairlearn/issues/472.

### before, 1000px width: the buttons overlap
image

### after, 1000px width: buttons are ok
image

### after, iPhone6 ish
Screen Shot 2020-06-11 at 3 56 58 PM


romanlutz:
Thanks for this! Never noticed it before.

Issue #478, opened 2020-06-12
Notebook: Examples of sociotechnical perspectives on pre-trial detention


kevinrobinson:
In the same spirit as https://github.com/fairlearn/fairlearn/issues/413, here's some examples of work on pre-trial detention that may be relevant if the project chooses to use the COMPAS dataset in any examples illustrating the value of its approach.  There's an existing [Binary Classification on COMPAS dataset](https://fairlearn.github.io/auto_examples/notebooks/plot_binary_classification_COMPAS.html) example, and discussion in https://github.com/fairlearn/fairlearn/issues/466#issuecomment-640854619 about potentially using the COMPAS dataset as part of an example or evaluation as well.

I've pulled out a quote from each to show the punchline.  I view the first one as the most important.

### [Understanding the Content and Consequences of Pre-Trial Detention (Lum, Bender and Wilkerson 2018)](https://facctconference.org/2018/livestream_vh210.html)
> Please, if you study algorithmic fairness & have used the COMPAS dataset as an example, watch our session. We need to hear the voices of those affected by our “fair” tech. 

### [An algorithm for removing sensitive information: application to race-independent recidivism prediction (Johndrow and Lum 2017)](https://arxiv.org/pdf/1703.04957.pdf):
> For drug crimes, African American drug users are arrested at a rate that is several times that of Caucasian drug users despite the fact that African American and Caucasian populations are estimated by public health researchers to use drugs at roughly the same rate [Langan,1995]... Given that the outcome variable is observed with bias with respect to the protected variable, we believe that the second set of approaches designed with the objective of achieving equivalent predictive accuracy by race are inappropriate for this particular application, as they ultimately rely upon comparing the  model’s  predictions  of  re-offense  to  a  fundamentally  flawed  and  biased  measure  of  re-offense:re-arrest. 

### [Fairness, Equality, and Power in Algorithmic Decision-Making (Kasy and Adebe 2020)](https://maxkasy.github.io/home/files/papers/fairness_equality_power.pdf)
> The first asks:what is the causal impact of the introduction of an algorithm on inequality, both within and between groups?  In contrast to fairness, this perspective is consequentialist.  It depends on the distribution of outcomes affected by the algorithm rather than treatment, and it does so for the full population rather than only for individuals who are part of the algorithm. 

and

[...]

PR #479, opened 2020-06-12
Website: Change top navbar to link to documentation

kevinrobinson:
The website has two links in the top navbar, but right now they just scroll down to a section on that page that doesn't correspond to what the links say.

This PR removes those links, and replaces them with links the User Guide and API Docs, with the assumption that people trying to get a sense of the project can scroll to scan, and that we sholud enable people who land here knowing what they to get to it directly.  Right now the only way to get to the website outside of the two main call-to-action buttons is the User Guide link buried in the footer.

### Now
"Fairness in AI" scrolls to "How Fairlearn Works" and "About Fairlearn" scrolls to "Why Fairlearn?"
Screen Shot 2020-06-12 at 3 08 29 PM

## After this pull request
Screen Shot 2020-06-12 at 3 08 24 PM

### DIFF
--- a/docs/_templates/index.html
+++ b/docs/_templates/index.html
@@ -27,10 +27,10 @@
       		<div class="collapse navbar-collapse" id="navbarResponsive">
 		        <ul class="navbar-nav ml-auto text-center">
 		          <li class="nav-item pr-0 pr-lg-4">
-		            <a class="nav-link" href="#fairness-in-ai">Fairness in AI</a>
+		            <a class="nav-link" href="https://fairlearn.github.io/user_guide/index.html">User Guide</a>
 		          </li>
 		          <li class="nav-item pr-0 pr-lg-4">
-		            <a class="nav-link" href="#about-fairlearn">About Fairlearn</a>
+		            <a class="nav-link" href="https://fairlearn.github.io/api_reference/index.html">API Docs</a>
 		          </li>
 		          <li class="nav-item pr-0 pr-lg-4">
 		            <a class="nav-link" href="#contribute">Contribute</a>


mesameki:
Thank you @kevinrobinson and @romanlutz for bringing this up. Right now, there are multiple repetitive terms (e.g., contribute) that are absolutely unnecessary. I'm fine with the "after" image. However, I feel we might not need the two buttons of "Get Started" and "API Docs" anymore. In my ideal world, On top, we will have "User Guide", "API Docs", and "GitHub". No need to put "contribute" there as it has a dedicated section on the landing page. Instead "contribute" can maybe be something like "Sample Notebooks" or "View Notebooks" or sth referring to the sample notebooks folder might be better? This way, we can only keep the "Get Started" button and remove the second one?

kevinrobinson:
@mesameki I agree about linking to the notebook directly, but see https://github.com/fairlearn/fairlearn/issues/478 for more on why my vote is to rework that notebook rather than call more attention to it right now.

@romanlutz I agree that the marketing page and website should be an integrated whole.  Over time, I'd love to be able to critique designs in terms of "will this help us reach our intended audience?" or "how well does this meet a real user need?"  I think that'll be hard to do without a deeper shared understanding of the project's roadmap, and its positioning within the broader landscape, etc. but if we can do that our conversations might go a lot farther a lot faster :)

romanlutz:
> @mesameki I agree about linking to the notebook directly, but see #478 for more on why my vote is to rework that notebook rather than call more attention to it right now.
> 
> @romanlutz I agree that the marketing page and website should be an integrated whole. Over time, I'd love to be able to critique designs in terms of "will this help us reach our intended audience?" or "how well does this meet a real user need?" I think that'll be hard to do without a deeper shared understanding of the project's roadmap, and its positioning within the broader landscape, etc. but if we can do that our conversations might go a lot farther a lot faster :)

Yes, roadmap's coming this week. It's really not about an individual notebook per se, rather linking to the section "example notebooks". I don't particularly care whether we include it now or at a later point, except that I'm sure we'll forget about it and it's simpler to just get things into the state that we want them to be in eventually rather than revisiting it. I'm not going to block the PR for that, though. Just wanted @MiroDudik to have a quick look because he had some thoughts about the navbar recently.

PR #488, opened 2020-06-17
Convert quickstart to use doctest

riedgar-ms:
The `quickstart.rst` document contains Python code which a user ought to be able to run. Use the `doctest` extension for Sphinx to enable this.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,8 +25,12 @@ jobs:
           command: sudo pip install -r requirements.txt
       
       - run:
-          name: 'Build'
-          command: python -m sphinx -b html -j auto docs docs/_build/html
+          name: 'Build doctest'
+          command: python -m sphinx -b doctest -n -j auto docs docs/_build/html
+
+      - run:
+          name: 'Build html'
+          command: python -m sphinx -b html -n -j auto docs docs/_build/html
 
       - persist_to_workspace:
           root: docs/_build
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -38,14 +38,14 @@ release = fairlearn.__version__
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
 extensions = [
-    "sphinx.ext.autodoc",
-    "sphinx.ext.napoleon",
-    "sphinx.ext.intersphinx",
-    "sphinx.ext.linkcode",
-    "sphinx.ext.extlinks",
-    "sphinx.ext.mathjax",
-    "sphinx_gallery.gen_gallery",
-    "sphinx.ext.doctest"
+    'sphinx.ext.autodoc',
+    'sphinx.ext.doctest',
+    'sphinx.ext.extlinks',
+    'sphinx.ext.intersphinx',
+    'sphinx.ext.linkcode',
+    'sphinx.ext.mathjax',
+    'sphinx.ext.napoleon',
+    'sphinx_gallery.gen_gallery'
 ]
 
 intersphinx_mapping = {'python3': ('https://docs.python.org/3', None),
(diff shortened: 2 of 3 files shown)


docs/conf.py
```diff
@@ -44,7 +44,8 @@
     "sphinx.ext.linkcode",
     "sphinx.ext.extlinks",
     "sphinx.ext.mathjax",
-    "sphinx_gallery.gen_gallery"
+    "sphinx_gallery.gen_gallery",
```
riedgar-ms, line 47: I am considering sorting these alphabetically

.circleci/config.yml
```diff
@@ -25,8 +25,12 @@ jobs:
           command: sudo pip install -r requirements.txt
       
       - run:
-          name: 'Build'
-          command: python -m sphinx -b html -j auto docs docs/_build/html
+          name: 'Build doctest'
+          command: python -m sphinx -b doctest -n -j auto docs docs/_build/html
```
riedgar-ms, line 29: The `-n` flag is for "nit-pick", which generates warnings about unresolved references and the like.

docs/quickstart.rst
```diff
@@ -82,32 +88,38 @@ definitions from
...
     >>> from sklearn.tree import DecisionTreeClassifier
     >>> 
     >>> classifier = DecisionTreeClassifier(min_samples_leaf=10, max_depth=4)
     >>> classifier.fit(X, y_true)
-    >>> 
+    DecisionTreeClassifier(...)
     >>> y_pred = classifier.predict(X)
     >>> group_summary(accuracy_score, y_true, y_pred, sensitive_features=sex)
-    {'overall': 0.8445686557538159, 'by_group': {'female': 0.9256336459010306, 'male': 0.8044974759063791}}
+    {'overall': 0.844..., 'by_group': {'female': 0.925..., 'male': 0.804...}}
```
riedgar-ms, line 102: The ellipsis is a text match, rather than a proper rounding. However, it feels wiser to do this than assume that the low level maths library never changes.

docs/conf.py
romanlutz: I'm very much for that

riedgar-ms: Done.

docs/quickstart.rst
```diff
@@ -61,12 +61,18 @@ For this example we use the
...
     >>> from shap.datasets import adult  # shap is only used its dataset utility
     >>> X, y_true = adult()
     >>> y_true = y_true * 1
     >>> sex = X['Sex'].apply(lambda sex: "female" if sex == 0 else "male")
+    >>> sex.value_counts()
+    male      21790
+    female    10771
+    Name: Sex, dtype: int64
 
 .. figure:: auto_examples/quickstart/images/sphx_glr_plot_adult_dataset_001.png
```
riedgar-ms, line 77: I hadn't spotted that this was from an `auto_example` .... I think that would complicate things.

romanlutz: I wonder whether there's a solution to this. I don't expect the dataset to change anytime soon, so I'm not super worried.

docs/quickstart.rst
```diff
@@ -82,32 +88,38 @@ definitions from
...
+.. doctest::
+
     >>> from fairlearn.widget import FairlearnDashboard
     >>> FairlearnDashboard(sensitive_features=sex,
-                           sensitive_feature_names=['sex'],
-                           y_true=y_true,
-                           y_pred={"initial model": y_pred})
+    ...                    sensitive_feature_names=['sex'],
+    ...                    y_true=y_true,
+    ...                    y_pred={"initial model": y_pred}) # doctest: +SKIP
```
romanlutz, line 122: why the dots?

romanlutz: This isn't output, but code

docs/quickstart.rst
riedgar-ms: I'll leave for now, then.

docs/quickstart.rst
riedgar-ms: Those are indicating continuation lines.

riedgar-ms: https://docs.python.org/3.8/library/doctest.html

PR #499, opened 2020-06-19
Run doctest build in parallel

riedgar-ms:
Experiment with running the doctest build in parallel to the main documentation build. This is due to the growing execution time of our examples.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,10 +23,6 @@ jobs:
       - run:
           name: 'Install required packages'
           command: sudo pip install -r requirements.txt
-      
-      - run:
-          name: 'Build doctest'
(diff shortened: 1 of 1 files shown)


riedgar-ms:
This is only going to be helpful if our walltime is limited, rather than our CPU time. Indeed, given the installation time requirements, it is rather unhelpful for CPU time.

romanlutz:
> This is only going to be helpful if our walltime is limited, rather than our CPU time. Indeed, given the installation time requirements, it is rather unhelpful for CPU time.

It's going to be helpful regardless. From what I've observed so far CircleCI allows us to use 4 pipelines in parallel, and no individual step can take more than 10 minutes unless you extend the timeout. So this will help in running the pipelines faster. The jobs run in different docker images, so they shouldn't be stealing CPU time from each other.

PR #490, opened 2020-06-18
Documentation: Add microrubric for critiques

kevinrobinson:
## Microrubric for critiques
This updates the page about contributing notebooks to include a short rubric for guiding critiques with five points:

1. Deployment context
2. For developers
3. Sociotechnical
4. Substantiated
5. Real harms

The intention is that this reflects the rough consensus about the team's aspirations from the meeting on 6/18 and discussion afterward.  While the [How to talk and write about fairness in AI](https://fairlearn.github.io/contributor_guide/how_to_talk_about_fairness.html) guide is awesome, in practice it's harder to refer to during a discussion.  So this minirubric is trying to be small enough that the team might practically use it within a meeting or discussion.

PR #502, opened 2020-06-20
Add notebook "Additive Counterfactual Fairness for Loan Approvals"

srayagarwal:


srayagarwal:
Updated the notebook: Additive Counterfactual Fairness for Loan Approvals.ipynb

romanlutz:
Some general thoughts below. These are just things that come to mind, and I don't expect you to have an answer to all of them right away. It's just things that we should think about.

- The notebook is still very long. I don't know whether readers will stick with it or drop off. One option to shorten it would be to focus on classification OR regression (but not both). 
    - given that this is a binary classification task, does it make sense to use it with regression techniques? Perhaps I missed something along the way.
- There are passages with a lot of code cells with no textual explanation in between. As a reader I prefer to have the text be descriptive enough that it would be sufficient even if the code wasn't there (and we could only see the output from the code). I do realize that's perhaps very subjective, so more opinions on this may be required.
- We just started a weekly deep dive session where we discuss example notebooks or applications that may become example notebooks at some point. The point of the session is to capture the sociotechnical context of the application, since it's never just about X, y, and sensitive features and how we're building a model. For this purpose @kevinrobinson has created a PR to add a microrubric for critiques https://github.com/fairlearn/fairlearn/pull/490 . If you're interested in picking this up in the deep dive session at some point let me know. Even otherwise, the points Kevin makes in his PR are very relevant. Questions like the following may need to be addressed
    - what differences are there between different kinds of lending (car loan, mortgage, credit card, ...)?
    - what does the regulatory landscape look like?
    - what do we know about the data? When/where was it collected? By whom? Does this affect the deployment context? 
    - The dataset seems to contain only loans that were granted. What about the people who never got a loan? How does this exclusion potentially affect people if we train a model on this?
    - The dataset seems to have Estonian and English speakers (and other). What's the distribution like? Are there significant differences between them? Should we perhaps look at these contexts separately?
    - What does the "cost" actually mean here? It seems like it's weighted with 0.7 * FP + 0.3 FN. FP would be that we predict someone to default but they actually wouldn't, whereas FN would be that we don't predict someone to default but they would. FN clearly affects the bottom line of the lending institution directly, and FP takes opportunity (a loan) away from a person who deserved to get it.
    - If I understand correctly you have 15 sensitive features, and you're assessing fairness for each of them. Have you considered intersectionality?
    - "It was observed that out of 15 sensitive features ($\mathbf{S}$) around 5 sensitive features are having bias as per statistical parity difference and disparate impact" - how are you setting the threshold for this? I saw the chart, but there was at least one more feature above the line which I guess is the threshold.
    As you can tell I drifted off into details again, but some of the big picture questions should be looked at first.
- New PRs should be in .py format (see examples/notebooks folder) so that they render on the webpage. Reformatting it may also make it easier to review because right now it doesn't even show up in the PR view, so it's impossible to provide comments on individual sections or lines. Looking at it is currently only possible by downloading or going to your branch: https://github.com/srayagarwal/fairlearn/blob/Usecase/notebooks/Additive%20Counterfactual%20Fairness%20for%20Loan%20Approvals.ipynb
  One caveat is that we use ReST instead of markdown in .py because that's what sphinx-gallery needs for rendering on the webpage. It shouldn't be too hard to get started, though. Visual Studio Code provides a "convert to python script" functionality if you right-click on the notebook.

Let me know what you think!

srayagarwal:
*   The notebook is still very long. I don't know whether readers will stick with it or drop off. One option to shorten it would be to focus on classification OR regression (but not both).
     *   given that this is a binary classification task, does it make sense to use it with regression techniques? Perhaps I missed something along the way. : I am trying to show two flavours of same algo. If it is still long, I may like to remove the first part showing disparate and parity differences
  *   There are passages with a lot of code cells with no textual explanation in between. As a reader I prefer to have the text be descriptive enough that it would be sufficient even if the code wasn't there (and we could only see the output from the code). I do realize that's perhaps very subjective, so more opinions on this may be required: I added lot of explanation, I would add more if needed, I just didn’t want to make it text heavy.
  *   We just started a weekly deep dive session where we discuss example notebooks or applications that may become example notebooks at some point. The point of the session is to capture the sociotechnical context of the application, since it's never just about X, y, and sensitive features and how we're building a model. For this purpose @kevinrobinson<https://github.com/kevinrobinson> has created a PR to add a microrubric for critiques #490<https://github.com/fairlearn/fairlearn/pull/490> . If you're interested in picking this up in the deep dive session at some point let me know. Even otherwise, the points Kevin makes in his PR are very relevant. Questions like the following may need to be addressed : Please add me to the meetings
     *   what differences are there between different kinds of lending (car loan, mortgage, credit card, ...)?
     *   what does the regulatory landscape look like?
     *   what do we know about the data? When/where was it collected? By whom? Does this affect the deployment context?
     *   The dataset seems to contain only loans that were granted. What about the people who never got a loan? How does this exclusion potentially affect people if we train a model on this?
     *   The dataset seems to have Estonian and English speakers (and other). What's the distribution like? Are there significant differences between them? Should we perhaps look at these contexts separately?
     *   What does the "cost" actually mean here? It seems like it's weighted with 0.7 * FP + 0.3 FN. FP would be that we predict someone to default but they actually wouldn't, whereas FN would be that we don't predict someone to default but they would. FN clearly affects the bottom line of the lending institution directly, and FP takes opportunity (a loan) away from a person who deserved to get it.
     *   If I understand correctly you have 15 sensitive features, and you're assessing fairness for each of them. Have you considered intersectionality? I am accessing over all bias/discrimination. In acf we can use multiple sensitive feature at one go, thus I avoided intersection features
     *   "It was observed that out of 15 sensitive features ($\mathbf{S}$) around 5 sensitive features are having bias as per statistical parity difference and disparate impact" - how are you setting the threshold for this? I saw the chart, but there was at least one more feature above the line which I guess is the threshold. In ACF model, we are taking multiple sensitive features and not just one, A user can add all the features he may deem fit. The choice of features are based on importance of it for the use case
As you can tell I drifted off into details again, but some of the big picture questions should be looked at first.
New PRs should be in .py format (see examples/notebooks folder) so that they render on the webpage. Reformatting it may also make it easier to review because right now it doesn't even show up in the PR view, so it's impossible to provide comments on individual sections or lines. Looking at it is currently only possible by downloading or going to your branch: https://github.com/srayagarwal/fairlearn/blob/Usecase/notebooks/Additive%20Counterfactual%20Fairness%20for%20Loan%20Approvals.ipynb
One caveat is that we use ReST instead of markdown in .py because that's what sphinx-gallery needs for rendering on the webpage. It shouldn't be too hard to get started, though. Visual Studio Code provides a "convert to python script" functionality if you right-click on the notebook. That is doable. Do you have a sample .py file with ReST that I can follow and rework the ipnby into .py. Even Jupyter gives an option of download as .py file, but the markdown will get lost

romanlutz:
@srayagarwal I've forwarded the invite for the deep dive session. We'd love to have you :-)

I actually like the analysis of disparities a lot, but I think it needs to go beyond just numbers and bar charts. They illustrate metrics, but these metrics actually mean something, and we need to highlight what that is, e.g., group A got a 30% higher false positive rate than group B. In this particular context this is relevant because FPR means that we predict that people will default when they actually won't and that may lead to denying them a loan that they otherwise deserve. Making the FPR similar between all groups at least ensures that we don't systematically disadvantage a group in this manner.

Now Equalized Odds goes further in that it considers FPR and FNR at the same time, so that's something to explain and why both are relevant.
Beyond that, I'm always hesitant with Equalized Odds because it assumes that the labels are fine as they are and bases the Equalized Odds difference (which uses FPR and FNR) on that. But if the underlying dataset is very skewed towards one group (e.g. group A has 80% with label 1 and group B has 10% with label 1) then - depending on the application - minimizing the equalized odds different just means that we're keeping this 80%-to-10% disparity in the selection rate. Perhaps that's fine in the application context, but it's something that probably needs to be discussed. Hopefully this kind of disparity doesn't exist to begin with, but without analyzing it we can't really be sure.

Happy to discuss this more.

romanlutz:
The last comment on this PR was over 6 months ago, so I'm assuming it's not active anymore. Please do correct me if this assumption is wrong. We can certainly talk about this in the weekly community call if it's still of interest.

PR #506, opened 2020-06-23
Add fetch_adult for UCI adult dataset and drop shap dependency

romanlutz:


fairlearn/datasets/_fetch_adult.py
```diff
@@ -0,0 +1,82 @@
...
+
+    as_frame : boolean, default=False
+        If True, the data is a pandas DataFrame including columns with
+        appropriate dtypes (numeric, string or categorical). The target is
+        a pandas DataFrame or Series depending on the number of target_columns.
+        The Bunch will contain a ``frame`` attribute with the target and the
+        data. If ``return_X_y`` is True, then ``(data, target)`` will be pandas
+        DataFrames or Series as describe above.
+
+    return_X_y : boolean, default=False.
```
riedgar-ms, line 48: Is this a common name for this argument? I don't think it makes much sense.

fairlearn/datasets/_fetch_adult.py
```diff
@@ -0,0 +1,82 @@
...
+            Array of ordered feature names used in the dataset.
+        DESCR : string
+            Description of the UCI Adult dataset.
+
+    Notes
+    -----
+    This dataset consists of 48842 samples and 14 features.
+    """
+    if not data_home:
+        data_home = pathlib.Path().home() / ".fairlearn-data"
```
riedgar-ms, line 74: `.fairlearn-data` is a string which feels like it ought to be in a `constants.py`

fairlearn/datasets/_fetch_adult.py
romanlutz: I copied the structure from fetch_boston. It does make sense to me what it means and does, although I would not really use it since it’s a subset of what you get otherwise.

romanlutz: ... and yes, this is the arg name in openml.

fairlearn/datasets/_common.py
```diff
@@ -0,0 +1,11 @@
+# Copyright (c) Microsoft Corporation and Fairlearn contributors.
+# Licensed under the MIT License.
+
+
+_DOWNLOAD_DIRECTORY_NAME = ".fairlearn-data"
```
MiroDudik, line 5: If you are adding this, you should add it to `_fetch_boston` too.

fairlearn/datasets/_common.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+class FairnessWarning(UserWarning):
```
MiroDudik, line 8: we don't need this exception.

fairlearn/datasets/_fetch_adult.py
```diff
@@ -0,0 +1,82 @@
...
+def fetch_adult(*, cache=True, data_home=None,
+                as_frame=False, return_X_y=False):
+    """Load the UCI Adult dataset (binary classification).
+
+    Download it if necessary.
+
+    ==============   ==============
+    Samples total             48842
+    Dimensionality               14
+    Features                   real
```
MiroDudik, line 19: Is that true? I don't think that the features are real valued: https://archive.ics.uci.edu/ml/datasets/Adult

fairlearn/datasets/_fetch_adult.py
```diff
@@ -0,0 +1,82 @@
...
+            a year (>50K) or not (<=50K).
+            If ``as_frame`` is True, ``target`` is a pandas object.
+        feature_names : list of length 14
+            Array of ordered feature names used in the dataset.
+        DESCR : string
+            Description of the UCI Adult dataset.
+
+    Notes
+    -----
+    This dataset consists of 48842 samples and 14 features.
```
MiroDudik, line 71: This information is redundant--maybe just drop it. I see that this was copied from `_fetch_boston`, but I would still drop it here. See: https://numpydoc.readthedocs.io/en/latest/format.html#sections

notebooks/Grid Search with Census Data.ipynb
```diff
@@ -19,7 +19,7 @@
    "source": [
     "## Load and preprocess the data set\n",
     "\n",
-    "For simplicity, we import the data set from the `shap` package, which contains the data in a cleaned format. We start by importing the various modules we're going to use:"
+    "For simplicity, we import the data set using the `fairlearn.datasets`, which contains the data in a cleaned format. We start by importing the various modules we're going to use:"
```
MiroDudik, line 22: Maybe we can simplify the first sentence:
```
We download the data set using `fetch_adult` function in `fairlearn.datasets`.
```

notebooks/Grid Search with Census Data.ipynb
```diff
@@ -141,7 +140,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "We can load this predictor into the Fairness dashboard, and examine how it is unfair (there is a warning about AzureML since we are not yet integrated with that product):"
+    "We can load this predictor into the Fairness dashboard, and examine how it is unfair:"
```
MiroDudik, line 143: "examine how it is unfair" -> "assess its fairness"

fairlearn/datasets/_common.py
MiroDudik: [i'm not 100% sure that this constant should be in `datastes/_common.py` as opposed to one level up, but since it's not part of the exported API, we can just leave things in `datastes/_common.py` for now]

romanlutz: I'm personally not a huge fan of having constants further "up" than they're needed, because it encourages people to "misuse" them for different purposes and then one ends up breaking things later or having to break things up again. You're right about _fetch_boston, of course!

fairlearn/datasets/_fetch_adult.py
romanlutz: The feature `fnlwgt` is real valued. Everything else isn't, although a few others are continuous.

fairlearn/datasets/_fetch_adult.py
romanlutz: That's fair.

fairlearn/datasets/_fetch_adult.py
```diff
@@ -0,0 +1,79 @@
...
+    year.
+
+    Parameters
+    ----------
+    cache : boolean, default=True
+        Whether to cache downloaded datasets using joblib.
+
+    data_home : optional, default: None
+        Specify another download and cache folder for the datasets. By default
+        all scikit-learn data is stored in '~/fairlearn_data' subfolders.
```
riedgar-ms, line 38: `~/.fairlearn_data` surely? I'd consider writing `$HOME` instead of `~` as well.

fairlearn/datasets/_fetch_adult.py
```diff
@@ -0,0 +1,79 @@
...
+        DataFrames or Series as describe above.
+
+    return_X_y : boolean, default=False.
+        If True, returns ``(data.data, data.target)`` instead of a Bunch
+        object.
+
+    Returns
+    -------
+    dataset : :class:`~sklearn.utils.Bunch`
+        Dictionary-like object, with the following attributes.
```
riedgar-ms, line 55: Shouldn't you mention that the return type is dependent on the `return_X_y` argument? And list out the alternatives.

fairlearn/datasets/_fetch_adult.py
romanlutz: Is `$HOME` cross-plat?
Because `~` works everywhere

fairlearn/datasets/_fetch_adult.py
romanlutz: Perhaps. Copied this from _fetch_boston

fairlearn/datasets/_fetch_adult.py
riedgar-ms: The exact syntax certainly isn't (although `~` doesn't work in `cmd.exe`, and that isn't quite dead yet). My concern is that `~` might get lost in the run of text. It's not critical, though.

The missing `.` OTOH..... and now I look closer, there's also `_` vs `-` in the directory name.

fairlearn/datasets/_fetch_adult.py
riedgar-ms: I think it would advisable.

romanlutz: Updated this with the various possible outcomes based on return_X_y and as_frame

fairlearn/datasets/_fetch_adult.py
romanlutz: `~` works in cmd.exe for me

romanlutz: fixed the dir name. Must have copied that from an earlier version of fetch_boston

docs/quickstart.rst
```diff
@@ -99,7 +101,7 @@ we can evaluate metrics to get a group summary as below:
     DecisionTreeClassifier(...)
     >>> y_pred = classifier.predict(X)
     >>> group_summary(accuracy_score, y_true, y_pred, sensitive_features=sex)
-    {'overall': 0.844..., 'by_group': {'female': 0.925..., 'male': 0.804...}}
+    {'overall': 0.8443552680070431, 'by_group': {'Female': 0.9251482213438735, 'Male': 0.8042879019908117}}
```
riedgar-ms, line 104: Doesn't the ellipsis force the match to be approximate when these are run as tests? Why give pointless precision?

riedgar-ms: Per:
https://github.com/fairlearn/fairlearn/pull/488
the ellipsis is a _text_ wildcard. So you can truncate the decimals, but don't round when doing so.

romanlutz: Thanks for the pointer! Updated accordingly.

PR #515, opened 2020-07-10
Connection update for PyPI Release Pipeline

riedgar-ms:
Updating the PyPI release pipeline to use the `TwineAuthenticate` task, rather than fetching things from Azure KeyVault.

### DIFF
--- a/devops/pypi-release-new.yml
+++ b/devops/pypi-release-new.yml
@@ -161,17 +161,12 @@ stages:
     vmImage: $(poolImage)
 
   variables:
-    # Following is actually the name of the service connection from the Project Settings page
-    kvSubscription: "Fairness - Automation (cecafb73-04ae-4432-9f96-d96925d28058)"
-    kvVaultName: fairlearndeploy
     ${{ if eq(parameters.releaseType, 'Test')}}:
-      kvUsername: usernametest
-      kvPassword: passwordtest
-      pypiUrl: https://test.pypi.org/legacy/
+      twineConnection: PyPI-Test
+      twineEndpoint: PyPITest
     ${{ if eq(parameters.releaseType, 'Production')}}:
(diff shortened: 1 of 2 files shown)


romanlutz:
> Updating the PyPI release pipeline to use the `TwineAuthenticate` task, rather than fetching things from Azure KeyVault.

Can you describe what's better about TwineAuthenticate? I'm not familiar with it. Does that mean we should delete the keyvault and it's credentials?

riedgar-ms:
The TwineAuthenticate task bundles together the credentials and endpoint for a PyPI upload (or to a private feed). Yes, it means that we would be able to delete the KeyVault.

romanlutz:
> The TwineAuthenticate task bundles together the credentials and endpoint for a PyPI upload (or to a private feed). Yes, it means that we would be able to delete the KeyVault.

I guess I'm still not sure where the secrets are actually stored then. It seems like we're doing away with the KeyVault, but where is the info stored. That's not clear to me from looking at the code.

riedgar-ms:
I imagine that the secrets are stored in a similar way to the secret which allows use of the service principal which connects to the KeyVault.

romanlutz:
> I imagine that the secrets are stored in a similar way to the secret which allows use of the service principal which connects to the KeyVault.

OK now I'm even more confused. Where are they actually stored? Is it in ADO? The current pipeline handles this by ADO having a service connection that has access to the KeyVault.

riedgar-ms:
> > I imagine that the secrets are stored in a similar way to the secret which allows use of the service principal which connects to the KeyVault.
> 
> OK now I'm even more confused. Where are they actually stored? Is it in ADO? The current pipeline handles this by ADO having a service connection that has access to the KeyVault.

But the service connection must contain some sort of secret for access to the service principal. Where's that stored?

PR #519, opened 2020-07-17
Add fetch_compas

hildeweerts:
This PR adds fetch_compas() and includes a small fix of the docs (fairlearn.datasets was inserted twice in the index). I didn't find any tests for the datasets, not sure what exactly should be tested. It feels a bit weird to have no tests at all though.

Edit: fixed the DCO by running the suggested commands but I'm not sure how I should do this directly in the future?

### DIFF
--- a/docs/api_reference/index.rst
+++ b/docs/api_reference/index.rst
@@ -6,7 +6,6 @@ API Reference
 .. toctree::
    :maxdepth: 1
 
-   fairlearn.datasets
    fairlearn.metrics
    fairlearn.postprocessing
    fairlearn.reductions
--- a/fairlearn/datasets/_fetch_compas.py
+++ b/fairlearn/datasets/_fetch_compas.py
@@ -1,0 +1,139 @@
+# Copyright (c) Microsoft Corporation and Fairlearn contributors.
+# Licensed under the MIT License.
+
+import pathlib
+import warnings
+
+from sklearn.datasets import fetch_openml
+from ._constants import _DOWNLOAD_DIRECTORY_NAME
+
+from fairlearn.exceptions import DataFairnessWarning
+
+
+def fetch_compas(*, cache=True, data_home=None,
+                 as_frame=False, return_X_y=False, warn=True):
+    """Load the ProPublica COMPAS dataset (binary classification).
+
+    Download it if necessary.
+
+    ==============   ===================
+    Samples total                   5278
+    Dimensionality                    13
+    Features         numeric and nominal
+    Classes                            2
+    ==============   ===================
+
+    See [1]_ for the original dataset and [2]_ for pre-processing steps of the current
+    dataset.
+
+    The prediction task is to determine whether a defendant will be charged again within two
+    years, which serves as a proxy for recidivism.
+
+    **This dataset has known fairness issues**. Fairness-unaware classification models are at risk
+    of incorrectly judging African-American defendants to be at a higher risk of recidivism and
+    incorrectly judging Caucasian defendants to be at lower risk of recidivism. See [3]_ for
+    more details.
+
+    Here's a table of all the variables in order:
+
+    =====================  ======================================================================
+    sex                    sex of the defendant; 1 if Male, 0 if Female
+    age                    age of the defendant
+    juv_fel_count          number of juvenile felony convictions
+    juv_misd_count         number of juvenile misdemeanor convictions
+    juv_other_count        number of other juvenile convictions
+    priors_count           number of prior convictions
+    age_cat_25-45          1 if defendant's age >=25 and <45, otherwise 0
+    age_cat_Greaterthan45  1 if defendant's age >=45, otherwise 0
+    age_cat_Lessthan25     1 if defendant's age <25, otherwise 0
+    race_African-American  1 if defendant's race African-American, otherwise 0
+    race_Caucasian         1 if defendant's race Caucasian, otherwise 0
+    c_charge_degree_F      1 if charged for felony, otherwise 0
+    c_charge_degree_M      1 if charged for misdemeanor, otherwise 0
+    =====================  ======================================================================
+
+    Parameters
+    ----------
+    cache : boolean, default=True
+        Whether to cache downloaded datasets using joblib.
+
+    data_home : optional, default: None
+        Specify another download and cache folder for the datasets. By default
+        all fairlearn data is stored in '~/.fairlearn-data' subfolders.
+
+    as_frame : boolean, default=False
+        If True, the data is a pandas DataFrame including columns with
+        appropriate dtypes (numeric, string or categorical). The target is
+        a pandas DataFrame or Series depending on the number of target_columns.
+        The Bunch will contain a ``frame`` attribute with the target and the
+        data. If ``return_X_y`` is True, then ``(data, target)`` will be pandas
+        DataFrames or Series as describe above.
+
+    return_X_y : boolean, default=False.
+        If True, returns ``(data.data, data.target)`` instead of a Bunch
+        object.
+
+    warn : boolean, default=True.
+        If True, it raises an extra warning to make users aware of the unfairness
+        aspect of this dataset.
+
+
+    Returns
+    -------
+    dataset : :class:`~sklearn.utils.Bunch`
+        Dictionary-like object, with the following attributes.
+
+        data : ndarray, shape (5278, 13)
+            Each row corresponding to the 13 feature values in order.
+            If ``as_frame`` is True, ``data`` is a pandas object.
+        target : numpy array of shape (5278,)
+            Each value indicates whether a defendant was charged again within two years after the
+            current charge.
+            If ``as_frame`` is True, ``target`` is a pandas object.
+        feature_names : list of length 13
+            Array of ordered feature names used in the dataset.
+        DESCR : string
+            Description of the COMPAS dataset.
+
+    (data, target) : tuple of (numpy.ndarray, numpy.ndarray) or (pandas.DataFrame, pandas.Series)
+        if ``return_X_y`` is True and ``as_frame`` is False
+
+    (data, target) : tuple of (pandas.DataFrame, pandas.Series)
+        if ``return_X_y`` is True and ``as_frame`` is True
+
+    See Also
+    --------
+    fairlearn.datasets.fetch_boston: Boston housing dataset (regression)
+    fairlearn.datasets.fetch_adult: UCI Adult dataset (binary classification)
+
+    Notes
+    -----
+    This dataset consists of 5278 samples and 13 features.
+
+    - During pre-processing only instances of defendants of the two most common races were kept,
+      which means that `race_African-American` and `race_Caucasian` are perfectly correlated.
+    - During pre-processing charges for ordinary traffic offenses were removed, which means that
+      c_charge_degree_F` and `c_charge_degree_M` are perfectly correlated.
+    - The features `age_cat_25-45`, `age_cat_Greaterthan45` and `age_cat_Lessthan25` are included
+      to allow linear classifiers to be non-linear in age.
+
+
+    References
+    ----------
+    .. [1] https://github.com/propublica/compas-analysis/
+    .. [2] https://www.openml.org/d/42193
+    .. [3] https://www.propublica.org/article/how-we-analyzed-the-compas-recidivism-algorithm/
+
+    """
+    if warn:
+        msg = "You are about to use a dataset with known fairness issues."
+        warnings.warn(DataFairnessWarning(msg))
+    if not data_home:
+        data_home = pathlib.Path().home() / _DOWNLOAD_DIRECTORY_NAME
+    return fetch_openml(
+        data_id=42193,
+        data_home=data_home,
+        cache=cache,
+        as_frame=as_frame,
+        return_X_y=return_X_y,
+    )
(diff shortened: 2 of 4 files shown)


docs/api_reference/index.rst
```diff
@@ -6,7 +6,6 @@ API Reference
 .. toctree::
    :maxdepth: 1
 
-   fairlearn.datasets
```
romanlutz, line 9: probably unintended?

fairlearn/datasets/_fetch_compas.py
```diff
@@ -0,0 +1,139 @@
...
+        DESCR : string
+            Description of the COMPAS dataset.
+
+    (data, target) : tuple of (numpy.ndarray, numpy.ndarray) or (pandas.DataFrame, pandas.Series)
+        if ``return_X_y`` is True and ``as_frame`` is False
+
+    (data, target) : tuple of (pandas.DataFrame, pandas.Series)
+        if ``return_X_y`` is True and ``as_frame`` is True
+
+    See Also
```
romanlutz, line 104: What's this for? I'm not sure I understand the purpose.

fairlearn/datasets/_fetch_compas.py
```diff
@@ -0,0 +1,139 @@
...
+    --------
+    fairlearn.datasets.fetch_boston: Boston housing dataset (regression)
+    fairlearn.datasets.fetch_adult: UCI Adult dataset (binary classification)
+
+    Notes
+    -----
+    This dataset consists of 5278 samples and 13 features.
+
+    - During pre-processing only instances of defendants of the two most common races were kept,
+      which means that `race_African-American` and `race_Caucasian` are perfectly correlated.
```
romanlutz, line 114: What do you mean by correlated? The first half sentence seems to indicate that it's just two races, so I don't know what that would correlate with.

docs/api_reference/index.rst
hildeweerts: Intended, fairlearn.datasets is currently twice in the index (first and last).

fairlearn/datasets/_fetch_compas.py
hildeweerts: I guess this is just a habit of mine when writing docstrings, but you can think of it as a recommender of "other functions you might like". 

Talking docs, I think they might be easier to read if we move towards overview table + separate pages for each class/function (i.e. like in scikit-learn's docs). What do you think?

fairlearn/datasets/_fetch_compas.py
hildeweerts: I guess my phrasing is not very clear. Because race_African-American and race-Caucasian are the only two possible races in the dataset, these two features have a perfect negative correlation (i.e. -1). Same for c_charge_degreeF and c_charge_degree_M. 

The reason I wrote this down explicitly is that including the redundant features influences the interpretation as well as the stability of the model you build. E.g. AFAIK in linear regression highly correlated features will mess up the variance of the weights resulting in an unstable model.

Alternatively, we can just remove the redundant features from the dataset and very clearly write down what the remaining variables mean; e.g. call it 'race' and explain that 1 means African-American and 0 Caucasian.

hildeweerts:
> I'd like to be more careful about adding COMPAS. If I remember correctly, there were some issues with the ProPublica's data set that were since addressed by some other folks, e.g.:
> https://github.com/beauCoker/age_of_unfairness
> (to accompany https://arxiv.org/pdf/1811.00731.pdf). But I don't remember the severity of the issues--we should review.

If I remember correctly, the main issue put forward by this work is with ProPublica's analysis of the COMPAS scores rather than the collected data. ProPublica assumed the COMPAS scores are linear in age (as stated by the creators) but this paper shows that they seem to be non-linear in age instead which leads to different conclusions about COMPAS scores.

That's not to say we shouldn't be careful! Are you aware of any other caveats we need to review?

If I think about it, calling this dataset COMPAS might even be a bit misleading given that the actual scores are not even part of the data.

MiroDudik:
The issue might have been partly about the meaning of the columns of the ProPublica datasets `*.csv` and how exactly they were generated from `compas.db`. Do we want to do any reproducibility check? Alternatively, we could just call the dataset `fetch_propublica_compas`.

Re. caveats, there was this rejoinder that provides a lot of context on how the risk assessment tools are used. We should use it to double-check the choice of the classification label as well as the subset of examples that should be used for training:
https://www.uscourts.gov/federal-probation-journal/2016/09/false-positives-false-negatives-and-false-analyses-rejoinder

docs/api_reference/index.rst
romanlutz: Ah, diff view doesn't show that!

fairlearn/datasets/_fetch_compas.py
romanlutz: Definitely prefer overview table! This approach won't scale long-term :-)
It's fine if we do that in a separate PR, though, I'm not too particular about it.

fairlearn/datasets/_fetch_compas.py
romanlutz: Got it! I somehow interpreted this as `race_African-American` being one value of the race column and `race-Caucasian` being another, but they're columns themselves with binary values. This makes sense. If you want you can add that disclaimer about redundancy. I think it's useful!

romanlutz:
> The issue might have been partly about the meaning of the columns of the ProPublica datasets `*.csv` and how exactly they were generated from `compas.db`. Please see the top of page 41 of [Rudin-Wang-Coker](https://arxiv.org/pdf/1811.00731.pdf):
> 
> > Note that while ProPublica publishes the code for their analysis and the raw data, they do not publish the code for processing the raw data. Thus we did that from scratch.
> 
> If we are comfortable with that caveat, we could possibly rename the dataset to something like `fetch_propublica_compas`.
> 
> Re. caveats, there was [this rejoinder](https://www.uscourts.gov/federal-probation-journal/2016/09/false-positives-false-negatives-and-false-analyses-rejoinder) that provides a lot of context on how the risk assessment tools are used. We should use it to double-check the choice of the classification labels as well as the restriction of examples that should be used for training.

I'm very much on board with naming it something more specific as indicated by @MiroDudik . The fact that it's a scenario that perhaps shouldn't be handled by an AI system is not a reason against including it IMO, for the same reasons we discussed with the Boston dataset before. We can document this and illustrate using an example notebook why it's not a good idea (if that's the conclusion). If anything that's a great reason to include it.

hildeweerts:
If we assume people will only use this data for educational purposes (like with the Boston housing dataset) I wouldn't worry too much about the caveats with ProPublica's data generation. On the other hand, the transparent approach taken by Rudin-Wang-Coker is more in line with the values of Fairlearn, so we can also opt to use their data instead.

fairlearn/datasets/_fetch_compas.py
riedgar-ms: Given the instability you mentioned, is there a particular reason for keeping these columns (rather than a single `race` column)? I agree that 'redundant' makes more sense than 'correlated' in this context.

hildeweerts: One reason may be to avoid confusion with what the column means if people don't read the documentation. But I'm also on board with removing one of them.

riedgar-ms: I'm sure people will manage to confuse themselves with or without documentation :-)

I suppose that one advantage of keeping this 'one-hot-encoding' model even for the binary feature is that it avoids the need to decide which group gets 0 and which gets 1. I'd like to understand that potential instability a little better; I am concerned that if the dataset is tricky to train a model on in the first place, users could get lost in those weeds, and think of the fairness issues in those terms, rather than the data themselves containing bias.

fairlearn/datasets/_fetch_compas.py
```diff
@@ -0,0 +1,139 @@
...
+    .. [3] https://www.propublica.org/article/how-we-analyzed-the-compas-recidivism-algorithm/
+
+    """
+    if warn:
+        msg = "You are about to use a dataset with known fairness issues."
+        warnings.warn(DataFairnessWarning(msg))
+    if not data_home:
+        data_home = pathlib.Path().home() / _DOWNLOAD_DIRECTORY_NAME
+    return fetch_openml(
+        data_id=42193,
```
riedgar-ms, line 134: Is the other version of this dataset mentioned in the discussion also part of openml, or would that require new download code?

hildeweerts: Not that I'm aware, but if we manage to get it in the right format we can just upload it :)

fairlearn/datasets/_fetch_compas.py
hildeweerts: It's not a huge issue necessarily. But what will happen is that for most ML models the usage of either `race_African-American`, `race-Caucasian`, or 'a little bit of both' will be arbitrary. Especially in the latter case, this might lead to confusion when interpreting the importance of features, e.g. in terms of the coefficients of a logistic regression model.

kevinrobinson:
If the idea is to use this for examples or educational materials then it would be great to hear thoughts on that (particularly in terms of a [real deployment context and real harms](https://fairlearn.github.io/contributor_guide/contributing_example_notebooks.html)).  It's already super exciting to see the discussion here about escaping the framing trap that comes with focusing on the dataset itself :)  There are highlights from several other references in https://github.com/fairlearn/fairlearn/issues/478 that might be helpful.  

As with the Boston housing dataset, if other folks disagree that's cool and I'll step back and let you all decide what makes sense from here.  Thanks for listening! 😄

fairlearn/datasets/_fetch_compas.py
romanlutz: The referenced one was actually uploaded by me :-D
I remember it being a huge pain since I had to convert it to some obscure format.

fairlearn/datasets/_fetch_compas.py
romanlutz: > If the idea is to use this for examples or educational materials then it would be great to hear thoughts on that (particularly in terms of a [real deployment context and real harms](https://fairlearn.github.io/contributor_guide/contributing_example_notebooks.html)). It's already super exciting to see the discussion here about escaping the framing trap that comes with focusing on the dataset itself :) There are highlights from several other references in #478 that might be helpful.
> 
> As with the Boston housing dataset, if other folks disagree that's cool and I'll step back and let you all decide what makes sense from here. Thanks for listening! 😄
> 
> EDIT: I also just saw #513 too, which says:
> 
> > Removing this accordingly until we
> > (a) figure out how to capture the sociotechnical context properly OR
> > (b) conclude that this is an area where AI systems should not be used at all and document that conclusion.
> > We should definitely avoid the solutionism trap from Selbst et al.'s paper "Fairness and Abstraction in Sociotechnical Systems".
> 
> Maybe that discussion has already happened offline?

@kevinrobinson there were no further discussions. @hildeweerts is just adding datasets, that's all :-)
Just didn't feel very good about having that notebook there, and the fact that people seem to be taking notebooks as best practices makes me feel very good about having removed it.

hildeweerts:
@kevinrobinson I was under the impression that it would be used for educational resources but I actually don't know if anybody is currently working on this. I think writing a notebook with a clear deployment context will also help in making decisions on the appropriateness of the target variable and other features. I'm okay with putting this on hold until then. I don't know if @romanlutz had other purposes for this dataset in mind that require it to be added right away?

romanlutz:
> @kevinrobinson I was under the impression that it would be used for educational resources but I actually don't know if anybody is currently working on this. I think writing a notebook with a clear deployment context will also help in making decisions on the appropriateness of the target variable and other features. I'm okay with putting this on hold until then. I don't know if @romanlutz had other purposes for this dataset in mind that require it to be added right away?

I didn't have anything specific in mind yet. The notebook that used it was recently removed.
I'm certainly interested in educational materials on this, but there's a lot of literature around COMPAS so I expect a certain ramp up time for whoever wants to do this.

fairlearn/datasets/_fetch_compas.py
MiroDudik: My suggestion: let's keep both columns, but update the description:
* perfectly correlated -> perfectly anticorrelated

fairlearn/datasets/_fetch_compas.py
```diff
@@ -0,0 +1,139 @@
...
+    fairlearn.datasets.fetch_adult: UCI Adult dataset (binary classification)
+
+    Notes
+    -----
+    This dataset consists of 5278 samples and 13 features.
+
+    - During pre-processing only instances of defendants of the two most common races were kept,
+      which means that `race_African-American` and `race_Caucasian` are perfectly correlated.
+    - During pre-processing charges for ordinary traffic offenses were removed, which means that
+      c_charge_degree_F` and `c_charge_degree_M` are perfectly correlated.
```
MiroDudik, line 116: Similarly as above:
* perfectly correlated -> perfectly anticorrelated

fairlearn/datasets/_fetch_compas.py
```diff
@@ -0,0 +1,139 @@
...
+import pathlib
+import warnings
+
+from sklearn.datasets import fetch_openml
+from ._constants import _DOWNLOAD_DIRECTORY_NAME
+
+from fairlearn.exceptions import DataFairnessWarning
+
+
+def fetch_compas(*, cache=True, data_home=None,
```
MiroDudik, line 13: As per the conversation elsewhere, let's change this:
* fetch_compas -> fetch_compas_propublica
(and similarly rename the file)

PR #526, opened 2020-07-22
Modify the logic for making copies of estimators using the `sklearn.base.clone` method.

hannanabdul55:
## What's there in this PR? 
- Use [`sklearn.base.clone`](https://github.com/scikit-learn/scikit-learn/blob/fd237278e/sklearn/base.py#L48) to make copies of estimators.


fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -149,7 +150,10 @@ def _call_oracle(self, lambda_vec):
                                          constant=redY_unique[0])
             self.n_oracle_calls_dummy_returned += 1
         else:
-            classifier = pickle.loads(self.pickled_estimator)
+            # use sklearn.base.clone to clone the estimator.
+            # Use safe=False to make a deep copy of the estimator if not an
+            # sklearn-based estimator.
```
riedgar-ms, line 155: Relieved to see that this will work with non-sklearn estimators. Can you let us know what this offers over `copy.deepcopy` (which would have been my first guess for an improvement over pickling)?

notebooks/tfco-fairlearn-experiment.ipynb
```diff
@@ -0,0 +1,414 @@
+{
```
romanlutz, line 1: please remove :-)

notebooks/tfco-fairlearn-experiment.ipynb
```diff
@@ -0,0 +1,414 @@
...
+ "cells": [
```
riedgar-ms, line 2: Can this be regular Python test files, instead of a notebook?

notebooks/tfco-fairlearn-experiment.ipynb
romanlutz: Oh this was intentional as I see now from the description. Hmm, I guess I have several reasons not to include it here at this point:

- new notebooks should go under examples/notebooks in .py format
- this introduces tensorflow as a dependency which we're definitely not going to add, so it would have to run in a separate pipeline with its own requirements.
- it's a separate contribution, best to keep the estimator copying as its own PR

Does that make sense?

hannanabdul55: Yes, I think this makes sense. maybe later we can have one consolidated PR with all the notebooks. I will remove this from here.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
hannanabdul55: It is actually just the same as `copy.deepcopy` for non-sklearn estimators. But for sklearn estimators, it would just be more efficient as it would not really do a complete `copy.deepcopy` for them. It would just instantiate a new estimator using the `get_params()` parameters internally. The [implementation](https://github.com/scikit-learn/scikit-learn/blob/fd237278e/sklearn/base.py#L48) of `sklearn.base.clone` will make it more clear.

notebooks/tfco-fairlearn-experiment.ipynb
hannanabdul55: I am going to remove this notebook in this PR as it is not pertaining to this implementation.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
riedgar-ms: Could you put that explanation in the description? I'm all for getting rid of the pickling.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -149,7 +149,12 @@ def _call_oracle(self, lambda_vec):
                                          constant=redY_unique[0])
             self.n_oracle_calls_dummy_returned += 1
         else:
-            classifier = pickle.loads(self.pickled_estimator)
+            # use sklearn.base.clone to clone the estimator.
+            # It is the same as copy.deepcopy for non-sklearn estimators (By using safe=False).
+            # For sklearn estimators, it is more efficient as it would not
+            # do a copy.deepcopy. Rather, instantiate a new estimator using the
+            # get_params() internally.
+            classifier = clone(estimator=self.estimator, safe=False)
```
romanlutz, line 157: Please rename this to estimator while you're at it. Classifier indicates that this can't cover other tasks, which I've tried to fix with renaming before, but evidently I missed this one.

test/unit/reductions/exponentiated_gradient/test_exponentiatedgradient_arguments.py
```diff
@@ -122,7 +122,7 @@ def Q(X): return expgrad._pmf_predict(X)[:, 1]
     @pytest.mark.parametrize("transformX", candidate_X_transforms)
     def test_input_X_unchanged(self, transformA, transformY, transformX, mocker):
         # The purpose of this test is to ensure that X is passed to the underlying estimator
-        # unchanged. For y and sample_weight ExponentiatedGradient makes certain transformations
+        # unchanged. For y and sample_weight ExponentiatedGradient makews certain transformations
```
romanlutz, line 125: probably unintended? Please undo

test/unit/reductions/exponentiated_gradient/test_exponentiatedgradient_arguments.py
```diff
@@ -132,30 +132,24 @@ def test_input_X_unchanged(self, transformA, transformY, transformX, mocker):
...
-        estimator = mocker.MagicMock()
-        estimator.predict = mocker.MagicMock(return_value=y)
-        # ExponentiatedGradient pickles and unpickles the estimator, which isn't possible for the
-        # mock object, so we patch import of pickle as well. It sets the result from pickle.loads
-        # as the estimator, so we can simply overwrite the return value to be our mocked estimator
-        # object.
-        mocker.patch('pickle.dumps')
-        mocker.patch('pickle.loads', return_value=estimator)
-
+        estimator = LeastSquaresBinaryClassifierLearner()
```
romanlutz, line 135: Wow I'm glad this logic is therefore gone! Thank you!

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
romanlutz: Also please add a note to CHANGES.md to indicate that we're using this instead of pickle.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
hannanabdul55: Sure 👍

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -123,7 +122,8 @@ def solve_linprog(self, nu):
         dual_c = np.concatenate((b_ub, -b_eq))
         dual_A_ub = np.concatenate((-A_ub.transpose(), A_eq.transpose()), axis=1)
         dual_b_ub = c
-        dual_bounds = [(None, None) if i == n_constraints else (0, None) for i in range(n_constraints + 1)]  # noqa: E501
+        dual_bounds = [(None, None) if i == n_constraints else (0, None) for i in
```
riedgar-ms, line 125: Was this split intentional?

hannanabdul55: Oops, I think my IDE automatically did this.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -155,21 +155,26 @@ def _call_oracle(self, lambda_vec):
...
         oracle_call_start_time = time()
-        classifier.fit(self.X, redY, sample_weight=redW)
+        estimator.fit(self.X, redY, sample_weight=redW)
         self.oracle_execution_times.append(time() - oracle_call_start_time)
         self.n_oracle_calls += 1
 
-        return classifier
+        return estimator
 
     def best_h(self, lambda_vec):
```
adrinjalali, line 179: should this (and other methods) also call `check_is_fitted`? Or do we rely on the underlying estimator to do that?

test/unit/reductions/exponentiated_gradient/simple_learners.py
```diff
@@ -34,3 +34,19 @@ def fit(self, X, Y, sample_weight):
...
+
+class MockEstimator:
+    def fit(self, X, y, sample_weight):
+        pass
+
+    def predict(self, X):
+        return np.ones(X.shape[0])
+
+
+class MockSKLearnEstimator(MockEstimator):
```
adrinjalali, line 47: Not sure why you need these, but you probably want to inherit from `BaseEstimator` if this is supposed to be `sklearn` compatible.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
hannanabdul55: I think the `_call_oracle` method calls `fit` here. Are you referring to invoking the `check_is_fitted` before that? I think we rely on the underlying estimator for this AFAIK. @romanlutz @riedgar-ms would know more.

test/unit/reductions/exponentiated_gradient/simple_learners.py
hannanabdul55: Yes, thanks for pointing this out. Will just remove this dummy class and use `BaseEstimator` directly!

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
romanlutz: @adrinjalali we're calling fit right before that. Any particular reason to call the check method here? I'm probably missing something.

adrinjalali: yeah, sorry, my bad. Ignore this one :D

romanlutz:
@MiroDudik unless you have comments let's complete this PR :-)

PR #527, opened 2020-07-22
Add missing over-/underprediction figures to dashboard to prevent it from crashing for regression tasks

chisingh:
Fixing the crash on regressions due to missing overprediction/underprediction figures.


romanlutz:
That code doesn't live here anymore, closing the PR

PR #531, opened 2020-07-27
Add GitHub Stats Page

riedgar-ms:
Add a simple GitHub stats page to the documentation. For now, this just creates a plots of stars over time.

To see the plots from a forked repo, we need to allow forks access to secrets in the "Advanced Settings" in CircleCI. We probably don't want that to be left on.


.circleci/config.yml
```diff
@@ -108,3 +108,19 @@ workflows:
           filters:
             branches:
               only: master  # don't deploy to GitHub pages from other branches
+
+  weekly-webpage:
```
romanlutz, line 112: doesn't this need to be listed under jobs?

docs/about/github_stars.py
```diff
@@ -0,0 +1,147 @@
+from bokeh.plotting import figure, show
```
romanlutz, line 1: copyright header missing

docs/about/github_stars.py
```diff
@@ -0,0 +1,147 @@
...
+import datetime
+from dateutil import relativedelta
+from dateutil.parser import isoparse
+import os
+
+from gql import gql, Client
+from gql.transport.requests import RequestsHTTPTransport
+
+_SECRET_ENV_VAR = "GITHUB_GRAPHQL"
```
romanlutz, line 10: Where did you get this from? Do we have an account or something?

requirements.txt
```diff
@@ -38,3 +38,7 @@ nbval
 sphinx
 sphinx-gallery
 pydata-sphinx-theme
+
+bokeh
```
romanlutz, line 42: This looks like it'll need an update after the split

docs/about/github_stars.py
```diff
@@ -0,0 +1,147 @@
...
+
+    show(p)
+
+
+print("Starting script {0}".format(__file__))
+if _SECRET_ENV_VAR in os.environ:
+    github_star_gazing()
+else:
+    print("Did not find environment variable {0}".format(_SECRET_ENV_VAR))
+    create_no_fetch_data_plot()
```
romanlutz, line 146: shouldn't this fail instead? 
Should we have retries?

docs/about/github_stars.py
riedgar-ms: It's using a token from my account. Set with only a few privileges, but not ideal.

requirements.txt
riedgar-ms: Good point....

docs/about/github_stars.py
riedgar-ms: The problem is that token. To get the sample webpage showing, I had to click on the "Allow secrets in builds for forks" option.... which is a security hole. So on forks, this script will fail (since I didn't leave that option on) without something like this.

.circleci/config.yml
riedgar-ms: It has jobs of its own, so.... how?

riedgar-ms: I suppose it might be simpler to add a scheduler to the existing job. The `doctest` part wouldn't be needed, but that might be better than having so much extra configuration which is largely a duplicate.

requirements.txt
riedgar-ms: Done

docs/about/github_stars.py
riedgar-ms: Fixed

riedgar-ms:
Latest rendering of the plot:
https://1020-133444044-gh.circle-artifacts.com/0/docs/_build/html/about/github-stats.html

docs/about/github_stars.py
MiroDudik: Our new header format should be:
```
```

Which reminds me--we should propagate this to the rest of the repository and say that that's our format.

docs/about/github_stars.py
MiroDudik: Ooh... this seems like a very opaque dependence, but I might not understand this right.

MiroDudik:
> Latest rendering of the plot:
> https://1020-133444044-gh.circle-artifacts.com/0/docs/_build/html/about/github-stats.html

I think that we should not show the current (incomplete) month at all, only the months that are complete. Otherwise the values in the columns are not comparable.

riedgar-ms:
I don't think this is needed

PR #533, opened 2020-07-28
Added unit tests for fetch functions in datasets package

gaugup:
Signed-off-by: gaugup <gaugup@microsoft.com>


test/unit/datasets/test_datasets.py
```diff
@@ -0,0 +1,57 @@
...
+import numpy as np
+
+from fairlearn.datasets import fetch_adult, fetch_boston
+
+# =============================================
+
+
+class TestFairlearnDataset:
+    def test_dataset_as_bunch_in_pandas_format(self):
+        fetch_function_list = [fetch_adult, fetch_boston]
```
riedgar-ms, line 14: Could you make these into [data-driven tests](https://docs.pytest.org/en/stable/parametrize.html), so that an early failure won't prevent testing of the other `fetch` functions.

test/unit/datasets/test_datasets.py
```diff
@@ -0,0 +1,57 @@
...
+            assert dataset['data'].shape is not None
+            assert isinstance(dataset['data'], np.ndarray)
+            assert dataset['target'].shape is not None
+            assert isinstance(dataset['target'], np.ndarray)
+            assert dataset['feature_names'] is not None
+            assert isinstance(dataset['feature_names'], list)
+            assert dataset['DESCR'] is not None
+            assert isinstance(dataset['DESCR'], str)
+
+    def test_dataset_as_data_target_in_pandas_format(self):
```
riedgar-ms, line 41: The name is not immediately making me realise that the key difference is `return_X_y=True`

gaugup: Changed the function names accordingly.

test/unit/datasets/test_datasets.py
gaugup: Thanks for the suggestion. Added the parameterization via pytest.

romanlutz: I think @riedgar-ms meant that fetch_adult and fetch_boston should also be parametrized.

riedgar-ms: Yes, that's actually what I was thinking of, although keeping the "as_frame" parameterisation also looks like it makes sense. An example of parameterising over a list of functions:
https://github.com/fairlearn/fairlearn/blob/8f087fbb0b27740d10b31d95706bb175a4b4581c/test/unit/metrics/test_metrics_engine.py#L31

Issue #466, opened 2020-06-08
Proposal: Port EqualOpportunityClassifier and DemographicParityClassifier


MBrouns:
I'm making this issue so that we can disucss how to move the `EqualOpportunityClassifier` and `DemographicParityClassifier from scikit-fairness to this project.

The methods are discussed in a notebook here: https://scikit-lego.readthedocs.io/en/latest/fairness.html

The `DemographicParityClassifier` is an implementation of the method described in M. Zafar et al. (2017), Fairness Constraints: Mechanisms for Fair Classification. The optimization itself is done using `cvxpy`. The `EqualOpportunityClassifier` is not described in a paper afaik, but is a simple modification on the constraint of the `DemographicParityClassifier`

Is this something you'd be open to having in this repo? If so, I can start working on a PR.

riedgar-ms:
I'll defer to @MiroDudik  on the question of the algorithm itself (but I expect a positive response - we'd like a richer library!), but I'm not a huge fan of the name. We've tended to have the name of the technique be the class name, and then specify the constraint as an argument. This is obviously true for things like reductions, but we do something similar for ThresholdOptimizer.

What sub-package were you thinking of using (creating, I expect)?

kevinrobinson:
It seems awesome to merge efforts! 👍   It'd be super helpful to understand why this approach is valuable, particularly in comparison to what is already in this repository, and other contributions being discussed.  If there's a way to show this in an example scenario or notebook, that'd be amazing.

re: fairness.html, I've shared some feedback on the fairness.html example already over in https://github.com/koaning/scikit-fairness/issues/31, and a specific critique https://github.com/fairlearn/fairlearn/issues/430#issuecomment-632690069, but to recap: if you are trying to predict the median house price in a census tract, I don't understand why you would want to enforce that those predictions are uncorrelated with information about the income of people living in that census tract.  It seems like this is what the [Demographic Parity Classifier](https://scikit-lego.readthedocs.io/en/latest/fairness.html#Demographic-Parity-Classifier) and [Equal Opportunity](https://scikit-lego.readthedocs.io/en/latest/fairness.html#Equal-opportunity) examples are doing, so I'd love to understand more about why someone would use this technique, or about how this would mitigate any real harms.

[...]

PR #536, opened 2020-07-30
Add bank marketing dataset to fairlearn

gaugup:
Signed-off-by: Gaurav Gupta <gaugup@microsoft.com>

### DIFF
--- a/fairlearn/datasets/__init__.py
+++ b/fairlearn/datasets/__init__.py
@@ -7,8 +7,10 @@
 
 from ._fetch_adult import fetch_adult
 from ._fetch_boston import fetch_boston
+from ._fetch_bank_marketing import fetch_bank_marketing
 
 __all__ = [
     "fetch_adult",
     "fetch_boston",
+    "fetch_bank_marketing",
 ]
--- a/fairlearn/datasets/_fetch_bank_marketing.py
+++ b/fairlearn/datasets/_fetch_bank_marketing.py
@@ -9,7 +9,7 @@ from ._constants import _DOWNLOAD_DIRECTORY_NAME
 
 def fetch_bank_marketing(*, cache=True, data_home=None,
                          as_frame=False, return_X_y=False):
-    """Load the UCI bank marketing dataset (binary classification)
+    """Load the UCI bank marketing dataset (binary classification).
 
     Download it if necessary.
 
--- a/fairlearn/datasets/_fetch_bank_marketing.py
+++ b/fairlearn/datasets/_fetch_bank_marketing.py
@@ -1,0 +1,87 @@
+# Copyright (c) Microsoft Corporation and Fairlearn contributors.
+# Licensed under the MIT License.
+
+import pathlib
+
+from sklearn.datasets import fetch_openml
+from ._constants import _DOWNLOAD_DIRECTORY_NAME
+
+
+def fetch_bank_marketing(*, cache=True, data_home=None,
+                         as_frame=False, return_X_y=False):
+    """Load the UCI bank marketing dataset (binary classification)
+
+    Download it if necessary.
+
+    ==============   ====================
+    Samples total                   45211
+    Dimensionality                     17
+    Features         numeric, categorical
+    Classes                             2
+    ==============   ====================
+
+    Source: http://archive.ics.uci.edu/ml/datasets/Bank+Marketing
+    The data is related with direct marketing campaigns of a Portuguese
+    banking institution. The marketing campaigns were based on phone calls.
+    Often, more than one contact to the same client was required,
+    in order to access if the product (bank term deposit) would be (or not) subscribed.
+
+    The classification goal is to predict if the client will subscribe a
+    term deposit (variable y).
+
+
+    Parameters
+    ----------
+    cache : boolean, default=True
+        Whether to cache downloaded datasets using joblib.
+
+    data_home : optional, default: None
+        Specify another download and cache folder for the datasets. By default
+        all scikit-learn data is stored in '~/.fairlearn-data' subfolders.
+
+    as_frame : boolean, default=False
+        If True, the data is a pandas DataFrame including columns with
+        appropriate dtypes (numeric, string or categorical). The target is
+        a pandas DataFrame or Series depending on the number of target_columns.
+        The Bunch will contain a ``frame`` attribute with the target and the
+        data. If ``return_X_y`` is True, then ``(data, target)`` will be pandas
+        DataFrames or Series as describe above.
+
+    return_X_y : boolean, default=False.
+        If True, returns ``(data.data, data.target)`` instead of a Bunch
+        object.
+
+    Returns
+    -------
+    dataset : :class:`~sklearn.utils.Bunch`
+        Dictionary-like object, with the following attributes.
+
+        data : ndarray, shape (48842, 14)
+            Each row corresponding to the 14 feature values in order.
+            If ``as_frame`` is True, ``data`` is a pandas object.
+        target : numpy array of shape (48842,)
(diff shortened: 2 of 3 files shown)


fairlearn/datasets/__init__.py
```diff
@@ -7,8 +7,10 @@
 
 from ._fetch_adult import fetch_adult
 from ._fetch_boston import fetch_boston
+from ._fetch_bank_marketing import fetch_bank_marketing
 
 __all__ = [
     "fetch_adult",
     "fetch_boston",
+    "fetch_bank_marketing",
```
riedgar-ms, line 15: Can you add this to the dataset tests?

gaugup: Yes, will do after https://github.com/fairlearn/fairlearn/pull/533 merges as this one has the tests.

riedgar-ms: I merged #533

gaugup: Added the tests

fairlearn/datasets/_fetch_bank_marketing.py
```diff
@@ -0,0 +1,87 @@
...
+    Classes                             2
+    ==============   ====================
+
+    Source: http://archive.ics.uci.edu/ml/datasets/Bank+Marketing
+    The data is related with direct marketing campaigns of a Portuguese
+    banking institution. The marketing campaigns were based on phone calls.
+    Often, more than one contact to the same client was required,
+    in order to access if the product (bank term deposit) would be (or not) subscribed.
+
+    The classification goal is to predict if the client will subscribe a
```
kevinrobinson, line 29: If the goal here is to use this for an example, there's some older comments on this here: https://github.com/fairlearn/fairlearn/issues/466#issuecomment-640852091, and I'll paste the relevant bit below here:

---
To try to help brainstorm another idea, I noticed that the Zafar paper runs its experiments on the UCI adult income dataset, and a "bank dataset."  Perhaps it might be helpful to explore if the original paper has example tasks that can show the value of these approaches?  While the URL in the paper to the bank dataset is broken, at first glance it seems like it refers to [a scenario](https://archive.ics.uci.edu/ml/datasets/Bank+Marketing) where a bank is trying to predict the success of a telemarking phone call campaign to get people to purchase 'term deposits' like bonds.  The paper discretizes age to "between 25 and 60 or not" and then shows the technique applied to a binary classifier.  I'm happy to help brainstorm but unfortunately it's also hard for me to understand how enforcing demographic parity on predictions about the success of a telemarketing campaign contributes to mitigating real harms.  I also don't understand paper's references to "disparate impact" in the context of marketing campaigns (eg, the company is deciding how many times to call each person and each row in the "bank dataset" encodes the bank's salespeople's decisions about how many times to call that home).  EDIT: Here's the references to the two papers on the UCI site for the "bank dataset": [A Data-Driven Approach to Predict the Success of Bank Telemarketing (Moro et al. 2014](http://repositorium.sdum.uminho.pt/bitstream/1822/30994/1/dss-v3.pdf) and [Using Data Mining for Bank Direct Marketing: An Application of the CRISP-DM Methodology (Moro et al. 2014)](https://repositorium.sdum.uminho.pt/bitstream/1822/14838/1/MoroCortezLaureano_DMApproach4DirectMKT.pdf).  In the second paper, they also describe how they are not using the dataset to train a model for prediction:

> The business goal is to find a model that can explain success of a contact, i.e. if the   client   subscribes   the   deposit.   Such   model can increase campaign efficiency by identifying the   main characteristics that affect success, helping in a better management of the available resources (e.g. human effort, phone  calls,  time) and selection of a high quality and affordable set of potential buying customers.

To be super clear, I'd love to see this kind of merging and collaboration, and I'm just a curious outsider :)  So I'm trying to share this feedback and asking these kinds of questions in that same spirit 😄 

---

If the intention is to use this for a kind of example, it'd be awesome to understand how this work can lead to the kinds of elements in [Contributing example notebooks](https://fairlearn.github.io/contributor_guide/contributing_example_notebooks.html).  Thanks! 👍

hildeweerts: I'm also wondering what's the use of adding data sets if there are no active efforts to create examples based on these data sets.

romanlutz: This is a dataset that's commonly used in the literature. As I've pointed out on various occasions I'm not super happy with some datasets used in the literature including the UCI adult one because it's not always super clear what the application context is related to fairness.

There seems to be a perception that adding a dataset means we're endorsing every use of the dataset, but that's not the case. This seems to keep coming up with every dataset that's added and I'm not sure whether there's general opposition to the idea of a dataset module or something else. Perhaps it needs more documentation in the user guide. If that's the core issue here we can open an issue for that.

In this particular case I have an application in mind that may not happen anytime soon. I'm in the process of writing a proposal but these things take time. Gaurav asked for a simple task to get started with the dev process, PRs on GitHub etc. Typically that means low priority items, and one of those is removing tempeh as a dependency for tests. To get closer to that we have to provide datasets from other sources such as openml. The one outstanding one (SEAPHE) is harder to replace since it's not in openml yet, so we started with something simpler, i.e. adding a different dataset that's commonly used in the literature.

hildeweerts: I fear my comment may have come across stronger than I intended. I am definitely not opposed to a dataset module! I just would prefer having fewer datasets that are well documented and used in examples or notebooks, as opposed to a bunch of datasets that are not really used anywhere in Fairlearn. I think the former approach aligns better with Fairlearn's values. I also think an advantage of Fairlearn is that data sets do not need to be transformed into some peculiar data structure. So having easy access to a bunch of datasets is not as crucial as it is for some other libraries, because it should be easy to get started with a data set of your own.

I totally get the rationale for a simple task to get started, though! We might want to put in some extra effort to add more 'good first issue'-s. I believe that fixing things in the documentation is usually seen as a good place to start, but perhaps we can come up with other things as well.

kevinrobinson: The first step of [Contributing example notebooks](https://fairlearn.github.io/contributor_guide/contributing_example_notebooks.html) says:

> **Deployment context**: Describes a real deployment context, not just a dataset.

That's why I've asked questions or repeated past critiques when there have been pull requests with no description that are just adding a dataset :)

romanlutz: Exactly my point: these are datasets, not deployment contexts. They don't tell you everything you need to know. That's why we're putting in all that effort into notebooks.

Definitely agree on better first issues. That's something we've been lacking.

Nothing in this module is really crucial as such, because evidently we can replace it with 1 line from sklearn (so far at least). To me this is a philosophical question on whether or not one wants a dataset module. Which datasets are in there doesn't really say anything to me at all, unless they're used in an irresponsible way. I realize there can be different opinions on this. I vaguely recall being part of the opposition of a dataset module in fact. Now that we have one (which we can totally revisit!) what's the harm in adding datasets that occur in the fairness literature, even if that ends up being for no other reason than pointing out that we don't want to use this for reasons X, Y, and Z? If the concerns are so strong I'd much rather get rid of the whole module and use the single line we're referencing when needed. Perhaps something to discuss in the next developer call so that we can get everyone's thoughts (?)

Unrelated: In general I believe in small PRs, so if splitting dataset additions into PRs makes it easier to review I'm all for it. That also gets smaller items like adding the dataset out of the way.

kevinrobinson:
@romanlutz I pinged you in chat; I think we are talking past each other and I'm not sure how else to bridge the gap at this moment, so will leave it be :)

PR #550, opened 2020-08-11
[WIP] Versioned documentation

riedgar-ms:
An in-progress experiment in creating versioned documentation


riedgar-ms:
I think the conclusion is that this is not the preferred pattern.

PR #551, opened 2020-08-11
Versioned documentation

riedgar-ms:
Use `sphinx-multiversion` to build versioned documentation. Unfortunately, the version picker is unable to cope with files which were renamed between different documentation versions.

Splits the static landing page off from the Sphinx documentation, and copies it into place as part of the overall build pipeline. The landing page points at the documentation for `master` for now. Once we have a new release, we should consider making the landing page point at the current PyPI version (the v0.4.6 docs are too sparse for that to be a good option at the current time).


kevinrobinson:
@riedgar-ms Given what I understand here, my vote would be to ship this immediately 👍 along with the manual build of v0.4.6 and a bullet-point doc of "how do I release a new version."  That being said, I'm not entirely sure I understand the full impact to the end user.  Apologies if I am missing this from the other docs in the proposals repo.

If someone goes to https://fairlearn.github.io/quickstart.html, what version of the docs would they see?  My understanding is that we'd want it to default to the last release, with links to jump to master or to older versions.  That way someone discovers the project, looks at the website, tries that code, and it works as expected.  FWIW, my vote is to fix that issue even if means accepting visual style regressions.  My understanding is that this would default to showing `master`, and so the root usability problem wouldn't be addressed.  Am I following?

Finally, it seems that if we want to ship this next we, it'd be good to cut a 0.4.7 release as well.  I can't tell what's needed to do that, but is that a prerequisite to using this approach to address the root usability issue?  Or do folks need to unblock https://github.com/fairlearn/fairlearn-proposals/pull/11 first?  I can't tell what's needed to address the root user issue, but would vote to punt everything else that is more forward-looking to get this shipped so people can type in the code on the website, and the library works as expected. 🚢

riedgar-ms:
@kevinrobinson All pages except the landing page will move under a directory corresponding to their branch. So anyone going to https://fairlearn.github.io/quickstart.html will find themselves contemplating a 404. The links from the landing page will point to the version from `master`. All pages would acquire the header inviting users to select their version, but it would still be a potential trap.

The PR I just put out may also remove the incompatibility preventing `sphinx-multiversion` from building the v0.4.6 docs... however, I don't want to pile multiversioning into that one as well. The change is worth making anyway.

kevinrobinson:
@riedgar-ms I see, thanks!  I'm assuming that fixing the first-time user experience is the top priority.  But if I am following you, this PR would still leave first time users with a bad experience whenever master has changes from the latest release.  Am I following?

riedgar-ms:
@kevinrobinson Yes, users would have to realise that they needed to look at a different documentation version. This PR is just about making sure that that documentation version is at least available.

If I can get this PR to build v0.4.6 as well, then we could change the main landing page to point at that documentation instead of `master`. Whatever you do though, _someone_ is going to get caught at some point (because people might clone and `pip install.` as well). Ultimately, minimising these issues will require us to stop breaking stuff so much.

adrinjalali:
Ideally, the website by default forwards to the documentation for the latest released version and not `master`. This way users see a documentation which corresponds to what they have if they've just installed the package from `pypi` or if they upgrade their package to the latest version. This usually minimizes confusion for users.

docs/static_landing_page/index.html
```diff
@@ -8,29 +8,29 @@
...
-      		<img class="img-fluid nav-logo" src="_static/images/fairlearn_full_color.png" alt="">
+      		<img class="img-fluid nav-logo" src="./images/fairlearn_full_color.png" alt="">
       		<button class="navbar-toggler navbar-light" type="button" data-toggle="collapse" data-target="#navbarResponsive" aria-controls="navbarResponsive" aria-expanded="false" aria-label="Toggle navigation">
         	<span class="navbar-toggler-icon icon-bar"></span>
       		</button>
       		<div class="collapse navbar-collapse" id="navbarResponsive">
 		        <ul class="navbar-nav ml-auto text-center">
 		          <li class="nav-item pr-0 pr-lg-4">
-		            <a class="nav-link" href="https://fairlearn.github.io/user_guide/index.html">User Guide</a>
+		            <a class="nav-link" href="./master/user_guide/index.html">User Guide</a>
```
riedgar-ms, line 30: Links like this would need to be updated to point to a different version as the default.

riedgar-ms:
@adrinjalali , point taken about the preference to default to the current version on PyPI. However, for right now, I can't get the `v0.4.6` documentation to build (too many incompatible changes to our `conf.py`). I can add that as a static directory in a subsequent PR, and update the links.

If anyone has any ideas as to how to fix up the layout of the version selector, please let me know. As noted in the comments, this has been noted as a defect by `pydata-sphinx-theme` (they don't have a hook to customise their sidebars), but there may be a way to do something about it within Sphinx itself.

.circleci/config.yml
```diff
@@ -26,7 +26,15 @@ jobs:
 
       - run:
           name: 'Build html'
-          command: python -m sphinx -b html -n -j auto docs docs/_build/html
+          command: sphinx-multiversion docs/ docs/_build/html # python -m sphinx -b html -n -j auto docs docs/_build/html
+
+      - run:
+          name: 'Copy static files'
+          command: cp -r docs/static_landing_page/* docs/_build/html
```
romanlutz, line 33: Doing these steps manually in the pipeline means it'll look different from doing it locally on my machine, right? Perhaps we should have a script?

docs/conf.py
```diff
@@ -63,6 +64,11 @@
 
 master_doc = 'contents'
 
+# Multiversion settings
+
+smv_tag_whitelist = None  # r'^v0\.4\.6$'
+smv_branch_whitelist = r'^master$'
```
romanlutz, line 70: Ouch, whitelist and master in the same line.... I hope that will change soon :-(
Unrelated to what you're adding, of course.

.circleci/config.yml
riedgar-ms: Since `shutil.copytree()` already exists, I'll look into that.

riedgar-ms:
The next issue in the `conf.py` is the `master_doc` which used to be `index` but is now `contents`. Unfortunately, the author of `sphinx-multiversion` has gone silent, since fixing the Windows-vs-Linux directory separators.

romanlutz:
> The next issue in the `conf.py` is the `master_doc` which used to be `index` but is now `contents`. Unfortunately, the author of `sphinx-multiversion` has gone silent, since fixing the Windows-vs-Linux directory separators.

Perhaps the branch from that version could be adjusted? Not sure if tags are "movable" in GitHub, though.

riedgar-ms:
Shortly after I wrote that, I got a response, and a fix (you'll have noted that the `requirement-dev.txt` is currently fetching `sphinx-multiversion` from GitHub). And then another response pointing out where our `conf.py` was causing trouble, along with a fix. I suggest taking a look at the latest build of the docs.

.circleci/config.yml
```diff
@@ -25,8 +25,8 @@ jobs:
           command: sudo python scripts/install_requirements.py --pinned False
 
       - run:
-          name: 'Build html'
-          command: python -m sphinx -b html -n -j auto docs docs/_build/html
+          name: "Build Documentation"
+          command: python scripts/build_documentation.py --documentation-path=docs --output-path=docs/_build/html
```
romanlutz, line 29: This should probably be updated in the Makefile as well!

requirements-dev.txt
```diff
@@ -33,4 +33,5 @@ nbval
 bokeh
 sphinx
 sphinx-gallery
+git+https://github.com/Holzhaus/sphinx-multiversion.git@env-vars#egg=sphinx-multiversion
```
romanlutz, line 36: Can we create an issue to move off of this and perhaps a separate one to point the docs at the last released version once v>0.4.6 is out?

scripts/build_documentation.py
```diff
@@ -0,0 +1,67 @@
...
+                        required=True)
+    return parser
+
+
+def main(argv):
+    _ensure_cwd_is_fairlearn_root_dir()
+    parser = _build_argument_parser()
+    args = parser.parse_args(argv)
+
+    with _LogWrapper("Copy static files"):
```
romanlutz, line 52: super-nit: `_LogWrapper` prints this as "start <string>" so this would be "start Copy static files", when "start copying static files" is probably easier to read. Yeah I know... that's why super-nit :-)

MiroDudik (approved):

PR #559, opened 2020-08-14
510 progress bar

### LINKED ISSUE #510: Show progress during unfairness mitigation
When running `GridSearch` or `ExponentiatedGradient` it may take a while to run all the iterations. It would be nice to have an idea about the progress. 

For `GridSearch` this could be something simple like `completed 7/100 iterations`, or perhaps something more sophisticated like `completed 7/100 iterations, projected to finish in 4:32`. For `ExponentiatedGradient` we can't know the number of iterations ahead of time, but we can tell the user how close we are to the stopping criterion as well as the current iteration number (in case we hit the maximum number of iterations).
We should perhaps also think about solutions other packages have for this.

Dref360:
**Summary**
Add a progress bar to GridSearch to inform the user on the progress.

Fixes #510 

For `ExponentiatedGradient`, we could do the same, but the progress bar would not be complete at the end. Should I add it there as well?

### DIFF
--- a/fairlearn/reductions/_grid_search/grid_search.py
+++ b/fairlearn/reductions/_grid_search/grid_search.py
@@ -9,6 +9,7 @@ from sklearn.base import BaseEstimator, MetaEstimatorMixin
 from sklearn.dummy import DummyClassifier
 from sklearn.utils.validation import check_is_fitted
 from time import time
+from tqdm import tqdm
 
 from fairlearn._input_validation import _validate_and_reformat_input, _KW_SENSITIVE_FEATURES
 from fairlearn.reductions._moments import Moment, ClassificationMoment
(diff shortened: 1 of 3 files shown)


fairlearn/reductions/_grid_search/grid_search.py
```diff
@@ -105,12 +106,24 @@ def fit(self, X, y, **kwargs):
...
         self.gammas_ = pd.DataFrame(dtype=np.float64)
         self.oracle_execution_times_ = []
+        verbose = kwargs.pop('verbose', False)
+
+        if verbose:
+            # We need tqdm to show the progress bar.
+            try:
+                from tqdm import tqdm
+            except ImportError:
+                raise RuntimeError(_TQDM_IMPORT_ERROR_MESSAGE)
```
riedgar-ms, line 126: Rather than raise an exception, since `verbose=True` is the default, I think it would be kinder to log a warning and behave as if `verbose` was false.

Tagging @adrinjalali , who first raised a caution around `tqdm`

Dref360: I set verbose default to False so that the user ia not forced to install tqdm.

romanlutz:
> Per the discussion on #517 , could you make the `tqdm` dependency soft?

Didn't that discussion #517  conclude we don't want this dependency at all?

Adding @adrinjalali as well.

adrinjalali:
Yes, the conclusion was to use `logger` instead.

Dref360:
Ah! I didn't see that PR, apologies.

I'll close this one then.

romanlutz:
@Dref360 no worries! If you're up for trying it that way (using logging) please let us know :-)

Issue #558, opened 2020-08-14
Interactive dashboard doesn't show on Azure Compute Instances, either JupyterLab or Jupyter Notebooks


maurominella:
If I setup a local Conda environment and run FairLearn dashboards, it works:

![image](https://user-images.githubusercontent.com/32428960/90260627-0f12ea00-de4c-11ea-9c2d-bbd1f1bd35ed.png)

If I run the same notebook -and libraries- on an Azure Compute Instance, it doesn't. More specifically:

- On JupyterLab, it reports "Error displaying widget: model not found"

![image](https://user-images.githubusercontent.com/32428960/90260504-df63e200-de4b-11ea-82d2-6b4bd0bf5946.png)

- On a Jupyter Notebook (running on Azure Compute Instance), the dashboard shows up but when I try using it, it gets stuck on this window:

![image](https://user-images.githubusercontent.com/32428960/90259875-066de400-de4b-11ea-8962-899c3a73ad48.png)

I checked the pip libraries, I simply have much more on Azure than locally but the common ones are the same (latest) version, installed today.
I also read that JupyterLab is still non supported, but I supposed that Jupyter Notebooks are supported, even on Azure, aren't they?

Thanks

romanlutz:
@maurominella you're right in that JupyterLab isn't support yet. It should work in Jupyter on Azure. You wrote "Azure Compute Instance" so I assume it's an [AzureML Compute Instance](https://docs.microsoft.com/en-us/azure/machine-learning/concept-compute-instance). The dashboard itself is showing up, just the calculation seems to not finish.

Sometimes it does take a little time to calculate if the dataset size is really large. I understand from your description that that can't be the case since it worked locally with the same data, right? Please confirm the size of the data (# of rows).

Something that jumped out at me from your code is that you're passing the `sensitive_feature_names` as an `np.array` as opposed to a list of strings. It doesn't make a difference when I'm running in v0.4.6, though.

Finally, I can't tell from your code what type `ys_pred` is. It should be a dictionary like `{"unmitigated": unmitigated_predictor.predict(X_test)}`, but I can't be sure from the partial snippet. If you can reproduce the same issue with a minimal example that would be very helpful.

maurominella:
Hi @romanlutz , thanks for your answer. I can't say why, but it works now. I just rebooted the VM.

I confirm it's this AzureML Compute Instance:
[...]

Issue #501, opened 2020-06-19
Fairlearn widget crashes when run inside of VS code


rchiodo:
#### Describe the bug

Fairlearn widget will crash when loaded into VS code in some circumstances. 

#### Steps/Code to Reproduce

1. Install VS code
2. Install the ms-python.python extension (it's used for opening notebooks)

Using the notebook in this zip file, opening the notebook and running it will cause a crash in the fairlearn javascript:
[sql_query2.zip](https://github.com/fairlearn/fairlearn/files/4806678/sql_query2.zip)

See this bug on the ms-python.python extension for VS code for more information:
https://github.com/microsoft/vscode-python/issues/12394

It should also be noted that VS code tries to load the fairlearn widget js from http://unpkg.com or http://jsdelivr.net. We're unable to find this js so we load the fairlearn js from the jupyter nbextensions folder. This might be what's causing the problem. JS loaded from the nbextensions folder can sometimes have implicit dependencies upon jupyter JS, whereas JS packaged on the CDN is explicit about those.

#### Screenshots
Here's a screenshot of the console log:

![image](https://user-images.githubusercontent.com/19672699/85179252-5e1a1580-b235-11ea-8cf9-eaec9c87226e.png)

#### Versions

System:
    python: 3.7.4 (default, Aug  9 2019, 18:34:13) [MSC v.1915 64 bit (AMD64)]
executable: C:\Users\rchiodo.REDMOND\AppData\Local\Continuum\miniconda3\envs\jupyter\python.exe
   machine: Windows-10-10.0.19041-SP0

Python dependencies:
       pip: 20.0.2
setuptools: 46.1.3.post20200330
   sklearn: 0.22.1
     numpy: 1.18.1
     scipy: 1.4.1
    Cython: None
    pandas: 0.25.1
matplotlib: 3.1.3
    tempeh: None

rihorn2:
Looked into this, its not related to VS Code or the environment, its a problem of datasize. Its seems at some point over 100,000 rows, the dashboard exceeds maximum call stack. We will look into reworking this with larger datasets in mind. Sorry for the bug, but it may still work in VS code with smaller data set sizes.

rihorn2:
The underlying issue is a call to Math.min. This javascript library function is recursive and fails at about this point. Will add a task and tests to make sure the dashboard avoids using Math.min/max in places that could be large arrays.

willgart:
[...]

Issue #484, opened 2020-06-16
Dashboard did not even show


yuanshi4:
#### Describe the issue linked to the documentation

when I was trying to access the dashboard, the user interactive dashboard did not show up, I only see [out]:<fairlearn.widget._fairlearn_dashboard.FairlearnDashboard at 0x7f0206c42cc0>

romanlutz:
Please run 
```python
from fairlearn import show_versions
show_versions()
```
and paste the output here. That should provide us with some information on your context.

It's likely that you need to reload the browser page with the jupyter notebook (assuming you're running one) and rerun the cell with the dashboard. This is an issue a few people have faced.

Tagging @rihorn2 for awareness.

yuanshi4:
![version](https://user-images.githubusercontent.com/60263695/84934672-4871e800-b09d-11ea-9b07-7507644b83ce.jpg)

yuanshi4:
I tried multiple attempts, using show( ) or display( ), the dashboard is still not showing up...Thank you for your help btw!

yuanshi4:
I am working on Jupyter Notebook btw

romanlutz:
Did you try reloading the page before rerunning it?

Can you also tell us which version of Fairlearn you're using? `pip freeze` should tell you

yuanshi4:
Yes, I did, tried reloading multiple times, and even restart kernel and re-run everything.

riedgar-ms:
Which browser/OS is this?

yuanshi4:
Chrome on Windows
Fairlearn version is 0.4.6

yuanshi4:
And the jupyter notebook is on AWS environment (doubt if that matters)

riedgar-ms:
You mean that you have a windows VM in AWS, and you're RDPing to it, and running the notebook there?

riedgar-ms:
Also, what version of Jupyter? As @romanlutz  said, if you could attach your `pip freeze` it would be very helpful.

riedgar-ms:
I've just used the following requirements file:
```
scikit-learn==0.23.1
numpy==1.18.5
scipy==1.4.1
pandas==1.0.4
Cython==0.28.4
matplotlib==3.0.3
fairlearn==0.4.6
```
and then doing a regular install of jupyter (and shap). With that, I can load one of our sample notebooks, and load the dashboard in Chrome.

[...]

PR #283, opened 2020-02-03
Adding the disparity tab back to the widget

chisingh:
The disparity calculation method can now be selected from the wizard.

PR #561, opened 2020-08-14
Plots: Add plot_disparities_in_performance and plot_disparities_in_metric

kevinrobinson:
## Rationale
This work is intended to help along the suggestion in https://github.com/fairlearn/fairlearn-proposals/pull/14#discussion_r470248943, with the assumption that the UI and Azure-specific bits of the project will be removed and placed in a separate repository with a different governance structure.

To illustrate that this is both viable and plays to the strengths of folks currently staffing the project, here is an example function that could be used in the quick start guide, removing the need for any Azure or UI code.  This is based on some key assumptions, including:
- to date, there has been no user research or feedback validating that users can use these visualizations to do meaningful work that assesses or reduces real world harm
- the current visualizations provide little value in making meaningful decisions about how to interpret the data, or how to take action on it
- the interactivity of the visualizations is low-value based on the team's current understanding of practitioner needs (eg, Holstein et al. 2019; Madaio et al. 2020)
- the overhead of making widgets and HTML/CSS/JS is a barrier for team members in maintaining, supporting and growing the project
- the current visualizations do not address any of the core challenges involved in fairness as sociotechnical work (eg, Selbst et al. 2019), and so reducing maintenance burden can free folks up to engage in higher-value work

If these assumptions are off, please comment and help validate them! 😄 

## What this adds
This pull request adds the `plot_disparities_in_performance` and `plot_disparities_in_selection_rate` functions, intended to replace those two visualizations built as widgets in HTML/CSS/JS.

### Disparity in selection rate
Note the change of bar color to grey, since blue was used to communicate something specific in the chart above and the meaning of the color is not the same in this chart.

#### now: +3 clicks in Azure UI
Screen Shot 2020-08-14 at 4 31 52 PM

#### added: plot_disparity_in_selection_rate
Screen Shot 2020-08-14 at 4 34 20 PM

### Disparity in accuracy
Note the overprediction and underprediction values within the charts are quite different in the new plot.  They reflect the values that the `false_positive_rate` and `false_negative_rate` functions return, but these don't really make sense to me.  I am mixing up something important here, but not sure what.  I'll take another look next week, but if other folks can spot what the issue is that'd be awesome and super helpful! 👍 

EDIT: copy-paste typo in underprediction/overprediction is fixed now.

#### now: +3 clicks in Azure UI
Screen Shot 2020-08-14 at 4 21 11 PM

#### added: plot_disparity_in_accuracy
Screen Shot 2020-08-14 at 4 03 45 PM

EDIT: as @romanlutz helpfully pointed out, this also supports multiple values for a single sensitive attribute.  This example uses `data.data['race']` in place of `data.data['sex']` below (these examples are all narrowly technical without sociotechnical context):
Screen Shot 2020-08-17 at 10 17 44 AM

## Other notes
- researchers interested in low-level HCI studies (eg, can people use this chart to do meaningful work?) could start those studies immediately to understand if there is any value here (or to do comparison studies of this versus elements from https://github.com/fairlearn/fairlearn/pull/283#issuecomment-662625530)
- this addresses issues like https://github.com/fairlearn/fairlearn/issues/558, https://github.com/fairlearn/fairlearn/issues/484, https://github.com/fairlearn/fairlearn/issues/270, https://github.com/fairlearn/fairlearn/issues/273, https://github.com/fairlearn/fairlearn/issues/501, etc.
- other folks looking for ways to contribute with code could add other functions (eg, a "model comparison" scatterplot), or others as-needed in the course of developing example notebooks (eg, for more realistic and useful subgroup analysis that a single binary sensitive attribute) or if there is validated demand for other visuals (eg, https://github.com/fairlearn/fairlearn/pull/283#issuecomment-662625530)
- other folks looking to help with user guide or documentation are welcome to update the related [user guide page](https://fairlearn.github.io/user_guide/assessment.html#fairlearn-dashboard)
- this doesn't address localization, but strings are factored out and if MS folks want to lend translation resources we could do that post-merge
- this doesn't address accessibility, I'm not sure how to do that with these kinds of charts but could partner with anyone to do this now or post-merge
- the quickstart example illustrates a narrowly technical perspective that doesn't follow the [contributing example notebooks](https://fairlearn.github.io/contributor_guide/contributing_example_notebooks.html), but this pull request does not address that
- there's no tests here, and I would vote to invest in user testing over unit testing or integration testing

## Code to test this in Jupyter on 0.4.6:
```
# copied verbatim from quickstart
import numpy as np
import pandas as pd
import matplotlib.pyplot as plt
from sklearn.datasets import fetch_openml
data = fetch_openml(data_id=1590, as_frame=True)
X = pd.get_dummies(data.data)
y_true = (data.target == '>50K') * 1
sex = data.data['sex']

from fairlearn.metrics import group_summary
from sklearn.metrics import accuracy_score
from sklearn.tree import DecisionTreeClassifier
classifier = DecisionTreeClassifier(min_samples_leaf=10, max_depth=4)
classifier.fit(X, y_true)
y_pred = classifier.predict(X)

# to use new functions
from fairlearn.plots import plot_disparities_in_performance, plot_disparities_in_selection_rate
plot_disparities_in_performance(y_true, y_pred, sex)
plot_disparities_in_selection_rate(y_true, y_pred, sex)
```


riedgar-ms:
As a warning, `matplotlib` only gets installed as part of `[customplots]`.

I think @romanlutz  came up with a way of testing plotting methods too?

romanlutz:
> As a warning, `matplotlib` only gets installed as part of `[customplots]`.
> 
> I think @romanlutz came up with a way of testing plotting methods too?

That's correct! I'm not too fond of that name `customplots`, so if people want to make that shorter (`plots`) I'm all in favor.

We'd also have to provide proper error messages telling users that they need matplotlib if it's not installed, just like with the postprocessing plots.

I don't quite agree with everything written in the description but that's immaterial. 

The more important question to me is what this means to UI development for the community and the proposal https://github.com/fairlearn/fairlearn-proposals/pull/14 . Does this mean you'd prefer not to have a separate repo elsewhere? On Thursday people were quite insistent that there should be a way for the community to write UI, so I'd rather get that settled before inventing new UIs.

Regarding testing: there's a pytest plugin for testing plots but it didn't work across platforms that we test on, so I never turned it on. It does work locally, though. See https://github.com/fairlearn/fairlearn/blob/master/test/unit/postprocessing/test_plots.py for more information.

Finally, this kind of change would need an update to CHANGES.md, API reference (docs folder), and user guide.

kevinrobinson:
@riedgar-ms @romanlutz Thanks for the comments! 👍

This gets started with the first step outlined in https://github.com/fairlearn/fairlearn-proposals/pull/14#discussion_r470248943, and I'm hoping to help move things forward by showing rather than telling :)  It also improves the UX for prospective users currently facing issues like https://github.com/fairlearn/fairlearn/issues/484 https://github.com/fairlearn/fairlearn/issues/501 and https://github.com/fairlearn/fairlearn/issues/558.

If folks want to merge this, I can work myself or with anyone else to finish this off.

romanlutz:
> @riedgar-ms @romanlutz Thanks for the comments! 👍
> 
> 
> 
> This gets started with the first step outlined in https://github.com/fairlearn/fairlearn-proposals/pull/14#discussion_r470248943, and I'm hoping to help move things forward by showing rather than telling :)  It also improves the UX for prospective users currently facing issues like https://github.com/fairlearn/fairlearn/issues/484 https://github.com/fairlearn/fairlearn/issues/501 and https://github.com/fairlearn/fairlearn/issues/558.
> 
> 
> 
> If folks want to merge this, I can work myself or with anyone else to finish this off.

Right. I think this is a fine first step. We sort of agreed to try and add documentation (API reference for new modules, CHANGES.md, user guide) whenever adding the code, although that's obviously not as strict as long as we have severe gaps around many parts of the codebase. Starting a new module strikes me as a great place to make sure we have some of these in place. I certainly won't insist on a comprehensive user guide in this PR, but the other parts are all fixable in a single line each. Let me know if you have questions about it.

The matplotlib concern should probably be addressed as well.

Other than that this looks great. Thanks for getting this started @kevinrobinson . 

Of course, this assumes that people in the community want this solution for the UI. It would certainly be nice to hear people's opinions, although in the current form matplotlib is optional and therefore this isn't that different of a setup from before. I know @adrinjalali talked about matplotlib plots before so this should be in line with his opinions hopefully (?)

kevinrobinson:
@romanlutz Thanks! 👍 

re: community consensus, my interpretation of https://github.com/fairlearn/fairlearn-proposals/pull/14#issuecomment-673332453 and https://github.com/fairlearn/fairlearn-proposals/pull/14#issuecomment-673307087 was that other folks have previously suggested this approach, which is part of why I'm trying to build on that consensus here.

But to shift gears, I think it is far more important whether **users** want this :)  I think all user-facing changes should primarily be evaluated on the value they create for users, be informed by user research, or other informal methods of validation and co-design.  So even the concerns in https://github.com/fairlearn/fairlearn-proposals/pull/14 are a primary consideration, hopefully on balance the discussion on this pull request also considers whether this change would serve users.  In that spirit, this PR improves the UX for people trying to visualize fairness metrics but facing issues like #484 #501 and #558.

---

re: documentation, I added API docs and a section to the user guide, mirroring what is already there.  I didn't make any edits to improve quality (eg, adding sociotechnical context, showing how these charts are relevant for assessing or reducing real harms, adding guidelines on how to interpret visuals, etc.)

re: code, I added commits removing `customplots` everywhere, that's ready for review.  I read https://github.com/fairlearn/fairlearn/pull/289, and so only added smoke tests for now.  The code in `plot_disparities_in_performance` uses metrics functions is the main place where code review would be really helpful! 👍

romanlutz:
Leaving matplotlib out of the main/default/base (not sure about terminology) package was a very conscious choice. If we want to undo that we should at least hear what @adrinjalali @MiroDudik @riedgar-ms think. That in itself will hold up this PR a bit because @MiroDudik isn't available for a few days.

I vaguely remember lots of issues testing this on different platforms. The separation into a base package and a plotting extension made that slightly harder, so the simplicity of having just one is something I appreciate. Perhaps the issues with including matplotlib go further than that, though? I'm sure @adrinjalali will know.

CHANGES.md
```diff
@@ -1,7 +1,7 @@
 # Changes
 
 ### v0.5.0
-
+* Add `plot_disparities_in_performance` and `plot_disparities_in_selection_rate` functions https://github.com/fairlearn/fairlearn/pull/561
```
romanlutz, line 4: nit: Add 'plots' module with ...

romanlutz: Oh just saw you renamed it ti metrics_plots. I liked the original name better but obviously that's very subjective

fairlearn/metrics_plots/plot_disparities_in_selection_rate.py
```diff
@@ -0,0 +1,83 @@
...
+from fairlearn.metrics import selection_rate_group_summary
+
+
+def plot_disparities_in_selection_rate(y_true, y_pred, sensitive_features):
+    """Plot disparities in selection for a binary classifier with a single binary sensitive feature.
+
+    This helps visualize differences in overall selection rate, and selection rates
+    for groups classified by sensitive attributes.
+
+    :param 1D-array y_true: Ground truth (correct) labels.
```
romanlutz, line 16: Please use numpydoc style instead

fairlearn/metrics_plots/plot_disparities_in_performance.py
```diff
@@ -0,0 +1,122 @@
...
+        false_negative_rate,
+        y_true,
+        y_pred,
+        sensitive_features=sensitive_features)
+    sensitive_values = sensitive_features.unique()
+    overpredictions = []
+    underpredictions = []
+    for sensitive_value in sensitive_values:
+        overpredictions.append(fp_summary['by_group'][sensitive_value])
+        underpredictions.append(fn_summary['by_group'][sensitive_value])
```
kevinrobinson, line 43: See the issue description for more, but a review of the code in this area that determines  overpredictions and underpredictions would be super helpful! 👍

fairlearn/metrics_plots/plot_disparities_in_performance.py
```diff
@@ -0,0 +1,122 @@
...
+
+
+import matplotlib.pyplot as plt
+import matplotlib.ticker as mtick
+from sklearn.metrics import accuracy_score
+from fairlearn.metrics import group_summary, false_positive_rate, false_negative_rate
+
+
+def plot_disparities_in_performance(y_true, y_pred, sensitive_features):
+    """Plot disparities in performance for a binary classifier with a single binary sensitive feature.
```
romanlutz, line 12: Why the restriction to a binary sensitive feature? Does this not generalize?

fairlearn/metrics_plots/plot_disparities_in_selection_rate.py
```diff
@@ -0,0 +1,94 @@
...
+# Licensed under the MIT License.
+
+
+import matplotlib.pyplot as plt
+import matplotlib.ticker as mtick
+from fairlearn.metrics import selection_rate_group_summary
+
+
+def plot_disparities_in_selection_rate(y_true, y_pred, sensitive_features):
+    """Plot disparities in selection for a binary classifier with a single binary sensitive feature.
```
romanlutz, line 11: Same here. Looks like this should generalize beyond binary sensitive features

fairlearn/metrics_plots/plot_disparities_in_performance.py
```diff
@@ -0,0 +1,122 @@
...
+
+    This helps visualize differences in overall prediction error, as well as false positive
+    and false negative error rates across the binary values of a sensitive attribute.
+
+    :param 1D-array y_true: Ground truth (correct) labels.
```
romanlutz, line 17: Numpydoc please

fairlearn/metrics_plots/plot_disparities_in_selection_rate.py
kevinrobinson: thanks!

adrinjalali:
Should we move this forward?

devops/nightly.yml
```diff
@@ -39,4 +39,3 @@ jobs:
 
 - template: templates/build-widget-job-template.yml
 
-- template: templates/limited-installation-job-template.yml
```
riedgar-ms, line 42: Why is this being removed?

fairlearn/metrics_plots/plot_disparities_in_performance.py
riedgar-ms: This feels like something that might belong in the metrics code (although that is in the process of being rewritten).

fairlearn/metrics_plots/plot_disparities_in_performance.py
```diff
@@ -0,0 +1,134 @@
...
+    underpredictions = []
+    for sensitive_value in sensitive_values:
+        overpredictions.append(fp_summary['by_group'][sensitive_value])
+        underpredictions.append(fn_summary['by_group'][sensitive_value])
+    disparity = abs(
+        accuracy_summary['by_group'][sensitive_values[0]]
+        - accuracy_summary['by_group'][sensitive_values[1]])
+
+    # chart text for localization
+    title_text = 'Disparity in performance'
```
riedgar-ms, line 61: These are not really localisable if they're embedded in the method. They need to be constants at the top of the file.

fairlearn/metrics_plots/plot_disparities_in_performance.py
```diff
@@ -0,0 +1,134 @@
...
+    xlabel_text = 'Prediction error rate'
+    ylabel_text = 'By group'
+    overall_performance_text = 'Overall performance'
+    accuracy_text = 'accuracy'
+    disparity_text = 'disparity'
+    underprediction_legend_text = "Underprediction\n  predicted=0\n  true=1"
+    overprediction_legend_text = "Overprediction\n  predicted=1\n  true=0"
+
+    # chart styles
+    figsize = (12, 4)
```
riedgar-ms, line 71: Shouldn't all of these be user-configurable?

fairlearn/postprocessing/_constants.py
```diff
@@ -9,8 +9,6 @@
 
 OUTPUT_SEPARATOR = "-"*65
 
-_MATPLOTLIB_IMPORT_ERROR_MESSAGE = "Please make sure to install fairlearn[customplots] to use " \
```
riedgar-ms, line 12: Why is this being removed?

romanlutz: I think because he's added matplotlib to the package as a "main" dependency, so this case doesn't happen anymore. The main question is whether we want that actually.

fairlearn/metrics_plots/plot_disparities_in_performance.py
romanlutz: Will there be localization? If so, I don't think it needs to be in the first version.

devops/nightly.yml
romanlutz: Because "limited" = no matplotlib, so since he added it as a "core" dependency it's not needed anymore.

fairlearn/postprocessing/_constants.py
adrinjalali: we could go either way. We could still have `matplotlib` as a soft dependency which is only needed when people use the plotting functionality, or have it as a core dependency. I would lean towards having it as a soft dependency.

romanlutz: Looks like @riedgar-ms, me, @MiroDudik (asked him separately), and @adrinjalali are all leaning towards soft dependency (i.e. status quo).

fairlearn/metrics_plots/plot_disparities_in_performance.py
romanlutz: That would be nice, although for a first version I won't insist on it.

fairlearn/metrics_plots/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+This module contains functions for visualizing metrics within
+notebooks.
+"""
+
+
+from .plot_disparities_in_performance import plot_disparities_in_performance
+from .plot_disparities_in_selection_rate import plot_disparities_in_selection_rate
+
+__all__ = [
+    "plot_disparities_in_performance",
```
romanlutz, line 16: Relaying some considerations from an earlier conversation with @MiroDudik :

We've been wanting to move away from these names since they were confusing to users according to feedback we received. 

Something like `plot_metric_by_group(metric, y_true, y_pred, *, sensitive_features)` would be more intuitive than the "disparity in predictions" used in the past and generally applicable for other metrics compared to `plot_disparities_in_selection_rate`. I understand if you just want to restrict it to the existing selection rate for this PR, though, especially considering that metrics are getting overhauled in parallel.

For the first one we're actually just plotting over- and underprediction, not any arbitrary performance metric. In theory, this could be done as follows
```
plot_error_rate_by_group(y_true, y_pred, *, sensitive_features, over_under_split=True)
```
where `over_under_split` decides whether it's just error rate or split into over- and underprediction (additional arg not required in this PR). However, over- and underprediction also turned out to be unnecessarily complicated for users, so we're planning to switch to FPR and FNR. The difference is that error rate = over + underprediction, whereas FPR and FNR have a different denominator (positives or negatives). Since they're more widely known and used that's more intuitive for users.

fairlearn/metrics_plots/plot_disparities_in_performance.py
romanlutz: I pointed this out in the other comment (today) but after consulting with @MiroDudik overprediction != FPR and underprediction != FNR, so we should change it so that it doesn't say over and underprediction anymore, but rather FPR and FNR.

fairlearn/metrics_plots/plot_disparities_in_selection_rate.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+    # compute
+    selection_rate_summary = selection_rate_group_summary(y_true,
+                                                          y_pred,
+                                                          sensitive_features=sensitive_features)
+    sensitive_values = sensitive_features.unique()
+    selection_rates = []
+    for sensitive_value in sensitive_values:
+        selection_rates.append(selection_rate_summary['by_group'][sensitive_value])
+    disparity = abs(selection_rates[0] - selection_rates[1])
```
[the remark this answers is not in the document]
romanlutz: You probably mean that it should be max and min as opposed to first and second element?
@riedgar-ms 's metrics API changes will allow for that to be done easily, although we don't have to wait for that necessarily.

fairlearn/metrics_plots/plot_disparities_in_performance.py
kevinrobinson: @romanlutz Thanks!  This is super helpful 👍 

As the [current plan for 0.5](https://hackmd.io/1b-RSHfQR221tokjL3j3JA?edit) is to ship the metrics changes, switch to Flask for the widget, and make documentation updates all around (eg, so the quickstart works), I'll leave this work alone so as to not slow down those other things from shipping first.

romanlutz:
@kevinrobinson just checking in on this: now that v0.5.0 is out we can go ahead with this as planned. Please let us know if you need any help on this.

romanlutz:
As discussed with @kevinrobinson I took over this PR with the following changes:
- move to new `MetricFrame` since this PR predates v0.5.0
- generalize the `plot_disparities_in_selection_rate` function to work with any metric (including binary classification & regression), hence renamed to `plot_disparities_in_metric`
- fix labels since we're showing FPR/FNR and the labels said over-/underprediction which isn't quite the same (denominator is different)

The reason there are two plots right now is that @kevinrobinson did an awesome job replicating the original plots from the Fairlearn dashboard. In reality, the result from `plot_disparities_in_performance` is just a combined FPR/FNR plot that shows accuracy through text. I'm fine with adding that for the sake of consistency with the existing dashboard, but theoretically it's possible to get the same information by calling `plot_disparities_in_metric` three times with the three individual metrics.

We can probably find better names (?) Suggestions welcome! Big ones:
- Should the module be `metrics_plots` or more generically `plots`, or `plotting`? We have plotting functionality for postprocessing currently half-hidden away in the postprocessing module. Perhaps that makes sense, but I'd like to hear other people's opinions.
- The function names are quite long.

@hildeweerts  @adrinjalali  @MiroDudik  @riedgar-ms 

I will also open a few issues with "help-wanted" to add plots for 
- model comparison similar to what we have in the Fairlearn dashboard
- passing in multiple metrics rather than just one, and plotting all of them in a subplots

adrinjalali:
I'll check the PR soon, in the meantime, could you please remind me why we have all those `.png` files in this PR? As in, why wouldn't they be generated in `.py` example files and then used in the user-guide?

romanlutz (commented):
> I'll check the PR soon, in the meantime, could you please remind me why we have all those `.png` files in this PR? As in, why wouldn't they be generated in `.py` example files and then used in the user-guide?

Which `.png` files are you referring to? Perhaps you're looking at an earlier version? The original PR had png files, but I replaced them with `.. plot::` routines provided by matplotlib to generate the plot when creating the webpage. IMO that's much preferable since we don't have to update the png whenever we make code changes.

docs/api_reference/index.rst
```diff
@@ -7,6 +7,7 @@ API Reference
    :maxdepth: 1
 
    fairlearn.metrics
+   fairlearn.metrics_plots
```
MiroDudik, line 10: I'm in favor of having plotting be always in the modules that they support. This is consistent with how [sklearn handles this](https://scikit-learn.org/stable/modules/classes.html#id3) and also how we have done it with `fairlearn.postprocessing.plot_threshold_optimizer` (I think we had a conversation about it back then).

devops/templates/limited-installation-job-template.yml
```diff
@@ -47,4 +47,4 @@ jobs:
 
   - task: PublishTestResults@2
     displayName: 'Publish Test Results **/TEST-*.xml'
-    condition: succeededOrFailed()
\ No newline at end of file
+    condition: succeededOrFailed()
```
adrinjalali, line 50: newline at the end of the file

docs/api_reference/index.rst
adrinjalali: I second this

docs/contributor_guide/development_process.rst
```diff
@@ -145,7 +145,7 @@ The prerequisites for Fairlearn are split between three separate files:
 
     -  `requirements.txt <https://github.com/fairlearn/fairlearn/blob/master/requirements.txt>`_
        contains the prerequisites for the core Fairlearn package
-
+
```
adrinjalali, line 148: whitespace at the end of the line

docs/user_guide/assessment.rst
```diff
@@ -395,6 +395,75 @@ see the :ref:`sphx_glr_auto_examples_plot_new_metrics.py` notebook in the
...
+
+Metrics plots
+-------------
+
+.. currentmodule:: fairlearn.metrics_plots
+
+The :py:mod:`fairlearn.metrics_plots` module visualizes fairness metrics from
+different perspectives.
+
+The examples below illustrate a scenario where *binary gender* is
```
adrinjalali, line 408: should we use a different example dataset for new examples we introduce?

fairlearn/metrics_plots/plot_disparities_in_metric.py
```diff
@@ -0,0 +1,86 @@
+# Copyright (c) Microsoft Corporation and Fairlearn contributors.
```
adrinjalali, line 1: remove _Microsoft_ from the new files?

fairlearn/metrics_plots/plot_disparities_in_metric.py
```diff
@@ -0,0 +1,86 @@
...
+# Licensed under the MIT License.
+
+
+import matplotlib.pyplot as plt
+import matplotlib.ticker as mtick
+from fairlearn.metrics import MetricFrame
+
+
+def plot_disparities_in_metric(metric, y_true, y_pred, sensitive_features, show_plot=True):
```
adrinjalali, line 10:
```suggestion
def plot_disparities_in_metric(metric, *, y_true, y_pred, sensitive_features, show_plot=True):
```

fairlearn/metrics_plots/plot_disparities_in_metric.py
```diff
@@ -0,0 +1,86 @@
...
+    Parameters
+    ----------
+    metric : callable
+        A metric function that takes arguments `y_true` and `y_pred`
+    y_true : array-like
+        The list of true values
+    y_pred : array-like
+        The list of predicted values
+    sensitive_features : array-like
+        the sensitive attributes
```
adrinjalali, line 25: do we want to also support conditional features in plots?

fairlearn/metrics_plots/plot_disparities_in_metric.py
```diff
@@ -0,0 +1,86 @@
...
+
+    """
+    # compute
+    metric_frame = MetricFrame(metric, y_true, y_pred,
+                               sensitive_features=sensitive_features)
+
+    # chart text for localization
+    metric_text = metric.__name__.replace('_', ' ')
```
adrinjalali, line 33: we could accept `metric_name` or `x_label`, `y_label` optionally if the `__name__` is not properly set or the user wants an alternative text.

fairlearn/metrics_plots/plot_disparities_in_metric.py
```diff
@@ -0,0 +1,86 @@
...
+    title_text = f'Disparity in {metric_text}'
+    ylabel_text = 'By group'
+    overall_performance_text = 'Overall'
+    diff_text = 'min-max difference'
+    ratio_text = 'min-max ratio'
+    # chart styles
+    figsize = (12, 4)
```
adrinjalali, line 40: this should also be an optional input parameter, or done via accepting `ax`

fairlearn/metrics_plots/plot_disparities_in_metric.py
adrinjalali, line 10: The plot functions should probably accept and return an `ax` parameter to receive and return the `matplotlib`'s `axis` object so that users can further customize the plots.

test/unit/metrics_plots/__init__.py
```diff
@@ -0,0 +1,2 @@
+# Copyright (c) Microsoft Corporation and Fairlearn contributors.
```
adrinjalali, line 1: not sure if this file is required.

test/unit/metrics_plots/conftest.py
```diff
@@ -0,0 +1,20 @@
+# Copyright (c) Microsoft Corporation and Fairlearn contributors.
```
adrinjalali, line 1: I'd rather have these bits in the test files themselves.

docs/api_reference/index.rst
romanlutz: Awesome, that works for me!

devops/templates/limited-installation-job-template.yml
romanlutz: added a newline now!

docs/user_guide/assessment.rst
romanlutz: I'm the first to criticize usage of the adult dataset since it's absolutely nonsensical and doesn't show a proper realistic task. After all, how is it useful to predict if someone earns more than the arbitrary cutoff amount of $50k? 

That means we have to have something better, though. The argument for adult so far was that it's widely used in the fairness literature when comparing results of mitigation techniques. When using another dataset we've always run into the problem that you'd need to contextualize this first. For example, it's questionable to use COMPAS or Boston housing datasets without at least pointing at the complicated and controversial background. We haven't written a section in the documentation about that yet, though.

I really want us to replace adult in every example eventually. I do think that it's a separate task from this PR, though. Perhaps the credit card default dataset would work, but again we should describe the context somewhere to avoid having people think that it's fine to just use such a dataset.

Wdyt?

fairlearn/metrics_plots/__init__.py
romanlutz: Closing this comment since I've taken over this PR and replaced this with `plot_disparities_in_metric`. Naming feedback is still welcome, of course!

fairlearn/metrics_plots/plot_disparities_in_metric.py
romanlutz: Added this in both functions. (didn't take suggestion to avoid DCO issues)

romanlutz: Sounds good. I compared with some of the functions in `scikit-learn`. They don't have a `show_plot` arg, but rather the caller is expected to call `plt.show()` themselves. Is that preferable? I find it kind of odd to call a function called `plot_*` and then there's no plot shown unless I do something more, but perhaps that's just me.

fairlearn/metrics_plots/plot_disparities_in_metric.py
romanlutz: Yes, but not in this PR. I've opened #668 for this purpose. Should be a good task for a new contributor, for example. Otherwise I'll see whether I can get to it at some point soon.

docs/user_guide/assessment.rst
adrinjalali: I agree that we should contextualize a data which we use in our docs, but this requirement is making us adding more documentation with a dataset which we know has these clear issues. I would be happier (I think) if we use another dataset which doesn't have these issues, and have the contextualization as a future work instead of adding more examples with dataset and having changing all of them later as a future work.

That said, I wouldn't veto this PR because of the dataset, it'd be just really nice if we moved away from this dataset, and not using it anew is a good place to start phasing it out I think.

fairlearn/metrics_plots/plot_disparities_in_metric.py
adrinjalali: In the context of matplotlib, I find it nice if the caller has the capability of having each plot in a subplot (hence passing `ax`), and makes it more configurable when the method returns an object which the user can change. In terms of the user having to call `plt.show()`, I'm agnostic. I don't mind having the `show_plot` as you have here. But it's basically replacing a single line of code. But I also don't mind having it `True` by default, which is kinda nice to the user I guess.

romanlutz: I'll leave it in there for now then, unless someone has stronger feelings about this :-)

fairlearn/metrics_plots/plot_disparities_in_metric.py
romanlutz: Agreed. The caller can adjust that if they want. I'm removing this line.

fairlearn/metrics_plots/plot_disparities_in_metric.py
romanlutz: Agreed! I added an optional argument `metric_name` and documented that we'll use `__name__` is no metric name is passed.

docs/user_guide/assessment.rst
romanlutz: I very much share your perspective. I'm not 100% happy with any of the three datasets currently in `fairlearn.datasets`, but I think the credit card default dataset from UCI could be great. We use it in this notebook:
https://github.com/fairlearn/fairlearn/blob/master/notebooks/Binary%20Classification%20with%20the%20UCI%20Credit-card%20Default%20Dataset.ipynb
For that, I'd create a separate PR to add it to the datasets module first, and then replace it for this example in a follow-up PR. We can discuss whether we want it elsewhere, too, for example in the quickstart. That dataset has its own issues (which were alluded to in #418 ), of course, but I think for the purpose of demonstrating the functionality of these plotting functions it'll be fine.

This would mean this stays as is in the current PR, though, because expanding the scope to adding another dataset, documenting that dataset, etc. doesn't seem right.

test/unit/metrics_plots/__init__.py
romanlutz: Turns out it is because we are importing from conftest which otherwise can't be found.

test/unit/metrics_plots/conftest.py
romanlutz: But that would mean we need to duplicate it if there are multiple files. Or did you mean "make it all a single file"? I'll go with that for now, but lmk if you meant something else.

test/unit/metrics_plots/__init__.py
romanlutz: This is now removable after merging all the content as indicated in https://github.com/fairlearn/fairlearn/pull/561/files/84ae2abcde883726545c2d11d3cd18a5b9b6633f#r551555912 which I now realize is probably what you wanted all along.

test/unit/metrics_plots/conftest.py
adrinjalali: I meant duplicating it in the file. It makes understanding a test file as a standalone file much easier. I'm not sure how common it is to load these datasets in the conftest, but I'm not used to it. And for a third person adding a test in a test file, they'd be a bit confused or not see that there are datasets available and they'd load their own in the tests.

adrinjalali (commented):
Let me know when you want another round of review :)

docs/user_guide/assessment.rst
MiroDudik: UCI credit data set is actually not ideal, because it doesn't exhibit equalized-odds disparity, so we have to partly synthesize it (which makes for a slightly odd notebook, but it would look even weirder here I think). But I don't remember whether it shows other kinds of disparity.

Two options:
* We use the original (untweaked) version of the UCI data set even if we don't have much disparity.
* We use a different data set that is known to exhibit disparity, say when using linear models.

That said. I don't think this should be a blocker on this PR. However, if we are okay just using the untweaked UCI credit data set, we can put `fetch_openml` in this sample code for now, and replace with the `fairlearn.datasets` equivalent once/if it's available.

@lejit, @hildeweerts: do you have any thoughts on this?

fairlearn/metrics/_plots/_plot_disparities_in_metric.py
```diff
@@ -0,0 +1,110 @@
+# Copyright (c) Microsoft Corporation and Fairlearn contributors.
+# Licensed under the MIT License.
+
+
+import matplotlib.pyplot as plt
+import matplotlib.ticker as mtick
+from fairlearn.metrics import MetricFrame
+
+
+def plot_disparities_in_metric(metric, *, y_true, y_pred, sensitive_features, show_plot=True,
```
MiroDudik, line 10: Two points re. API here:
* I think this should be called `plot_metric_by_group`
* instead of `metric_name`, we could just allow the format where `metric` is a single item dictionary? this would be more in line with how you would get this done with `MetricFrame`.

Another idea (for another PR) is to have a method `MetricFrame.plot(show_plot=True, ax=None)` and this would be just a wrapper for that? This would be analogous to [pd.DataFrame.plot](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.plot.html). Although there's an obvious wrinkle re. what to do if you have multiple metrics and/or control features, which we probably don't need to resolve in this PR :-)

docs/user_guide/assessment.rst
LeJit: With regards to other data sets in a similar domain that we could use for examples, we could potentially use the German Credit dataset or the Lending Club dataset. Both are commonly used benchmark datasets in fairness papers, and German Credit is available through OpenML (Lending Club isn't though). We should eventually add the German Credit dataset to `fairlearn.datasets` anyways because it is a commonly used benchmark.

I have already written the `fairlearn.datasets` function for the UCI credit card data set as part of a sub-task to redo the associated Jupyter notebook. If we want to use the untweaked UCI credit card dataset for now, I can create a small pull request to get that code merged in.

test/unit/metrics_plots/conftest.py
romanlutz: That's fair. We've been doing it like that throughout the test/unit directory so far, so that's probably why Kevin originally copied this behavior. I assume the current structure is in line with your expectations. If not, please let me know!

docs/user_guide/assessment.rst
hildeweerts: Whether we have to add a particular data set to `fairlearn.datasets` depends quite a bit on the purpose of that part of the repo. We've actually had discussions about this in the past and I'm not sure whether a consensus was reached (see e.g., #583).

On a more practical note, I'm not familiar with the Lending Club data but the fact it isn't on OpenML at the moment shouldn't be a problem, because we can just add it ourselves! Is this the one you refer to: https://www.kaggle.com/wordsforthewise/lending-club? From the description it doesn't seem like this data was collected with permission from Lending Club, though...

hildeweerts: I was just wondering, perhaps concerns re. proper contextualization can be partly solved by taking the effort to create data sheets (see https://arxiv.org/abs/1803.09010) for the data sets in `fairlearn.datasets`. We probably won't be able to fill out everything, but it seems to me that if we want to practice what we preach, this would be a minimal suggestion.

fairlearn/metrics/_plots/_plot_disparities_in_metric.py
hildeweerts: I like the idea for `MetricFrame.plot()`! Re. multiple metrics/control features: I personally would probably go for something like a heat map.

hildeweerts:
I know I'm a bit late to the party and I haven't run the latest code myself, but I have to admit that I personally don't find the `plot_disparities_in_performance()` plot very intuitive - although displaying FPR/FNR is already makes more sense to me than underprediction/overprediction.

I'm probably biased though because I've mostly worked with imbalanced datasets in the past (fraud detection, predictive maintenance, etc.), in which case accuracy doesn't really make sense as a performance metric, nor does plotting FPR/FNR on the same scale.

romanlutz:
> I know I'm a bit late to the party and I haven't run the latest code myself, but I have to admit that I personally don't find the `plot_disparities_in_performance()` plot very intuitive - although displaying FPR/FNR is already makes more sense to me than underprediction/overprediction.
> 
> I'm probably biased though because I've mostly worked with imbalanced datasets in the past (fraud detection, predictive maintenance, etc.), in which case accuracy doesn't really make sense as a performance metric, nor does plotting FPR/FNR on the same scale.

@MiroDudik mentioned something similar and I totally agree. IMO the ideal state would be to cut that function and if someone wants to see both they can use the other function to plot FPR and FNR. The only reason it's there is that Kevin replicated the existing `FairlearnDashboard` in this PR, and I felt bad about just cutting it. If we all agree that it's best not to have it let's do that!

Note that there are a few follow-up tasks that I created already ( #668 #667 #666 ) so those sorts of things are beyond the scope of this PR :-)

MiroDudik:
I was working on a tutorial and I realized that our `MetricFrame` **already supports** plotting functionality to an extent. This is a small example demonstrating it:

```python
# Imports
import pandas as pd
import fairlearn.metrics as flm
import sklearn.metrics as skm

# Create a toy data set
df = pd.DataFrame(
    {'sex':  ['F', 'M', 'F', 'M', 'M', 'F', 'M', 'F', 'M', 'M'],
     'race': ['White', 'White', 'White', 'Black', 'Black', 'Black', 'Hispanic', 'Hispanic', 'White', 'Black'],
     'y_true': [0, 1, 1, 0, 1, 1, 1, 1, 1, 0],
     'y_pred': [1, 1, 1, 0, 0, 1, 0, 1, 1, 1]})

# Define metrics
metrics = {
    'accuracy': skm.accuracy_score,
    'precision': skm.precision_score,
    'recall': skm.recall_score,
    'false_positive_rate': flm.false_positive_rate,
    'true_positive_rate': flm.true_positive_rate,
    'avg_true': lambda y_true, y_pred: np.mean(y_true),
    'avg_pred': lambda y_true, y_pred: np.mean(y_pred),
    'count': lambda y_true, y_pred: y_true.shape[0],
}

# Set up metric frame
mf = MetricFrame(metrics, df['y_true'], df['y_pred'], sensitive_features=df[['sex','race']])

# Plots
mf.by_group.plot.bar(y=['false_positive_rate','true_positive_rate'],
                     title='Compare multiple disaggregated metrics in a single plot');

mf.by_group.plot.bar(subplots=True, layout=[3,3], legend=False, figsize=[12,8],
                     title='Show all metrics');
```

image

[tagging @adrinjalali , @hildeweerts , @romanlutz ]

adrinjalali:
It'd be nice to have some of these examples in our documentation.

romanlutz:
@MiroDudik this is awesome! I assume this is because `MetricFrame`'s `by_group` is returned as a `DataFrame` which has `plot`: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.plot.html

@rishabhsamb FYI this could be useful!

romanlutz:
Closing this PR since #766 covers this with much less custom code. Thanks @kevinrobinson for getting the ball rolling on this! I strongly doubt we'd be this far in the transition away from the dashboard if not for this work.

PR #566, opened 2020-08-31
Sociotechnical meeting notes, 8/27/20

kevinrobinson:
Adding these in for reference.  My vote is to not put these on the website, but just to leave them in GitHub for if folks want to refer back, so this PR doesn't update the website structure.


romanlutz:
@kevinrobinson  I'm not a huge fan of using PRs as a storage for notes. We can put such things in the "Wiki" section, perhaps, or just keep them in hackpad if people don't like the wiki.

kevinrobinson:
@romanlutz ah I was just trying to follow what you have done for the developer calls, I'll close and keep in email then.

PR #568, opened 2020-09-02
Remove Windows 3.8 from Nightly-Fixed

riedgar-ms:
For reasons which are somewhat obscure but appear [related to a recent `setuptools` update](https://github.com/pypa/setuptools/issues/2368), the `Nightly-Fixed` build for Python 3.8 on Windows is failing to install. Since placing an upper bound on `setuptools` does not mitigate the issue, temporarily remove the `Nightly-Fixed` builds for Python 3.8 on Windows.


riedgar-ms:
Looks like `setuptools` has been updated to address the issue. This is no longer needed

PR #579, opened 2020-09-18
Testing Infrastructure for other ML Packages

riedgar-ms:
Starting work on the infrastructure required for testing other machine learning packages following the [document in the Fairlearn Proposals repository](https://github.com/fairlearn/fairlearn-proposals/blob/master/documentation/OtherMLPackages.md).

The tests are very much smoke tests, which mainly check that no exceptions are thrown when the Fairlearn mitigation algorithms use estimators from other packages.


riedgar-ms:
This only has smoke tests for LightGBM right now; I'd like to get this checked in, and then start getting the pipelines set up. Those require me to work from the Fairlearn repo itself, rather than a fork, so I wanted to do this 'conventionally' first.

setup.cfg
```diff
@@ -27,4 +27,6 @@ per-file-ignores =
     scripts/*.py:D100,D101,D102,D103,D104,D107,D205,D400,D401,RST201,RST205,RST301
     test/*.py:D104
     test/*/*.py:D100,D101,D102,D103,D104,D105,D107
+    test_othermlpackages/*.py:D104
```
romanlutz, line 30: not very fond of this name... I presume it's not test/othermlpackages because that would mean you can't run `pytest test/` in the pipelines?

test_othermlpackages/ReadMe.md
```diff
@@ -0,0 +1,18 @@
+# Tests for other Machine Learning Packages
+
+The main test suite for Fairlearn uses estimators from SciKit-Learn.
+However, we wish to ensure that Fairlearn works with other machine learning packages, provided they implement the SciKit-Learn Estimator interface (with `sample_weight=` on the `fit()` method).
+This directory contains tests which demonstrate Fairlearn working with various other packages.
+
+We have made this a separate directory because machine learning packages tend to be heavyweight, and difference packages may have conflicting dependencies.
```
romanlutz, line 7: typo "different" instead of "difference"

test_othermlpackages/lightgbm/conda-env.yaml
```diff
@@ -0,0 +1,12 @@
+name: fairlearn-lightgm
```
romanlutz, line 1: typo: gbm

test_othermlpackages/lightgbm/test_lightgbm.py
```diff
@@ -0,0 +1,41 @@
...
+import pytest
+
+import test_othermlpackages.package_test_common as ptc
+
+from fairlearn.reductions import DemographicParity
+
+try:
+    import lightgbm as lgb
+except ImportError:
+    pytestmark = pytest.mark.skip(reason="LightGBM not installed")
```
romanlutz, line 13: Won't these be executed deliberately? We may just accidentally skip them and not notice with this code.

test_othermlpackages/package_test_common.py
```diff
@@ -0,0 +1,59 @@
...
+    expgrad = ExponentiatedGradient(
+        estimator,
+        constraints=moment)
+    expgrad.fit(X, Y, sensitive_features=A)
+
+    assert expgrad.n_oracle_calls_ > 1
+    assert len(expgrad.predictors_) > 1
+
+
+def run_gridsearch_classification(estimator, moment):
```
romanlutz, line 48: ThresholdOptimizer?

test_othermlpackages/lightgbm/test_lightgbm.py
riedgar-ms: I'm not sure what you mean. This lets you run `python -m pytest test_othermlpackages` and it will automatically skip any tests which don't have their dependencies installed.

test_othermlpackages/lightgbm/conda-env.yaml
riedgar-ms: Thanks

test_othermlpackages/ReadMe.md
riedgar-ms: Thanks

setup.cfg
riedgar-ms: Yes, although.... with the `ImportException` guards they should be harmless now.

test_othermlpackages/ReadMe.md
```diff
@@ -0,0 +1,18 @@
+# Tests for other Machine Learning Packages
+
+The main test suite for Fairlearn uses estimators from SciKit-Learn.
+However, we wish to ensure that Fairlearn works with other machine learning packages, provided they implement the SciKit-Learn Estimator interface (with `sample_weight=` on the `fit()` method).
```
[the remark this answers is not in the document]
riedgar-ms: Fixed

setup.cfg
romanlutz: Then... should they just be under `test/` instead?

test_othermlpackages/lightgbm/test_lightgbm.py
romanlutz: ... which means it'll pass even if something isn't set up right and the installation doesn't have the right packages (for whatever reason). What we really want is to deliberately run it, and then it should never skip, right?

riedgar-ms: Point taken. This was my workaround to allow sharing of the common code while putting tests in their own directory. But yes, as you say, if the installation is bad on the test machine, then we will have spuriously green test runs.

setup.cfg
riedgar-ms: This goes with the other comment.... something has to give. If we keep the `ImportException` guards, then all these can certainly move under `test/`. However, I am more persuaded by your point that those guards could get us spuriously green builds; I'd rather take them out, work around the common code issue, and then keep the separate directory.

romanlutz: Indeed! So ignore this comment then :-)

test_othermlpackages/lightgbm/test_lightgbm.py
romanlutz: Agreed! I've used this workaround plenty of times, but in this case it may be better to fail explicitly :-)

romanlutz (approved):
Thanks for setting this up!

test_othermlpackages/lightgbm/test_lightgbm.py
riedgar-ms: Directory structure flattened, and this `try` block removed.

test_othermlpackages/package_test_common.py
riedgar-ms: Smoke test added

PR #582, opened 2020-09-22
Re-revised metrics API

riedgar-ms:
A new API for metrics, implementing support for multiple sensitive features, multiple metrics, and also adding control features. The internals might not be the most efficient, but the tests should define a stable API which can be used in refactoring.


riedgar-ms:
Tagging @adrinjalali @hildeweerts @amueller for information

fairlearn/metrics/experimental/_grouped_metric.py
```diff
@@ -0,0 +1,224 @@
...
+            for i in range(len(metric_functions)):
+                fc = FunctionContainer(metric_functions[i], None, sp_list[i], p_list[i])
+                assert fc.name_ not in func_dict
+                func_dict[fc.name_] = fc
+        else:
+            fc = FunctionContainer(metric_functions, None, sample_params, params)
+            func_dict[fc.name_] = fc
+        return func_dict
+
+    def _process_features(self, base_name, features, expected_length):
```
adrinjalali, line 172: You probably can use sklearn's `check_consistent_length` around here a bit.

adrinjalali (commented):
skimming through the code, a few things I noticed.

fairlearn/metrics/experimental/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+from ._derived_metrics import make_derived_metric
+from ._grouped_metric import GroupedMetric
+from . import _function_container
+from . import _group_feature
+
+__all__ = [
+    'GroupedMetric',
+    'make_derived_metric',
+    '_function_container',
+    '_group_feature'
```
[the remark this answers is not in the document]
riedgar-ms: Don't I have to do that, in order to be able to test them?

fairlearn/metrics/experimental/_function_container.py
```diff
@@ -0,0 +1,78 @@
...
+        for name, value in self.sample_params_.items():
+            curr_params[name] = value[mask]
+        return curr_params
+
+    def evaluate(self, y_true, y_pred, mask):
+        """Read a placeholder comment."""
+        assert isinstance(y_true, np.ndarray)
+        assert isinstance(y_pred, np.ndarray)
+        assert len(y_true) == len(y_pred)
+        assert len(y_true) == len(mask)
```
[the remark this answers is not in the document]
riedgar-ms: Yes, these `assert` statements would need to be converted.

fairlearn/metrics/experimental/_grouped_metric.py
riedgar-ms: Thanks - I'll look into it.

riedgar-ms: Have switched over, thanks!

fairlearn/metrics/experimental/_grouped_metric.py
```diff
@@ -0,0 +1,208 @@
...
+        if self._cf_names is None:
+            result = pd.Series(index=self.by_group.columns, dtype='object')
+            for m in result.index:
+                min_val = self.by_group[m].min()
+                result[m] = min_val
+        else:
+            result = self.by_group.groupby(level=list(range(len(self._cf_names)))).min()
+        return result
+
+    def difference(self, method='minmax'):
```
riedgar-ms, line 110: Suggestion from @romanlutz  to not have a default.

fairlearn/metrics/_metric_frame.py
```diff
@@ -0,0 +1,477 @@
...
+    This class also supports the concept of 'control features.' Like the sensitive
+    features, control features identify subgroups within the data, but
+    aggregations are not performed over the control features. Instead, the
+    aggregations produce a result for each subgroup identified by the control
+    feature(s). The name 'control features' refers to the statistical practice
+    of 'controlling' for a variable.
+
+    Parameters
+    ----------
+    metric_functions : callable or dict
```
MiroDudik, line 62: let's rename to `metric` this will allow more flexibility in future.

fairlearn/metrics/_metric_frame.py
```diff
@@ -0,0 +1,477 @@
...
+        The underlying metric functions which are to be calculated. This
+        can either be a single metric function or a dictionary of functions.
+        These functions must be callable as
+        ``fn(y_true, y_pred, **sample_params)``.
+        If there are any other arguments required (such as ``beta`` for
+        :func:`sklearn.metrics.fbeta_score`) then
+        :func:`functools.partial` must be used.
+
+    y_true : array_like
```
MiroDudik, line 71: I think we should be a bit more careful here. Do you mean [sklearn's array-like](https://scikit-learn.org/stable/glossary.html#term-array-like)? Or do you mean [numpy array_like](https://numpy.org/doc/stable/user/basics.creation.html#converting-python-array-like-objects-to-numpy-arrays) which is somewhat [vacuous](https://stackoverflow.com/questions/40378427/numpy-formal-definition-of-array-like-objects). I would prefer to be more precise and say (for example) `list`, `1d array`, `pandas.Series`. The reason for nit-picking here is that I'd like to understand what's our compatibility contract for future. I'd like to avoid being in the situation where we define our contract based on "what our code happens to support currently" (which could be too idiosyncratic to maintain).

fairlearn/metrics/_metric_frame.py
```diff
@@ -0,0 +1,477 @@
...
+        The true values
+
+    y_pred : array_like
+        The predicted values
+
+    sensitive_features : It's complicated
```
MiroDudik, line 77: List the actual types here and then give logical constraints in the description.

fairlearn/metrics/_metric_frame.py
```diff
@@ -0,0 +1,477 @@
...
+        These functions must be callable as
+        ``fn(y_true, y_pred, **sample_params)``.
+        If there are any other arguments required (such as ``beta`` for
+        :func:`sklearn.metrics.fbeta_score`) then
+        :func:`functools.partial` must be used.
+
+    y_true : array_like
+        The true values
+
+    y_pred : array_like
```
MiroDudik, line 74: See above.

fairlearn/metrics/_metric_frame.py
```diff
@@ -0,0 +1,477 @@
...
+        cases, names of the form ``sensitive_feature_[n]`` will be automatically
+        generated. If the input type is :class:`np.ndarray` and there are
+        two dimensions (after squeezing) then multiple sensitive features
+        are generated. Acceptable 'named' types are :class:`pd.Series`,
+        :class:`pd.DataFrame` and dictionaries of 1d array-likes.
+        In all cases, the names *must* be strings.
+        We also forbid DataFrames with column names of ``None``.
+
+    control_features : It's complicated
+        Control features are similar to sensitive features, in that they
```
MiroDudik, line 91: See above

fairlearn/metrics/_metric_frame.py
```diff
@@ -0,0 +1,477 @@
...
+    y_true : array_like
+        The true values
+
+    y_pred : array_like
+        The predicted values
+
+    sensitive_features : It's complicated
+        The sensitive features which should be used to create the subgroups.
+        At least one sensitive feature must be provided. There are several
+        possible input types. Acceptable 'unnamed' types are lists (of
```
MiroDudik, line 80: I think that "named" and "unnamed" distinction is not too relevant here. I think it would be clearer to instead separate out cases how to provide a single feature:
* `array1d-like`, which is one of: `list`, `array1d`, `pd.Series`

Or multiple features:
* `array2d`, `pd.DataFrame`, `dict` of array1d-like

I would talk about naming below that (unless there's a better place?).

fairlearn/metrics/_metric_frame.py
riedgar-ms: It's the same as it's always been - we feed it into `numpy.asarray()` and then `numpy.squeeze()`. I can document that that is what happens.

fairlearn/metrics/_metric_frame.py
riedgar-ms: What flexibility do you think we're missing with `metric_functions` ?

fairlearn/metrics/_metric_frame.py
riedgar-ms: Redrafted.

fairlearn/metrics/_metric_frame.py
riedgar-ms: Redrafted

fairlearn/metrics/_metric_frame.py
riedgar-ms: Redrafted

fairlearn/metrics/_metric_frame.py
riedgar-ms: Redrafted

fairlearn/metrics/_metric_frame.py
```diff
@@ -0,0 +1,477 @@
...
+_DUPLICATE_FEATURE_NAME = "Detected duplicate feature name: '{0}'"
+_TOO_MANY_FEATURE_DIMS = "Feature array has too many dimensions"
+_SAMPLE_PARAM_KEYS_NOT_IN_FUNC_DICT = \
+    "Keys in 'sample_params' do not match those in 'metric_functions'"
+
+
+class MetricFrame:
+    """Contains a collection of group fairness metrics.
+
+    This class holds group fairness metrics calculated for any number of underlying
```
MiroDudik, line 32: Let's call this "data structure to store and manipulate disaggregated metrics". I think that using the word `class` is actually a bit confusing. You can take a quick look at [pandas.DataFrame doc](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.html).

riedgar-ms: Changed. And I eliminated a 'class' later in the documentation block as well.

fairlearn/metrics/_metric_frame.py
MiroDudik: putting in strings or scorer objects?

riedgar-ms: OK. I'll change it.

fairlearn/metrics/_metric_frame.py
```diff
@@ -0,0 +1,482 @@
...
+
+        If a particular combination of classes was not present in the dataset
+        (likely to occur as more sensitive and control features
+        are specified), then the corresponding entry in the DataFrame will
+        be NaN.
+        """
+        return self._by_group
+
+    @property
+    def control_features(self) -> List[str]:
```
MiroDudik, line 230: this should be either `control_feature_names` or `control_levels`. i much prefer latter for conciseness and also for the context where it is used.

fairlearn/metrics/_metric_frame.py
```diff
@@ -0,0 +1,482 @@
...
+
+                # Only call the metric function if the mask is non-empty
+                if sum(mask) > 0:
+                    curr_metric = func_dict[func_name].evaluate(y_true, y_pred, mask)
+                    result[func_name][row_curr] = curr_metric
+        return result
+
+    @property
+    def overall(self) -> Union[pd.Series, pd.DataFrame]:
+        """Return the underlying metrics evaluated on the whole dataset.
```
MiroDudik, line 198: For the time being, maybe we can still follow the convention https://numpydoc.readthedocs.io/en/latest/format.html#method-docstrings
It might be nice to have a standard  `Returns` section with the return types. The current rendering is:
`Union[pandas.core.series.Series, pandas.core.frame.DataFrame]`
which is not very user friendly.

fairlearn/metrics/_metric_frame.py
```diff
@@ -0,0 +1,482 @@
...
+            result = pd.Series(index=self.by_group.columns, dtype='object')
+            for m in result.index:
+                min_val = self.by_group[m].min()
+                result[m] = min_val
+        else:
+            result = self.by_group.groupby(level=self.control_features).min()
+        return result
+
+    def difference(self,
+                   method: str) -> Union[pd.Series, pd.DataFrame]:
```
MiroDudik, line 297: Let's add a standard `Returns` section. The current rendering is:
`Union[pandas.core.series.Series, pandas.core.frame.DataFrame]`
which is not very user friendly.

fairlearn/metrics/_metric_frame.py
riedgar-ms: I prefer `control_levels`; I started off calling it `control_feature_names` but was asked to change.

fairlearn/metrics/_metric_frame.py
riedgar-ms: Added them.

fairlearn/metrics/_metric_frame.py
```diff
@@ -0,0 +1,515 @@
...
+
+            If a particular combination of classes was not present in the dataset
+            (likely to occur as more sensitive and control features
+            are specified), then the corresponding entry in the DataFrame will
+            be NaN.
+        """
+        return self._by_group
+
+    @property
+    def control_levels(self) -> List[str]:
```
MiroDudik, line 225: Right now we are returning `[]` when no control features are provided. I think that we should return `None` in that case, so we can write things like
`self.by_group.max(level=self.control_levels)`.

fairlearn/metrics/_metric_frame.py
```diff
@@ -0,0 +1,515 @@
...
+        Returns
+        -------
+        pandas.Series or pandas.DataFrame
+            If there are control features, then the result is a
+            :class:`pandas.DataFrame` with columns corresponding to the
+            underlying metrics, and rows indexed by the classes identified
+            by the control features. Otherwise, a :class:`pandas.Series`
+            indexed by the names of the underlying metrics.
+        """
+        if not self.control_levels:
```
MiroDudik, line 274: this should simplify to `return self.by_group.max(level=control_levels)` once we update the default `control_levels`.

fairlearn/metrics/_metric_frame.py
```diff
@@ -0,0 +1,515 @@
...
```
MiroDudik, line 302: simplify to one line

fairlearn/metrics/_metric_frame.py
```diff
@@ -0,0 +1,515 @@
...
+        pandas.Series or pandas.DataFrame
+            If there are control features, then the result is a
+            :class:`pandas.DataFrame` with columns corresponding to the
+            underlying metrics, and rows indexed by the classes identified
+            by the control features. Otherwise, a :class:`pandas.Series`
+            indexed by the names of the underlying metrics.
+        """
+        subtrahend = np.nan
+        if method == 'between_groups':
+            subtrahend = self.group_min()
```
MiroDudik, line 346: simplify to `return self.group_max() - self.group_min()`?

fairlearn/metrics/_metric_frame.py
```diff
@@ -0,0 +1,515 @@
...
+        elif method == 'to_overall':
+            subtrahend = self.overall
```
MiroDudik, line 348: simplify to `return (self.by_group - self.overall).abs().max(level=control_levels)`?

fairlearn/metrics/_metric_frame.py
```diff
@@ -0,0 +1,515 @@
...
+        any two pairs of groups in the ``by_groups`` DataFrame (i.e.
+        ``group_max() - group_min()``). Alternatively, ``to_overall``
+        computes the difference between each subgroup and the
+        corresponding value from ``overall`` (if there are control
+        features, then ``overall`` is multivalued for each metric).
+        The result is the absolute maximum of these values.
+
+        Returns
+        -------
+        pandas.Series or pandas.DataFrame
```
MiroDudik, line 337: what are the return types if only one metric is provided? i'm assuming that for one metric we return `pandas.Series` if there are some control features, and a scalar if there are no control features?

riedgar-ms: The return types are exactly the same as if two or more metrics are provided - a Series or a DataFrame

fairlearn/metrics/_metric_frame.py
```diff
@@ -0,0 +1,515 @@
...
+        """
+        return self._overall
+
+    @property
+    def by_group(self) -> pd.DataFrame:
+        """Return the collection of metrics evaluated for each subgroup.
+
+        Returns
+        -------
+        pandas.DataFrame
```
MiroDudik, line 208: I'm now really wondering whether we shouldn't return `pandas.Series` when we have only one metric. This would be consistent with returning just a scalar with `overall`, `difference`, `ratio`, `group_min`, `group_max` in that case as well.

riedgar-ms: That's quite a change you're proposing.

MiroDudik: i resurfaced this on gitter. i think that there are good usability reasons for the change.

fairlearn/metrics/_metric_frame.py
riedgar-ms: That's an interesting point actually. The following two work:
```python
mf.by_group.max(level=None)
mf.by_group.unstack(level=[])
```
The following two fail:
```python
mf.by_group.max(level=[])
mf.by_group.unstack(level=None)
```
So, the question is more 'which set of methods do we prioritise?'

fairlearn/metrics/_metric_frame.py
riedgar-ms: Also, having `.group_max()` give a scalar if there is only one result would be inconsistent with plain pandas:
```python
>>> import pandas as pd
>>> df = pd.DataFrame(data=[0,1,2,3,4])
>>> df.max()
0    4
dtype: int64
>>> type(df.max())
<class 'pandas.core.series.Series'>
```
I'm not a fan of return types changing with input data; the reason I tolerate it to some extent here is that for the no-control-features case, I agree that there's no good name for the single row of the DataFrame. However, contracting to a scalar for things like `.max()` when pandas itself does not seems unwise. And given that, I think that leaving `.overall` as a single element Series is better for consistency that quietly dropping it to a scalar.

fairlearn/metrics/_metric_frame.py
MiroDudik: I know. I think that it's better to enable `max`, `min`, `mean`, ... than `unstack`.

fairlearn/metrics/_metric_frame.py
MiroDudik: If `by_group` is a `pandas.Series` (as I suggest for this case) then `by_group.max()` is a scalar.

MiroDudik: For `by_group`, I think that we should have the following behavior:

* if `metric` is a callable then `by_group` is `pandas.Series` without any name (i.e., `by_group.name == None`)
* if `metric` is a dictionary then `by_group` is `pandas.DataFrame` (even if the dictionary has only one element)

I think that this would nicely handle both the use case when the data scientist wants to analyze a single metric as well as when `MetricFrame` is used as part of another system. It is also consistent with the pandas behavior:
```python
scores = pd.Series(data=[0, 1, 1], index=['a','b','a'])

scores.groupby(level=0).agg('mean').__class__   # pandas.Series
scores.groupby(level=0).agg(['mean','min']).__class__   # pandas.DataFrame
scores.groupby(level=0).agg(['mean']).__class__   # pandas.DataFrame
```

fairlearn/metrics/_metric_frame.py
MiroDudik: Resolving to `None`.

fairlearn/metrics/_metric_frame.py
MiroDudik: Resolved to accept this proposal.

fairlearn/metrics/_metric_frame.py
riedgar-ms: Some simplifications done, but internals can wait on a refactor.

fairlearn/metrics/_metric_frame.py
riedgar-ms: Some simplifications done, but internals can wait on a refactor

fairlearn/metrics/_metric_frame.py
riedgar-ms: There has been some simplification :-)

fairlearn/metrics/_metric_frame.py
riedgar-ms: Have done some simplification.

fairlearn/metrics/_metric_frame.py
riedgar-ms: Adjusted as agreed at the meeting.

fairlearn/metrics/_metric_frame.py
```diff
@@ -0,0 +1,552 @@
...
+_FEATURE_LIST_NONSCALAR = "Feature lists must be of scalar types"
+_FEATURE_DF_COLUMN_BAD_NAME = "DataFrame column names must be strings. Name '{0}' is of type {1}"
+_DUPLICATE_FEATURE_NAME = "Detected duplicate feature name: '{0}'"
+_TOO_MANY_FEATURE_DIMS = "Feature array has too many dimensions"
+_SAMPLE_PARAM_KEYS_NOT_IN_FUNC_DICT = \
+    "Keys in 'sample_params' do not match those in 'metric'"
+
+
+class MetricFrame:
+    """Contains a collection of group fairness metrics.
```
MiroDudik, line 30: -> `Collection of disaggregated metric values.`

fairlearn/metrics/_metric_frame.py
```diff
@@ -0,0 +1,552 @@
...
+    This data structure stores and manipulates disaggregated values for any number of underlying
+    metrics. At least one sensitive feature must be supplied, which is used
+    to split the data into subgroups. The underlying metric(s) is(are) calculated
+    across the entire dataset (made available by the :attr:`.overall` property) and
+    for each identified subgroup (made available by the :attr:`.by_group` property).
+
+    The only limitations placed on the metric functions are that:
+
+    * The first two arguments they take must be ``y_true`` and ``y_pred`` arrays
+    * Any other arguments must be *sample* based, and will be split up along with
```
MiroDudik, line 41: ->
Any other arguments must correspond to sample properties (such as sample weights), meaning that their first dimension is the same as that of `y_true` and `y_pred`.

fairlearn/metrics/_metric_frame.py
```diff
@@ -0,0 +1,552 @@
...
+      the ``y_true`` and ``y_pred`` arrays
+
+    The interpretation of the ``y_true`` and ``y_pred`` arrays is up to the
+    underlying metric - it is perfectly possible to pass in lists of class
+    probability tuples. The return value of the metric is also very flexible;
```
MiroDudik, line 46: Are we relying anywhere on the non-numeric scalar return values? If we are not, I would just say that if the metric is returning something else than int or float, it is currently supported (modulo that some aggregators are not well defined) but that the behavior might change in future.

The reason for this is that there might be some good use-cases when metric returns `pd.Series` or other 1d-array-like, say the metric value + standard error, in which case we may want to create a hierarchical column index rather than keep the series inside the cell.

fairlearn/metrics/_metric_frame.py
```diff
@@ -0,0 +1,552 @@
...
+    there is no problem around returning confusion matrices.
+
+    Once calculated, scalar group fairness metrics can be aggregated over
```
MiroDudik, line 49: ->
Group fairness metrics are obtained by methods that implement
various aggregators over group-level metrics, such such as the
maximum, minimum, or the worst-case difference or ratio.

fairlearn/metrics/_metric_frame.py
```diff
@@ -0,0 +1,552 @@
...
+        :func:`sklearn.metrics.fbeta_score`) then
+        :func:`functools.partial` must be used.
+
+        **Note** that the values returned by various members of the class change
+        based on whether this argument is a callable or a dictionary of
+        callables. This distinction remains *even if* the dictionary only
+        contains a single entry.
+
+    y_true : List, pandas.Series, numpy.ndarray, pandas.DataFrame
+        The true values.
```
MiroDudik, line 77: values -> labels

fairlearn/metrics/_metric_frame.py
```diff
@@ -0,0 +1,552 @@
...
+
+    y_pred : List, pandas.Series, numpy.ndarray, pandas.DataFrame
+        The predicted values.
```
MiroDudik, line 80: predicted values -> predictions

fairlearn/metrics/_metric_frame.py
```diff
@@ -0,0 +1,552 @@
...
+        Control features can be specified similarly to the sensitive features.
+        However, their default names (if none can be identified in the
+        input values) are of the format ``control_feature_[n]``.
+
+        **Note** the types returned by members of the class vary based on whether
+        control features are present.
+
+    sample_params : dict
+        Parameters for the metric function(s). If there is only one metric function,
+        then this is a dictionary of strings and array_likes, which are split
```
MiroDudik, line 105: array_likes -> array-like objects
(i think we moved away from array-likes? but i'm also okay to keep as is...)

fairlearn/metrics/_metric_frame.py
```diff
@@ -0,0 +1,552 @@
...
+        alongside the ``y_true`` and ``y_pred`` arrays, and passed to the metric function.
+        If there are multiple metric functions (passed as a dictionary), then this is
+        a nested dictionary, with the first set of string keys identifying the
+        metric function name, with the values being the string-to-array_like dictionaries.
```
MiroDudik, line 109: just flagging `array_like` in case you'd like to change it, but i'm actually fine keeping as is.

docs/user_guide/mitigation.rst
```diff
@@ -527,16 +555,25 @@ Group :code:`"a"` has an average loss of :math:`0.05`, while group
...
-    >>> from fairlearn.metrics import mean_absolute_error_group_summary
+    >>> from sklearn.metrics import mean_absolute_error
     >>> bgl = BoundedGroupLoss(ZeroOneLoss(), upper_bound=0.1)
     >>> X                  = np.array([[0], [1], [2], [3]])
     >>> y_true             = np.array([0.3, 0.5, 0.1, 1.0])
     >>> y_pred             = np.array([0.3, 0.6, 0.6, 0.5])
     >>> sensitive_features = np.array(["a", "a", "b", "b"])
-    >>> mean_absolute_error_group_summary(y_true, y_pred, sensitive_features=sensitive_features)
-    {'overall': 0.275, 'by_group': {'a': 0.0499..., 'b': 0.5}}
+    >>> mae_summary = MetricFrame(mean_absolute_error,
```
romanlutz, line 567: Nit: do we still want to call it summary even though the name changed?

fairlearn/metrics/__init__.py
```diff
@@ -30,62 +42,28 @@
...
 
 _extra_metrics = [
     "true_positive_rate",
     "true_negative_rate",
     "false_positive_rate",
     "false_negative_rate",
-    "balanced_root_mean_squared_error",
     "mean_prediction",
     "selection_rate",
-    "_mean_overprediction",
```
romanlutz, line 56: What happened to over and underprediction?

romanlutz (approved):

fairlearn/metrics/__init__.py
romanlutz: Ah looks like it just moved a bit.

fairlearn/metrics/_metric_frame.py
riedgar-ms: Done

fairlearn/metrics/_metric_frame.py
```diff
@@ -0,0 +1,552 @@
+# Copyright (c) Microsoft Corporation and Fairlearn contributors.
+# Licensed under the MIT License.
+
+import copy
+import logging
+import numpy as np
+import pandas as pd
+from typing import Any, Callable, Dict, List, Optional, Union
```
MiroDudik, line 8: `Callable`, `Dict` and `List` are going to be [deprecated](https://docs.python.org/3/library/typing.html#module-contents). Okay to keep for now, but maybe pro-actively create an issue?

fairlearn/metrics/_metric_frame.py
riedgar-ms: Updated.

fairlearn/metrics/_metric_frame.py
riedgar-ms: I can tweak that. There is a test which makes sure everything works with confusion matrices (see `test_metricframe_nonscalar.py`).

fairlearn/metrics/_metric_frame.py
riedgar-ms: Updated.

fairlearn/metrics/_metric_frame.py
```diff
@@ -0,0 +1,552 @@
...
+                    mask = self._mask_from_tuple((row_curr,), rows)
+
+                # Only call the metric function if the mask is non-empty
+                if sum(mask) > 0:
+                    curr_metric = func_dict[func_name].evaluate(y_true, y_pred, mask)
+                    result[func_name][row_curr] = curr_metric
+        return result
+
+    @property
+    def overall(self) -> Union[pd.Series, pd.DataFrame]:
```
MiroDudik, line 197: type descriptor is incorrect. also it's incorrect for other aggregators.

fairlearn/metrics/_metric_frame.py
```diff
@@ -0,0 +1,552 @@
...
+        if self._user_supplied_callable:
+            if self.control_levels:
+                return self._overall.iloc[:, 0]
+            else:
+                return self._overall.iloc[0]
+        else:
+            return self._overall
+
+    @property
+    def by_group(self) -> pd.DataFrame:
```
MiroDudik, line 236: type descriptor is incorrect

fairlearn/metrics/_metric_frame.py
riedgar-ms: Changed.

fairlearn/metrics/_metric_frame.py
riedgar-ms: Switched.

fairlearn/metrics/_metric_frame.py
riedgar-ms: Switched anyway.

fairlearn/metrics/_metric_frame.py
riedgar-ms: Fixed, thanks.

fairlearn/metrics/_metric_frame.py
riedgar-ms: Fixed.

fairlearn/metrics/_metric_frame.py
riedgar-ms: I've fixed `Dict` and `List`... will create issue for Callable, since that could get interesting.

docs/user_guide/mitigation.rst
riedgar-ms: Adjusted.

fairlearn/metrics/_metric_frame.py
riedgar-ms: Will create the issue once this is in. Fixing isn't as easy as I'd hoped.

fairlearn/metrics/_metric_frame.py
riedgar-ms: Updated as agreed

CHANGES.md
```diff
@@ -44,6 +44,7 @@
 * Change the method to make copies of the estimator in `ExponentiatedGradient` from `pickle.dump` to `sklearn.clone`. 
 * Add an argument to GridSearch and ExponentialGradient `sample_weight_name` to control
   how `sample_weight` is supplied to `estimator.fit`.
+* Large changes to the metrics API. Please see the documentation and examples for more information.
```
MiroDudik, line 47: if we are dropping functions like all of the automatically derived metrics and all of the `make_*`, we should say so!

riedgar-ms: I can add the derived metrics we generated back in, using the old backend. Adding a 'how to transition' document is a follow up.

test/unit/metrics/test_metricframe_process_features.py
```diff
@@ -0,0 +1,168 @@
...
+# Licensed under the MIT License.
+
+import numpy as np
+import pandas as pd
+import pytest
+
+import fairlearn.metrics as metrics
+from .utils import _get_raw_MetricFrame
+
+
```
MiroDudik, line 11: i would prefer if we were able to write down these tests as tests of the **external** API, rather than an internal API, which i don't expect to be stable. but for now, happy to keep things as is.

riedgar-ms: The cases are also covered through external tests, but this gives an easier way of doing them systematically. Since we only really have one entry point (the constructor), lots can go wrong in a test of that, whereas this lets us target specific portions of it. I believe that `_process_features` could technically be moved out of `MetricFrame` and made a 'module private' rather than 'class private' if that's a particular concern.

fairlearn/metrics/__init__.py
```diff
@@ -10,14 +10,29 @@
 arguments may be present (most often sample weights), which will
 affect how the metric is calculated.
 
-The group metrics in this module have signatures
-``g(y_true, y_pred, group_membership, ...)``
-where ``group_membership`` is an array of values indicating
-a group to which each pair of true and predicted values belong.
-The metric is evaluated for the entire set of data, and also
-for each subgroup identified in ``group_membership``.
+This module provides the concept of a *grouped metric*.
```
MiroDudik, line 13: grouped metric -> disaggregated metric

fairlearn/metrics/__init__.py
```diff
@@ -10,14 +10,29 @@
...
+This is a metric where in addition to ``y_true`` and ``y_pred``
+values, the user provides information about group membership
+for each sample.
+For example, a user could provide a 'Gender' column, and the
+grouped metric would contain separate results for the subgroups
```
MiroDudik, line 18: grouped metric -> disaggregated metric

fairlearn/metrics/__init__.py
riedgar-ms: Changed. And in the single line headline at the start of the documentation block.

fairlearn/metrics/__init__.py
riedgar-ms: Fixed.

CHANGES.md
MiroDudik: ok. as long as we have the "transitioning section" i'm good.

MiroDudik: actually--i think it would still be good to mention: which functions we are dropping, and which functions are changing their API. so for now say: "MetricFrame is introduced, and group_summary variants are being removed."

riedgar-ms: Done

fairlearn/metrics/_metric_frame.py
MiroDudik: thanks for updating re. aggregators. do you think we need to say anything about how we might treat array-like types in future?

riedgar-ms: I think that 'at the current time' implies that.

MiroDudik (approved):

test/unit/metrics/test_metricframe_process_features.py
MiroDudik: I understand where this comes from. Just flagging, but no need to address.

PR #584, opened 2020-09-23
Pipeline for testing other ML packages

riedgar-ms:
Add a pipeline for running the new `test_othermlpackages/` directory. This is separate from the main test suite, since each package has its own conda environment file, and hence each has its own separate job.


riedgar-ms:
A sample run of the pipeline:
https://dev.azure.com/responsibleai/fairlearn/_build/results?buildId=8630&view=results

romanlutz (approved):

PR #586, opened 2020-09-25
Smoke tests for TensorFlow

riedgar-ms:
Add TensorFlow to the `test_othermlpackages/` directory. Some product code modifications were required, since it appears that:

- `KerasClassifier.fit()` doesn't return `self`
- `KerasClassifier.predict()` returns an array of shape (n,1) and not of shape (n). This is important when subtracting from a pandas Series

Also tweak the pipeline so that changes to the `test_othermlpackages/` directory should trigger it as a gate.

### DIFF
--- a/test_othermlpackages/conda-tensorflow.yaml
+++ b/test_othermlpackages/conda-tensorflow.yaml
@@ -1,0 +1,9 @@
+name: fairlearn-tensorflow
+channels:
+- defaults
+- conda-forge
+dependencies:
+- python==3.8
+- pip>=19.1.1
+- pytest
+- tensorflow
--- a/test_othermlpackages/conda-tensorflow.yaml
+++ b/test_othermlpackages/conda-tensorflow.yaml
@@ -3,7 +3,7 @@ channels:
 - defaults
 - conda-forge
 dependencies:
-- python==3.8
+- python==3.6
 - pip>=19.1.1
 - pytest
 - tensorflow
--- a/test_othermlpackages/package_test_common.py
+++ b/test_othermlpackages/package_test_common.py
@@ -6,6 +6,7 @@
 import pandas as pd
 
 from sklearn.datasets import fetch_openml
+from sklearn.model_selection import train_test_split
 from sklearn.preprocessing import LabelEncoder, StandardScaler
 
 from fairlearn.postprocessing import ThresholdOptimizer
@@ -29,7 +30,13 @@ def fetch_adult():
     X_scaled = sc.fit_transform(X)
     X_scaled = pd.DataFrame(X_scaled, columns=X.columns)
 
-    return X_scaled, Y, A
+    X_train, X_test, Y_train, Y_test, A_train, A_test = \
+        train_test_split(X_scaled, Y, A,
+                         test_size=0.3,
+                         random_state=12345,
+                         stratify=Y)
+
+    return X_train, Y_train, A_train
 
 
 def run_expgrad_classification(estimator, moment):
--- a/test_othermlpackages/package_test_common.py
+++ b/test_othermlpackages/package_test_common.py
@@ -3,6 +3,8 @@
 
 """Common testing methods for use with other ML packages."""
 
+import copy
+
 import pandas as pd
 
 from sklearn.datasets import fetch_openml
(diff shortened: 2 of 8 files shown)


test_othermlpackages/test_tensorflow.py
```diff
@@ -0,0 +1,42 @@
...
+from fairlearn.reductions import DemographicParity
+
+from tensorflow.keras.layers import Dense
+from tensorflow.keras.models import Sequential
+from tensorflow.keras.wrappers.scikit_learn import KerasClassifier
+
+
+def create_model():
+    # create model
+    model = Sequential()
```
riedgar-ms, line 15: This is largely copied from a tutorial online, so please let me know if there's something better to do.

arjsingh: The structure of the test and the model architecture seem to be fine, but what is this test actually checking? I think you're showing that even with a KerasClassifier, which in turn is a wrapper wrapping a network with fully connected layers and cross entropy loss, the run completes. However, the test would be more robust if you could get the returned model and assert that the shape, of the output tensor after every layer in the neural net, is as expected.

riedgar-ms: What would the output tensor after every layer be expected to be? Sorry, I'd not touched TensorFlow before this morning.

I have just added some extra tests in the common file which verify that the disparities have been reduced as expected.

test_othermlpackages/package_test_common.py
```diff
@@ -29,42 +33,88 @@ def fetch_adult():
...
+    gs.fit(X_train, Y_train, sensitive_features=A_train)
 
     assert len(gs.predictors_) == num_predictors
 
+    verification_moment.load_data(X_test, Y_test, sensitive_features=A_test)
+    gamma_unmitigated = verification_moment.gamma(lambda x: unmitigated.predict(x))
+    gamma_mitigated = verification_moment.gamma(lambda x: gs.predict(x))
+
+    for idx in gamma_mitigated.index:
+        assert abs(gamma_mitigated[idx]) <= abs(gamma_unmitigated[idx]), "Checking {0}".format(idx)
```
riedgar-ms, line 97: @MiroDudik is this a sensible way to check that the reductions have actually done what they're supposed to?

test_othermlpackages/test_tensorflow.py
arjsingh: Could use a combination of tf.get_layer and tf.shape to assert on the shape of a given layer, or it could also be some simple assertions like the ones mentioned here:
https://www.tensorflow.org/api_docs/python/tf/keras/layers/Layer

Not necessary, but only if you'd want to add any tf specific unit testing.

riedgar-ms: I could add that later, but I don't want to test TensorFlow itself - I want to make sure that TensorFlow can be used with Fairlearn.

test_othermlpackages/conda-tensorflow.yaml
```diff
@@ -0,0 +1,9 @@
+name: fairlearn-tensorflow
+channels:
+- defaults
+- conda-forge
+dependencies:
+- python==3.6
+- pip>=19.1.1
+- pytest
+- tensorflow
```
romanlutz, line 9: I presume tensorflow > 2.0?

test_othermlpackages/package_test_common.py
romanlutz: You could run expgrad until convergence and check if the violation is below the bound (assuming that it actually converges as opposed to hitting the iteration limit)

For grid search there are no guarantees like that.

test_othermlpackages/package_test_common.py
```diff
@@ -29,42 +33,88 @@ def fetch_adult():
...
+    dp_diff_unmitigated = demographic_parity_difference(Y_test,
+                                                        unmitigated_predictions,
+                                                        sensitive_features=A_test)
 
-    results = to.predict(X, sensitive_features=A)
-    assert results is not None
+    dp_diff_mitigated = demographic_parity_difference(Y_test,
+                                                      mitigated_predictions,
+                                                      sensitive_features=A_test)
+    assert dp_diff_mitigated <= dp_diff_unmitigated
```
romanlutz, line 120: You can also assert that the disparity on the training data is 0 (or rather super close)

romanlutz (approved):

PR #592, opened 2020-10-02
Start adding PEP484 annotations

riedgar-ms:
Experimenting with [PEP 484](https://www.python.org/dev/peps/pep-0484) annotations to some routines within Fairlearn. Eventually, this will allow us to run a tools such as [`mypy`](https://mypy.readthedocs.io/en/stable/) which can perform static type checking, and could be run as part of our gates.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -43,6 +43,7 @@ release = fairlearn.__version__
 extensions = [
     'bokeh.sphinxext.bokeh_plot',
     'sphinx.ext.autodoc',
+    'sphinx_autodoc_typehints',
     'sphinx.ext.doctest',
     'sphinx.ext.extlinks',
     'sphinx.ext.intersphinx',
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -43,14 +43,14 @@ release = fairlearn.__version__
 extensions = [
     'bokeh.sphinxext.bokeh_plot',
     'sphinx.ext.autodoc',
-    'sphinx_autodoc_typehints',
     'sphinx.ext.doctest',
     'sphinx.ext.extlinks',
     'sphinx.ext.intersphinx',
     'sphinx.ext.linkcode',
     'sphinx.ext.mathjax',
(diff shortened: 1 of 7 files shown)


romanlutz:
> Experimenting with [PEP 484](https://www.python.org/dev/peps/pep-0484) annotations to some routines within Fairlearn. Eventually, this will allow us to run a tools such as [`mypy`](https://mypy.readthedocs.io/en/stable/) which can perform static type checking, and could be run as part of our gates.

A few thoughts:
- If we want to go this route we should document that we want new code to have these annotations, and changes to existing code should add them (just like with numpydoc).
- Is this already used in other packages and what have people's experiences been like? I personally haven't used it yet so I'm very curious. I also want to avoid adding a lot of obstacles or issues with this in case it's not as mature as we'd hope. Just tagging @adrinjalali  @koaning  @hildeweerts  @mbrouns @MiroDudik  @dref360 @amueller
- I'm not super excited about how this looks in the API reference page but perhaps that's more a matter of taste (?) - picture taken from https://1507-133444044-gh.circle-artifacts.com/0/docs/_build/html/api_reference/fairlearn.metrics.html
  
image

adrinjalali:
Some related discussion: https://github.com/scikit-learn/scikit-learn/issues/16705

riedgar-ms:
Thanks @adrinjalali , I think I came across that thread last Friday when I was first playing around with these.... and discovering how slippery the argument types to `numpy.asarray()` really are. The point from @romanlutz  about the ugliness of the `Union` type is well taken, athough... Sphinx ought to be able to recognise that and translate it into something more human-readable.

I certainly don't think that this should be a 'do all at once' effort. It's more a 'convert if you're doing something else with the function' sort of thing. And the note from the Pandas developer on that thread about bugs in MyPy is another concern.

It was someone else at Microsoft who mentioned this to me, and they've found that adding the typing is incrementally useful (at least if MyPy is running regularly).

riedgar-ms:
Leaving be for now.

PR #596, opened 2020-10-04
Add sample_weight_name argument to Reductions

### LINKED ISSUE #580: Can't use sklearn.Pipeline with GridSearch.
Hello,

I may be wrong, but GridSearch does not allow [sklearn.Pipeline](https://scikit-learn.org/stable/modules/generated/sklearn.pipeline.Pipeline.html) for its estimator.
This would be a useful feature have!

## How to reproduce.
In `test/unit/reductions/grid_search/test_grid_search_arguments.py` add:

```python
# Additional Imports
from sklearn.pipeline import Pipeline
from sklearn.preprocessing import StandardScaler

# Set up Pipeline estimator
class TestPipelineEstimator(ConditionalOpportunityTests):
    def setup_method(self, method):
        self.estimator = Pipeline([('scaler', StandardScaler()), ('logistic', LogisticRegression(solver='liblinear'))])
        self.disparity_criterion = DemographicParity()
```

Then : `pytest test/unit/reductions/grid_search/test_grid_search_arguments.py` would fail with:

```
ValueError: Pipeline.fit does not accept the sample_weight parameter. You can pass parameters to specific steps of your pipeline using the stepname__parameter format, e.g. `Pipeline.fit(X, y, logisticregression__sample_weight=sample_weight)`.
```

### Proposed solution

In GridSearch constructor, we could allow specifying the key for `sample_weight`.

Ex. `GridSearch(self.estimator, self.disparity_criterion, grid_size=2, sample_weight_key='logistic__sample_weight')`

I already made the test so I can add the solution too if it is approved.

Dref360:
Allow the user to supply sample_weight_key to better control the call to estimator.fit.

Fixes #580

### DIFF
--- a/fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
+++ b/fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
@@ -38,9 +38,12 @@ class _Lagrangian:
     opt_lambda : bool
         indicates whether to optimize lambda during the calculation of the
         Lagrangian; optional with default value True
+    sample_weight_key : str
+        Argument to supply `sample_weight` in `estimator.fit`.
     """
 
-    def __init__(self, X, sensitive_features, y, estimator, constraints, B, opt_lambda=True):
+    def __init__(self, X, sensitive_features, y, estimator, constraints, B, opt_lambda=True,
+                 sample_weight_key='sample_weight'):
         self.X = X
         self.n = self.X.shape[0]
         self.y = y
@@ -61,6 +64,7 @@ class _Lagrangian:
         self.n_oracle_calls_dummy_returned = 0
         self.last_linprog_n_hs = 0
         self.last_linprog_result = None
+        self.sample_weight_key = sample_weight_key
 
     def _eval(self, Q, lambda_vec):
         """Return the value of the Lagrangian.
@@ -170,7 +174,7 @@ class _Lagrangian:
             estimator = clone(estimator=self.estimator, safe=False)
 
         oracle_call_start_time = time()
-        estimator.fit(self.X, redY, sample_weight=redW)
+        estimator.fit(self.X, redY, **{self.sample_weight_key: redW})
         self.oracle_execution_times.append(time() - oracle_call_start_time)
         self.n_oracle_calls += 1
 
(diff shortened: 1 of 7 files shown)


fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -38,9 +38,12 @@ class _Lagrangian:
     opt_lambda : bool
         indicates whether to optimize lambda during the calculation of the
         Lagrangian; optional with default value True
+    sample_weight_key : str
+        Argument to supply `sample_weight` in `estimator.fit`.
```
riedgar-ms, line 42: Can I suggest "Name of the argument to `estimator.fit()` which supplies the sample weights (defaults to `sample_weight`)"

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -51,17 +51,21 @@ class ExponentiatedGradient(BaseEstimator, MetaEstimatorMixin):
         if True each step of exponentiated gradient is followed by the saddle
         point optimization over the convex hull of classifiers returned so
         far; default True
+    sample_weight_key : str
+        Argument to supply `sample_weight` in `estimator.fit`.
```
riedgar-ms, line 55: See above for suggested change

riedgar-ms (commented):
Should be reviewed by @MiroDudik  too.

riedgar-ms (approved):

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
MiroDudik: I think we should change `sample_weight_key` to `sample_weight_name`.

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
MiroDudik: Similarly, I think we should change `sample_weight_key` to `sample_weight_name`.

romanlutz (approved):
I'm very excited to get these improvements in! There's nothing I want changed in the code, but you should add your name to the AUTHORS.md and add a short description to CHANGES.md

Dref360:
I added my name and a small description. :)

MiroDudik (approved):

PR #597, opened 2020-10-06
Correct random_state behavior in ThresholdOptimizer and ExponentiatedGradient

### LINKED ISSUE #588: Add random_state to ThresholdOptimizer

The results of `ThresholdOptimizer` are currently not reproducible. Although `random_state` in `ThresholdOptimizer.predict()` helps to ensure reproducibility in the predictions, running `fit()` twice will result in a different optimizer object (and hence different predictions).

I'd like a `random_state` argument for the `ThresholdOptimizer` object itself.

romanlutz:
closes #588 

@hildeweerts FYI


test/unit/test_random_state.py
```diff
@@ -0,0 +1,75 @@
...
+
+    # score groups
+    y_pred_test = to.predict(X_test, sensitive_features=race_test, random_state=0)
+    for _ in range(100):
+        assert (y_pred_test ==
+                to.predict(X_test, sensitive_features=race_test, random_state=0)).all()
+    assert (y_pred_test != to.predict(X_test, sensitive_features=race_test, random_state=1)).any()
+
+
+def test_random_state_reductions():
```
riedgar-ms, line 38: Why 'reductions' and not 'exponentiatedgradient'?

MiroDudik (approved):
I've checked the bugfix re. random_state, and the API extension in `exponentiade_gradient.py`. Both look good to me!

test/unit/test_random_state.py
romanlutz: At first I thought both, but then remembered that Grid Search has no `random_state` :-) While cleaning up I must have overlooked this.

riedgar-ms (approved):

Issue #610, opened 2020-10-25
DemographicParity init parameters


chsafouane:
The API reference page for [fairlearn.reductions](https://fairlearn.github.io/api_reference/fairlearn.reductions.html#fairlearn.reductions.DemographicParity) indicates that DemographicParity can take multiple parameters when instatiating and the [code on github](https://github.com/fairlearn/fairlearn/blob/master/fairlearn/reductions/_moments/utility_parity.py#L224-L248) indicates that it does as its base class is `UtilityParity`. Yet, when I try to execute the line instantiating dp in the example in the [user guide](https://fairlearn.github.io/user_guide/mitigation.html)  `dp = DemographicParity(difference_bound=0.01)`, I get this error
`TypeError: __init__() got an unexpected keyword argument 'difference_bound'`.

I installed the package using both pip and conda and in both cases, I get the same error. Is there anything I'm doing wrong ? If it's not the case, where would the problem possibly come from ?

riedgar-ms:
It's been a long time since the last release, and we don't have versioned docs. I believe you've come across one of the places where the docs are far ahead of the PyPI installed version. Does it work if you install Fairlearn from git?

chsafouane:
After installing the package directly from github, the example in the user guide works as a charm. Thanks

Issue #556, opened 2020-08-13
"GroupLossMoment" versus "BoundedGroupLoss"


nina-hua:
Hello, I've installed fairlearn using pip and in the init script, "GroupLossMoment" is present while "BoundedGroupLoss" isn't. Comparing this script to the one found in the Github, they're not consistent, which I'm assuming is because I've installed fairlearn with pip instead of pulling from the repo. Does  "GroupLossMoment" work comparably to "BoundedGroupLoss"?

romanlutz:
Hi @nina-hua ! Your assessment sounds correct :-)
`GroupLossMoment` has been renamed to `BoundedGroupLoss`, but it's also a little bit more general now. GLM is only usable with `GridSearch`, while BGL is usable with `GridSearch` and `ExponentiatedGradient`. [When using BGL with `ExponentiatedGradient` you need to provide the `upper_bound` argument.] Otherwise they should behave the same way.

Please let me know if there are other questions! We definitely need to write a set of changes from v0.4.6 to v-next and publish them on the webpage (besides CHANGES.md). Since it's also related: the multi-version support on the webpage that @riedgar-ms  is working on will be very useful.

nina-hua:
Thanks Roman! Do you have a link to older documentation on how to use GroupLossMoment?

romanlutz:
I don't unfortunately, but you can use it just like `BoundedGroupLoss` except that there's no `upper_bound` argument: https://fairlearn.github.io/user_guide/mitigation.html#bounded-group-loss

It will only work with `GridSearch`, though.

nina-hua:
Thanks Roman! I'm actually trying out GridSearch with the Boston Housing Dataset. But I'm running into an error setting the grid size. I've tried using 10 versus 1000 estimators, but I'm not sure how to mitigate this issue?
`Message:'Generating a grid with {} grid points. It is recommended to use at least {} grid points. Please consider increasing grid_size.'
Arguments: (10, 31828687130226345097944463881396533766429193651030253916189694521162207808802136034115584)`

nina-hua:
Is there a tutorial on how to use GridSearch with regression problems since I've only stumbled upon classification problems?

romanlutz:
We don't have one yet. Wow that's a big number. How many groups do you have in that example? The recommended grid size is exponential in the number of groups.

I also want to point out the fairness issues with the Boston housing dataset. More context is available at this link: https://github.com/koaning/scikit-fairness/issues/31

[...]

Issue #552, opened 2020-08-12
'GridSearch' object has no attribute 'predictors_'


ancibira:
Hi there, I am running the notebook: Binary Classification with the UCI Credit-Card Default Dataset.
When running the gridsearch component I get an error when trying to get the scores. I have made no alterations to the notebook. 

![image](https://user-images.githubusercontent.com/48265599/89964665-2015fc80-dc8e-11ea-979b-1a01feee2409.png)

#### Versions
- OS: Windows
- Browser: Edge
- Python version: 3.7.4
- Fairlearn version: 0.4.6

riedgar-ms:
You may have run into one of our backwards compatibility problems.... did you `pip install` Fairlearn, but download the notebook from the website? If so, then I'm afraid there have been breaking changes (we need to do a new release to fix this, and also provide versioned documentation).

To use the latest notebooks, you need to clone the repository, and install Fairlearn via `pip install -e .`

To use notebooks with the latest Fairlearn release, you need to go to tag `v0.4.6` of the repository, and get the notebooks from that.

ancibira:
Yes I did pip install Fairlearn, but download the notebook from the website. Thank you, I will do as you suggested.

PR #612, opened 2020-10-29
Installation note for Quickstart

riedgar-ms:
Update the Quickstart page with a note about installing Fairlearn directly from GitHub. This should ease issues with examples not working with broken APIs (and this will be helpful information even with versioned documentation).


docs/quickstart.rst
```diff
@@ -23,6 +23,24 @@ Fairlearn is also available on
...
+.. note::
+
+    The Fairlearn API is not completely stable, so example code in 
+    this documentation may not work with every version of Fairlearn.
+    In particular, it may be necessary to install Fairlearn directly
+    from GitHub via:
+
+    .. code-block:: bash
+
+        pip install git+https://github.com/fairlearn/fairlearn.git
```
kevinrobinson, line 35: This seems like a good note, and great to include at the top of the API doc pages.  But for a quickstart, my vote is that we set new users up for success - if they blindly copy the code and run it, the code we show them will work, and that code will show them the value of the library.

Until documentation and release processes are more stable (eg, https://github.com/fairlearn/fairlearn/issues/482, https://github.com/fairlearn/fairlearn/pull/551, https://github.com/fairlearn/fairlearn-proposals/pull/13 and https://github.com/fairlearn/fairlearn-proposals/pull/11), what do you think about making lines 9-14 recommend installing from master, and move the other options with this great note into the existing "Advanced installation instructions" page that the quickstart already links to?

I'm all for the more thorough versioning and release process, and I get that this might not feel like a great long-term solution too.  But this isn't a hypothetical problem - it keeps tripping people up (https://github.com/fairlearn/fairlearn/issues/610, https://github.com/fairlearn/fairlearn/issues/530, https://github.com/fairlearn/fairlearn/pull/532, other folks in gitter).  If the quickstart recommends installing from master, it may also set folks up for success with then navigating through the web documentation and being able to try other parts of the API (eg, https://github.com/fairlearn/fairlearn/issues/448, https://github.com/fairlearn/fairlearn/issues/495, https://github.com/fairlearn/fairlearn/issues/552 and https://github.com/fairlearn/fairlearn/issues/556).

romanlutz: The reason it trips up folks is that they install with pip and quickstart shows the development version's quickstart. That should be fixed with versioned docs, right? [To my knowledge @riedgar-ms  was planning to default to the latest stable version in the docs once v0.5.0 is out.]

@riedgar-ms if you check this in, then the quickstart of every following version will have this at the top. Does that make sense?

Isn't what you're describing the case from lines 23-24?

riedgar-ms: I agree that we should probably make the default-from-homepage docs be the latest release (and automate that update). The quickstart would then match the version users pip installed.

I think that it does make sense to have some sort of note at the top of every quickstart, warning people to match versions. We get too many bugs of this.

As for the pip install... what I wrote is slightly different, since it's pip that does the git clone into one of its own directories. However, I hadn't noticed that bit, and will redraft to make what it point at those advanced instructions.

docs/quickstart.rst
```diff
@@ -23,6 +23,16 @@ Fairlearn is also available on
 For checking out the latest version in our repository check out our
 :ref:`advanced_install`.
 
+.. note::
+
+    The Fairlearn API is not completely stable, so example code in
```
romanlutz, line 28: Two thoughts:
- "not stable" tells me as a user that I don't want to use this. It makes it seem like more of somebody's pet project than what we're doing here. 
- When running an example one should always match versions, that's not caused by Fairlearn being less mature than scikit-learn, for example. If I install a scikit-learn release version and go to their "master" doc page things can break. That's not weird at all, it's expected. That's why their docs point at the last release version. Similarly, I can't expect to install version X and take notebook from version Y and expect that it'll work flawlessly. So the comment is great, but it's not caused by our API being "not completely stable", it's merely making it a bit more frequent.

examples/README.rst
```diff
@@ -8,6 +8,11 @@ examples soon. If you're interested in contributing to existing notebooks or
 adding new ones please consult the guide on
 :ref:`contributing_example_notebooks`.
 
+.. note::
+
+    The Fairlearn API is not completely stable, so if you want to run these
```
romanlutz, line 13: Same comment as above. You should match versions, but not because this isn't stable, but because one always matches versions.

docs/quickstart.rst
riedgar-ms: Yes and no..... I take your point about perhaps avoiding the word 'stable' for the reasons you mention. However, I disagree with your statement about always matching versions. I would expect old versions of SciKit-Learn examples to work with newer versions. And the new version of those examples to work with the old SciKit-Learn. If they don't (at least for a couple of versions back), then that's telling me that I'm getting on an update treadmill, and that would not fill me with enthusiasm either.

I'm not suggesting we go for `debian stable`, but we're seeing a lot of users tripped up by this.

docs/quickstart.rst
```diff
@@ -23,6 +23,16 @@ Fairlearn is also available on
 For checking out the latest version in our repository check out our
 :ref:`advanced_install`.
 
+.. note::
+
+    The Fairlearn API is still evolving, so example code in 
+    this documentation may not work with every version of Fairlearn.
+    Please use the version selector to get to the instructions for
```
romanlutz, line 30: nit: I'd be very specific and say it's in the "top left corner".

I also found that it's apparently possible to update tags: https://git-scm.com/book/en/v2/Git-Basics-Tagging

Given that we're soon on a new version I don't know if it's worth the effort. We should just make sure that the default website points to v0.5.0 once it's out.

romanlutz (approved):

docs/quickstart.rst
romanlutz: Thanks for updating. I really like the way it is right now. One could still try and split hairs over whether it's because of the evolving API or whether that's just a factor that makes it worse, but I think that's besides the point and won't get us anywhere. I can very well live with this :-)

docs/quickstart.rst
riedgar-ms: Not sure what you mean about tagging (although they are certainly mutable in Git). I did consider directing people to the location of the version picker, but that could change in future....

mesameki (approved):

docs/quickstart.rst
```diff
@@ -23,6 +23,16 @@ Fairlearn is also available on
...
+    the appropriate version (unfortunately there are no instructions
```
MiroDudik, line 31: i would just remove the parenthesis starting with `(unfortunately ...`. it's too confusing.

riedgar-ms: Done

PR #614, opened 2020-10-30
Add user guide for fairlearn.postprocessing.ThresholdOptimizer

romanlutz:
This PR finally adds a user guide for `ThresholdOptimizer` which explains some of the pitfalls that users should be aware of when using this. If this reads like you shouldn't really use it other than perhaps as a benchmark to compare with, then I've done this right.


docs/user_guide/mitigation.rst
```diff
@@ -68,6 +68,216 @@ Postprocessing
...
 
+Fairlearn currently supports one postprocessing technique that is based on the
+paper *Equality of Opportunity in Supervised Learning* [#3]_. Unlike other
+mitigation techniques :code:`ThresholdOptimizer` is built to satisfy the
+specified fairness criteria exactly and with no remaining disparity.
+In many cases this comes at the expense of performance, for example, with
+significantly lower accuracy. Regardless, it is a useful data point to compare
+results with.
+
+:code:`ThresholdOptimizer` expects an estimator that provides it with scores.
```
riedgar-ms, line 79: Why not `:class:`.ThresholdOptimizer` so that you get the link? In other places too.

docs/user_guide/mitigation.rst
```diff
@@ -68,6 +68,216 @@ Postprocessing
...
+    >>> y = (data.target == '>50K') * 1
+    >>> sex = data.data['sex']
+    >>> logistic_regression = LogisticRegression()
+    >>> threshold_optimizer = ThresholdOptimizer(estimator=logistic_regression, constraints="equalized_odds", objective="accuracy_score")
+    >>> threshold_optimizer.fit(X, y, sensitive_features=sex)
+    ThresholdOptimizer(constraints='equalized_odds', estimator=LogisticRegression())
+    >>> threshold_optimizer.interpolated_thresholder_._print_thresholds()
+    {
+        "Female": {
+            "p_ignore": 0.13186000824177083,
```
riedgar-ms, line 113: Are these all checked in the `doctest` build?

docs/user_guide/mitigation.rst
```diff
@@ -68,6 +68,216 @@ Postprocessing
...
+
+In other words, we create a linear combination of the binary thresholding rule
+outputs. :math:`p_0` and :math:`p_1` indicate how close we are to the points
+representing the two thresholding rules, and :math:`p_0+p_1=1`. `p_ignore` is
+:math:`0` if the curve does not need to be "pulled" towards the diagonal,
+i.e., when the selected solution lies on the curve itself as opposed to below.
+Above, the curve for "Male" is strictly below the curve for "Female", so
+:math:`p_{ignore}=0` for "Male" and non-zero for "Female".
+
+Importantly, :math:`p_{ignore}` is only required for equalized odds. Equalized
```
riedgar-ms, line 169: Don't you need an \mbox or something equivalent for the subscripted text? Otherwise it will be rendered as algebra?

docs/user_guide/mitigation.rst
```diff
@@ -68,6 +68,216 @@ Postprocessing
...
+- predict 1 if the score is greater than :math:`0.5`.
+- predict 0 with probability :math:`0.999926` if the score is less or equal to
+  :math:`0.5`. This means there is a very low probability of getting label 1
+  regardless of the features.
+
+For "Female" the result includes the :math:`p_{ignore}` term, but one of the
+thresholds is set to always be true
+
+- predict 1 with probability
+  :math:`0.13186 \cdot 0.035 + (1-0.13186) * 0.00563 = 0.0095` if the score is
```
riedgar-ms, line 190: A \cdot and * in the same equation?

riedgar-ms:
Can you add a description to the PR, please?

romanlutz:
> Can you add a description to the PR, please?

It's would be exactly identical to the title. I'm a big fan of descriptions, too, whenever there's anything to say beyond the title.

docs/user_guide/mitigation.rst
```diff
@@ -68,6 +68,218 @@ Postprocessing
 
 .. currentmodule:: fairlearn.postprocessing
 
+Fairlearn currently supports one postprocessing technique that is based on the
+paper *Equality of Opportunity in Supervised Learning* [#3]_. Unlike other
+mitigation techniques :class:`ThresholdOptimizer` is built to satisfy the
+specified fairness criteria exactly and with no remaining disparity.
+In many cases this comes at the expense of performance, for example, with
+significantly lower accuracy. Regardless, it is a useful data point to compare
+results with.
```
riedgar-ms, line 77: Do you want to mention that it also works with a pretrained estimator here? It's sort of implicit in the name, but might be worth mentioning.

docs/user_guide/mitigation.rst
```diff
@@ -68,6 +68,218 @@ Postprocessing
...
+While the output of :class:`ThresholdOptimizer` is binary, the input need not
+be. In fact, real valued input, e.g. from a regressor, provides it with many
+more options to create thresholds. For :math:`n` input data points with
+:math:`m \leq n` different score values it has :math:`m+1` different
+thresholds. At each threshold one can create one of two thresholding rules,
+i.e. functions that indicate which data points get label :code:`1` based on
+their score.
+
+We can think of each thresholding rule as a binary classifier and determine
+their true positive rate and false positive rate.
```
riedgar-ms, line 89: 'their' does not match 'each' - I think you want "its"

docs/user_guide/mitigation.rst
```diff
@@ -68,6 +68,218 @@ Postprocessing
...
+The code below also visualizes the threshold selection process with an ROC
```
riedgar-ms, line 90: Don't need 'also'

docs/user_guide/mitigation.rst
```diff
@@ -68,6 +68,218 @@ Postprocessing
...
+curve. The ROC curves consists of the true and false positive rates for each
```
riedgar-ms, line 91: "The ROC curves *consist* " (sorry, I'm being picky this evening)

docs/user_guide/mitigation.rst
```diff
@@ -68,6 +68,218 @@ Postprocessing
...
+            "operation1": "[>-inf]"
+        }
+    }
+    >>> plot_threshold_optimizer(threshold_optimizer)
+
+The printed thresholding rules influence the predictions, so we need to first
+understand the formula that puts all the parts together.
+Instead of selecting a particular thresholding rule we can actually select
+any point between any two thresholding rules.
+This is a very important insight and occurs in almost every case.
```
riedgar-ms, line 135: I can't quite figure out what this sentence means. I'm not sure you need it at all.

docs/user_guide/mitigation.rst
```diff
@@ -68,6 +68,218 @@ Postprocessing
...
+whether :class:`ThresholdOptimizer` should be used. Additionally, it requires
+the `sensitive_features` not only at training time, but also at deployment
+time, i.e., when calling :code:`predict()`.
+
+Due to the separate thresholding rules per sensitive feature value, one might
+argue that this constitutes
+`disparate treatment <https://en.wikipedia.org/wiki/Disparate_treatment>`_ in
+certain contexts where this is legally relevant.
+
+For comparison, let's also examine what this looks like for demographic parity
```
riedgar-ms, line 208: Since this is simpler, would it make more sense to put it first?

docs/user_guide/mitigation.rst
romanlutz: added that as a new section

docs/user_guide/mitigation.rst
riedgar-ms: OK

docs/user_guide/mitigation.rst
```diff
@@ -68,6 +68,253 @@ Postprocessing
...
+    >>> from sklearn.linear_model import LogisticRegression
+    >>> from fairlearn.postprocessing import ThresholdOptimizer, plot_threshold_optimizer
+    >>> from sklearn.datasets import fetch_openml
+    >>> import pandas as pd
+    >>> data = fetch_openml(data_id=1590, as_frame=True)
+    >>> X = pd.get_dummies(data.data)
+    >>> y = (data.target == '>50K') * 1
+    >>> sex = data.data['sex']
+    >>> logistic_regression = LogisticRegression()
+    >>> threshold_optimizer = ThresholdOptimizer(estimator=logistic_regression, constraints="equalized_odds", objective="accuracy_score")
```
riedgar-ms, line 107: Could you rewrap lines like this, so that users don't get horizontal scroll bars?

riedgar-ms (approved):

docs/user_guide/mitigation.rst
romanlutz: Good point! I thought I did that, but must have been another place in the code. :-)

docs/user_guide/mitigation.rst
```diff
@@ -68,6 +68,259 @@ Postprocessing
...
+results with.
+
+:class:`ThresholdOptimizer` expects an estimator that provides it with scores.
+While the output of :class:`ThresholdOptimizer` is binary, the input need not
+be. In fact, real valued input, e.g. from a regressor, provides it with many
+more options to create thresholds. For :math:`n` input data points with
+:math:`m \leq n` different score values it has :math:`m+1` different
+thresholds. At each threshold one can create one of two thresholding rules,
+i.e. functions that indicate which data points get label :code:`1` based on
+their score.
```
adrinjalali, line 86: I think it'd be easier to follow if you explain the thresholding thing before you say why soft values are preferred.

docs/user_guide/mitigation.rst
```diff
@@ -68,6 +68,259 @@ Postprocessing
...
+
+.. plot::
+    :include-source:
+
+    >>> from sklearn.linear_model import LogisticRegression
+    >>> from fairlearn.postprocessing import ThresholdOptimizer, plot_threshold_optimizer
+    >>> from sklearn.datasets import fetch_openml
+    >>> import pandas as pd
+    >>> data = fetch_openml(data_id=1590, as_frame=True)
+    >>> X = pd.get_dummies(data.data)
```
adrinjalali, line 103: could we use a `Pipeline` instead of `get_dummies`? See #649

docs/user_guide/mitigation.rst
```diff
@@ -68,6 +68,259 @@ Postprocessing
...
+      - :code:`accuracy_score`, :code:`balanced_accuracy_score`
+   *  - :code:`demographic_parity`, :code:`false_positive_rate_parity`, :code:`false_negative_rate_parity`, :code:`true_positive_rate_parity`, :code:`true_negative_rate_parity`
+      - :code:`selection_rate`, :code:`true_positive_rate`, :code:`true_negative_rate`, :code:`accuracy_score`, :code:`balanced_accuracy_score`
+
+Using a pre-trained estimator
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By default, :class:`ThresholdOptimizer` trains the passed estimator using its
+:code:`fit()` method. In some cases we already have a trained estimator, for
+example if we want to compare it with others. For this use case
```
adrinjalali, line 293: not sure what "for example if we want to compare it with others" means here.

docs/user_guide/mitigation.rst
```diff
@@ -68,6 +68,259 @@ Postprocessing
...
+    >>> threshold_optimizer.fit(X, y, sensitive_features=A)
+    ThresholdOptimizer(estimator=LogisticRegression(), prefit=True)
+
+If it detects that the estimator
+may not be fitted as defined by
+`scikit-learn's check <https://scikit-learn.org/stable/modules/generated/sklearn.utils.validation.check_is_fitted.html>`_
+it prints a warning, but still proceeds with the assumption that the user
+knows what they are doing. Not every machine learning package adheres to
+scikit-learn's convention of setting all members with trailing underscore
+during :code:`fit()`, so this is unfortunately an imperfect check.
```
adrinjalali, line 322: This would change once we are compatible enough to put this estimator in a `GridSearchCV`, since that would mean calling `clone` on the estimator and the child estimator would not be fit then.

docs/user_guide/mitigation.rst
```diff
@@ -68,6 +68,259 @@ Postprocessing
...
+    >>> y = (data.target == '>50K') * 1
+    >>> sex = data.data['sex']
+    >>> logistic_regression = LogisticRegression()
+    >>> threshold_optimizer = ThresholdOptimizer(
+    ...     estimator=logistic_regression,
+    ...     constraints="equalized_odds",
+    ...     objective="accuracy_score")
+    >>> threshold_optimizer.fit(X, y, sensitive_features=sex)
+    ThresholdOptimizer(constraints='equalized_odds', estimator=LogisticRegression())
+    >>> threshold_optimizer.interpolated_thresholder_._print_thresholds()
```
adrinjalali, line 113: an example shouldn't be using private API. We either need to make it public, or instead write a little function in this example which does the printing for us (my preferred way)

fairlearn/postprocessing/_interpolated_thresholder.py
```diff
@@ -129,3 +131,13 @@ def predict(self, X, *, sensitive_features, random_state=None):
         positive_probs = self._pmf_predict(
             X, sensitive_features=sensitive_features)[:, 1]
         return (positive_probs >= random_state.rand(len(positive_probs))) * 1
+
+    def _print_thresholds(self):
```
LeJit, line 135: Do we want to comment this function before we merge the change? It sticks out a bit in the code base because all the other functions in the file have a docstring.

docs/user_guide/mitigation.rst
romanlutz: Sure, copied that from another place in the repo, but I'm all in favor of moving everything to better practices.

docs/user_guide/mitigation.rst
romanlutz: I wrote that function specifically for this user guide actually. Even if I wrote it just in this snippet it would still have to access private members. If people are fine with it I can make `_print_thresholds()` public by removing the underscore. Wdyt?

docs/user_guide/mitigation.rst
romanlutz: I rephrased that a little and removed this particular sentence. I think what I meant originally was that one might want to compare several mitigation techniques and the pretrained model, but that's not a great justification for this either.

docs/user_guide/mitigation.rst
```diff
@@ -68,6 +68,259 @@ Postprocessing
...
+
+- predict 1 if the score is greater than :math:`0.5`.
+- predict 0 with probability :math:`0.999926` if the score is less or equal to
+  :math:`0.5`. This means there is a very low probability of getting label 1
+  regardless of the features.
+
+For "Female" the result includes the :math:`p_{\text{ignore}}` term, but one of the
+thresholds is set to always be true
+
+- predict 1 with probability
```
LeJit, line 193: Should we include the "predict 0" rule for "Female" to make it consistent with the "Male" section?

romanlutz: Consistency is worth rewriting this. I've thought about it a bit and the "Female" case is the much more complicated one due to the extra term with `p_ignore`, so I've instead adjusted the "Male" section to be consistent. I hope that makes sense for you.

fairlearn/postprocessing/_interpolated_thresholder.py
romanlutz: Did this in a follow-up commit, and made it "public" by removing the underscore since Adrin commented elsewhere that we shouldn't use private methods in the user guide.

docs/user_guide/mitigation.rst
romanlutz: Actually, come to think of it, it seems like this code could be reused in many different places. Is there a reason why there isn't a shorter / one-line version of this? Even if it's just done in every example where we use `fetch_adult` it feels like we'd simplify the shown code a lot by putting that code in a utility function.

romanlutz: After spending about an hour on this I'd much rather not do this in this PR. This is consistent with other places such as quickstart, and changing all of them goes way beyond the scope of this PR. Nevertheless, I want to address it at some point. This means that we need to be able to pass `Pipeline`s into mitigation techniques and from my experiment that doesn't necessarily work since the feature encoding (e.g. one hot encoding of categorical string features) only happens after our mitigation techniques do some of their own validation (namely `check_X_y`) which results in errors. The cleaner way would probably be to create a `Pipeline` around the mitigation technique, so that the mitigation technique never needs to see the "un-preprocessed" `X`, but I remember that we need to do something more to get that to work. I'd really like to prioritize some of this since it keeps coming up and people expect this to work.

Does this make sense @adrinjalali ? I'll open an issue for this.

adrinjalali: sure, happy to have it tackled in a different issue.

fairlearn/postprocessing/_interpolated_thresholder.py
adrinjalali, line 135: what about a public `get_thresholds()` method which returns a `dict` and if we wanna print it nicely, we can use `pprint` or `print(json.dumps())`

romanlutz: I like that for another reason: someone asked me recently if they could get the thresholds that it comes up with and change them as needed. For example, this could be used to simplify them a bit and then plug it right back in. Perhaps that's a bit more than you thought, but this would make it easier :-) Will work this in right away

romanlutz: I just realized that `InterpolatedThresholder`, the class on which I wrote the `thresholds()` method, is currently private since it's only used internally. I'm sort of torn between making it public by adding it to the `__init__.py` file of the postprocessing module and moving the `thresholds()` method to `ThresholdOptimizer` since that's basically the main class people are interacting with. This will also need slight adjustments for the case we discussed in #660 if we go with the latter, but nothing dramatic.

docs/user_guide/mitigation.rst
romanlutz: This doesn't use the private API anymore since I used your suggestion from another comment. Thanks!

docs/user_guide/mitigation.rst
romanlutz: Awesome! --> #665

docs/user_guide/mitigation.rst
romanlutz: I very much agree after re-reading this. Must have been terribly confusing early on without even understanding the high level points... Moved it almost all the way to the end into a `.. note:`

docs/user_guide/mitigation.rst
romanlutz: True. Do you want me to mention that in some way? I feel like we can inform users that this option won't work inside a `GridSearchCV` when we are actually compatible, but I'm happy to add whatever you think makes sense.

fairlearn/postprocessing/_interpolated_thresholder.py
romanlutz: Went with the latter option for now since it strikes me as more intuitive for users.

romanlutz:
This is complete as far as I am concerned. I'm checking on the build failures right now since something unrelated fails with scikit-learn 0.24.0, but I should have that resolved shortly. Besides that, feel free to complete the PR if you don't have any other comments @adrinjalali (or others) since Richard already signed off.

docs/user_guide/mitigation.rst
```diff
@@ -68,6 +68,263 @@ Postprocessing
...
+.. list-table::
+   :header-rows: 1
+   :stub-columns: 1
+
+   *  - fairness criteria
+      - objectives
+   *  - :code:`equalized_odds`
+      - :code:`accuracy_score`, :code:`balanced_accuracy_score`
+   *  - :code:`demographic_parity`, :code:`false_positive_rate_parity`, :code:`false_negative_rate_parity`, :code:`true_positive_rate_parity`, :code:`true_negative_rate_parity`
+      - :code:`selection_rate`, :code:`true_positive_rate`, :code:`true_negative_rate`, :code:`accuracy_score`, :code:`balanced_accuracy_score`
```
adrinjalali, line 293: could these lines be broken?

test/unit/preprocessing/linear_dep_remover/test_sklearn_compat.py
```diff
@@ -30,7 +30,6 @@
         estimator_checks.check_dtype_object,
         estimator_checks.check_sample_weights_pandas_series,
         estimator_checks.check_sample_weights_list,
-        estimator_checks.check_sample_weights_invariance,
```
adrinjalali, line 33: not sure why this needs to be removed here.

adrinjalali (commented):
Otherwise looks pretty good to me, thanks @romanlutz

test/unit/preprocessing/linear_dep_remover/test_sklearn_compat.py
adrinjalali: Ah I saw your commit message. Fair enough (I would have made it a separate PR, but happy to have it here)

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -228,6 +228,33 @@ def fit(self, X, y, *, sensitive_features, **kwargs):
             sensitive_feature_vector, y, scores)
         return self
 
+    def thresholds(self):
```
MiroDudik, line 231: I'm not convinced that this change belongs into this PR.

There are two main issues:
1. This should not be a method, but an [estimated attribute](https://scikit-learn.org/stable/developers/develop.html#estimated-attributes). Its name should match the name of the corresponding attribute of `InterpolatedThresholder`, which is `interpolation_dict` (that is not a great name and changing it to something like `thresholds` would be fine with me).
2. One of the use cases that came up with how folks have been using `ThresholdOptimizer` is that they need access to not just one set of thresholds, but to multiple--each corresponding to a different trade-off between the objective and the value of the constraint. For example, instead of simply minimizing error subject to false_positive_rate_parity, you may want to examine several solutions that achieve `false_positive_rate_parity` at different values of the `false_positive_rate`. This means that we will need a mechanism similar to the one we have right now in `GridSearch`. With that in mind, it might be confusing to have a field called `thresholds_` that captures just the "best set of thresholds" rather than the thresholds along the trade-off curve.

I think that (1) above would be very simple to address even in this PR. I'm not sure whether we need to address (2) right now or later, but I wanted to raise it for your awareness since it might inform how we rename the fields.

docs/user_guide/mitigation.rst
```diff
@@ -68,6 +68,263 @@ Postprocessing
 
 .. currentmodule:: fairlearn.postprocessing
 
+Fairlearn currently supports one postprocessing technique that is based on the
+paper *Equality of Opportunity in Supervised Learning* [#3]_. Unlike other
+mitigation techniques :class:`ThresholdOptimizer` is built to satisfy the
+specified fairness criteria exactly and with no remaining disparity.
+In many cases this comes at the expense of performance, for example, with
```
MiroDudik, line 75: The main point with usability is actually that it requires access to the sensitive feature at deployment time. We should mention it somewhere early on.

docs/user_guide/mitigation.rst
```diff
@@ -68,6 +68,263 @@ Postprocessing
...
+significantly lower accuracy. Regardless, it is a useful data point to compare
+results with.
+
+We can think of each thresholding rule as a binary classifier and determine
```
MiroDudik, line 79: This description is only true if you select "equalized_odds" as the constraints. Equalized odds are a special case.

I would begin with the simple cases:
* for each sensitive feature, consider classifiers obtained by thresholding its prediction and plot the curve corresponding to the upper envelope of all achievable combinations of the objective and the constraint metric
* simultaneously walk along all the curves--so as to maintain the same value of the constraint--while monitoring the objective value, and pick the combination of thresholds that optimizes the objective
* that's exactly what's depicted if you run `plot_threshold_optimizer`
* To achieve the precise value of the constraint along the curve, you typically need to randomize between two thresholds, hence the need for `p0` and `p1`
* Flip tells you whether you're allowed to flip the monotonicity... the need for flipping may suggest that the underlying classifier is not behaving as expected and any such instance should be inspected.

For equalized odds, we proceed differently:
* Here we again consider all possible threshold rules for each sensitive feature, and plot the achievable set of true positive rate / false positive rate combinations. (I'm not sure that calling it ROC curve is actually helpful here. Is it?)
* We take overlap among all these plots to get the set of TPR/FPR combinations achievable by all sensitive features.
* We select the combination that maximizes the objective (equalized odds will be satisfied automatically).
* The operation of taking an overlap may require that we step below the Pareto curve for a given feature, and it means that with a certain probabily `p_ignore` we may need to toss a biased coin, with the bias indicated by `prediction_constant`

test/unit/preprocessing/linear_dep_remover/test_sklearn_compat.py
romanlutz: @riedgar-ms is tackling this in #669 which is very likely going in before this, so it'll be removed in the next sync. I just wanted to see if the builds pass.

fairlearn/postprocessing/_threshold_optimizer.py
romanlutz: (2) is definitely not in scope of this PR which is about documentation. This uncertainty about where this might eventually go if someone has time to get to it is what led me to make the original method start with an underscore to give us flexibility going forward. @adrinjalali 's comment about not using underscore-prefixed methods in the public docs made me change that.

(1) `InterpolatedThresholder` is internal. It doesn't even show up in the docs https://fairlearn.github.io/v0.5.0/api_reference/fairlearn.postprocessing.html
It's not a big deal to change the name, so I can change it there, too.


Given these points should it seems like it's best to keep it private AND still use it in the docs. @adrinjalali what do you think?

docs/user_guide/mitigation.rst
romanlutz: How do you mean? It renders like this (as intended):
image

MiroDudik: i'm assuming @adrinjalali is asking for shorter lines in `.rst`

fairlearn/postprocessing/_threshold_optimizer.py
MiroDudik: Actually, if we use the name that takes form of a singular, then we could export it and be still future-proof re. (2) above. For example we could use something like `threshold_info_` or even `thresholder_info_`? We would similarly change the name of the corresponding field of `InterpolatedThresholder`.

adrinjalali: I'm usually wary of `attribute_` which is indeed a getter property instead of an actual attribute.

I would still find it odd to use private API in the docs. People copy/paste from the docs, and we should be stricter about changing what people use than what's internal. Putting something private there makes it used.

As an intermediary solution, I'd put the code for this function in the example and let people copy/paste the code while we figure the API. Or have a separate PR and figure the API out and then have this example merged.

docs/user_guide/mitigation.rst
romanlutz: I've reordered it as you said and adjusted the description. The original idea was to start with the key idea of the paper (eq. odds postprocessing), but obviously that's less relevant for users. If they want to read the paper they can always do that.

docs/user_guide/mitigation.rst
romanlutz: Added that comment.

fairlearn/postprocessing/_threshold_optimizer.py
romanlutz: IMO `InterpolatedThresholder` is actually a great name and much better than `Thresholder`. I'd expect something very different from a class called `Thresholder`, namely a very simple thresholder. I can certainly just use the code in the example like @adrinjalali suggests. That would be
```
threshold_optimizer.interpolated_thresholder_.interpolation_dict
```
or perhaps `threshold_info_` instead of `interpolation_dict` as @MiroDudik suggests.

romanlutz: Update: I've removed the method from `ThresholdOptimizer` and renamed `interpolation_dict` to `threshold_info_` on the `InterpolatedThresholder` class. In the example it just directly uses it now as mentioned above (except for the rename). This allows us to make this decision later on without requiring people to access private methods or attributes, and we can still show it. I hope that works for everyone.

fairlearn/postprocessing/_interpolated_thresholder.py
romanlutz: closing this thread since it doesn't exist anymore (see other thread)

docs/user_guide/mitigation.rst
romanlutz: Ah, didn't realize that I can move it over several lines! Done!

test/unit/preprocessing/linear_dep_remover/test_sklearn_compat.py
romanlutz: #669 is completed and this is now gone from this PR. Closing

romanlutz:
I think I've addressed everything now. @riedgar-ms already signed off, so unless there's anything else @adrinjalali or @MiroDudik feel free to merge the PR.

adrinjalali (commented):
Thanks @romanlutz

fairlearn/postprocessing/_interpolated_thresholder.py
```diff
@@ -20,13 +20,13 @@ class InterpolatedThresholder(BaseEstimator, MetaEstimatorMixin):
...
     Based on the values of sensitive features, it then applies a randomized thresholding
-    transformation according to the provided `interpolation_dict`.
+    transformation according to the provided `threshold_info`.
 
     Parameters
     ----------
     estimator :
         base estimator
-    interpolation_dict : dict
+    threshold_info : dict
```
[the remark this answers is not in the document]
romanlutz: Depends on what you think of as public. It's not in the API docs, but it is, of course, not prefixed with underscore. I would call that non-public.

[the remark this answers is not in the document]
romanlutz: It's not quite the thresholds only, there's some additional info. That's why I originally used the name interpolation dict. It contains info about how to do the interpolation between the two thresholds.

adrinjalali (approved):
Thanks @romanlutz

adrinjalali:
Tests fail

docs/user_guide/mitigation.rst
```diff
@@ -68,6 +68,320 @@ Postprocessing
...
+    >>> y = (data.target == '>50K') * 1
+    >>> sex = data.data['sex']
+    >>> logistic_regression = LogisticRegression()
+    >>> threshold_optimizer = ThresholdOptimizer(
+    ...     estimator=logistic_regression,
+    ...     constraints="equalized_odds",
+    ...     objective="accuracy_score")
+    >>> threshold_optimizer.fit(X, y, sensitive_features=sex)
+    ThresholdOptimizer(constraints='equalized_odds', estimator=LogisticRegression())
+    >>> print(json.dumps(threshold_optimizer.interpolated_thresholder_.threshold_info_, default=str, indent=4))
```
adrinjalali, line 248: while you're at fixing the tests, you could also shorten this line :D

romanlutz:
@MiroDudik wants to make a pass over this and directly make changes. I'll wait until that's done and fix the tests afterwards.

docs/user_guide/mitigation.rst
```diff
@@ -137,7 +450,7 @@ provide :code:`signed_weights` that are used to relabel and reweight samples.
 .. _constraints_binary_classification:
 
 Fairness constraints for binary classification
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
hildeweerts, line 453: [Are these changes on purpose?]

hildeweerts (approved):
There's a few changes in symbols used for the headers for which I'm not sure whether they're intentional but other than that I'm happy to get this merged!

requirements-dev.txt
```diff
@@ -32,7 +32,7 @@ nbval
 bokeh
 pypandoc
 sphinx==4.0.3  # a bug in 4.1.0 prevents the logo from being rendered
-sphinx-gallery
+sphinx-gallery  # TODO merge required change first
```
hildeweerts, line 35: [I assume this TODO can be removed?]

hildeweerts:
I've attempted to resolve the merge conflict. Feel free to merge when the checks pass :) @fairlearn/fairlearn-maintainers

romanlutz:
> I've attempted to resolve the merge conflict. Feel free to merge when the checks pass :) @fairlearn/fairlearn-maintainers 

@hildeweerts this isn't done. The feedback I received was that I need ti make extensive changes to the PR to switch out the dataset which effectively means adjusting a lot of the text. Plus, there was a technical issue that prompted me to submit a PR with sphinx-gallery. It's still open on their side. Perhaps it's possible to do this without that, but I don't quite remember as it's been several months. I'll have to read through this all once more... What I definitely remember is the dataset being the blocker.

riedgar-ms:
@romanlutz any plans to reanimate this PR?

romanlutz:
I will focus on this in the coming weeks. I would like to avoid celebrating the PR's 3rd birthday next year 😆

docs/user_guide/mitigation/postprocessing.rst
```diff
@@ -14,69 +14,135 @@ Threshold optimizer
...
+Unlike other mitigation techniques :class:`ThresholdOptimizer` is built to
+satisfy the specified fairness criteria exactly and with no remaining
+disparity.
 In many cases this comes at the expense of performance, for example, with
-significantly lower accuracy. Regardless, it is a useful data point to compare
-results with. Importantly, :class:`ThresholdOptimizer` requires the sensitive
-features to be available at deployment time (i.e., for the :code:`predict`
-method).
+significantly lower accuracy.
+Regardless, it is a useful data point to compare results with.
```
riedgar-ms, line 24: Nitpick: Sentence ends with a preposition, and I don't think it's making a dramatic effect.

riedgar-ms (commented):
Did you mean to include `filename.png` ?

docs/user_guide/mitigation/postprocessing.rst
romanlutz: rephrased. LMK if that's better

romanlutz:
> Did you mean to include `filename.png` ?

I did not. Removed it now, though! Thanks.

romanlutz:
Still not ready to properly review. I got the code running properly including the plot output which was one of the main obstacles in the past. I also switched to the diabetes hospital dataset, generated the charts and then realized that I can't use demographic parity there which kind of defeats my purpose. I'd also have to do a lot more data wrangling than I want to do here, so either we have to have a preprocess function somewhere for that dataset to do the things we're doing in the SciPy tutorial or I can't use it.

hildeweerts:
@romanlutz do you think it would be possible to get this user guide in for the next release (I currently put this on June 1)? I think it would be great if we could include this (or a minimal version of it) on the website sooner rather than later!

hildeweerts:
Alright! I've added it to the milestone. It would be great if we can get it in, but no worries if you can't find the time in May.

hildeweerts:
@romanlutz do you have a better idea of your availability to work on this in May?

romanlutz:
With two weeks to go before the 3rd anniversary of this PR I have finally gotten it into a reviewable state!!!

@fairlearn/fairlearn-maintainers please take a look. Would love to get this merged... The main change is that I replaced the adult dataset with diabetes hospital which meant changing practically everything as demographic parity makes no sense there, but FPR parity and EO do. LMK what you think 😄

docs/user_guide/mitigation/postprocessing.rst
```diff
@@ -0,0 +1,448 @@
...
+In our example above, this might mean that members of one group might have a
+base probability of being predicted to be readmitted that is higher than
+the base probability for members of another group.
+We want to emphasize that a non-zero probability to get label 1 is not
+acceptable in some application contexts, so it needs to be carefully evaluated
+whether :class:`ThresholdOptimizer` should be used.
+
+Due to the separate thresholding rules per sensitive feature value, one might
+argue that this constitutes
+`disparate treatment <https://en.wikipedia.org/wiki/Disparate_treatment>`_ in
```
hildeweerts, line 388: under US anti-discrimination law or direct discrimination under EU non-discrimination law (which tragically does not have it's own Wikipedia page)

hildeweerts (commented):
Woooh! 🥳 

I only have a few small comments!

docs/user_guide/mitigation/postprocessing.rst
```diff
@@ -0,0 +1,448 @@
...
+.. _threshold_optimizer:
+
+Threshold optimizer
+^^^^^^^^^^^^^^^^^^^
+
+:class:`ThresholdOptimizer` is based on the paper
+*Equality of Opportunity in Supervised Learning*
+:footcite:`hardt2016equality`.
+Unlike other mitigation techniques :class:`ThresholdOptimizer` is built to
+satisfy the specified fairness criteria exactly and with no remaining
```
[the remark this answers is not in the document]
romanlutz: Indeed! We'll have to update the docs in #1248.

docs/conf.py
```diff
@@ -196,6 +197,14 @@
     "member-order": "groupwise"
 }
 
+# Options for the `::plot` directive
+# ----------------------------------
+# https://matplotlib.org/stable/api/sphinxext_plot_directive_api.html
```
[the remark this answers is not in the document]
romanlutz: You can have plots as output 😄 It builds upon doctest which does not include this functionality. The only other way to get plots as output is to have an example notebook but this is a user guide, of course.

docs/user_guide/mitigation/postprocessing.rst
```diff
@@ -0,0 +1,448 @@
...
+    The :code:`flip` argument on the constructor indicates whether flipped
+    thresholds can be used. Flipped constraints such as `"<0.6"`
+    indicate that the underlying estimator's scores do not exhibit the
+    expected monotonicity and any such instance should be inspected.
+
+.. note::
+
+    :class:`ThresholdOptimizer` expects an estimator that provides it with
+    scores. While the output of :class:`ThresholdOptimizer` is binary, the
+    input need not be. In fact, real valued input, e.g. from a regressor,
```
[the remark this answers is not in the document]
romanlutz: adopted 😄

docs/user_guide/mitigation/postprocessing.rst
```diff
@@ -0,0 +1,448 @@
...
+representing the two thresholding rules, and :math:`p_0+p_1=1`.
+:math:`p_{\text{ignore}}` is :math:`0` if the curve does not need to be "pulled"
+towards the diagonal, i.e., when the selected solution lies on the curve
+itself as opposed to below.
+Above, the curve for "Asian" is strictly above the other curves, so
+:math:`p_{\text{ignore}}` is highest for "Asian" and lower for other groups.
+
+Importantly, :math:`p_{\text{ignore}}` is only required for equalized odds.
+We want to stress on the fact that it is crucial to understand the
+implications of such a result.
```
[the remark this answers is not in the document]
romanlutz: added

docs/user_guide/mitigation/postprocessing.rst
```diff
@@ -0,0 +1,457 @@
...
+Predictions of the resulting :class:`ThresholdOptimizer` model are determined
+randomly based on the probabilities to predict label 1 that the formula
+produces.
+Another important implication is that using randomized thresholds in this way
+results in what philosophers refer to as "leveling down", i.e., decreasing
+performance for the better-off group without making the worse-off group any
+better off in absolute terms.:footcite:`weerts2022does` :footcite:`mittelstadt2023unfairness`
```
hildeweerts, line 382:
```suggestion
better off in absolute terms. :footcite:`weerts2022does` :footcite:`mittelstadt2023unfairness`
```

docs/user_guide/mitigation/postprocessing.rst
```diff
@@ -0,0 +1,457 @@
...
+the base probability for members of another group.
+We want to emphasize that a non-zero probability to get label 1 is not
+acceptable in some application contexts, so it needs to be carefully evaluated
+whether :class:`ThresholdOptimizer` should be used.
+
+Due to the separate thresholding rules per sensitive feature value, one might
+argue that this constitutes
+`disparate treatment <https://en.wikipedia.org/wiki/Disparate_treatment>`_
+under US anti-discrimination law or direct discrimination under
+EU non-discrimination law.
```
hildeweerts, line 398:
```suggestion
EU non-discrimination law, in contexts covered by these laws.
```

I did like your specification that it applies in some contexts but not in others. I don't know as much about US law, but at least in the EU, protected characteristics differ across domains and sometimes apply only to public bodies etc.

PR #616, opened 2020-10-30
Fix widgets docs by making class attributes private, convert to numpydoc

romanlutz:
The public (not prefixed with underscore) attributes of the `FairlearnDashboard` class showed up in the documentation, and evidently the values from some use cases carried over, leading to a massive unintended data output. This PR fixes this problem by making the class attributes private. At the same time I took the liberty to convert the docs to numpydoc format.

### DIFF
--- a/fairlearn/widget/fairness_dashboard.py
+++ b/fairlearn/widget/fairness_dashboard.py
@@ -32,16 +32,16 @@ class FairlearnDashboard(object):
 
     Parameters
     ----------
-    sensitive_features : {:class:`numpy.ndarray`, list[][], :class:`pandas.DataFrame`, :class:`pandas.Series`}  # noqa: E501
+    sensitive_features : numpy.ndarray, list[][], pandas.DataFrame, pandas.Series
         A matrix of feature vector examples (# examples x # features),
         these can be from the initial dataset, or reserved from training.
-    y_true : {:class:`numpy.ndarray`, list[]}
+    y_true : numpy.ndarray, list[]
         The true labels or values for the provided dataset.
-    y_pred : {:class:`numpy.ndarray`, list[][], list[], dict {string: list[]}}
+    y_pred : numpy.ndarray, list[][], list[], dict {string: list[]}
         Array of output predictions from models to be evaluated. Can be a single
         array of predictions, or a 2D list over multiple models. Can be a dictionary
         of named model predictions.
-    sensitive_feature_names : {:class:`numpy.ndarray`, list[]}
+    sensitive_feature_names : numpy.ndarray, list[]
         Feature names
     """
 
--- a/fairlearn/widget/fairness_dashboard.py
+++ b/fairlearn/widget/fairness_dashboard.py
@@ -28,26 +28,28 @@ from scipy.sparse import issparse
 
 
 class FairlearnDashboard(object):
-    """The dashboard class, wraps the dashboard component.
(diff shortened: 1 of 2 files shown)


fairlearn/widget/fairness_dashboard.py
```diff
@@ -28,26 +28,28 @@
...
+    sensitive_features : {:class:`numpy.ndarray`, list[][], :class:`pandas.DataFrame`, :class:`pandas.Series`}  # noqa: E501
+        A matrix of feature vector examples (# examples x # features),
         these can be from the initial dataset, or reserved from training.
-    :type sensitive_features: numpy.ndarray or list[][] or pandas.DataFrame or pandas.Series
-    :param y_true: The true labels or values for the provided dataset.
-    :type y_true: numpy.ndarray or list[]
-    :param y_pred: Array of output predictions from models to be evaluated. Can be a single
+    y_true : {:class:`numpy.ndarray`, list[]}
+        The true labels or values for the provided dataset.
+    y_pred : {:class:`numpy.ndarray`, list[][], list[], dict {string: list[]}}
```
riedgar-ms, line 40: I don't think you need :class: here - only in the main run of the text.

riedgar-ms (approved):

fairlearn/widget/fairness_dashboard.py
```diff
@@ -28,26 +28,28 @@
 
 
 class FairlearnDashboard(object):
-    """The dashboard class, wraps the dashboard component.
+    r"""The dashboard class, wraps the dashboard component.
 
-    :param sensitive_features: A matrix of feature vector examples (# examples x # features),
+    Parameters
+    ----------
+    sensitive_features : {:class:`numpy.ndarray`, list[][], :class:`pandas.DataFrame`, :class:`pandas.Series`}  # noqa: E501
```
MiroDudik, line 35: `# noqa` [renders!](https://1786-133444044-gh.circle-artifacts.com/0/docs/_build/html/master/api_reference/fairlearn.widget.html)
also, please get rid of enclosing `{}` around listing of types

MiroDudik (approved):

fairlearn/widget/fairness_dashboard.py
romanlutz: It does render differently. If I don't put this then only one of the types gets the link, the others are just text.

fairlearn/widget/fairness_dashboard.py
romanlutz: Saw this `{}` elsewhere in the repo and assumed that's how we want to express this. I suppose that's tricky because I can't spread it across multiple lines (renders the wrong way). Perhaps if I don't use `:class:` it will be short enough. Will try this...

fairlearn/widget/fairness_dashboard.py
MiroDudik: i'm not sure if those are the right type descriptors, also lists of types seem to work fine in metrics: https://fairlearn.github.io/master/api_reference/fairlearn.metrics.html

romanlutz: I'll just go back to what we had before. Not going to waste much time on this, since this will be deleted shortly. We should get it figured out in the new repository, though.

fairlearn/widget/fairness_dashboard.py
romanlutz: removed

PR #622, opened 2020-11-04
Set default of MetricFrame difference and ratio to be 'between_groups'

riedgar-ms:
Adding a default method of `between_groups` to `MetricFrame.difference()` and `MetricFrame.ratio()`. This had been previously discussed, but not implemented. Also make sure that the `method=` argument has documentation.


MiroDudik (approved):

romanlutz (approved):
That is a default that I can get behind :-)

adrinjalali (approved):

Issue #570, opened 2020-09-08
Consider changing logo link in the documentation page from contents to index


leonarduschen:
#### Describe the issue linked to the documentation

Clicking the logo leads to contents instead of index. This is because `pydata-sphinx-theme` sets logo href to `master_doc` (which in `fairlearn` case is contents).

#### Suggest a potential alternative/fix
I'm not sure if linking the logo to `index` instead of `content` is really what `fairlearn` wants, but an extra opinion from `fairlearn` maintainers would be awesome.

I have created an issue in `pydata-sphinx-theme` and will mention this issue over there (https://github.com/pandas-dev/pydata-sphinx-theme/issues/245)

romanlutz:
This is mostly self-inflicted since we wanted to have a custom title page, so the docs "root" is `contents` as you're describing, and not index. I don't consider it to be a big issue to be honest. Adding a link from contents back to the `index` page may be a useful "band aid" solution until there's perhaps a solution from pydata-sphinx-theme. Feel free to open a PR!

PR #618, opened 2020-10-30
MAINT: Bump pydata-sphinx-theme; redirect logo link from ToC to landing page

leonarduschen:
Closes #570 

WIP: Need to wait for `pydata-sphinx-theme` new release. Opening a PR now to keep me from forgetting.

PR #623, opened 2020-11-04
Documentation fixes

riedgar-ms:
Various updates to the documentation:

- Rename `contents.html` to `index.html`
- Move migration guides under the user guide
- Link to migration guide from quickstart
- Other small edits


docs/about/index.rst
```diff
@@ -31,7 +31,7 @@ Migrating from Previous Versions
 --------------------------------
```
MiroDudik, line 31: i don't think we need this paragraph here now that we have a note in quickstart and user guide.

MiroDudik (approved):

MiroDudik:
Please also check: https://github.com/fairlearn/fairlearn/pull/612/files#r517659422

docs/conf.py
```diff
@@ -85,12 +85,12 @@ def check_if_v046():
 # This pattern also affects html_static_path and html_extra_path.
 exclude_patterns = ['_build', 'Thumbs.db', '.DS_Store']
 
-master_doc = 'contents'
+master_doc = 'index'
```
romanlutz, line 88: What changed? This was intentional to my knowledge since index is our custom landing page. 

[There's a PR out #618 that seems related to me, just like the related issue.]

docs/user_guide/migrating_versions/index.rst
```diff
@@ -1,7 +1,7 @@
 .. _version_migration_guide:
 
-Version Migration Guide
-=======================
+Migrating from Prior Versions
```
romanlutz, line 3: nit: lowercase p and v

docs/conf.py
MiroDudik: I didn't catch the conversation on #570 (also tagging @leonarduschen). Basically, I chatted with @riedgar-ms and we thought that it's better if the root of the version `<version>` is: `https://fairlearn.github.io/<version>/index.html`. 

(Committing to `https://fairlearn.github.io/<version>/contents.html` as the root seems very unusual.)

Also take a look at the version selectors on:
https://fairlearn.github.io/master/contents.html
https://fairlearn.github.io/v0.4.6/index.html

Versus what you get with the current PR:
https://1871-133444044-gh.circle-artifacts.com/0/docs/_build/html/master/index.html
https://1871-133444044-gh.circle-artifacts.com/0/docs/_build/html/v0.4.6/index.html

Do you see any issues with this update?

**Re. what changed**: My guess is that previously we didn't have a versioned documentation, so there was a conflict between the "landing page" https://fairlearn.github.io/ and the "root of the documentation". That conflict has now disappeared. The final tweak that we will need to do is to redirect https://fairlearn.github.io/ to point at `https://fairlearn.github.io/<most-recent-version>/index.html`--but I think we should hold off with that until the landing page redesign (and that redesigned landing page then becomes the new "version root"). At that point, we will follow a similar pattern as https://scikit-learn.org/

riedgar-ms: I'm pretty sure @MiroDudik  correctly outlined the reason why we had `contents.html` before - I think you did that bit @romanlutz ? Now that we have versioned documentation, and each version is its own mini-website with the  custom-designed landing page separate and non-versioned, so we can go back to `index.html`.

As for changing the default links from the landing page, I do have an item on the release checklist to update the links to point at the latest release. In due course, we'll need to write a script to do that automatically as part of the release process.

docs/user_guide/migrating_versions/index.rst
riedgar-ms: We haven't been entirely consistent on that sort of point, but I can change this one.

docs/conf.py
leonarduschen: As far as my understanding goes, setting `master_doc` to `contents` is just a (somewhat hacky) way to make a custom landing page. The side effect of this is - clicking the logo will then leads to `contents` instead of `index` because `pydata-sphinx-theme` sets it to `master_doc` by default.

#618 is supposed to fix this issue by bumping `pydata-sphinx-theme`. The new feature would allow customizing where the logo leads when clicked (internal HTML page or external link). That said, if `index` is indeed the desired logo link, then this PR should fix #570

MiroDudik: @leonarduschen: Thanks for clarifying!

I think that a more basic issue that is being addressed by this PR is that we had no `index.html` at the root level.

romanlutz: Thanks for the description, @MiroDudik ! I had indeed not realized that this adds a separate one per version. In fact, that solves more issues than I had initially realized! Wonderful :-)

romanlutz (approved):

PR #625, opened 2020-11-07
Revert dashboard to ipywidgets

romanlutz:
Due to issues with the flask-based version in some of the previously supported environments we'll keep the ipywidgets-based version that was part of v0.4.6. The dashboard will be removed after the upcoming v0.5.0 release anyway, so the flask-based version (with improvements to support these environments) will be released from the new location as `raiwidgets`.

This is a fairly straightforward revert back to v0.4.6 with a few exceptions:
- adjusted to use `MetricFrame` instead of the previous group metrics
- adjusted serialization format to dict since `MetricFrame` returns `pandas.Series` and that causes a `ValueError` from ipywidgets
- added warning that the dashboard will be removed from Fairlearn after v0.5.0
- removed `rai_core_flask` as dependency (since it won't be needed anymore for v0.5.0 or after v0.5.0)

This also includes a few minor CHANGES.md adjustments (typos and line length) for the upcoming release.


riedgar-ms (approved):
I will try out locally

riedgar-ms:
I've just confirmed that with a local install of this branch, the dashboard appears under Python 3.6 using one of our example notebooks.

MiroDudik (approved):

PR #627, opened 2020-11-09
Preparations for v0.5.0

riedgar-ms:
Some small changes:

- Bump version in `__init__.py` to `v0.5.1.dev0` (which is what our development trunk will be once v0.5.0 is out)
- Update the 'link to current version' in the main ReadMe to v0.5.0

I have left the remainder of the documentation links point to `master` for now, since they will break until v0.5.0 ships otherwise (I'm prepared to tolerate that one 'current' version at the top of the ReadMe, but not every link on the page).

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Website: https://fairlearn.github.io/
 
 ## Current release
 
-- The current stable release is available at [Fairlearn v0.4.6](https://github.com/fairlearn/fairlearn/tree/v0.4.6).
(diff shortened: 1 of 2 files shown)


romanlutz (approved):
We'll also need a new entry in CHANGES.md for the next version.

Does the doc page need to be switched to default to 0.5.0 instead of master?

riedgar-ms:
Changes.md can follow on later. Which doc page do you mean 'switch to v0.5.0' ? If you mean the main website, that can happen after the release (we'll break a lot of things otherwise).

MiroDudik (commented):
agreed with @romanlutz re. redirecting all of the links from the landing page to 0.5.0 rather than master.

PR #631, opened 2020-11-11
Add control features to moments

riedgar-ms:
A control features to the classification moments. Control features stratify the dataset, with fairness constraints applied within each stratum, but not between strata.


fairlearn/_input_validation.py
```diff
@@ -79,14 +80,27 @@ def _validate_and_reformat_input(X, y=None, expect_y=True, enforce_binary_labels
...
+    # Handle the control features
+    control_features = kwargs.get(_KW_CONTROL_FEATURES)
+    if control_features is not None:
+        check_consistent_length(X, control_features)
+        control_features = check_array(control_features, ensure_2d=False, dtype=None)
+
+        # compress multiple control features into a single column
+        if len(control_features.shape) > 1 and control_features.shape[1] > 1:
+            control_features = \
+                _compress_multiple_sensitive_features_into_single_column(control_features)
```
riedgar-ms, line 92: This function would obviously be renamed

fairlearn/reductions/_moments/utility_parity.py
```diff
@@ -3,16 +3,20 @@
...
 from fairlearn._input_validation import _MESSAGE_RATIO_NOT_IN_RANGE
 from .error_rate import ErrorRate
 
+
 _UPPER_BOUND_DIFF = "upper_bound_diff"
 _LOWER_BOUND_DIFF = "lower_bound_diff"
 _MESSAGE_INVALID_BOUNDS = "Only one of difference_bound and ratio_bound can be used."
 _DEFAULT_DIFFERENCE_BOUND = 0.01
 
+_KW_CONTROL_FEATURES = "control_features"
```
riedgar-ms, line 18: I know this is a duplicate... would need to think about injecting this in a sensible way

fairlearn/reductions/_moments/utility_parity.py
```diff
@@ -245,7 +249,11 @@ class DemographicParity(UtilityParity):
 
     def load_data(self, X, y, **kwargs):
         """Load the specified data into the object."""
-        super().load_data(X, y, event=_ALL, **kwargs)
+        control_features = kwargs.get(_KW_CONTROL_FEATURES)
+        if control_features is None:
+            super().load_data(X, y, event=_ALL, **kwargs)
+        else:
+            super().load_data(X, y, event=control_features, **kwargs)
```
riedgar-ms, line 256: This is the key point - is this all I need to do to add control features. Basically, create new events for each?

test/unit/reductions/exponentiated_gradient/test_control_features.py
```diff
@@ -0,0 +1,100 @@
...
+        constraints=DemographicParity(),
+        eps=0.01)
+    expgrad_control.fit(X_dummy, y,
+                        sensitive_features=X['sens'],
+                        control_features=X['ctrl'])
+    y_pred_control = expgrad_control.predict(X_dummy)
+    mf_control = MetricFrame(selection_rate, y, y_pred_control,
+                             sensitive_features=X['sens'],
+                             control_features=X['ctrl'])
+    print(mf_control.by_group)
```
riedgar-ms, line 100: Comparing these print statements, I think it's doing what I want, but.... I'm not completely sure, and I could be misunderstanding something relevant to the synthetic data.

fairlearn/_input_validation.py
```diff
@@ -45,10 +46,11 @@ def _validate_and_reformat_input(X, y=None, expect_y=True, enforce_binary_labels
     :param enforce_binary_labels: if True raise exception if there are more than two distinct
         values in the `y` data; default False
     :type enforce_binary_labels: bool
-    :return: the validated and reformatted X, y, and sensitive_features; note that certain
-        estimators rely on metadata encoded in X which may be stripped during the reformatting
-        process, so mitigation methods should ideally use the input X instead of the returned X
-        for training estimators and leave potential reformatting of X to the estimator.
+    :return: the validated and reformatted X, y, sensitive_features and control_features; note
```
romanlutz, line 49: Didn't we want to update documentation to numpydoc whenever we touch it? :-)

test/unit/input_convertors.py
```diff
@@ -4,7 +4,7 @@
 import numpy as np
```
romanlutz, line 4: nit: should the file name be "converters" with E instead of O?

romanlutz (approved):
Looks great to me! Only very minor comments. This is exactly what I had expected.

fairlearn/_input_validation.py
```diff
@@ -45,10 +46,11 @@ def _validate_and_reformat_input(X, y=None, expect_y=True, enforce_binary_labels
...
+        that certain estimators rely on metadata encoded in X which may be stripped during
+        the reformatting process, so mitigation methods should ideally use the input X instead
+        of the returned X for training estimators and leave potential reformatting of X to the
+        estimator.
     :rtype: (pandas.DataFrame, pandas.Series, pandas.Series)
```
romanlutz, line 54: I believe that the return type needs updating.

fairlearn/_input_validation.py
riedgar-ms: Done

fairlearn/_input_validation.py
riedgar-ms: Done

fairlearn/_input_validation.py
```diff
@@ -10,6 +10,7 @@
 logger = logging.getLogger(__file__)
 
 _KW_SENSITIVE_FEATURES = "sensitive_features"
+_KW_CONTROL_FEATURES = "control_features"
```
MiroDudik, line 13: I don't think that we need these two constants. We are very strongly locked to "sensitive_features" and "control_features" and using `_KW_*` makes things look complicated.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -44,12 +44,15 @@ class _Lagrangian:
     """
 
     def __init__(self, X, sensitive_features, y, estimator, constraints, B, opt_lambda=True,
-                 sample_weight_name='sample_weight'):
+                 sample_weight_name='sample_weight',
+                 control_features=None):
```
MiroDudik, line 48: I think that this is becoming an anti-pattern here. Maybe we can change the signature to:
```python
    def __init__(self, X, y, estimator, constraints, B, opt_lambda=True,
                 sample_weight_name='sample_weight',
                 **kwargs)
```

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -44,12 +44,15 @@ class _Lagrangian:
...
         self.X = X
         self.n = self.X.shape[0]
         self.y = y
         self.constraints = constraints
-        self.constraints.load_data(X, y, sensitive_features=sensitive_features)
+        self.constraints.load_data(X, y,
+                                   sensitive_features=sensitive_features,
+                                   control_features=control_features)
```
MiroDudik, line 55: And this should be:
```python
self.constraints.load_data(X, y, **kwargs)
```

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -90,8 +90,10 @@ def fit(self, X, y, **kwargs):
             logger.debug("Regression problem detected")
             is_classification_reduction = False
 
-        _, y_train, sensitive_features = _validate_and_reformat_input(
-            X, y, enforce_binary_labels=is_classification_reduction, **kwargs)
+        _, y_train, sensitive_features, control_features = \
```
MiroDudik, line 93: We should move the call to `_validate_and_reformat_input` into `Moment.load_data`. Exponentiated gradient is just a reduction, so it should be touching the data only minimally.

fairlearn/reductions/_grid_search/grid_search.py
```diff
@@ -126,10 +127,14 @@ def fit(self, X, y, **kwargs):
...
-        _, y_train, sensitive_features_train = _validate_and_reformat_input(
-            X, y, enforce_binary_labels=is_classification_reduction, **kwargs)
+        _, y_train, sensitive_features_train, control_features_train = \
+            _validate_and_reformat_input(X, y,
+                                         enforce_binary_labels=is_classification_reduction,
+                                         **kwargs)
 
         kwargs[_KW_SENSITIVE_FEATURES] = sensitive_features_train
+        if control_features_train is not None:
+            kwargs[_KW_CONTROL_FEATURES] = control_features_train
```
MiroDudik, line 137: After `_validate_and_reformat_input` is moved into `Moment.load_data`, we should be able to get rid of this chunk of the code.

fairlearn/reductions/_moments/bounded_group_loss.py
```diff
@@ -36,6 +38,8 @@ def default_objective(self):
 
     def load_data(self, X, y, **kwargs):
         """Load data into the moment object."""
+        if kwargs.get(_KW_CONTROL_FEATURES) is not None:
```
MiroDudik, line 41: Let's drop this. If you're worried about using unsupported `kwargs` (a good worry), we should change the signature of `load_data` to
```python
    def load_data(self, X, y, *, sensitive_features):
```

fairlearn/reductions/_moments/bounded_group_loss.py
```diff
@@ -5,7 +5,9 @@
 import numpy as np
 from .moment import LossMoment
 from .moment import _GROUP_ID, _LABEL, _LOSS, _PREDICTION, _ALL
-from fairlearn._input_validation import _KW_SENSITIVE_FEATURES
+from fairlearn._input_validation import _KW_SENSITIVE_FEATURES, _KW_CONTROL_FEATURES
+
+_NO_CONDITIONAL_FEATURES = "Conditional features are not supported for regression moments"
```
MiroDudik, line 10: Lines 8-10 should be removed (see comments below).

fairlearn/reductions/_grid_search/grid_search.py
```diff
@@ -10,7 +10,8 @@
 from sklearn.utils.validation import check_is_fitted
 from time import time
 
-from fairlearn._input_validation import _validate_and_reformat_input, _KW_SENSITIVE_FEATURES
+from fairlearn._input_validation import _validate_and_reformat_input
```
MiroDudik, line 13:
```suggestion
```

fairlearn/reductions/_grid_search/grid_search.py
```diff
@@ -10,7 +10,8 @@
...
+from fairlearn._input_validation import _KW_SENSITIVE_FEATURES, _KW_CONTROL_FEATURES
```
MiroDudik, line 14:
```suggestion
```

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -100,7 +102,8 @@ def fit(self, X, y, **kwargs):
         B = 1 / self.eps
         lagrangian = _Lagrangian(X, sensitive_features, y_train, self.estimator,
                                  self.constraints, B,
-                                 sample_weight_name=self.sample_weight_name)
+                                 sample_weight_name=self.sample_weight_name,
+                                 control_features=control_features)
```
MiroDudik, line 106: Update to something like:
```python
        lagrangian = _Lagrangian(X, y, self.estimator,
                                 self.constraints, B,
                                 sample_weight_name=self.sample_weight_name,
                                 **kwargs)
```

fairlearn/_input_validation.py
riedgar-ms: I'm not committed to the `_KW` prefix, but in general, it's better to keep things like this as separate constants.

fairlearn/reductions/_moments/utility_parity.py
```diff
@@ -285,9 +312,13 @@ class TruePositiveRateParity(UtilityParity):
...
-        super().load_data(X, pd.Series(y),
-                          event=pd.Series(y).apply(lambda y: _LABEL + "=" + str(y)).where(y == 1),
-                          **kwargs)
+        base_event = pd.Series(y).apply(lambda y: _LABEL + "=" + str(y)).where(y == 1)
+        control_features = kwargs.get(_KW_CONTROL_FEATURES)
+        if control_features is not None:
+            event = base_event.combine(pd.Series(control_features), _combine_event_and_control)
+        else:
+            event = base_event
+        super().load_data(X, pd.Series(y), event=event, **kwargs)
```
MiroDudik, line 321: There's too much copy-pasting happening here. Also `**kwargs` doesn't help. We should have something like
```python
    def load_data(self, X, y, *, sensitive_features, control_features=None):
```

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
riedgar-ms: There are.... issues with that. For example, `ExponentiatedGradient.fit()` uses `y_train.shape[0]` in its error estimates. That's relying on the data conversion being done in `_validate_and_reformat_input()`. I'm squinting at the usage of `y` inside `Lagrangian`, and I think that will be OK if it's with the unprocessed `y` .... but it's hard to be sure without a bunch of testing. There's a similar consideration in GridSearch.

I'm not saying that this isn't a good idea - and `_validate_and_reformat_input()` is turning into a bit of a monster too - but this is non-trivial.

riedgar-ms: I think I have a way forward, but the surgery is going to be moderately major.

adrinjalali: You probably want something like `sklearn.utils.validation._num_samples`

riedgar-ms:
@MiroDudik , I've done some refactoring of the `Moments` code. Please can you take another look.

fairlearn/reductions/_moments/utility_parity.py
riedgar-ms: There has been some heavy refactoring of this code. Let me know what you think.

fairlearn/_input_validation.py
MiroDudik: why?

fairlearn/reductions/_moments/utility_parity.py
```diff
@@ -282,12 +374,30 @@ class TruePositiveRateParity(UtilityParity):
...
+            The feature array as a :class:`pandas.DataFrame`
+        y : pd.Series
+            The label vector as a :class:`pandas.Series`
+
+        Returns
+        -------
+        pandas.Series
+            The event vector
+        """
+        assert isinstance(X, pd.DataFrame)
```
MiroDudik, line 395: I think that at some point we might have decided not to use `assert` in our libraries and only in tests? @romanlutz @adrinjalali ?

fairlearn/reductions/_moments/bounded_group_loss.py
```diff
@@ -34,12 +35,17 @@ def default_objective(self):
         """Return a default objective."""
         return MeanLoss(self.reduction_loss)
 
-    def load_data(self, X, y, **kwargs):
+    def load_data(self, X, y, *, sensitive_features):
         """Load data into the moment object."""
-        kwargs_mod = kwargs.copy()
+        X_train, y_train, sf_train, _ = \
```
MiroDudik, line 40:
```suggestion
       _, _, sf_train, _ = \
```

fairlearn/reductions/_moments/moment.py
```diff
@@ -25,31 +24,50 @@ class Moment:
     def __init__(self):
         self.data_loaded = False
 
-    def load_data(self, X, y, **kwargs):
+    def load_data(self,
+                  X,
+                  y: pd.Series,
```
MiroDudik, line 29: okay for now. in future, this should be the same as `y_true` and `y_pred` in `MetricFrame`

fairlearn/reductions/_moments/moment.py
```diff
@@ -25,31 +24,50 @@ class Moment:
...
         self.tags = pd.DataFrame({_LABEL: y})
-        if _KW_SENSITIVE_FEATURES in kwargs:
-            self.tags[_GROUP_ID] = kwargs[_KW_SENSITIVE_FEATURES]
+        if sensitive_features is not None:
+            self.tags[_GROUP_ID] = sensitive_features
         self.data_loaded = True
         self._gamma_descr = None
 
+    @property
+    def X(self):
```
MiroDudik, line 57: let's rename this to something that begins with `_` so that this is not part of the exported interface.

fairlearn/reductions/_moments/moment.py
```diff
@@ -25,31 +24,50 @@ class Moment:
...
+        """Return the feature array as supplied to :meth:`load_data`."""
+        return self._X
+
     @property
     def total_samples(self):
         """Return the number of samples in the data."""
         return self.X.shape[0]
 
+    @property
+    def y_reformat(self):
```
MiroDudik, line 67: let's rename this to something that begins with `_` so that this is not part of the exported interface.

fairlearn/reductions/_moments/utility_parity.py
```diff
@@ -3,16 +3,30 @@
...
 _LOWER_BOUND_DIFF = "lower_bound_diff"
 _MESSAGE_INVALID_BOUNDS = "Only one of difference_bound and ratio_bound can be used."
 _DEFAULT_DIFFERENCE_BOUND = 0.01
 
+_CTRL_EVENT_FORMAT = "control={0},{1}"
+
+
+def _combine_event_and_control(event: str, control: str) -> str:
+    if pd.notnull(control):
+        if event != _ALL:
```
MiroDudik, line 23: i don't think that this is necessary. the naming convention of the lambda index is right now only **internal** interface, so i think that you can always use `_CTRL_EVENT_FORMAT.format(control, event)`

fairlearn/reductions/_moments/utility_parity.py
```diff
@@ -88,26 +102,77 @@ def default_objective(self):
         """Return the default objective for moments of this kind."""
         return ErrorRate()
 
-    def load_data(self, X, y, event=None, utilities=None, **kwargs):
-        """Load the specified data into this object.
+    def compute_base_event(self, X: pd.DataFrame, y: pd.Series) -> pd.Series:
```
MiroDudik, line 105: What's the need for this? Why do we need this more complicated mechanism and cannot just handle everything on the level of `load_data`? if you prefer, you could change the signature to something like:
```python
def load_data(self, X, y, *, sensitive_features, event=None, utilities=None, control_features=None)
```

fairlearn/reductions/_moments/utility_parity.py
MiroDudik: [but this is not a biggie... i'm fine leaving as is for now]

fairlearn/reductions/_moments/utility_parity.py
```diff
@@ -282,12 +374,30 @@ class TruePositiveRateParity(UtilityParity):
 
     short_name = "TruePositiveRateParity"
 
-    def load_data(self, X, y, **kwargs):
```
MiroDudik, line 285: We should be able to change this signature as well as other moments derived from `UtilityRate` to something like:
```python
    def load_data(self, X, y, *, sensitive_features, control_features=None):
```

fairlearn/reductions/_moments/utility_parity.py
romanlutz: Not a great way to do error handling... at least there should be a message to be shown if the assert fails. What's the argument against raising exceptions?

riedgar-ms: I used `assert` because this is not an expected error condition, but an internal sanity check. This should never be called unless the arguments have been converted.

fairlearn/reductions/_moments/utility_parity.py
riedgar-ms: I came up with this pattern to address the following comment:
https://github.com/fairlearn/fairlearn/pull/631#discussion_r540527969
That version of the code left `UtilityParity` unchanged, which meant that the code to handle `control_features` was appearing in each `load_data()` method.

fairlearn/reductions/_moments/moment.py
riedgar-ms: I only did this to make `X` read-only. I can change it back.

fairlearn/reductions/_moments/moment.py
riedgar-ms: I believe that by the time this call is made, `y` should have been turned into a `Series` already. Having each method have to accept all types of input feels like a good way of doing a lot of unnecessary conversions.

fairlearn/reductions/_moments/bounded_group_loss.py
riedgar-ms: Why? The Moments don't call anything else, and the code in them has always assumed that they are dealing with a Series for `y` (hence the `y_reformat` property which I added).

MiroDudik: For example, our `ConditionalLossMoment` should very easily work with multi-class classification. In that case it's often useful to represent `y` as a two-dimensional array.

For the time being, as I suggest below, we can keep `y_train` here and rename the property to something like `_y_reformat` until we figure out whether we really want to export it. My hunch is that we shouldn't export it and we just currently have some abstraction leakage [to be fixed in a future PR].

fairlearn/reductions/_moments/utility_parity.py
MiroDudik: I see. I think that a better way to avoid copy-pasting would be to go essentially with your previous implementation, but create a helper function to handle combining `base_event` and `control_features` into `event`.

The basic issue that I see with this proposal is that `UtilityRate` should not be dealing with `control_features` at all. (In fact, it shouldn't be even dealing with `y`, but that's for another PR.)

fairlearn/_input_validation.py
```diff
@@ -86,28 +107,37 @@ def _validate_and_reformat_input(X, y=None, expect_y=True, enforce_binary_labels
     else:
         result_y = pd.Series(dtype="float64")
 
-    return pd.DataFrame(X), result_y, pd.Series(sensitive_features.squeeze())
+    return pd.DataFrame(X), result_y, pd.Series(sensitive_features.squeeze()), control_features
```
romanlutz, line 110: Do we want to either do `squeeze` for both `sensitive_features` and `control_features` here, or both earlier?

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -188,7 +191,10 @@ def fit(self, X, y, *, sensitive_features, **kwargs):
         else:
             raise ValueError(NOT_SUPPORTED_CONSTRAINTS_ERROR_MESSAGE)
 
-        _, _, sensitive_feature_vector = _validate_and_reformat_input(
+        if kwargs.get(_KW_CONTROL_FEATURES) is not None:
+            raise ValueError(NO_CONTROL_FEATURES)
```
romanlutz, line 195: Is this something we want at some point? If so, let's briefly describe this and put it in an issue with "help-wanted"? It should be simple enough I think. You'd just create individual objects per control group.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -154,9 +151,9 @@ def _call_oracle(self, lambda_vec):
         if isinstance(self.constraints, ClassificationMoment):
             redY = 1 * (signed_weights > 0)
         else:
-            redY = self.y
+            redY = self.constraints.y_reformat
         redW = signed_weights.abs()
-        redW = self.n * redW / redW.sum()
+        redW = self.constraints.total_samples * redW / redW.sum()
```
romanlutz, line 156: I have to say this usage of a different `y_reformat` and different `n` confuses me. Can we at least document what this is supposed to do?

fairlearn/reductions/_moments/moment.py
romanlutz: `reformat` is a verb, so this indicates to me that it does something, rather than just return the reformatTED `y`.

fairlearn/reductions/_moments/utility_parity.py
```diff
@@ -3,16 +3,30 @@
...
+from fairlearn._input_validation import _MESSAGE_RATIO_NOT_IN_RANGE, _validate_and_reformat_input
 from .error_rate import ErrorRate
 
+
 _UPPER_BOUND_DIFF = "upper_bound_diff"
 _LOWER_BOUND_DIFF = "lower_bound_diff"
 _MESSAGE_INVALID_BOUNDS = "Only one of difference_bound and ratio_bound can be used."
 _DEFAULT_DIFFERENCE_BOUND = 0.01
 
+_CTRL_EVENT_FORMAT = "control={0},{1}"
```
romanlutz, line 18: What if either one of them already has a comma? How do we distinguish between control and event? Or is that something we don't care about?

fairlearn/reductions/_moments/utility_parity.py
```diff
@@ -239,13 +304,40 @@ class DemographicParity(UtilityParity):
     The :meth:`signed_weights` method will compute the costs according
     to Example 3 of
     `Agarwal et al. (2018) <https://arxiv.org/abs/1803.02453>`_.
+
+    This Moment also supports control features, which can be used to
```
romanlutz, line 308: I'd suggest staying away from the term `Moment` here as well, unless you link to the base class using ReST

fairlearn/reductions/_moments/utility_parity.py
```diff
@@ -239,13 +304,40 @@ class DemographicParity(UtilityParity):
...
+      P[h(X) = 1 | A = a, C = c] = P[h(X) = 1 | C = c] \; \forall a, c
     """
 
     short_name = "DemographicParity"
 
-    def load_data(self, X, y, **kwargs):
-        """Load the specified data into the object."""
-        super().load_data(X, y, event=_ALL, **kwargs)
+    def compute_base_event(self, X: pd.DataFrame, y: pd.Series) -> pd.Series:
+        """Compute the base event for demographic parity.
```
romanlutz, line 320: nit: we have a lot of different ways of capitalizing {D|d}emographic {P|p}arity. Perhaps worth standardizing at some point, not just for DP, but all kinds of others, too.

test/unit/reductions/moments/test_moments_demographic_parity.py
```diff
@@ -140,8 +140,8 @@ def test_signed_weights():
     a0_threshold = 0.2
     a1_threshold = 0.7
 
-    a0_label = "OneThing"
-    a1_label = "AnotherThing"
+    a0_label = 0xDEAD
```
romanlutz, line 143: What's up with these? :O

romanlutz:
Unfortunately there seems to be no way for me to comment directly on this, so here's the screenshot:
image

I actually disagree with this. If an arg is not explicitly named in the signature I don't really realize it exists. Having `arg=None` there makes it clear that I can provide it, and it doesn't really cost us anything. I'm curious why that's called out as an anti-pattern. It strikes me as a commonly used pattern, see for example LogisticRegression in sklearn https://github.com/scikit-learn/scikit-learn/blob/0fb307bf3/sklearn/linear_model/_logistic.py#L1011

fairlearn/reductions/_moments/moment.py
```diff
@@ -25,31 +24,50 @@ class Moment:
...
+            The feature array
+        y : pandas.Series
+            The label vector
+        sensitive_features : pandas.Series
+            The sensitive feature vector (default None)
         """
         assert self.data_loaded is False, \
             "data can be loaded only once"
-        self.X = X
+        assert isinstance(y, pd.Series)
```
romanlutz, line 45: Can't seem to comment further on the existing thread:
image

I have yet to see a good reason for using asserts in actual shipped code like this, although I'm ready to be convinced otherwise.

For me it's either
- This condition can happen and we should have proper try-except with a descriptive error message OR
- This condition can't happen, in which case we don't need the assert either.

docs/user_guide/mitigation.rst
```diff
@@ -492,6 +492,39 @@ constraints:
                  b          -0.24
     dtype: float64
 
+
+Control features
+^^^^^^^^^^^^^^^^
+
+The above moments (:ref:`demographic_parity`,
```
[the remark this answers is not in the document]
riedgar-ms: When talking about objects, they should be called what they're called in code, I think. However, some redrafting along the lines you suggest can be done.

docs/user_guide/mitigation.rst
```diff
@@ -492,6 +492,39 @@ constraints:
...
+:ref:`True and False Positive Rate Parity <true_positive_rate_parity>`,
+:ref:`equalized_odds` and :ref:`error_rate_parity`) all support the concept
+of *control features*.
+A control feature stratifies the dataset, and applies the fairness constraint
+within each stratum, but not between strata.
+An example were this might be useful is a loan scenario, where we might want
```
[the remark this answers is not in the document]
riedgar-ms: Thatnks - that's too many 'where's in a sentence anyway.

docs/user_guide/mitigation.rst
```diff
@@ -492,6 +492,39 @@ constraints:
...
+to apply a mitigation for the sensitive features while controlling for some
+other feature(s).
+This should be done with caution, since the control features may have a
+correlation with the sensitive features due to historical biases.
+In the loan scenario, we might choose to control for income level, on the
+grounds that higher income individuals are more likely to be able to repay
+a loan.
+However, due to historic bias, there is a correlation between the income level
```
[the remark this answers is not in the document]
riedgar-ms: Agreed.

docs/user_guide/mitigation.rst
```diff
@@ -492,6 +492,39 @@ constraints:
...
+of individuals and their race and gender.
+
+
+Control features modify the above equations.
+Consider a control feature :math:`c \in \mathcal{C}`.
```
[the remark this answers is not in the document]
riedgar-ms: Tweaking the text a bit to make that more obvious.

fairlearn/_input_validation.py
riedgar-ms: I don't think it really matters. I think there are larger issues with this function anyway.

fairlearn/postprocessing/_threshold_optimizer.py
riedgar-ms: I can certainly file an issue for it.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
riedgar-ms: It's doing exactly what it was before, except that knowledge of the data has been pushed into the Moment, per @MiroDudik 's suggestion.

fairlearn/reductions/_moments/bounded_group_loss.py
riedgar-ms: I've renamed `y_reformat` to `_y_as_series`. Not that I'm a huge fan of objects accessing each other's innards either. But then again, 'private' is merely a convention in Python.

fairlearn/reductions/_moments/moment.py
riedgar-ms: "Can't" is a very strong word, especially in Python.

I think this is partly because it's not clear to me what the public API of the `Moment` objects is supposed to be. I'm pretty sure this method is only used 'internally' (rather than externally by users), but I'm not certain. And while I agree that it's not going to give the prettiest of error messages, it is at least going to be clear as to what's wrong. Rather than blowing up with something more obscure later in the method - or worse, not blowing up for some time.

fairlearn/reductions/_moments/moment.py
riedgar-ms: Reverted to how it was before.

fairlearn/reductions/_moments/moment.py
riedgar-ms: As noted above, this is now `_y_as_series` to be more descriptive.

fairlearn/reductions/_moments/utility_parity.py
riedgar-ms: I could make it `event={1}` too, I suppose. But so long as this generates unique strings, I don't believe we ever care.

fairlearn/reductions/_moments/utility_parity.py
riedgar-ms: Added link

fairlearn/reductions/_moments/utility_parity.py
riedgar-ms: Agreed, but not now.

fairlearn/reductions/_moments/utility_parity.py
riedgar-ms: See note above about what the public API is supposed to be. This particular point is moot, though since I'm about to remove this method.

test/unit/reductions/moments/test_moments_demographic_parity.py
riedgar-ms: The moving around of the input conversions has resulted in a requirement for numerical labels. As for the specific values themselves.... `0xDEADBEEF` was used a a 'fill' value for memory, to make it easier to catch uninitialised access:
https://en.wikipedia.org/wiki/Magic_number_(programming)#Magic_debug_values

fairlearn/reductions/_moments/utility_parity.py
riedgar-ms: Actually, things we have generally set the event to be `Label=<STRING>`

fairlearn/reductions/_moments/utility_parity.py
riedgar-ms: Done

fairlearn/reductions/_moments/moment.py
MiroDudik: agreed...

fairlearn/_input_validation.py
MiroDudik: I'm fine keeping as is for now. But this was my reasoning:
* we use the optional argument `sensitive_features` all over the code and documentation, so if we decide to change that convention, we need to do some form of extensive Find/Replace anyway.
* having a constant called `_KW_SENSITIVE_FEATURES` slightly simplifies Find/Replace exercise, but introduces the need to include `import _KW_SENSITIVE_FEATURES` whenever I want to use `'sensitive_features'` string. That's a rather non-trivial dependency. It also seemingly results in a harder to read code. Do we want to have a master file `fairlearn/_constants.py` with all of keyword attributes that we use?
* at most places we should use explicit keyword arguments, so the use case of `_KW_SENSITIVE_FEATURES` is very limited (we might be able to remove it altogether).

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
MiroDudik: This is the right behavior for now. In a future PR we may want to push the construction of `redY` and `redW` into `signed_weights`. But please keep as is for now.

fairlearn/_input_validation.py
riedgar-ms: After the latest revisions, `_KW_SENSITIVE_FEATURES` is only use in `_input_validation.py` and `_KW_CONTROL_FEATURES` only pops up when `ThresholdOptimizer` says "can't to that."

fairlearn/reductions/_moments/utility_parity.py
riedgar-ms: Updated.

fairlearn/reductions/_moments/utility_parity.py
riedgar-ms: Done

fairlearn/reductions/_moments/moment.py
MiroDudik: I'm generally in agreement with @romanlutz's  observation, but I think that @riedgar-ms has a point. I have definitely used asserts for what's effectively "consistency checks of data structures". They are only effective [in the debug mode](https://docs.python.org/3/reference/simple_stmts.html#assert), so if somebody executes python with `-O`, they are ignored.

That said, we have places in the codebase where asserts should be probably replaced by exceptions.

fairlearn/reductions/_moments/utility_parity.py
```diff
@@ -239,13 +264,29 @@ class DemographicParity(UtilityParity):
...
+    def load_data(self, X, y, *, sensitive_features, control_features=None):
         """Load the specified data into the object."""
-        super().load_data(X, y, event=_ALL, **kwargs)
+        _, y_train, sf_train, cf_train = \
+            _validate_and_reformat_input(X, y,
+                                         enforce_binary_labels=True,
+                                         sensitive_features=sensitive_features,
+                                         control_features=control_features)
+
+        base_event = pd.Series(data=_ALL, index=range(y_train.shape[0]))
```
MiroDudik, line 287:
```suggestion
        base_event = pd.Series(data=_ALL, index=y_train.index)
```

fairlearn/reductions/_moments/utility_parity.py
```diff
@@ -383,6 +442,18 @@ class ErrorRateParity(UtilityParity):
...
+    def load_data(self, X, y, *, sensitive_features, control_features=None):
         """Load the specified data into the object."""
-        super().load_data(X, y, event=_ALL, utilities=np.vstack([y, 1-y]).T, **kwargs)
+        _, y_train, sf_train, cf_train = \
+            _validate_and_reformat_input(X, y,
+                                         enforce_binary_labels=True,
+                                         sensitive_features=sensitive_features,
+                                         control_features=control_features)
+        utilities = np.vstack([y_train, 1-y_train]).T
+        base_event = pd.Series(data=_ALL, index=range(y_train.shape[0]))
```
MiroDudik, line 453:
```suggestion
        base_event = pd.Series(data=_ALL, index=y_train.index)
```

MiroDudik (approved):
As far as I'm concerned, this is good to go as is. Please review the remaining comments!

test/unit/input_convertors.py
riedgar-ms: That's a question for another time.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
riedgar-ms: This is perhaps an area we should take some time on in future. I think that part of the back-and-forth we've been having with this PR are due to us not being quite sure what should be where.

fairlearn/reductions/_moments/moment.py
riedgar-ms: Also, in this case, I'm only asserting the PEP484 annotation :-)

Since Moments are not really user-facing (apart from the constructor), I'm going to leave it in, but I agree that things we expect users to see should be proper exceptions.

fairlearn/reductions/_moments/utility_parity.py
riedgar-ms: Taking this suggestion.

fairlearn/reductions/_moments/utility_parity.py
riedgar-ms: Taking this.

fairlearn/reductions/_grid_search/grid_search.py
riedgar-ms: Done

fairlearn/reductions/_moments/bounded_group_loss.py
riedgar-ms: Done

fairlearn/_input_validation.py
romanlutz: More of a consistency thing, but I'm not going to block on this.

fairlearn/reductions/_moments/moment.py
romanlutz: Don't want to drag this further, but them only running in debug mode is another very good reason not to use them if you ask me...

CHANGES.md
```diff
@@ -1,5 +1,10 @@
 # Changes
 
+### v0.5.1
+
+* Add `control_features` to the classification moments
```
romanlutz, line 5: As a user that doesn't really tell me anything. Before releasing v0.5.1 we should at least write an example for how to use this.

romanlutz (approved):
Other than the few things we said we'd discuss later on (supporting control features in postprocessing, capitalization of constraint names, broader refactoring of internals of reductions methods) I'm mostly concerned that people won't know how to use this. The CHANGES.md description isn't very descriptive if you don't know what moments are. I'll sign off assuming that we'll add some sort of example. If that assumption is wrong please let me know.

fairlearn/reductions/_moments/moment.py
riedgar-ms: OK. I'll remove. It was mainly so that when putting this together I could be sure that all paths ending up here had been through input validation and reforming.

CHANGES.md
riedgar-ms: I'll expand it a bit. There are examples in the user guide, although there isn't a notebook. We really need some decent case studies, rather than manufacturing new examples off the Adult Census data....

PR #634, opened 2020-11-12
Predictive Policing Study

michaelamoako:
Linter issues: Line too long, blank line required between summary and description, missing docstring

Signed-off-by: Michael Amoako <miamoako@microsoft.com>


michaelamoako:
Remaining questions:

- How do I add colors to table in reST format (desired colors indicated in parentheses)
- Should **Exploration** be a separate section in Contents, a subsection of Fairlearn results, or left as is?

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+import numpy as np
+import pandas as pd
+from fairlearn.metrics import MetricFrame
+from fairlearn.metrics import selection_rate
+from fairlearn.reductions import DemographicParity, ExponentiatedGradient
+from fairlearn.datasets import fetch_boston
+from sklearn.metrics import accuracy_score
+from sklearn.tree import DecisionTreeClassifier
+from statistics import mean, median
+from fairlearn.widget import FairlearnDashboard
```
romanlutz, line 126: The dashboard will be removed shortly. We'll need to replace whatever you're using it for with the new plots #561 or custom ones.

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+# The case study on predictive policing and applications of machine
+# learning techniques to making decision-making in the criminal justice
+# system more ‘objective’ reveals that this socio-technical problem
+# requires extensive thoughtful consideration. The findings show how
+# biased data and lack of transparency in models can result in ‘unfair’
+# models that disproportionately oppress marginalized communities -
+# including low income communities as demonstrated in this case.
+#
+# This case study is also an interesting illustration of the general harms
+# of ‘algorithmic decision-making’ systems in various socio-technical
```
romanlutz, line 351: I believe we have different spellings of sociotechnical with and without hyphen. Let's be consistent. I think elsewhere it's without hyphen.

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+# In this document we outline a sample sociotechnical case study - using a
+# hypothetical predictive policing scenario - based on data found in the
+# `Boston housing dataset <https://www.kaggle.com/c/boston-housing>`_
+# which we use in our scenario to predict per capita crime rate.
+# Discussion around the known fairness issues associated with this dataset
+# is included `here
+# <https://medium.com/@docintangible/racist-data-destruction-113e3eff54a8>`_.
+# For further questions about this document please contact Bruke Kifle
+# (@brkifle) or Michael Amoako (@miamoako). The full notebook containing
+# the code below is available upon request.
```
romanlutz, line 34: what does this mean? What's missing here?

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+# 1. Overview
+# 2. Background
+# 3. Fairlearn Results
+# 4. Problem area & Sociotechnical context
+# 5. Conclusion
+#
+# Overview
+# --------
+#
+# In this document we outline a sample sociotechnical case study - using a
```
romanlutz, line 25: I would feel more comfortable if we could mention the purpose of this study from the start. I think we should express that we do not endorse predictive policing and point to the extensive research on the topic that outlines related issues.

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+# hypothetical predictive policing scenario - based on data found in the
+# `Boston housing dataset <https://www.kaggle.com/c/boston-housing>`_
```
romanlutz, line 27: There's quite a bit of a jump from "we're using a housing dataset" to "we predict crime rate". As a reader I'm confused by this. How does that connection make sense? Does it at all? I think it's worth pointing out how such a design choice can affect the outcome.

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+#       - 1000(Bk - 0.63)^2 where Bk is the proportion of blacks by town
+#    *  - LSTAT (BLUE)
+#       - % lower status of the population
+#    *  - MEDV (GREEN)
+#       - Median value of owner-occupied homes in $1000's
+# *Figure A: Dataset variables*
+#
+# Where MEDV (colored green in Figure A above) is the target variable
+# (that a model built on this dataset is intended to predict). We make
+# modifications below with the understanding that LSTAT is a parameter of
```
romanlutz, line 112: If we want to be consistent with Fairlearn's documentation let's call it "sensitive feature". You can link to the user guide intro section for a description.

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+#       - pupil-teacher ratio by town
+#    *  - B
+#       - 1000(Bk - 0.63)^2 where Bk is the proportion of blacks by town
+#    *  - LSTAT (BLUE)
+#       - % lower status of the population
+#    *  - MEDV (GREEN)
+#       - Median value of owner-occupied homes in $1000's
+# *Figure A: Dataset variables*
+#
+# Where MEDV (colored green in Figure A above) is the target variable
```
romanlutz, line 110: I'm confused. Why is predicting value of homes equal to predicting crime rates? If you meant to say that that's the original purpose of the dataset please make this clearer.

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+#
+#
+# **After Mitigation - 74% and 42% respectively**
+#
+#
+# The significant drop in accuracy indicates that the underlying issue may
+# not be at the model level but rather a reflection of the data (and
+# sociotechnical context itself). In the given data, LSTAT is a reliable
+# predictor of the crime rate threshold condition, which is a reflection
+# of systemic biases leading to this observed trend. This is why accuracy
```
romanlutz, line 313: I think I see where you're trying to go with this, but can we be more specific on what the systemic biases are that you're referring to?

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+# problems. In this scenario, we’ve seen the grave consequences of failing
+# to acknowledge and ensure such considerations are a part of the design
+# and deployment of such ML systems. The result? A feedback loop of
+# discrimination and injustice due to lack of transparency and biased
+# training data. Some key takeaways for this case include:
+#
+#
+# * Algorithms / Algorithmic decision-making systems are not ‘objective’
+# * Lack of transparency and biased historical data leads to discriminatory models
+# * Biased data -> Biased predictions = Feedback loop of discrimination & injustice
```
romanlutz, line 331: @hannawallach will have thoughts on this. Perhaps "data with fairness issues -> unfair predictions"

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+#   of disparity would we be looking to reduce? (Explored in the section
+#   to follow)
+#
+# Another important question to consider: should the described problem
+# (policing) be framed as an ML task at all?
+#
+# **Exploration: What if we attempt to apply Fairlearn's mitigation functionality?**
+#
+#
+# *Quickstart - Mitigating Disparity*
```
romanlutz, line 263: Why are we calling this quickstart?

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+#
+#
+# If we observe disparities between groups we may want to create a new
+# model while specifying an appropriate fairness constraint. Note that the
+# choice of fairness constraints is crucial for the resulting model, and
+# varies based on application context. If selection rate is highly
+# relevant for fairness in this contrived example, we can attempt to
+# mitigate the observed disparity using the corresponding fairness
+# constraint called Demographic Parity. In real world applications we need
```
romanlutz, line 272: selection rate isn't known to everyone. Can you describe what that means in this context?

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+mean_crimerate = mean(crimerate_data)
+median_crimerate = median(crimerate_data)
+threshold_crime_rates = [mean_crimerate, median_crimerate]
+min_crimerate = min(crimerate_data)
+max_crimerate = max(crimerate_data)
+"""
+[0] = Mean
+[1] = Median
+"""
+y_true = (crimerate_data > threshold_crime_rates[0]) * 1
```
romanlutz, line 150: Interesting design choice. Why binary classification and not regression? I think this goes back to describing what would happen to the model outputs. Does 1 mean we'll have police presence there? It's not 100% clear to me. If it's a regression task then I can proportionately assign people to go to certain areas, but if I have way too many 1s then I don't know how I'd decide.
[I wouldn't want pred. policing to be used at all, so this is purely hypothetical, but I think it needs a bit more detail to make it clear what harms this introduces]

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+# **After Mitigation - 74% and 42% respectively**
+#
+#
+# The significant drop in accuracy indicates that the underlying issue may
+# not be at the model level but rather a reflection of the data (and
+# sociotechnical context itself). In the given data, LSTAT is a reliable
+# predictor of the crime rate threshold condition, which is a reflection
+# of systemic biases leading to this observed trend. This is why accuracy
+# for the higher bins falls off drastically if we try to achieve similar
+# selection rates across LSTAT bins.
```
romanlutz, line 315: This somehow gives me the impression that if we had higher accuracy it would be fine. I don't think that's true, though. The fundamental problem is that historical overpolicing of certain areas (and thereby people) has created a situation where `y_true` is really not "true", but rather the recorded policing data. I'm sure a study on pred. policing can say this much more eloquently than my comment :-)

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+#
+# Problem Area & Sociotechnical Context
+# -------------------------------------
+#
+# The case study above highlights the importance of considering the
+# sociotechnical aspects of tasks traditionally deemed to be solely ML
+# problems. In this scenario, we’ve seen the grave consequences of failing
+# to acknowledge and ensure such considerations are a part of the design
+# and deployment of such ML systems. The result? A feedback loop of
```
romanlutz, line 324: I understand what you mean, but I think we need to be much more explicit in pointing out what the feedback loop is here and how it gets created. We can't assume everyone has the background to infer it.

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+# %%
+#
+#
+# Contents
+# ========
+#
+# 1. Overview
+# 2. Background
+# 3. Fairlearn Results
+# 4. Problem area & Sociotechnical context
```
romanlutz, line 19: I feel like the context is by far the most important. I'm not sure it even needs to be separate from background (?). The use of Fairlearn and demonstrating why that's not a solution to the problem is a great bonus, but the focus is elsewhere (at least in my mind).

romanlutz (commented):
@miamoako @brkifle thanks for getting this started! You've clearly put in some major effort and I think it shows in the thoughtfulness of many passages. I've tried to be a bit more critical than usual since I consider this a high stakes scenario and want to make absolutely sure we don't misrepresent anything or imply support for technical solutions to this complex issue.

I'm sure our community has plenty of perspectives on this as well. You may want to spread the news on Gitter as well to get more input. Most people don't check GitHub on a daily basis so your feedback could take a while to trickle in.

@mmadaio  @MiroDudik  @hannawallach  @riedgar-ms

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+# characteristics of the scenario:
+#
+#
+# * Stakeholders: police officers, members of the community (especially
+#   people who may be more affected by predictive policing than others),
+#   contractor / data scientist
+#
+# * Harms: overpolicing of neighborhoods can
+#   lead to disproportionate effect on communities in these neighborhoods
+#   (perhaps exacerbated by feedback loop)
```
riedgar-ms, line 59: I think that the 'bad' feedback loops  should be highlighted a bit more. I know that the next two bullet points expound a bit, but I think that the feedback is a big part of why we don't think this is a good idea.

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+#
+# * Notes: Feedback loops!
+#   Prediction on behavior based on circumstances; perhaps useful for
+#   aggregate observations about behavior but less so for individual
+#   predictions
+#
+# * Dangerous to assume that ‘crime’ is a single phenomenon.
+#   There are different kinds of crime, so also need to make sure that data
+#   match the crime being predicted.
```
riedgar-ms, line 68: This bit is very important to the feedback - something like a broken taillight is only going to be caught if you have a police officer present. Burglary will tend not to have that dependency.

Also related to this are the 'no failure' criteria from the article Minsoo shared in the reading group: if the number of reported crimes goes down, then predictive policing must be working! If the number of arrests go up, then predictive policing must be working!

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+median_lstat = median(lstat)
+max_lstat = max(lstat)
+min_lstat = min(lstat)
+
+print('Mean LSTAT ', mean_lstat)
+print('Median LSTAT ', median_lstat)
+print('Max LSTAT ', max_lstat)
+print('Min LSTAT ', min_lstat)
+
+bins = [0, 10, 20, 30, 40]
```
riedgar-ms, line 193: Can these be computed based on the min and max LSTAT? Otherwise these look a bit like magic numbers.

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+#    *  - RAD
+#       - index of accessibility to radial highways
+#    *  - TAX
+#       - full-value property-tax rate per $10,000
+#    *  - PTRATIO
+#       - pupil-teacher ratio by town
+#    *  - B
+#       - 1000(Bk - 0.63)^2 where Bk is the proportion of blacks by town
+#    *  - LSTAT (BLUE)
+#       - % lower status of the population
```
riedgar-ms, line 105: What does this mean? It's used as the sensitive feature below, so I'd like to understand it better.

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+
+rawData = bostonDS.data
+print(rawData)
+print('Min crime rate', min_crimerate)
+print('Max crime rate', max_crimerate)
+print('Mean crime rate', mean_crimerate)
+print('Median crime rate', median_crimerate)
+# %%
+#
+# **2. We build a classifier using CRIM as the target value that the model ultimately predicts, and also removing CRIM in the data used to train the model.**
```
riedgar-ms, line 160: It would be interesting to see whether you get the same results if you drop your sensitive feature from the dataset as well.

examples/plot_pp_study.py
riedgar-ms: Given that two of the LSTAT bins end up with 100% accuracy I very much agree. Results like that are suspicious :-)

examples/plot_pp_study.py
riedgar-ms: Very much agree.

riedgar-ms (commented):
Thanks for starting this!

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+# * Notes: Feedback loops!
+#   Prediction on behavior based on circumstances; perhaps useful for
+#   aggregate observations about behavior but less so for individual
+#   predictions
+#
+# * Dangerous to assume that ‘crime’ is a single phenomenon.
+#   There are different kinds of crime, so also need to make sure that data
+#   match the crime being predicted.
+#
+# For this case study, we are utilizing the Boston housing value dataset
```
mmadaio, line 70: Still feels like more justification is needed for why this dataset is used.

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+# (described here). Originally, this dataset includes the following variables:
+#
+# .. list-table::
+#    :header-rows: 1
+#    :widths: 5 20
+#    :stub-columns: 1
+#
+#    *  - Variable
+#       - Description
```
mmadaio, line 79: For most of these variables, it's not clear why these might reasonably be used to predict crime

examples/plot_pp_study.py
mmadaio: Agreed

examples/plot_pp_study.py
mmadaio: Agreed, and more generally, I think from what you and Bruke have said in the presentations, @michaelamoako, you're trying to do something more critical with the scenario than is coming across here. Specifically around choices of dataset, sensitive features, etc.

examples/plot_pp_study.py
mmadaio: Might be worth looking at how this task is traditionally framed also. The CAPP-PGH primer might be helpful here. Depending on your goal, you may want to replicate what those approaches do, or take a different approach.

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+from fairlearn.widget import FairlearnDashboard
+
+# %%
+#
+# **1. We devise a new target column based on CRIM (colored blue in Figure A above), indicating whether a certain per capita crime rate threshold is met.**
+#
+#    While we calculate and print Mean, Median, Max, and Min crime
+#    rates below, the analysis to follow is specifically for the mean crime
+#    rate as the chosen threshold. Similar results are observed for median
+#    crime rate as the chosen threshold as well (which one can observe by
```
mmadaio, line 135: Why this choice? Is this a common practice in predictive policing?

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+bins = [0, 10, 20, 30, 40]
+print(bins)
+labels = [1, 2, 3, 4]
+
+lstat_binned = pd.cut(lstat, bins, labels=labels)
+# %%
+#
+# **4. We treat lstat_binned (defined in step 3 above) as the sensitive feature on which to assess fairness metrics.**
+#
+accuracy_frame = MetricFrame(accuracy_score, y_true, y_pred, sensitive_features=lstat_binned)
```
mmadaio, line 202: Why was accuracy the metric you decided to use here?

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+
+# %%
+#
+
+classifier.fit(data_without_crimerate, y_true)
+
+y_pred = classifier.predict(data_without_crimerate)
+# %%
+#
+# **3. We create bins for LSTAT (colored blue in Figure A above), to understand the models performance on different LSTAT value intervals.**
```
mmadaio, line 176: As @riedgar-ms said above, I think we need some rationale for why "lower status" (is that socioeconomic status?) is the relevant sensitive feature. Has this been shown by others to be relevant?

examples/plot_pp_study.py
michaelamoako: Once the dashboard is removed, my understanding is I would do the following to add (as @kevinrobinson suggested) the new plots: 

**At the top** 
from fairlearn.plots import plot_disparities_in_performance, plot_disparities_in_selection_rate

**In the sections where I produced plots**
plot_disparities_in_performance(y_true, y_pred, _sensitive_feature_)

plot_disparities_in_selection_rate(y_true, y_pred, _sensitive_feature_)

Is my understanding correct? When should I anticipate his PR being merged?

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+#
+# *Figure C: Disparity in selection rate across lstat bins*
+#
+# Here are some key questions to think about given the results above:
+#
+#
+# * Using mean or median crime rate as the threshold value, the overall
+#   accuracy of the model is 95%. Does that mean the model is good to go?
+#   Can there still be issues despite what seems like a properly working
+#   model? Tradeoffs between accuracy and fairness?
```
mmadaio, line 246: I think this scenario should probably take a stance rather than just raising these questions. At least citing and discussing prior research (e.g., Richardson et al., 2019) or news articles that discuss the limits and harms of predictive policing even with "accurate" models.

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+#
+# * What does it mean that the last bin has the highest selection rate?
+#   Given the sociotechnical context, what would this imply in terms of
+#   the societal impact and outcome?
```
mmadaio, line 250: Again, I think it's worth having _some_ sort of answer to these, even if we ultimately leave it up to the reader

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+#
+# * What would a ‘mitigation’ mean in this scenario? Assuming no
+#   constraints on the mitigations offered in Fairlearn modules, what type
+#   of disparity would we be looking to reduce? (Explored in the section
+#   to follow)
+#
+# Another important question to consider: should the described problem
+# (policing) be framed as an ML task at all?
```
mmadaio, line 258: This feels like a pretty key question, and something I think we should either raise at the outset of the scenario, or address more prominently here...

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+# constraint called Demographic Parity. In real world applications we need
+# to be mindful of the sociotechnical context when making such decisions.
+# The Exponentiated Gradient mitigation technique used fits the provided
+# classifier using Demographic Parity as the objective, leading to a
+# vastly reduced difference in selection rate.
+#
+# *Hypothesis*
+#
+#
+# In this sociotechnical context, the disparity is inherent in the data
```
mmadaio, line 281: I think we need a bit more detail here - is that specifically the Boston housing dataset? (and thus readers should just use a different dataset?) Or is that about policing and crime data more generally?

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+# a significant difference in the accuracy for the 3rd and 4th bins:
+#
+# **Before Mitigation - 95%+ accuracy on both bins**
+#
+#
+# **After Mitigation - 74% and 42% respectively**
+#
+#
+# The significant drop in accuracy indicates that the underlying issue may
+# not be at the model level but rather a reflection of the data (and
```
mmadaio, line 310: I agree, but why does that drop in accuracy suggest that? Will that be obvious to the reader?

examples/plot_pp_study.py
mmadaio: Agreed - see some of the links I provided above

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+# of systemic biases leading to this observed trend. This is why accuracy
+# for the higher bins falls off drastically if we try to achieve similar
+# selection rates across LSTAT bins.
+#
+# Problem Area & Sociotechnical Context
+# -------------------------------------
+#
+# The case study above highlights the importance of considering the
+# sociotechnical aspects of tasks traditionally deemed to be solely ML
+# problems. In this scenario, we’ve seen the grave consequences of failing
```
mmadaio, line 322: What are the consequences that we've seen in this scenario? I think this should be more explicit throughout..

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+# to acknowledge and ensure such considerations are a part of the design
+# and deployment of such ML systems. The result? A feedback loop of
+# discrimination and injustice due to lack of transparency and biased
+# training data. Some key takeaways for this case include:
+#
+#
+# * Algorithms / Algorithmic decision-making systems are not ‘objective’
```
mmadaio, line 329: Again, I agree with this, but has this scenario demonstrated that?

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+# * Lack of transparency and biased historical data leads to discriminatory models
```
mmadaio, line 330: I'm not sure I see how the analyses in the scenario show this

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+# are racist (MIT Technology Review)
+#
+# Conclusion
+# ----------
+#
+# The case study on predictive policing and applications of machine
+# learning techniques to making decision-making in the criminal justice
+# system more ‘objective’ reveals that this socio-technical problem
+# requires extensive thoughtful consideration. The findings show how
+# biased data and lack of transparency in models can result in ‘unfair’
```
mmadaio, line 346: I think we should be drawing from the [fairness style guide](https://fairlearn.github.io/v0.5.0/contributor_guide/how_to_talk_about_fairness.html) when we talk about this piece. I also think these points about problems with objectivity and unfairness should be made more explicit earlier and throughout.

examples/plot_pp_study.py
```diff
@@ -0,0 +1,362 @@
...
+# models that disproportionately oppress marginalized communities -
+# including low income communities as demonstrated in this case.
```
mmadaio, line 348: And just to recap, the evidence here is that there were differences in accuracy across LSTAT (SES?) bins, and even using a mitigation approach to constrain selection rate disparity across bins did not reduce that disparity, is that right?

michaelamoako: Correct, though as @romanlutz  pointed out, there might be an (unintended) takeaway that if the accuracy were higher, there wouldn't be an issue. Adding commentary on the fact that the fundamental issue in the scenario is that 'y_true' is a product of historical overpolicing (which is the case for data in this space generally, and not specific to this scenario), and adding the links you suggested could help ensure the reader walks away with the intended takeaway.

kevinrobinson:
If the intention here is to demonstrate the value of Fairlearn 0.50, it might be productive to try using the guidelines in [Contributing example notebooks](https://fairlearn.github.io/v0.5.0/contributor_guide/contributing_example_notebooks.html).

If there's a different intention - eg, advocating for not building particular kinds of ML systems, illustrating a concept for educational purposes, reaching an audience other than potential Fairlearn users - it might help to articulate that intention in the pull request description.  That can help sort out feedback on the intention itself vs. feedback on execution.

examples/plot_pp_study.py
romanlutz: Exactly. I think either way is fine for demonstrating that it's not good, we just want to be very clear :-)

examples/plot_pp_study.py
romanlutz: That's right! At this point we can merge it, but there are a few comments on the PR. I expect that we'll see some progress there before the end of the month, but need to check on it now.

michaelamoako:
@mmadaio should I close this PR?

michaelamoako:
Replaced with new PR around Construct Validity

Issue #641, opened 2020-11-20
citation info, bibtex


hbaniecki:
Hi, great work with the package, especially the interactive dashboard.
Is there any citation info for the software, or a Bibtex to copy?
https://github.com/fairlearn/fairlearn/search?q=citation

riedgar-ms:
For the entire package, there is a whitepaper:
https://www.microsoft.com/en-us/research/uploads/prod/2020/05/Fairlearn_whitepaper.pdf
The two main techniques have papers:
https://arxiv.org/pdf/1803.02453.pdf
https://papers.nips.cc/paper/6374-equality-of-opportunity-in-supervised-learning.pdf

I'm not sure of where to get a standardised BibTeX format for these.... @MiroDudik ?

hbaniecki:
Thanks

riedgar-ms:
@MiroDudik just sent this:
```
@techreport{bird2020fairlearn,
author = {Bird, Sarah and Dud{\'i}k, Miro and Edgar, Richard and Horn, Brandon and Lutz, Roman and Milan, Vanessa and Sameki, Mehrnoosh and Wallach, Hanna and Walker, Kathleen},
title = {Fairlearn: A toolkit for assessing and improving fairness in {AI}},
institution = {Microsoft},
year = {2020},
month = {May},
url = "https://www.microsoft.com/en-us/research/publication/fairlearn-a-toolkit-for-assessing-and-improving-fairness-in-ai/",
number = {MSR-TR-2020-32},
}
```

hbaniecki:
Will use

Dref360:
I find it complicated to find this information on the website.

1. Go to fairlearn.io
2. Click on API Docs
3. Click on "About"
4. Find the section https://fairlearn.org/v0.6.0/about/index.html#citing-fairlearn

PR #642, opened 2020-11-23
Add citation information

riedgar-ms:
Provide a BibTeX entry for the whitepaper which describes the Fairlearn package. As we shift to a more open governance model, this should be updated with a more general publication.

Inspired by #641

### DIFF
--- a/docs/about/index.rst
+++ b/docs/about/index.rst
@@ -31,7 +31,7 @@ Governance
 Citing Fairlearn
 ----------------
 
-If you wish to add a citation to Fairlearn, please use the following:
+If you wish to cite Fairlearn in your work, please use the following:
 
 .. code ::
 
@@ -44,6 +44,8 @@ If you wish to add a citation to Fairlearn, please use the following:
         url = "https://www.microsoft.com/en-us/research/publication/fairlearn-a-toolkit-for-assessing-and-improving-fairness-in-ai/",
         number = {MSR-TR-2020-32},
     }
+
+As we shift to a more open governance model, we expect to update this citation.
 
 Frequently asked questions
 --------------------------
--- a/docs/about/index.rst
+++ b/docs/about/index.rst
@@ -28,6 +28,23 @@ Governance
 ----------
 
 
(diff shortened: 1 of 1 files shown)


docs/about/index.rst
```diff
@@ -28,6 +28,25 @@ Governance
...
+        author = {Bird, Sarah and Dud{\'i}k, Miro and Edgar, Richard and Horn, Brandon and Lutz, Roman and Milan, Vanessa and Sameki, Mehrnoosh and Wallach, Hanna and Walker, Kathleen},
+        title = {Fairlearn: A toolkit for assessing and improving fairness in {AI}},
+        institution = {Microsoft},
+        year = {2020},
+        month = {May},
+        url = "https://www.microsoft.com/en-us/research/publication/fairlearn-a-toolkit-for-assessing-and-improving-fairness-in-ai/",
+        number = {MSR-TR-2020-32},
+    }
+
+As we shift to a more open governance model, we expect to update this citation.
```
romanlutz, line 48: I'm not sure I understand what this means. In my mind, it seems reasonable to write a follow-up at some point (I have no timeline in mind) when we expand the focus beyond binary classification & regression to more complex tasks, or perhaps when we manage to get a few examples worked out. That said, I am not aware of plans and this seems to be hinting at something, so I just want to clarify :-)

romanlutz (approved):

docs/about/index.rst
MiroDudik: Agreed with @romanlutz. we will add a new citation if there's a new paper! and we should write a new paper when there's a good reason to write a new paper--so when we go neutral, we should reflect if there's enough for a new paper.

For now, let's just drop "As we shift ..."

MiroDudik (approved):
Good to go once "As we shift ..." sentence is removed.

docs/about/index.rst
riedgar-ms: I had it in to minimise the appearance of excessive MS influence. I'll remove.

PR #646, opened 2020-12-01
Start testing with Python 3.9

riedgar-ms:
Now that Python 3.9 is available on the ADO agents, we can start testing against it.


devops/PR-Gate.yml
```diff
@@ -16,7 +16,7 @@ jobs:
   parameters:
     platforms:  { Linux: ubuntu-latest }
     installationType: 'PipLocal'
-    pyVersions: [3.6]
+    pyVersions: [3.9]
```
romanlutz, line 19: Why remove 3.6? Shouldn't it be additional? FWIW I'm way more concerned about 3.6 breaking since that's on so many machine vs. 3.9 which almost nobody uses yet

riedgar-ms: Because the Windows build is doing 3.6 as well, and this is the gate.

The point is currently moot, since `numpy` is not installing successfully under 3.9. For some reason it's being compiled from source, and there's an error during compilation (it looks like a structure has changed).

romanlutz (approved):

adrinjalali (approved):

PR #649, opened 2020-12-03
Updates for make_derived_metric

riedgar-ms:
I had overlooked some comments on the previous PR for `make_derived_metric()`. This is to address those points.

The largest change is to the data preprocessing, which has been adjusted to avoid the problem of Data Leakage between the training and test sets (credit to @adrinjalali  for the actual `Pipeline` code). This fix has been ported to the other metrics example.

### DIFF
--- a/examples/plot_make_derived_metric.py
+++ b/examples/plot_make_derived_metric.py
@@ -35,10 +35,12 @@ import numpy as np
 import sklearn.metrics as skm
 from sklearn.compose import ColumnTransformer
 from sklearn.datasets import fetch_openml
+from sklearn.impute import SimpleImputer
 from sklearn.linear_model import LogisticRegression
 from sklearn.model_selection import train_test_split
 from sklearn.preprocessing import StandardScaler, OneHotEncoder
-
+from sklearn.compose import make_column_selector as selector
+from sklearn.pipeline import Pipeline
 from fairlearn.metrics import MetricFrame, make_derived_metric
 from fairlearn.metrics import accuracy_score_group_min
 
@@ -46,11 +48,9 @@ from fairlearn.metrics import accuracy_score_group_min
 # Next, we import the data, dropping any rows which are missing data:
 
 data = fetch_openml(data_id=1590, as_frame=True)
-X_raw = data.data.dropna()
-# Figure out which rows were dropped
-mask = data.data.notna().all(axis=1)
-y = (data.target.loc[mask] == '>50K') * 1
-A = X_raw[['race', 'sex']]
+X_raw = data.data
+y = (data.target == ">50K") * 1
+A = X_raw[["race", "sex"]]
 
 # %%
 # We are now going to preprocess the data. Before applying any transforms,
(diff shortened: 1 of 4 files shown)


examples/plot_make_derived_metric.py
```diff
@@ -62,12 +62,14 @@
 X_scaled = sc.fit_transform(X_dummies)
 X_scaled = pd.DataFrame(X_scaled, columns=X_dummies.columns)
 
-X_train, X_test, Y_train, Y_test, A_train, A_test = train_test_split(X_scaled, Y, A,
+X_train, X_test, y_train, y_test, A_train, A_test = train_test_split(X_scaled, y, A,
```
riedgar-ms, line 65: @adrinjalali I'm aware that your biggest concern was about this line, and that I'm not addressing it here. Could you explain again, please?

riedgar-ms: Ping.... @adrinjalali . I'd like to make sure I address your point this time, but I'm not quite sure what you meant.

adrinjalali: Sorry about my lack of clarity, I'll try to explain better. I'm basically referring to this concept: https://en.wikipedia.org/wiki/Leakage_(machine_learning), which in short is:

>In statistics and machine learning, leakage (also data leakage, or target leakage) is the use of information in the model training process which would not be expected to be available at prediction time, causing the predictive scores (metrics) to overestimate the model's utility when run in a production environment.

The _mistake_ I was referring to, is when something is done to the data, before train, test split, which would mean that that _something_ has used information which exists in train AND test data. The test data should be treated as if it didn't exist at the time of the training the model.

Now in this example, we do a `le.fit_transform(...)` on the whole data, so what's the issue?

The issue arises if there happen to be labels in the test data, which don't exist in the training data. In that case, doing a `fit_transform` on the train data, would lead a different result compared to doing that on the whole dataset. This issue becomes much more serious if people start doing transformations such as certain normalizations which would take data distributions into account, since the test data would affect those distribution parameter estimates.

This is why we either need to put every preprocessing step in a `Pipeline` and treat it as one single estimator, or at least first split the data, then do `fit_transform` on the train data, and `transform` on the test data.

riedgar-ms: I think the problem is more that you're conversing with a non-data scientist. I think I understand now.... the point is that the transforms should all be 'fit' on the training data, and then applied to the test data. The `StandardScaler` in particular is going to 'leak' data between the two halves of the split.

Looks like I'll have to acquaint myself with `OneHotEncoder` in place of `get_dummies()`.

riedgar-ms: OK, can you take another look @adrinjalali . I know I could use a pipeline as well, but am I at least doing the right steps now? If so, I can apply this change to the other examples as well (since this preprocessing has been cut-and-pasted a few times).

adrinjalali:
It was easier for me to implement the changes in a PR @riedgar-ms , which you can see here: https://github.com/riedgar-ms/fairlearn/pull/1

examples/plot_new_metrics.py
```diff
@@ -123,37 +127,80 @@ def occupation_transform(occ_string):
...
+# we need to split the data into training and test sets before applying
+# any transforms or scaling:
 
-le = LabelEncoder()
 
-sc = StandardScaler()
-X_dummies = pd.get_dummies(X_raw)
-X_scaled = sc.fit_transform(X_dummies)
-X_scaled = pd.DataFrame(X_scaled, columns=X_dummies.columns)
+(X_train, X_test, y_train, y_test, A_train, A_test) = train_test_split(
```
riedgar-ms, line 137: This section is copied from the other notebook @adrinjalali

adrinjalali (approved):
Thanks @riedgar-ms , LGTM.

mesameki (approved):

adrinjalali:
woohoo, two approvals, merging :)

adrinjalali:
oh, I can't lol

riedgar-ms:
> oh, I can't lol

That's for our ToDo list - make sure you have the right permissions @adrinjalali (and @hildeweerts too). Although in this case, since there's a DCO mess up (I think I did this when I squashed the PR you made against my fork, rather than just merging it), it might take owner permissions. At least until we kill off DCO....

Issue #601, opened 2020-10-20
Improving API Reference format


hildeweerts:
I think our current API Reference could benefit from restructuring. Some ideas:

#### Structure API Reference in tables
The current format puts the documentation of all functions/classes of a module on a single page. As the pages are getting quite long and often contain different types of functionality (e.g., "base" metrics versus fairness metrics in the `metrics` module), it can be difficult to get a good overview of the available functionality. This also makes it harder than necessary to find the documentation of a particular function/class.

Instead, I'd suggest to list the functions/classes in tables with a short description of each item. We can put the actual documentation on a separate page for each function/class. We can put all tables on a single page, similar to [scikit-learn](https://scikit-learn.org/stable/modules/classes.html) or on a separate page for each module, similar to [pandas](https://pandas.pydata.org/docs/reference/index.html). My personal preference would go to the latter.

#### Ordering Attributes/Methods
The attributes and methods of a class are currently ordered alphabetically. 

I'd suggest to group them separately instead, similar to e.g., [this](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.html#pandas.DataFrame) or [this](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html#sklearn.linear_model.LogisticRegression)

#### Make References Consistent
The usage of references seems inconsistent. In `fetch_boston()` the *References* docstring is used, in `fetch_adult()` the reference (including the full url and explicit *'Source:'*) is in the description, in `ThresholdOptimizer` a reference is linked as a clickable *name (year)* in the *Notes*, and all functions/classes in the `reductions` module link the reference as a clickable *name (year)* in the description. 

I'd suggest to follow the [docstring guide](https://numpydoc.readthedocs.io/en/latest/format.html) and put implementation details/background theory in the *Notes* and proper references in the *References* (i.e. not just links, as they might break).

#### See Also
I always like the *See Also* sections of classes/functions, as they can lead you to different parts of the package you may not have considered yourself, especially if you're new to the field. This is probably something that's not as relevant with the current size of the package, but might be nice to consider as Fairlearn grows.

---

Do other people have thoughts on this? Tagging @riedgar-ms @MiroDudik @romanlutz

romanlutz:
Couldn't agree more! I haven't looked into how scikit-learn achieves this yet, but I'm hoping that it's compatible with our docs generation setup. 

Would you like to try this out @hildeweerts ? If not I'll mark it as "help-wanted".

[...]

PR #657, opened 2020-12-15
Change Autodoc sort order

riedgar-ms:
As noted by @hildeweerts  in #601 sorting members strictly alphabetically makes little sense. It turns out that there is an `autodoc` option to group things (e.g. properties, methods) together, so turn this on.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -153,6 +153,10 @@ html_sidebars = {
     "**": ["version-sidebar.html", "sidebar-search-bs.html", "sidebar-nav-bs.html"],
 }
 
+# Change the ordering of the member documentation
+
+autodoc_member_order = 'groupwise'
+
 # The following is used by sphinx.ext.linkcode to provide links to github
 # based on pandas doc/source/conf.py
 
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -153,14 +153,18 @@ html_sidebars = {
     "**": ["version-sidebar.html", "sidebar-search-bs.html", "sidebar-nav-bs.html"],
 }
 
-# Change the ordering of the member documentation
+# Auto-Doc Options
+# ----------------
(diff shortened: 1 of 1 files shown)


riedgar-ms:
This is an improvement, but I'm still not completely convinced. For example, on the metrics page, `make_derived_metric()` is still buried in the middle of the metric functions. To work around this, I believe we'd need to stop using `automodule` and list out the individual members in the order we want. That could get tedious.

romanlutz (approved):
I'm fine with this as an improvement over what we had before. What state are you trying to get to exactly (since you aren't quite happy yet)?

riedgar-ms:
As I said, we sometimes have groupings of functionality within the same 'kind' of object. So this still has `make_derived_metric()` listed in the middle of other actual metrics, when it should probably be listed separately. However, to do that, we'd probably have to curate the list manually, and that has its own pitfalls.

hildeweerts:
If manual curation is required to structure the API in tables (as suggested in #601), I am all for it. What pitfalls are you referring to exactly, @riedgar-ms?

hildeweerts (approved):

riedgar-ms:
@hildeweerts basically my concern is that as soon as you have manual processes (e.g. listing out the members manually), things start to be overlooked.

If you're happy with this, are you able to merge, in addition to approving? I think that you should have gotten that ability, but @adrinjalali  had trouble with a different PR last week.

hildeweerts:
@riedgar-ms that's a very valid concern. I don't think we will be continuously adding many user-facing features, though. To me, the risk of forgetting to add some new functionality to the docs does not outweigh having unstructured docs for all existing functionalities. But I'm happy to hear other people's thoughts on this.

It seems like I was able to wield my newfound powers and merge!

PR #661, opened 2020-12-21
Restructure mitigations section of user guide

riedgar-ms:
The Mitigations section of the user guide is growing rapidly, so put it into its own directory and split into separate files. These are brought in using the `.. include::` directive, and have an `.inc` suffix to avoid double processing.

Also fix some capitalisation issues noted on other PRs.


docs/quickstart.rst
```diff
@@ -172,10 +172,10 @@ while specifying an appropriate fairness constraint. Note that the choice of
 fairness constraints is crucial for the resulting model, and varies based on
 application context. If selection rate is highly relevant for fairness in this
 contrived example, we can attempt to mitigate the observed disparity using the
-corresponding fairness constraint called Demographic Parity. In real world
+corresponding fairness constraint called demographic parity. In real world
```
riedgar-ms, line 175: With apologies to @romanlutz 's teutonic sensibilities ;-)

docs/user_guide/index.rst
```diff
@@ -9,6 +9,6 @@ User Guide
 
    fairness_in_machine_learning
    assessment
-   mitigation
+   mitigation/index
```
romanlutz, line 12: Out of curiosity: will this wreak havoc with multi-version?

docs/user_guide/mitigation/index.rst
```diff
@@ -0,0 +1,81 @@
+.. _mitigation:
```
romanlutz, line 1: Any changes here or just copy-paste?

docs/user_guide/mitigation/index.rst
```diff
@@ -0,0 +1,81 @@
...
+   :code:`predict` method on identical data. For each of our algorithms, we provide
+   explicit access to the probability distribution used for randomization.
+
+
+.. include:: reductions.inc
+.. include:: postprocessing.inc
+
+.. topic:: References:
+
+   .. [#1] Agarwal, Dudik, Wu `"Fair Regression: Quantitative Definitions and
```
romanlutz, line 70: Can these still be referenced from the `.inc` files?

docs/user_guide/mitigation/postprocessing.inc
```diff
@@ -0,0 +1,6 @@
+.. _postprocessing:
+
+Postprocessing
+--------------
+
+.. currentmodule:: fairlearn.postprocessing
```
romanlutz, line 6: I was shocked here that my user guide is gone until I realized that the PR hasn't been completed yet. We should  get to that as well.

docs/user_guide/index.rst
riedgar-ms: It will mean that users can't just click between versions, since the files will have moved.

docs/user_guide/mitigation/index.rst
riedgar-ms: Should just be copy-pasting into three separate files.

docs/user_guide/index.rst
adrinjalali: as long as we have _a_ copy of the docs for each version, I don't think we should worry about how users can find a file in different versions.

PR #563, opened 2020-08-20
Regression branch leijie

Acornagain:
Add CDF_DemographicParity and other relevant components to the fairlearn library.

PR #662, opened 2020-12-21
Ensure Reductions can be pickled

### LINKED ISSUE #569: Picklable objects (currently throws an error)
```python
from sklearn.tree import DecisionTreeClassifier
from fairlearn.reductions import TruePositiveRateDifference, ExponentiatedGradient
clf = ExponentiatedGradient(DecisionTreeClassifier(), TruePositiveRateDifference())
clf.fit(X, y, sensitive_features=A)

import joblib
joblib.dump(clf, 'file.pkl')

# or
# import pickle; pickle.dumps(clf)
```
Throws the error: ```AttributeError: Can't pickle local object '_Lagrangian.best_h.<locals>.h'```

I'd like to be able to pickle a fitted classifier/reductions object, or persist it in some other way.
I have searched the API docs and this doesn't seem to be a feature.

#### Additional context
It seems that the problem is some object that was defined locally and thus cannot be imported.
Maybe exposing this object in the package's namespace would be a solution (?)

riedgar-ms:
Move the inline function in `_Lagrangian.best_h()` into a new callable class, so that the object becomes pickable. Closes #569 .

Also add similar tests for `GridSearch` (which didn't actually suffer from the problem).


fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -16,6 +16,19 @@
 logger = logging.getLogger(__name__)
 
 
+class _h:
```
romanlutz, line 19: @MiroDudik let's please find a slightly more descriptive name? :-)

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -16,6 +16,19 @@
...
+    def __init__(self, classifier):
+        self._classifier = classifier
+
+    def __call__(self, X):
```
romanlutz, line 23: It's always been sort of annoying that we don't call predict, but rather have a callable since it's inconsistent with scikit-learn. Maybe this is a good time to change that?

romanlutz: We're already storing the classifier so perhaps we don't need more?

riedgar-ms: That's a sentiment I have shared. This is a minimal change with the existing pattern, but.... I don't recall if there was a particular reason we kept to this when making all the other changes to ExpGrad.

romanlutz: For me, the issue was that I didn't quite have enough time to sort out everything I didn't like :-) This is definitely one of those. @MiroDudik may have thoughts.

MiroDudik:
Mhmm... can we do this after #563? Especially since `cloudpickle` work to mitigate #569?

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
adrinjalali: yeah by looking at the code, it doesn't seem like something we need.

adrinjalali (commented):
regarding the test files, it seems to me this is a good place to start a sort of a common test directory. We want all estimators to be pickleable, so we can put that under the common tests.

adrinjalali:
`clouldpickle` works, but there are issues with interoperability with dask and other backends which means joblib would still use pickle which means the cloudpickle trick is not relevant there. It's still nice to have the objects pickleable if possible. And this doesn't seem too hard here.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
riedgar-ms: If the reasons for needing are only historical, I'll happily change it to passing an estimator.... @MiroDudik ?

MiroDudik:
> Mhmm... can we do this after #563? Especially since `cloudpickle` works to mitigate #569?

I have checked with @Acornagain and we'll try to get #563 in by the end of February. If it's still not in by the end of February, then you're free to merge this PR (and we'll incorporate those changes into #563).

hildeweerts:
The end of February has passed, are there still things that need to be resolved for this to be merged? @MiroDudik @riedgar-ms @adrinjalali @romanlutz

riedgar-ms:
As @hildeweerts points out, not mere February but the Ides of March have come and gone. I have just updated the name of the class I extracted to be a little more descriptive.

What do we want to do with this (particularly @MiroDudik )? I believe that this the smallest change to make things pickleable. However, as @romanlutz  and @adrinjalali  point out, should we be restructuring things so that this isn't needed at all?

adrinjalali:
I'd be happier with the little restructuring if it's not too hard.

riedgar-ms:
> I'd be happier with the little restructuring if it's not too hard.

I'd be happy to do that (although I wouldn't hold up v0.6.1 for it).... @MiroDudik this was your original pattern. Are you happy with me changing things around so that we pass the estimator, rather than a callable object? I'm not sure why it was done that way originally.

JorickvdHoeven:
Hi Everyone, cloudpickle is causing some frankly weird errors when we're migrating models built in OSX to Linux. It looks like this pull request will help mitigate that issue. What still needs to be done to merge it?

MiroDudik (approved):
This looks fine to me--just one suggestion re. changing the name of the helper class to make it less confusing.

MiroDudik:
@riedgar-ms are you unblocked on this?

MiroDudik:
This is now completed by #1167. Closing.

Issue #4, opened 2019-03-13
Potential restriction of fairlearn?


guanqun-yang:
I am wondering if there are any underlying requisites to apply **any** sklearn model, when I applied `sklearn.neural_network.MLPClassifier()` it says
![fairlearn](https://user-images.githubusercontent.com/36497361/54322048-dca6f280-45b0-11e9-940e-c400c4017215.png)
But when I applied `sklearn.svm.SVC()`, this issues **doe not** arise. 
It seems to me that the input object should have `class_weight` attribute, which `MLPClassifier()` does not have. So is this true?

MiroDudik:
That is indeed correct. See the documentation of expgrad:

    Required input arguments:
      ...
      learner -- a learner implementing methods fit(X,Y,W) and predict(X),
                 where X is the DataFrame of covariates, and Y and W
                 are the Series containing the labels and weights,
                 respectively; labels Y and predictions returned by
                 predict(X) are in {0,1}

But, it's a good idea to provide a more informative error message. Also,
maybe at some point in future, it might be nice to have wrapper that would
convert a standard (non-weighted) learner into a weighted learner.

romanlutz:
I marked it as a feature-request. I cannot provide an estimate at this point, but will update as soon as this gets scheduled.

romanlutz:
@Pooja1301 is looking into this now for `ExponentiatedGradient`, and then same resampling should be applicable to `GridSearch` subsequently. Stay tuned :-)

Pooja1301:
I have started looking into it.

romanlutz:
@violazhong @riedgar-ms and I discussed this earlier today. Capturing some of our notes.

The most important aspect of this work is passing in the resampling method. @riedgar-ms pointed out that this should probably be in the constructor or the `fit` method. Since we're passing the `estimator` in the constructor that seems to be the most suitable place IMO. Once we have that on the object we'd have to replace this line https://github.com/fairlearn/fairlearn/blob/d1581b4576fa5c4dbb0fb99bf5548c74ba05b0ac/fairlearn/reductions/_exponentiated_gradient/_lagrangian.py#L178
with something like this (pardon the pseudocode at times):
```
if estimator has fit function and fit function has sample_weight arg:
     estimator.fit(self.X, redY, **{self.sample_weight_name: redW})
elif user has provided resampling_method:
     resampled_X, resampled_Y = resampling_method(X, redY, redW)
[...]

PR #663, opened 2020-12-24
Adding resampling process in case some classifier fit function can't take weights 

violazhong:
Hello team, 

I created a new branch which is a quick implementation regarding this issue: https://github.com/fairlearn/fairlearn/issues/4.

**What I did here:** 
1) modify the estimator.fit(), so now it could handle those classifiers that don't take sample weights. 
2) add 'resampling.py' which now only has one resampling method: deterministic and keeping the sample size.

**What will be done in the future:**
1) we'll let the user pass their own resampling methods.
2) we'll find more resampling methods and add them to 'resampling.py'.
3) I'll work on a better description/documentation for the changes that I made. 

As usual, please feel free to let out any ideas/suggestions/concerns. We'll do a few unit tests shortly and will keep everyone updated. Yay:-)

### DIFF
--- a/fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
+++ b/fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
@@ -2,12 +2,15 @@
 # Licensed under the MIT License.
 
 import logging
+import math
 import numpy as np
 import pandas as pd
 import scipy.optimize as opt
 from sklearn import clone
 from sklearn.dummy import DummyClassifier
 from time import time
+import inspect
+import resampling
 
 from ._constants import _PRECISION, _INDENTATION, _LINE
 
@@ -62,7 +65,6 @@ class _Lagrangian:
         self.n_oracle_calls_dummy_returned = 0
         self.last_linprog_n_hs = 0
         self.last_linprog_result = None
-        self.sample_weight_name = sample_weight_name
 
     def _eval(self, Q, lambda_vec):
         """Return the value of the Lagrangian.
@@ -172,7 +174,31 @@ class _Lagrangian:
             estimator = clone(estimator=self.estimator, safe=False)
 
         oracle_call_start_time = time()
-        estimator.fit(self.constraints.X, redY, **{self.sample_weight_name: redW})
+
+        # The line that is being replaced
+        # estimator.fit(self.constraints.X, redY, **{self.sample_weight_name: redW})
+
+        # ------- change made starting from here -------- #
+
+        # historical version: try and except, downside: too expensive
+        # try:
+        #     estimator.fit(self.constraints.X, redY, **{self.sample_weight_name: redW})
+        # except TypeError:
+        #     resampled_X, resampled_Y = self.resampling_method(self.constraints.X, redY, redW)
+        #     estimator.fit(resampled_X, resampled_Y)
+
+        # proposing version: if check args
+        if estimator.fit() and 'sample_weight' in inspect.getfullargspec(estimator.fit).args:
+            estimator.fit(self.constraints.X, redY, **{self.sample_weight_name: redW})
+        # To do 1: add other resampling methods in fairlearn.reductions.resampling
+        # To do 2: let users pass their own resampling methods, and we orchestrate here
+        # now we only consider this resampling method：fixed_size_deterministic
+        else:
+            resampled_X, resampled_y = resampling.resample_fixed_size_deterministic(self.constraints.X, redY, redW)
+            estimator.fit(resampled_X, resampled_y)
+
+        # ------- change made ending from here -------- #
+
         self.oracle_execution_times.append(time() - oracle_call_start_time)
         self.n_oracle_calls += 1
 
--- a/fairlearn/reductions/resampling.py
+++ b/fairlearn/reductions/resampling.py
@@ -1,0 +1,55 @@
+# Copyright (c) Microsoft Corporation and Fairlearn contributors.
+# Licensed under the MIT License.
+
+import logging
+import math
+# import numpy as np
+# import pandas as pd
+# import scipy.optimize as opt
+# from sklearn import clone
+# from sklearn.dummy import DummyClassifier
+# from time import time
+# import inspect
+# import resampling
+
+
+logger = logging.getLogger(__name__)
+
+
+def resample_fixed_size_deterministic(X, y, weights):
+    """ Resampling methods: Deterministic and keep the sample size.
+
+        Parameters
+        ----------
+        X : {numpy.ndarray, pandas.DataFrame}
+            the training features
+        sensitive_features : {numpy.ndarray, pandas.Series, pandas.DataFrame, list}
+            the sensitive features to use for constraints
+        y : {numpy.ndarray, pandas.Series, pandas.DataFrame, list}
+            the training labels
+        weight : {numpy.ndarray, pandas.Series, pandas.DataFrame, list}
+            the resampling weights
+        """
+
+    factor = len(X) / sum(weights)
+    resampled_X, resampled_y = [], []
+
+    scaled_weight = [weight * factor for weight in weights]
+    for i in range(len(X)):
+        if scaled_weight[i] >= 1 and len(resampled_X) < len(X):
+            add_num = math.floor(scaled_weight[i])
+        resampled_X += [X[i]] * add_num
+        resampled_y += [y[i]] * add_num
+    scaled_weight[i] -= add_num
+
+    remain_slots = len(X) - len(resampled_X)
+
+    if remain_slots > 0:
+        scale_tuple = list(enumerate(scaled_weight))
+        scale_tuple.sort(key=lambda x: x[1], reverse=True)
+        scale_tuple = scale_tuple[:remain_slots]
+        for index, scale in scale_tuple:
+            resampled_X += [X[index]]
+            resampled_y += [y[index]]
+
+        return resampled_X, resampled_y


fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -172,7 +174,31 @@ def _call_oracle(self, lambda_vec):
             estimator = clone(estimator=self.estimator, safe=False)
 
         oracle_call_start_time = time()
-        estimator.fit(self.constraints.X, redY, **{self.sample_weight_name: redW})
+
+        # The line that is being replaced
```
romanlutz, line 178: Thanks for providing the context! Instead of putting it in the code, do you mind pasting it in the PR description with triple backticks/ ` ? Then it render like this:
```
this is code
```

fairlearn/reductions/resampling.py
```diff
@@ -0,0 +1,55 @@
...
+# import resampling
+
+
+logger = logging.getLogger(__name__)
+
+
+def resample_fixed_size_deterministic(X, y, weights):
+    """ Resampling methods: Deterministic and keep the sample size.
+
+        Parameters
```
romanlutz, line 22: nit: I think we can unindent the docstring starting with this line

fairlearn/reductions/resampling.py
```diff
@@ -0,0 +1,55 @@
...
+        ----------
+        X : {numpy.ndarray, pandas.DataFrame}
+            the training features
+        sensitive_features : {numpy.ndarray, pandas.Series, pandas.DataFrame, list}
+            the sensitive features to use for constraints
+        y : {numpy.ndarray, pandas.Series, pandas.DataFrame, list}
+            the training labels
+        weight : {numpy.ndarray, pandas.Series, pandas.DataFrame, list}
+            the resampling weights
```
romanlutz, line 31: please also add `Returns`

fairlearn/reductions/resampling.py
```diff
@@ -0,0 +1,55 @@
+# Copyright (c) Microsoft Corporation and Fairlearn contributors.
+# Licensed under the MIT License.
+
+import logging
+import math
+# import numpy as np
```
romanlutz, line 6: nit: please remove unused imports

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -62,7 +65,6 @@ def __init__(self, X, y, estimator, constraints, B, opt_lambda=True,
         self.n_oracle_calls_dummy_returned = 0
         self.last_linprog_n_hs = 0
         self.last_linprog_result = None
-        self.sample_weight_name = sample_weight_name
```
romanlutz, line 65: we shouldn't delete this since it's actually used in the fit call

romanlutz (commented):
Thanks for submitting this @violazhong ! I left a few comments, but the overall approach looks good to me. 
@MiroDudik and @riedgar-ms may want to take a look, too, since we've talked about this a long time ago.

I'll take another look once your "To do 2: let users pass their own resampling methods" is implemented. Right now the user wouldn't realize that resampling is happening.

Additionally, for this PR we'll need tests (as discussed) and perhaps a small example for the webpage. Let me know if you need any help with either of them :-)

Finally, make sure to add a line to CHANGES.md to point out that this new functionality is added.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -172,7 +174,31 @@ def _call_oracle(self, lambda_vec):
...
 
         oracle_call_start_time = time()
-        estimator.fit(self.constraints.X, redY, **{self.sample_weight_name: redW})
+
+        # The line that is being replaced
+        # estimator.fit(self.constraints.X, redY, **{self.sample_weight_name: redW})
+
+        # ------- change made starting from here -------- #
+
+        # historical version: try and except, downside: too expensive
```
riedgar-ms, line 183: In general, you don't need to put things like this into a comment.

For this specific case, I would ask for a quantification of 'too expensive' but... a general rule to follow is that exceptions should be truly exceptional anyway :-)

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -2,12 +2,15 @@
...
 import logging
+import math
 import numpy as np
 import pandas as pd
 import scipy.optimize as opt
 from sklearn import clone
 from sklearn.dummy import DummyClassifier
 from time import time
+import inspect
+import resampling
```
riedgar-ms, line 13: Since `resampling` is part of Fairlearn..... we usually do those imports with a prefixed `.` to emphasise this (like in the `._constants` below). That bit of coding style could be up for debate, though.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -172,7 +174,31 @@ def _call_oracle(self, lambda_vec):
...
+        #     estimator.fit(resampled_X, resampled_Y)
+
+        # proposing version: if check args
+        if estimator.fit() and 'sample_weight' in inspect.getfullargspec(estimator.fit).args:
+            estimator.fit(self.constraints.X, redY, **{self.sample_weight_name: redW})
+        # To do 1: add other resampling methods in fairlearn.reductions.resampling
+        # To do 2: let users pass their own resampling methods, and we orchestrate here
+        # now we only consider this resampling method：fixed_size_deterministic
+        else:
+            resampled_X, resampled_y = resampling.resample_fixed_size_deterministic(self.constraints.X, redY, redW)
```
riedgar-ms, line 197: I think that this shouldn't happen by default - the resampling method should be an argument to the constructor, and we should throw an exception if there's no `sample_weight` argument and no resampler (this is sort of implicit in your To Dos, but even if there's only one way available, I don't think we should do it "behind the user's back").

Of course... what to do if there is a resampler, but sample weights are available in the estimator.... my preference would be to throw an exception there (since the user may be confused), but opinions may well differ on that.

fairlearn/reductions/resampling.py
```diff
@@ -0,0 +1,55 @@
...
+
+    if remain_slots > 0:
+        scale_tuple = list(enumerate(scaled_weight))
+        scale_tuple.sort(key=lambda x: x[1], reverse=True)
+        scale_tuple = scale_tuple[:remain_slots]
+        for index, scale in scale_tuple:
+            resampled_X += [X[index]]
+            resampled_y += [y[index]]
+
+        return resampled_X, resampled_y
```
riedgar-ms, line 55: Does this have the right indentation? What is returned if `remain_slots <= 0` ?

fairlearn/reductions/resampling.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def resample_fixed_size_deterministic(X, y, weights):
+    """ Resampling methods: Deterministic and keep the sample size.
+
+        Parameters
+        ----------
+        X : {numpy.ndarray, pandas.DataFrame}
+            the training features
+        sensitive_features : {numpy.ndarray, pandas.Series, pandas.DataFrame, list}
```
riedgar-ms, line 26: I don't see sensitive_features in the argument list?

fairlearn/reductions/resampling.py
```diff
@@ -0,0 +1,55 @@
...
+# from sklearn.dummy import DummyClassifier
+# from time import time
+# import inspect
+# import resampling
+
+
+logger = logging.getLogger(__name__)
+
+
+def resample_fixed_size_deterministic(X, y, weights):
```
riedgar-ms, line 19: Can you add some tests of this function, to make it easier to see what's going on, please?

fairlearn/reductions/resampling.py
```diff
@@ -0,0 +1,55 @@
...
+        weight : {numpy.ndarray, pandas.Series, pandas.DataFrame, list}
+            the resampling weights
+        """
+
+    factor = len(X) / sum(weights)
+    resampled_X, resampled_y = [], []
+
+    scaled_weight = [weight * factor for weight in weights]
+    for i in range(len(X)):
+        if scaled_weight[i] >= 1 and len(resampled_X) < len(X):
```
riedgar-ms, line 39: Doesn't this `if` statement need an `else` ? I'm trying to figure out what's going to happen if there's an extremely 'heavy' element (or two) on the end of the array. Can `add_num` end up causing the size of `resampled_X` to exceed `X` ?

fairlearn/reductions/resampling.py
riedgar-ms: Yes.... I'm also not sure what this is going to return. It looks like it might not preserve the input types?

fairlearn/reductions/resampling.py
```diff
@@ -0,0 +1,55 @@
...
+            add_num = math.floor(scaled_weight[i])
+        resampled_X += [X[i]] * add_num
+        resampled_y += [y[i]] * add_num
+    scaled_weight[i] -= add_num
```
riedgar-ms, line 43: Is this meant to be part of the `for` loop? The indentation makes it look like it's outside (and I'm not sure offhand what Python leaves the loop variable equal to on exit).

riedgar-ms (commented):
Thanks for getting started on this!

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -172,7 +174,31 @@ def _call_oracle(self, lambda_vec):
...
+        # historical version: try and except, downside: too expensive
+        # try:
+        #     estimator.fit(self.constraints.X, redY, **{self.sample_weight_name: redW})
+        # except TypeError:
+        #     resampled_X, resampled_Y = self.resampling_method(self.constraints.X, redY, redW)
+        #     estimator.fit(resampled_X, resampled_Y)
+
+        # proposing version: if check args
+        if estimator.fit() and 'sample_weight' in inspect.getfullargspec(estimator.fit).args:
+            estimator.fit(self.constraints.X, redY, **{self.sample_weight_name: redW})
```
adrinjalali, line 192: it's more pythonic to put this in a try/except block instead of checking the signature of a function. A method can also accept `sample_weight` through `**kwargs` instead of `sample_weight.

adrinjalali (commented):
I don't think resampling is in the scope of an estimator. We should raise a meaningful error when the underlying estimator doesn't support `sample_weight` and then have a meta-estimator which does the resampling before calling the estimator, and point users to use that meta-estimator instead. This would also avoid the need to the duplication of this code when the other estimators in `fairlearn` encounter an underlying estimator which doesn't support `sample_weight`.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
riedgar-ms: Good point. I forgot about `**kwargs` :-/

romanlutz: I actually recommended not using try/except because it's more expensive when you hit the exception, so this one's on me. I suppose we don't expect this to happen terribly often anyway, and when it happens it happens once rather than thousands of times (at which point the "more expensive" argument would count more).

romanlutz:
> I don't think resampling is in the scope of an estimator. We should raise a meaningful error when the underlying estimator doesn't support `sample_weight` and then have a meta-estimator which does the resampling before calling the estimator, and point users to use that meta-estimator instead. This would also avoid the need to the duplication of this code when the other estimators in `fairlearn` encounter an underlying estimator which doesn't support `sample_weight`.

Good point! I hadn't actually thought much beyond what @MiroDudik outlined in the issue. Perhaps even there it wasn't meant to be this specific to `ExponentiatedGradient`. The way you outline it it should probably live in a separate module, right? Maybe `fairlearn.resampling`? Or just a place for utilities, `fairlearn.utils`?
@violazhong I'm afraid I didn't quite think "big enough" about this. What do you think about @adrinjalali 's suggestion? That would make this independent from the `ExponentiatedGradient` code. Instead, we'd put this logic into a wrapper class to which one can pass an estimator, and the wrapper has the suitable API `fit(X, y, sample_weight=...)`.

If I think a little bit bigger: would such resampling methods be better suited in a more general package like `scikit-learn`? Not that I want to abandon this PR, but I'm wondering whether this belongs here long-term.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
riedgar-ms: It's going to happen each time through the loop (for non-sample_weight estimators).... however, it should be a fair bet that `fit()` is vastly more expensive than catching an exception.

I too was of the school that exceptions should be truly exceptional, though.

riedgar-ms:
Expanding the scope to being a more general approach is certainly a good idea... and something other packages might want to pick up on. Until they did, I agree that we'd need to keep our own version.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
adrinjalali: In python we use exception handling as a way to flow-control, which is not necessarily the case in many other programming languages, here you have a nice summary of it: https://stackoverflow.com/questions/16138232/is-it-a-good-practice-to-use-try-except-else-in-python

adrinjalali:
Resamplers have been a tricky topic in the context of scikit-learn's API (ref: https://github.com/scikit-learn/enhancement_proposals/pull/15), but we're leaning towards having meta-estimators handling it. I wouldn't say this is a high priority here, we can simply just support estimators which support sample_weight for now. But if there's will and people are happy to tackle the issue, I think it should go into a `fairlearn.resampler` module.

MiroDudik:
I agree that a resampling wrapper belongs into a module that is separate from what we're currently housing in `fairlearn.reductions`. While resampling is technically a reduction it is a different kind of reduction than we have in `fairlearn.reductions`. What we currently have in `fairlearn.reductions` could be more precisely called a Lagrangian reduction.

I guess my only question is whether we want to have a highly specific module `fairlearn.resampler` or slightly more generic module `fairlearn.wrappers` (to include some additional simple reductions that don't do any "training" and just transform one kind of an input into a different kind of input), or like @romanlutz says `fairlearn.utils` or some similar catch-all module (I'm not too excited about this idea, but open to it).

If we decide to have `fairlearn.wrappers`, do we need to rename `fairlearn.reductions` to something like `fairlearn.moment_reductions` or `fairlearn.lagrangian_reductions` (I'd love to avoid renaming!).

adrinjalali (changes requested):
I think the best way to move forward here would be to have a `MetaEstimator` which would accept a pipeline or another estimator and do the resampling for that estimator.

You can check [here](https://github.com/scikit-learn/scikit-learn/blob/8ea176ae0ca535cdbfad7413322bbc3e54979e4d/sklearn/calibration.py#L43) for an example of a meta-estimator.

romanlutz:
> I think the best way to move forward here would be to have a `MetaEstimator` which would accept a pipeline or another estimator and do the resampling for that estimator.
> 
> You can check [here](https://github.com/scikit-learn/scikit-learn/blob/8ea176ae0ca535cdbfad7413322bbc3e54979e4d/sklearn/calibration.py#L43) for an example of a meta-estimator.

Agreed. @violazhong is actively looking into this and adjusting according to the comments right now (based on Gitter chat conversations with me). Thanks again @violazhong for leading this effort :-) Please do ping us when you want further feedback after your next set of updates.

violazhong:
@adrinjalali thank you for the reference. and yes, we're writing a new class for this ```MetaEstimator```. 

Here I have a question about resampling strategies. @MiroDudik had a good explanation about a randomized strategy in https://github.com/fairlearn/fairlearn/issues/4, and we'll use that as our default strategy, while also allow the user to pass the resampling strategy at their choices. So we'll add more resampling strategies in the future, for this, my current thought is to write a new class for all the resampling strategies. Would you think this is a good approach? In terms of resampling strategies, could you point to some references, so I could study and implement them as well? 

@romanlutz absolutely happy to collaborate! And yeah, I should post more updates here on Github.

adrinjalali:
I think `impalanced-learn` is a good place for some resampling strategies:

https://imbalanced-learn.org/stable/api.html

MiroDudik:
My suggestion is to start with a small and targeted implementation. The goal of the strategy that I'm proposing in #4 is very specific: convert estimators that operate by minimizing an empirical loss function (but don't allow sample weights) into estimators that allow sample weights. With that goal in mind, we might not need that many alternative sampling strategies.

I think that this is distinct from the goal of dealing with imbalanced data sets.

PR #669, opened 2021-01-04
Restore Green Builds

riedgar-ms:
Address build breaking issues:
- The latest version of `scikit-learn` was causing trouble with the tests for `CorrelationRemover` due to changes in the behaviour of the `check_sample_weights_invariance()` test. Since `CorrelationRemover` doesn't actually permit sample weights, remove this check from the list of checks run
- Something had changed how Pandas was printing out `DataFrame` objects, causing the expected output not to match in the docs. Force a precision to address this

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -76,7 +76,6 @@ than $50,000 a year.
 
     >>> import numpy as np 
     >>> import pandas as pd
-    >>> pd.set_option('display.precision', 4)
     >>> import matplotlib.pyplot as plt 
     >>> from sklearn.datasets import fetch_openml
     >>> data = fetch_openml(data_id=1590, as_frame=True)
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -112,11 +112,11 @@ we can evaluate metrics for subgroups within the data as below:
     >>> y_pred = classifier.predict(X)
     >>> gm = MetricFrame(accuracy_score, y_true, y_pred, sensitive_features=sex)
     >>> print(gm.overall)
-    0.8443552680070431
+    0.8443...
     >>> print(gm.by_group)
     sex
-    Female    0.9251
-    Male      0.8043
+    Female    0.9251...
+    Male      0.8042...
     Name: accuracy_score, dtype: object
 
 Additionally, Fairlearn has lots of other standard metrics built-in, such as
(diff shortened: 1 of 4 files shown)


docs/quickstart.rst
```diff
@@ -76,6 +76,7 @@ than $50,000 a year.
 
     >>> import numpy as np 
     >>> import pandas as pd
+    >>> pd.set_option('display.precision', 4)
```
romanlutz, line 79: Why don't we keep it as is? Ideally such code shouldn't be in there as it's unrelated to the example.

riedgar-ms: From the PR description 'Something had changed about how Pandas was printing out DataFrame objects'

I'm not sure what had changed, although the recent release of pandas 1.2.0 is a strong suspect. The point is `doctest` does string matching, so it's either do this, or scatter ellipses throughout the results (which then act as wildcards). I believe that we could move directives like this into a hidden code block, but I think I'd prefer to leave the change out in the open.

adrinjalali: I've mentioned before that we should keep the precision to 3 or 4 and then `...` in the doctests, so I'd say we should go with that and not include this display option in the example.

riedgar-ms: I'm not quite sure what you mean. It appears that pandas has changed its default display for DataFrames. At least some of the updates I had to make involved changing a `0` or `1` to `0.0` or `1.0`. While I suppose those could be handled by the `...` wildcard, I don't think it would look good in the docs.

riedgar-ms: I suppose that, since we don't build the docs with old Pandas versions, I could just update to the new formatting (and truncate decimals at 3 or 4 places).... is that what you meant @adrinjalali ?

adrinjalali: That's definitely a solution. I'm a bit surprised about 1.0 vs 1 though. That means there's a dtype change between versions? If that's the case, we can force the dtype and make sure it's always int or float, whichever it needs to be.

riedgar-ms: It might also be related to our dtype being 'object' a lot of the time. Anyway, I've just pushed the update.

romanlutz: My main point was that I don't want this particular line in all examples, and that's addressed, so I'll sign off happily :-) Thanks!

romanlutz (approved):

docs/quickstart.rst
riedgar-ms: @adrinjalali does this look better to you?

MiroDudik (approved):

adrinjalali (approved):
Thanks!

Issue #655, opened 2020-12-11
Support fort streaming metrics


Dref360:
#### Is your feature request related to a problem? Please describe.

I am using fairlearn to evaluate deep learning models. In this context, we often compute metrics per batch due to the large size of the dataset. In addition, some tasks such as semantic segmentation are so expensive that we can't hold all the predictions in memory.

#### Describe the solution you'd like
I would like if `MetricFrame` would accept an `update` method where we can add more data. For the first version, we can simply add more data to the internal representation of `ytrue, ypred, sensitive_attribute`. But then I would like it if we could keep a confusion matrix and the `update` method would update this confusion matrix.

#### Describe alternatives you've considered, if relevant

I made a wrapper around Fairlearn metrics that keep tracks of `ytrue, ypred, sensitive_attribute` and then call the underlying metric when needed.

#### Additional context

Example of a metric in Pytorch Lightning, a fairly popular DL framework.
https://pytorch-lightning.readthedocs.io/en/stable/metrics.html#implementing-a-metric

Keras has a similar API

https://keras.io/api/metrics/#as-subclasses-of-metric-stateful

riedgar-ms:
As discussed in our call, this certainly sounds like something Fairlearn should support. However, `MetricFrame` doesn't keep track of anything - in fact, it doesn't really care what the contents of `y_true` and `y_pred` really are, and certainly doesn't require that the incoming data can be formed into a confusion matrix. That would be something for the individual metrics themselves to support.

The way I see this working is that each metric would be responsible for its own accumulations, and we'd make a copy of the relevant object for each detected sensitive feature combination. The `update(...)` method on the `MetricFrame` would in turn call the update method on each of those underlying metric objects.

MiroDudik:
**(1) Re. "wrapper" implementation** where all the incoming data is kept around:
* This should not be default, but only enabled by setting a Boolean flag in the constructor.

**(2) To incorporate some initial efficient implementations** (say for the metrics based on keeping track of the contingency table), we could allow strings as metrics, similarly to [pd.SeriesGroupBy.aggregate](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.core.groupby.SeriesGroupBy.aggregate.html), e.g.,
```python
mf = flm.MetricFrame('accuracy_score', y_true, y_pred, sensitive_features=sf)
mf.add_data(y_true_new, y_pred_new, sensitive_features=sf_new)

[...]

PR #670, opened 2021-01-07
Initial implementation of streaming metric

Dref360:
Summary:

In #655, we propose streaming metrics so that we can use Fairlearn metrics with batches as it is common in Deep learning. The naive implementation is to simply accumulate the values before computing the metric when asked.

**NOTE** This is WIP, open to suggestions and I wanted some guidance to be sure I didn't miss something obvious. I will add unit tests before making it ready for review.

Related issues
* Related to #655 (does not fully fix it).

Quick test:

```python
import numpy as np
import sklearn.metrics as skm
classes = [0,1]
sensitive = ['a', 'b']
y_true = np.random.choice(classes, [100], replace=True)
y_pred = np.random.choice(classes, [100], replace=True)
sensitive_attrs = np.random.choice(sensitive, [100], replace=True)
met = MetricFrame(skm.recall_score, y_true=y_true, y_pred=y_pred, sensitive_features=sensitive_attrs)
met.overall


# Batched, we simulate 2 batches.
yt1,yt2 = y_true[:40], y_true[40:]
yp1,yp2 = y_pred[:40], y_pred[40:]
sa1,sa2 = sensitive_attrs[:40], sensitive_attrs[40:]


met_strt = MetricFrame(skm.recall_score, y_true=[], y_pred=[], sensitive_features=[], streaming=True)
met_strt.add_data(yt1, yp1, sensitive_features=sa1)
met_strt.add_data(yt2, yp2, sensitive_features=sa2)

# Check that it's equal.
assert met_strt.overall == met.overall
```

Signed-off-by: Dref360 <frederic.branchaud-charron@usherbrooke.ca>


fairlearn/metrics/_metric_frame.py
```diff
@@ -117,24 +126,127 @@ def __init__(self,
...
+            self._y_p = []
+            self._overall = None
+            self._by_group = None
+
+    def add_data(self,
+                 y_true,
+                 y_pred, *,
+                 sensitive_features,
+                 control_features: Optional = None,
+                 ):
```
MiroDudik, line 166: Do we want to add `sample_params` so as to pretty much follow the same format as constructor minus `metric`?

fairlearn/metrics/_metric_frame.py
```diff
@@ -226,6 +338,9 @@ def overall(self) -> Union[Any, pd.Series, pd.DataFrame]:
             interface when calling programatically, while also reducing
             typing for those using Fairlearn interactively.
         """
+        if self._overall is None:
```
riedgar-ms, line 341: Does this mean that once a user has invoked one of these summary functions, they can no longer keep streaming data into the object?

riedgar-ms (commented):
Could you start adding some tests, please? The sample code in the PR description is helpful, but it's much better to have something which is running in the automation.

fairlearn/metrics/_metric_frame.py
Dref360: We can still keep it, but we will wipe `_overall` and `by_group` so the next call will recompute it.

fairlearn/metrics/_metric_frame.py
```diff
@@ -117,24 +126,127 @@ def __init__(self,
...
         check_consistent_length(y_true, y_pred)
+        check_consistent_length(y_true, sensitive_features)
+
         y_t = _convert_to_ndarray_and_squeeze(y_true)
         y_p = _convert_to_ndarray_and_squeeze(y_pred)
 
-        func_dict = self._process_functions(metric, sample_params)
+        self._y_t.append(y_t)
+        self._y_p.append(y_p)
+        self._s_f.append(sensitive_features)
```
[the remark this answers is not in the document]
Dref360: I will add support for more formats, but these accumulators accumulate batches and we concatenate them when needed.

Dref360:
I will add tests later this week.

fairlearn/metrics/_metric_frame.py
```diff
@@ -117,24 +126,127 @@ def __init__(self,
                  y_pred, *,
                  sensitive_features,
                  control_features: Optional = None,
-                 sample_params: Optional[Union[Dict[str, Any], Dict[str, Dict[str, Any]]]] = None):
+                 sample_params: Optional[Union[Dict[str, Any], Dict[str, Dict[str, Any]]]] = None,
+                 streaming=False):
         """Read a placeholder comment."""
+        if len(y_true) == 0 and not streaming:
+            raise ValueError(_BAD_INIT_ERROR)
+
```
MiroDudik, line 134: Folks, I'd like to hit a pause and open a quick discussion about what this should look like internally. My suggestion is that we should create an intermediate `DataFrame` representation. Specifically:

**1. We create an internal function of the following form:**

```python
def _pandify(y_true, y_pred, sensitive_features, control_features, sample_params) -> pd.DataFrame:
    
    # My sample code below assumes that y_pred and y_true are 1D array-like,
    # sensitive_features and control_features are dicts of 1D array-like,
    # sample_params are dicts of dicts of 1D array-like.
    # 
    # The actual code would first need to do validation and conversion
    # including the generation of names of control features and sensitive features.
    
    shared_inputs_df = pd.concat(
        {'dummy': pd.DataFrame({'y_true': y_true, 'y_pred': y_pred})}, axis=1)
    sensitive_features_df = pd.concat(
        {'dummy': pd.DataFrame(sensitive_features)}, axis=1)     
    control_features_df = pd.concat(
        {'dummy': pd.DataFrame(control_features)}, axis=1)           
    sample_params_df = pd.concat(
        {key: pd.DataFrame(sample_params[key]) for key in sample_params.keys()}, axis=1)

    combined_df = pd.concat({'shared_inputs': shared_inputs_df,
        'sensitive_features': sensitive_features_df,
        'control_features': control_features_df, 
        'sample_params': sample_params_df}, axis=1)
    return combined_df
```

**2. For non-streaming implementation:**
We call `_pandify` in `__init__` and then calculate `self._overall` and `self._by_group`.

**3. For streaming implementation:**
We call `_pandify` in `add_data`. If this is not the first call, we check that the columns of the new data frame agree with the columns of the old data frame and if they do, we call `pd.concat` on them.

What do you think? (The reason for my micromanaging this is that there are a few extensions of `MetricFrame` in the pipeline and I think they'll be much smoother if we have the intermediate `DataFrame` representation.)

MiroDudik: There would be some updating in other internal processing functions--we would no longer be passing multiple arguments, e.g., we would have
```python
def _compute_overall(self, func_dict, combined_df):
```

Dref360: That is a great idea! It would simplify a lot of things :) I'll work on it this week.

fairlearn/metrics/_metric_frame.py
```diff
@@ -117,24 +126,127 @@ def __init__(self,
...
+            However, their default names (if none can be identified in the
+            input values) are of the format ``control_feature_[n]``.
+
+            **Note** the types returned by members of the class vary based on whether
+            control features are present.
+        """
+        if not self._streaming:
+            raise Exception("This MetricFrame does not support adding data.")
         check_consistent_length(y_true, y_pred)
+        check_consistent_length(y_true, sensitive_features)
```
adrinjalali, line 202: I _think_ you can pass multiple arrays to the same function call here

adrinjalali (commented):
Here's a few notes :) Thanks @Dref360

fairlearn/metrics/_metric_frame.py
Dref360: I think this would lead to errors.
Example:
```
met = MetricFrame(skm.recall_score, y_true=[], y_pred=[], sensitive_features=[], streaming=True, sampling_params={'average':'micro'})

met_strt.add_data(..., sample_params={'average':'macro'})
....
met_strt.add_data(..., sample_params={'average':'micro'})
```

Then we compute the value differently for different batches.

Dref360:
I added some tests and implemented most of the comments. Thank you everyone!

As for _pandify, @MiroDudik I'm not sure how to handle "metric-wide" sample_params like `average` in `skm.recall_score` using this method. I would need some guidance.

fairlearn/metrics/_metric_frame.py
```diff
@@ -147,8 +288,9 @@ def __init__(self,
                 raise ValueError(_DUPLICATE_FEATURE_NAME.format(name))
             nameset.add(name)
 
-        self._overall = self._compute_overall(func_dict, y_t, y_p, cf_list)
-        self._by_group = self._compute_by_group(func_dict, y_t, y_p, sf_list, cf_list)
+        self.func_dict = self._process_functions(self.metric, s_p)
```
riedgar-ms, line 291: Why has the processing of the functions been moved out of the constructor?

test/unit/metrics/test_metricframe_streaming.py
```diff
@@ -0,0 +1,131 @@
...
+@pytest.mark.parametrize("transform_y_p", conversions_for_1d)
+@pytest.mark.parametrize("transform_y_t", conversions_for_1d)
+@pytest.mark.parametrize("num_batches", [1, 3, 5])
+def test_sample_params(transform_y_t, transform_y_p, num_batches):
+    sample_weights = np.random.rand(len(y_t))
+    batch_generator = batchify(num_batches, y_t, y_p, g_4, sample_weights)
+    target = metrics.MetricFrame(skm.recall_score, [], [], sensitive_features=[],
+                                 sample_params=None, streaming=True)
+
+    with pytest.raises(ValueError) as excinfo:
```
riedgar-ms, line 24: I think this should be a separate test.

test/unit/metrics/test_metricframe_streaming.py
```diff
@@ -0,0 +1,131 @@
...
+                    sensitive_features=g_f,
+                    sample_params={'sample_weight': sw_batches})
+    assert target._overall is None
+    assert target._by_group is None
+
+
+@pytest.mark.parametrize("transform_y_p", conversions_for_1d)
+@pytest.mark.parametrize("transform_y_t", conversions_for_1d)
+@pytest.mark.parametrize("num_batches", [1, 3, 5])
+def test_basic_streaming(transform_y_t, transform_y_p, num_batches):
```
riedgar-ms, line 79: Thanks for adding these. Could you also have a test where the batches change type each time? e.g. the first `add_data()` call uses a list, the next a Series, the next an ndarray etc.?

test/unit/metrics/test_metricframe_streaming.py
```diff
@@ -0,0 +1,131 @@
...
+    assert target.control_levels is None
+    assert isinstance(target.sensitive_levels, list)
+    assert (target.sensitive_levels == ['My feature'])
+
+    # Check we have correct return types
+    assert isinstance(target.overall, float)
+    assert isinstance(target.by_group, pd.Series)
+
+    # Check we have expected number of elements
+    assert len(target.by_group) == 2
```
riedgar-ms, line 45: Given that the non-streaming tests already do this sort of checking, would it be easier to just put the data into a non-streaming `MetricFrame` and then check for equality?

test/unit/metrics/test_metricframe_streaming.py
```diff
@@ -0,0 +1,131 @@
...
+    assert target._overall is None
+    assert target._by_group is None
+
+
+@pytest.mark.parametrize("transform_y_p", conversions_for_1d)
+@pytest.mark.parametrize("transform_y_t", conversions_for_1d)
+@pytest.mark.parametrize("num_batches", [1, 3, 5])
+def test_basic_streaming(transform_y_t, transform_y_p, num_batches):
+    batch_generator = batchify(num_batches, y_t, y_p, g_4)
+    target = metrics.MetricFrame(skm.recall_score, [], [], sensitive_features=[],
```
riedgar-ms, line 81: Could we also have a case where some subgroups of the sensitive feature(s) are not present in the initial batch?

riedgar-ms: And another case where there are intermixed calls to `add_data()` and `.overall` and `.by_group` etc.?

test/unit/metrics/test_metricframe_streaming.py
riedgar-ms: And on a related point, what happens if you call `.overall` before calling `add_data()` when streaming?

fairlearn/metrics/_metric_frame.py
MiroDudik: I can see that the internal representation becomes more obscure, but I don't understand the column consistency issue. If we have the current data in `df` and new data in `df_new` we can just check consistency by calling `df.columns.symmetric_difference(df_new.columns).empty` and we can just update with `df = df.append(df_new)`.

adrinjalali: These operations you're mentioning are quite expensive. And in terms of API design, it's much easier to understand what the input to a method is, if it gets things it needs in separate arguments compared to accepting a single dataframe which the user of the method would need to know how to prepare.

fairlearn/metrics/_metric_frame.py
Dref360: Because the process_functions is dependent on the sample_params.

test/unit/metrics/test_metricframe_streaming.py
Dref360: It will raise `ValueError('No data to process.')` Exception message to be reworded.

test/unit/metrics/test_metricframe_streaming.py
Dref360: Hmm my current implementation does not support it. I'll think on how to support it.

test/unit/metrics/test_metricframe_streaming.py
Dref360: Right!

Dref360:
Hello!
Sorry for the delay.
I think I answered most questions and fixed most issues. So I would welcome a new round of review.

Of course, if we want to change the design I am open to suggestions!

fairlearn/metrics/_metric_frame.py
```diff
@@ -117,24 +129,156 @@ def __init__(self,
...
+            # constructor.
+            self._y_true = None
+            self._y_pred = None
+            check_consistent_length(y_true, y_pred)
+            y_t = _convert_to_ndarray_and_squeeze(y_true)
+            y_p = _convert_to_ndarray_and_squeeze(y_pred)
+            self._compute_metric(y_t, y_p, sensitive_features, control_features, sample_params)
+        else:
+            # If we are streaming, we wait, but initialize accumulators.
+            assert y_true == [] and y_pred == [] and sensitive_features == [], _NON_EMPTY_INIT_ERR
```
riedgar-ms, line 160: This will eventually need to get changed to a ValueError, but that can wait until we're on to the polishing stage.

fairlearn/metrics/_metric_frame.py
```diff
@@ -117,24 +129,156 @@ def __init__(self,
...
+        sample_params : dict
+            Parameters for the metric function(s). If there is only one metric function,
+            then this is a dictionary of strings and array-like objects, which are split
+            alongside the ``y_true`` and ``y_pred`` arrays, and passed to the metric function.
+            If there are multiple metric functions (passed as a dictionary), then this is
+            a nested dictionary, with the first set of string keys identifying the
+            metric function name, with the values being the string-to-array-like dictionaries.
+        """
+        if not self._streaming:
+            raise Exception("This MetricFrame does not support adding data.")
```
riedgar-ms, line 214: Again, not vital quite yet, but could we raise a more specific exception? Although I'm not sure that `ValueError` is appropriate here.

fairlearn/metrics/_metric_frame.py
```diff
@@ -117,24 +129,156 @@ def __init__(self,
...
+        check_consistent_length(y_true, y_pred, sensitive_features)
```
riedgar-ms, line 215: Shouldn't there be a check on `sample_params` for consistent lengths as well?

riedgar-ms: Not to mention `control_features`

fairlearn/metrics/_metric_frame.py
```diff
@@ -117,24 +129,156 @@ def __init__(self,
...
+        if self._c_f:
+            self._c_f.append(control_features)
+
+        if sample_params:
+            if self._s_p is None:
+                self._s_p = [sample_params]
+            elif isinstance(self._s_p, list):
+                self._s_p.append(sample_params)
+            else:
+                raise ValueError("MetricFrame was initialized with `sample_params` already set.")
```
riedgar-ms, line 232: I don't understand this error. Aren't there two cases which should be being caught:
- User provides `sample_params` on the first call to this method, but then forgets on a subsequent call
- User _doesn't_ provide `sample_params` on the first call, but tries to on some later call

riedgar-ms (commented):
Thanks for keeping at this!

fairlearn/metrics/_metric_frame.py
Dref360: There is also the case where the user supplies `sample_params` at initialization. For example, to provide `average='binary'` to `skm.f1_score`. Then, the user is not allowed to supply `sample_params` after.

fairlearn/metrics/_metric_frame.py
Dref360: done

fairlearn/metrics/_metric_frame.py
Dref360: done

fairlearn/metrics/_metric_frame.py
riedgar-ms: Surely `average='binary'` isn't a _sample_ parameter? By definition, sample parameters have a unique value for each sample - things like sample weights.

fairlearn/metrics/_metric_frame.py
riedgar-ms: @adrinjalali  would have better thoughts than I do on that.

adrinjalali: Isn't this a `NotImplementedError`?

test/unit/metrics/test_metricframe_streaming.py
```diff
@@ -148,3 +154,72 @@ def test_exception_on_no_data():
...
+        y_t_batches, y_p_batches, g_4_batches = next(batch_generator)
+        g_f = pd.DataFrame(data=g_4_batches, columns=['My feature'])
+        target.add_data(transform_y_t(y_t_batches),
+                        transform_y_p(y_p_batches),
+                        sensitive_features=g_f,
+                        control_features=g_f)
+
+    batch_generator = batchify(num_batches, y_t, y_p, g_4)
+    # We initialize with CF
+    target = metrics.MetricFrame(skm.recall_score, [], [], sensitive_features=[],
```
riedgar-ms, line 179: Can you make this 'success' case a separate test?

test/unit/metrics/test_metricframe_streaming.py
```diff
@@ -148,3 +154,72 @@ def test_exception_on_no_data():
...
+
+    # We forget to supply it the next time, it should raise.
+    with pytest.raises(ValueError):
+        target.add_data(transform_y_t(y_t_batches),
+                        transform_y_p(y_p_batches),
+                        sensitive_features=g_f)
+
+    # Version with sample_weights supplied at the start.
+    target = metrics.MetricFrame(skm.recall_score, [], [], sensitive_features=[],
+                                 sample_params={'sample_weight': sample_weights},
```
riedgar-ms, line 214: Shouldn't this be `sample_params={'sample_weight': [] }` here - the sample weights would need to be streamed as well?

riedgar-ms (commented):
Sorry for the delay in reviewing.

riedgar-ms:
@Dref360 did you have any questions about my review comments? Have I missed something else you've changed?

Dref360:
Sorry about that, I'll work on the comments this week.

riedgar-ms:
> Sorry about that, I'll work on the comments this week.

No problem - we know this is volunteer work!

fairlearn/metrics/_metric_frame.py
```diff
@@ -119,24 +133,162 @@ def __init__(self,
                  y_pred, *,
```
riedgar-ms, line 133: As a warning #822 is changing this signature slightly. There may be merge conflicts.

fairlearn/metrics/_metric_frame.py
riedgar-ms: Checked below, I see.

fairlearn/metrics/_metric_frame.py
```diff
@@ -119,24 +133,162 @@ def __init__(self,
...
+        batch_type = type(batches[0])
+        if not all([type(arr) is batch_type for arr in batches]):
+            raise ValueError("Can't concatenate arrays of different types.")
+        if batch_type is np.ndarray:
+            result = np.concatenate(batches)
+        elif batch_type is list:
+            result = sum(batches, [])
+        elif batch_type is pd.DataFrame:
+            col_nums = batches[0].columns
+            assert all([col_nums == df.columns for df in batches]), 'Need same columns'
```
riedgar-ms, line 264: Can this raise a `ValueError` too?

fairlearn/metrics/_metric_frame.py
riedgar-ms: Ahh yes, I see above.

test/unit/metrics/test_metricframe_streaming.py
riedgar-ms: I think we should

test/unit/metrics/test_metricframe_streaming.py
```diff
@@ -0,0 +1,215 @@
...
+@pytest.mark.parametrize("transform_y_t", conversions_for_1d)
+@pytest.mark.parametrize("num_batches", [1, 3, 5])
+def test_cf_argument_mismatch(transform_y_t, transform_y_p, num_batches):
+    batch_generator = batchify(num_batches, y_t, y_p, g_4)
+
+    # We initialize without CF
+    target = metrics.MetricFrame(skm.recall_score, [], [], sensitive_features=[],
+                                 streaming=True)
+    # We supply CF in the call, it should fail.
+    with pytest.raises(ValueError):
```
riedgar-ms, line 169: Should there be a check on the exception message, so you can be sure that it raised on the right statement?

fairlearn/metrics/_metric_frame.py
```diff
@@ -119,24 +133,162 @@ def __init__(self,
...
+                check_consistent_length(y_true, *sample_params.values())
+                self._s_p = self._concat_batches([self._s_p, sample_params])
+            else:
+                raise ValueError("MetricFrame expected `sample_params` to be supplied.")
+
+        # Reset metric
+        self._overall = None
+        self._by_group = None
+
+    def _concat_batches(self, batches):
```
riedgar-ms, line 251: I know that it's technically private, but it might be worth writing tests of this specific method. It may be easier to make sure that each branch is checked that way.

riedgar-ms (commented):
Thank you for keeping at this!

riedgar-ms (commented):
Tagging @adrinjalali  and @MiroDudik  ... @Dref360 deserves a break from my reviews :-)

fairlearn/metrics/_metric_frame.py
adrinjalali: we really need to have that "patch" check @riedgar-ms , it'd be easier to review here then.

fairlearn/metrics/_metric_frame.py
```diff
@@ -195,24 +209,166 @@ def __init__(self,
...
+        """Concatenate a list of items together."""
```
adrinjalali, line 328: please explain in the docstring what this function's supposed to do exactly, the code seems a bit not so nice here.

fairlearn/metrics/_metric_frame.py
```diff
@@ -195,24 +209,166 @@ def __init__(self,
...
+        if len(batches) == 0:
+            raise ValueError(_EMPTY_BATCHES_ERR)
```
adrinjalali, line 330: instead of raise, I'd just return None.

fairlearn/metrics/_metric_frame.py
```diff
@@ -195,24 +209,166 @@ def __init__(self,
...
+        self._compute_metric(self._concat_batches(self._y_true),
+                             self._concat_batches(self._y_pred),
+                             self._concat_batches(self._s_f),
+                             self._concat_batches(self._c_f) if self._c_f else None,
+                             self._s_p)
+
+    def _compute_metric(self, y_t, y_p, s_f, c_f, s_p):
         # Now, prepare the sensitive features
-        sf_list = self._process_features("sensitive_feature_", sensitive_features, y_t)
+        sf_list = self._process_features("sensitive_feature_", s_f, y_t)
```
adrinjalali, line 363: could we please not have the very shortened version of the variable names?

adrinjalali (commented):
Thanks for all the work so far @Dref360 , it's certainly getting there.

fairlearn/metrics/_metric_frame.py
Dref360: I'm afraid that this could lead to issues in the future.
Other libraries also raise:

Numpy: `ValueError: need at least one array to concatenate`
Pandas: `ValueError: No objects to concatenate`

fairlearn/metrics/_metric_frame.py
riedgar-ms: Do you mean the code coverage check?

fairlearn/metrics/_metric_frame.py
riedgar-ms: I'm happy with raising here.

fairlearn/metrics/_metric_frame.py
```diff
@@ -171,6 +179,13 @@ class MetricFrame:
         a nested dictionary, with the first set of string keys identifying the
         metric function name, with the values being the string-to-array-like dictionaries.
 
+    streaming : bool
+        If True, will modify the behavior to accept the accumulation of values before computing
+        the score. One can append data with ``add_data``. The user can't supply data in the
+        constructor if ``streaming=True``.
+
+        **Note** currently, it only stores the values before computing the metric.
```
riedgar-ms, line 187: I'm not quite sure that this note is supposed to mean.

fairlearn/metrics/_metric_frame.py
riedgar-ms: Is this issue still outstanding? It does seem quite major based on @adrinjalali 's comments.

fairlearn/metrics/_metric_frame.py
```diff
@@ -195,24 +210,176 @@ def __init__(self,
...
+            y_true_processed = _convert_to_ndarray_and_squeeze(y_true)
+            y_pred_processed = _convert_to_ndarray_and_squeeze(y_pred)
+            self._compute_metric(y_true_processed, y_pred_processed,
+                                 sensitive_features, control_features, sample_params)
+        else:
+            # If we are streaming, we wait, but initialize accumulators.
+            accumulators = [y_true, y_pred, sensitive_features]
+            # CF must be None or an empty list.
+            cf_validated = control_features is None or len(control_features) == 0
+            # Same for SP
```
riedgar-ms, line 243: Nit: can we write out 'SP' and 'CF' in the comment at least.

test/unit/metrics/test_metricframe_streaming.py
```diff
@@ -0,0 +1,255 @@
...
+    for y_t_batches, y_p_batches, g_4_batches in batch_generator:
+        g_f = pd.DataFrame(data=g_4_batches, columns=['My feature'])
+        target.add_data(y_true=transform_y_t(y_t_batches),
+                        y_pred=transform_y_p(y_p_batches),
+                        sensitive_features=g_f)
+    assert len(target._y_true) == 2 * num_batches
+    assert target._overall is None  # has been reset
+    assert target._by_group is None  # has been reset
+
+    # We can recompute it with the new batches.
```
riedgar-ms, line 45: You mean that invoking `by_group` implicitly recomputes?

test/unit/metrics/test_metricframe_streaming.py
```diff
@@ -0,0 +1,255 @@
...
+    # If we continue to add data, it resets the metrics
+    g_f = pd.DataFrame(data=g_4_batches, columns=['My feature'])
+    target.add_data(y_true=transform_y_t(y_t_batches),
+                    y_pred=transform_y_p(y_p_batches),
+                    sensitive_features=g_f)
+    assert target._overall is None
+    assert target._by_group is None
+
+    # We can't initialize streaming metrics with data
+    with pytest.raises(ValueError) as excinfo:
```
riedgar-ms, line 150: This looks like it should be a separate test.

test/unit/metrics/test_metricframe_streaming.py
riedgar-ms: Have you added these tests? I seem to be overlooking them.

riedgar-ms (commented):
Thank you for the update!

fairlearn/metrics/_metric_frame.py
adrinjalali: yes

fairlearn/metrics/_metric_frame.py
adrinjalali: If I'm not mistaken, we are NOT going down the pandas way, so I'd say we're good?

fairlearn/metrics/_metric_frame.py
Dref360: In #655, we mentioned that in the future we would like to support object that have an "update function. Like [Keras Stateful Metrics](https://keras.io/api/metrics/) or [BaaL Metrics](https://github.com/ElementAI/baal/blob/master/src/baal/utils/metrics.py#L22).

In this PR, we simply accumulate the predictions and we run the metrics on all values.

riedgar-ms: Isn't that always going to be required?

Dref360: Yeah I guess, unless we have multiple metrics object at the same time for each combination of sensitive attribute.

I'll remove for now and we can decide what to do later.

fairlearn/metrics/_metric_frame.py
```diff
@@ -3,32 +3,40 @@
...
 _FEATURE_LIST_NONSCALAR = "Feature lists must be of scalar types"
 _FEATURE_DF_COLUMN_BAD_NAME = "DataFrame column names must be strings. Name '{0}' is of type {1}"
 _DUPLICATE_FEATURE_NAME = "Detected duplicate feature name: '{0}'"
 _TOO_MANY_FEATURE_DIMS = "Feature array has too many dimensions"
-_SAMPLE_PARAM_KEYS_NOT_IN_FUNC_DICT = \
-    "Keys in 'sample_params' do not match those in 'metric'"
+_SAMPLE_PARAM_KEYS_NOT_IN_FUNC_DICT = ("Keys in 'sample_params'"
+                                       " do not match those in 'metric'")
+_BAD_INIT_ERROR = """Non streaming metrics require data to process.
+ Set streaming=True for streaming metrics"""
```
adrinjalali, line 33: we should be consistent, either have newlines in all messages or none, I prefer none.

fairlearn/metrics/_metric_frame.py
```diff
@@ -194,24 +207,181 @@ def __init__(self,
...
+            If there are multiple metric functions (passed as a dictionary), then this is
+            a nested dictionary, with the first set of string keys identifying the
+            metric function name, with the values being the string-to-array-like dictionaries.
+        """
+        if not self._streaming:
+            raise NotImplementedError(_NOT_STREAMING_ERR)
+        check_consistent_length(y_true, y_pred, sensitive_features)
+
         y_t = _convert_to_ndarray_and_squeeze(y_true)
         y_p = _convert_to_ndarray_and_squeeze(y_pred)
```
adrinjalali, line 298: @riedgar-ms how does this work with the arbitrary y_true, y_pred shapes as you have it in the example we just merged?

fairlearn/metrics/_metric_frame.py
```diff
@@ -194,24 +207,181 @@ def __init__(self,
...
+                                 control_features,
+                                 sample_params)
+        else:
+            # If we are streaming, we wait, but initialize accumulators.
+            accumulators = [y_true, y_pred, sensitive_features]
+            # control_features must be None or an empty list.
+            cf_validated = control_features is None or len(control_features) == 0
+            # Same for sample_params
+            sp_validator = sample_params in (None, {})
+            if any(acc != [] for acc in accumulators) or not cf_validated or not sp_validator:
```
adrinjalali, line 244:
```suggestion
            if any(acc for acc in accumulators) or not cf_validated or not sp_validator:
```

fairlearn/metrics/_metric_frame.py
```diff
@@ -194,24 +207,181 @@ def __init__(self,
...
+            self._overall = None
+            self._by_group = None
+
+    def add_data(self, *, y_true, y_pred, sensitive_features, control_features: Optional = None,
+                 sample_params: Optional[Union[Dict[str, Any], Dict[str, Dict[str, Any]]]] = None):
+        """Add data to the MetricFrame.
+
+        Parameters
+        ----------
+        y_true : List, pandas.Series, numpy.ndarray, pandas.DataFrame
```
adrinjalali, line 257:
```suggestion
        y_true : array-like
```

?

fairlearn/metrics/_metric_frame.py
```diff
@@ -194,24 +207,181 @@ def __init__(self,
...
+        The ground-truth labels (for classification) or target values (for regression).
+
+        y_pred : List, pandas.Series, numpy.ndarray, pandas.DataFrame
+                The predictions.
+
+        sensitive_features : List, pandas.Series, dict of 1d arrays, numpy.ndarray,
+         pandas.DataFrame
```
adrinjalali, line 264:
```suggestion
        sensitive_features : array-like, or dict of array-like
```

?

fairlearn/metrics/_metric_frame.py
```diff
@@ -194,24 +207,181 @@ def __init__(self,
...
+                raise ValueError(f"Column mismatch expected {col_nums},"
+                                 f" got {[df.columns for df in batches]}")
+            result = pd.concat(batches)
+        elif batch_type is dict:
+            # Create a new dict that merges values of all keys.
+            all_keys = set(sum((list(b.keys()) for b in batches), []))
+            return {k: self._concat_batches([b[k] for b in batches if k in b])
+                    for k in all_keys}
+        else:
+            raise ValueError(f"Can't concatenate {batches}")
```
adrinjalali, line 359: I think I'd be happier if we just convert everything to numpy arrays and concatenate them instead of all these different cases. Do we need them to stay the same type?

adrinjalali (commented):
Thanks for all the work so far @Dref360 

One general concern I have is that this is touching a lot of the MetricFrame. The question is, do we want it to be the same class, or rather have a `StreamingMetricFrame` where everything is just assumed to be streamed? We can share the common code in a common parent class.

We should also add this to our user guides.

fairlearn/metrics/_metric_frame.py
riedgar-ms: Aren't these lines unchanged? Now, it's true that doing the squeeze might have unpleasant effects if the user supplied metric functions were expecting the extra dimensions. I'm not sure that's an issue for this PR, though.

adrinjalali:
Maybe with a bit of refactoring I'd be happier with having only one class, but right now reviewing the PR gives me the impression that it's not very maintainable.

fairlearn/metrics/_metric_frame.py
romanlutz: Doesn't sphinx complain about this since it's not technically a "type"?

adrinjalali: I don't think so: https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.FastICA.html#sklearn.decomposition.FastICA.fit_transform

adrinjalali: This is the guidelines sklearn has: https://scikit-learn.org/stable/developers/contributing.html?highlight=contribute#guidelines-for-writing-documentation

Dref360:
I agree that simply making a wrapper class around MetricFrame would be easier, so I'll close this PR and open another later with a new implementation.

Thank you for your reviews!

PR #671, opened 2021-01-07
Fix problems with TensorFlow

riedgar-ms:
The `Other-ML-Packages` pipeline was having trouble with TensorFlow and the `ThresholdOptimizer` test. This was ultimately traced to a problem calling `LabelEncoder()` with a pandas `Series` of strings. For some reason, a quite old version of pandas was being installed, and the recent update to SciKit-Learn didn't work with this. Updating the entire build to a more recent Python version (pandas has recently dropped support for Python 3.6) appears to solve the problem.


romanlutz (approved):
py3.6 is meant to be deprecated anyway ;-) Thanks for fixing this!

MiroDudik (approved):

PR #678, opened 2021-01-14
FAQ Update

riedgar-ms:
Add some more questions to the FAQ, based on a training course run at Microsoft.

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -54,6 +54,8 @@ Why am I seeing fairness issues, even though my data are reflective of the gener
     Machine learning models often perform poorly for subgroups which are poorly
     represented.
     For this reason, balanced sampling is generally better for ML than population sampling.
+    On a related point, this is why it is important to consider multiple fairness metrics,
+    and how they vary across different subgroups.
 
 Won't making a model fairer reduce its accuracy?
     There are often many machine learning models that achieve similar levels of accuracy
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -70,6 +70,18 @@ Can the mitigation algorithms in Fairlearn make my model fair?
     However, to assess whether the new model is fairer, it is important to consider
     not only the fairness metrics, but also the societal and technical context in which
     the model is applied.
+
+I've got improved data, trained and mitigated new models, checked all the metrics... am I done?
+    The ML lifecycle doesn't end when a model is deployed.
+    Models need to be monitored in production.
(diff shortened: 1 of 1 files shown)


docs/faq.rst
```diff
@@ -65,6 +72,18 @@ Can the mitigation algorithms in Fairlearn make my model fair?
     However, to assess whether the new model is fairer, it is important to consider
     not only the fairness metrics, but also the societal and technical context in which
     the model is applied.
+
+I've got improved data, trained and mitigated new models, checked all the metrics... am I done?
+    The ML lifecycle doesn't end when a model is deployed.
```
romanlutz, line 77: I would expand a little on the statement in the beginning about metrics. Metrics don't tell the whole story. Fairness is sociotechnical. So this may not even have been sufficient at deployment time. I do like your points about data drift etc., of course.

docs/faq.rst
```diff
@@ -65,6 +72,18 @@ Can the mitigation algorithms in Fairlearn make my model fair?
...
+    Models need to be monitored in production.
+    On a technical level, this means checking for data drift within the vulnerable
+    subgroups identified during the fairness analysis.
+    However the societal aspects need to be considered as well, for example:
+    
+    - Are the actual harms as expected, both in the nature of the harms and their
+      distribution?
+    - Have the users of the model (who may not be the subjects of the model)
```
romanlutz, line 85: The term I keep reading about in the literature is "strategic behavior" in case you want to use it.

docs/faq.rst
riedgar-ms: Good point - even though I wrote the question, I didn't notice that it implicitly included a deploy/no-deploy choice. I've added a note to highlight this.

docs/faq.rst
riedgar-ms: And at the cost of some personal mental trauma, I'll even use that spelling....

docs/faq.rst
```diff
@@ -50,6 +50,13 @@ The model is unfair because the data are biased. Isn't it better to get better d
     each. In some of these cases, obtaining 'better data' may not be practical, but it
     might still be possible to use some mitigation algorithms.
 
+Why am I seeing fairness issues, even though my data are reflective of the general population?
+    Machine learning models often perform poorly for subgroups which are poorly
```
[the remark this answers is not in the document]
riedgar-ms: I've tried to work that in. Let me know what you think.

romanlutz (approved):

PR #686, opened 2021-01-28
Updates for next release

riedgar-ms:
Making sure our various files are up-to-date, in preparation for the next release.


romanlutz:
> Making sure our ReadMe is up-to-date, in preparation for the next release.

Do you want to check CHANGES.md as well? It's definitely missing preprocessing, perhaps more :-(

riedgar-ms:
> > Making sure our ReadMe is up-to-date, in preparation for the next release.
> 
> Do you want to check CHANGES.md as well? It's definitely missing preprocessing, perhaps more :-(

I should read what I'm writing more carefully.... this PR currently only updates the changes file. I'll take a look at the ReadMe too

romanlutz:
Since this is in anticipation of a release.... the dashboard warning says that after 0.5.0 the dashboard will go away. Should instead say "after the next release"? That way we don't need to update it and we're eventually right :-) Also, people can prepare by switching to `raiwidgets` starting tomorrow.

riedgar-ms:
> Since this is in anticipation of a release.... the dashboard warning says that after 0.5.0 the dashboard will go away. Should instead say "after the next release"? That way we don't need to update it and we're eventually right :-) Also, people can prepare by switching to `raiwidgets` starting tomorrow.

I've gone through and adjusted the text to make the ultimate removal more ambiguous. Please let me know if I've missed any.

romanlutz (approved):
Thanks! `raiwidgets` is now released as well :-)

docs/quickstart.rst
```diff
@@ -142,7 +142,7 @@ While this page shows only screenshots, the actual dashboard is interactive.
 .. note::
 
     The :code:`FairlearnDashboard` will move from Fairlearn to the
-    :code:`raiwidgets` package after the v0.5.0 release. Instead, Fairlearn
+    :code:`raiwidgets` package in the near future. Instead, Fairlearn
```
adrinjalali, line 145: we can point users to the dashboard package already since it's released. We should also raise a warning when they use the dashboard and point them to the other package.

riedgar-ms: In the code, we're doing that already:
https://github.com/fairlearn/fairlearn/blob/44244c6a513eb9b8446369d3c92dce4adf1e4f7e/fairlearn/widget/_fairlearn_dashboard.py#L47
Let me redraft these now that raiwidgets is published to PyPi

adrinjalali (approved):

PR #691, opened 2021-02-03
Release v0.6.0

riedgar-ms:
Updates to the development trunk for v0.6.0 (these occur immediately after the tag/branch for v0.6.0:

- Bumps development version to v0.6.1.dev0
- Updates ReadMe to have link to v0.6.0
- Updates landing page to point to v0.6.0 docs by default
- Ensures that v0.6.0 docs are built

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Website: https://fairlearn.github.io/
 
 ## Current release
 
-- The current stable release is available at [Fairlearn v0.5.0](https://github.com/fairlearn/fairlearn/tree/release/v0.5.0).
+- The current stable release is available at [Fairlearn v0.5.0](https://github.com/fairlearn/fairlearn/tree/release/v0.6.0).
 
 - Our current version differs substantially from version 0.2 or earlier. Users of these older versions should visit our [onboarding guide](https://fairlearn.github.io/master/contributor_guide/development_process.html#onboarding-guide).
 
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Website: https://fairlearn.github.io/
 
 ## Current release
 
-- The current stable release is available at [Fairlearn v0.5.0](https://github.com/fairlearn/fairlearn/tree/release/v0.6.0).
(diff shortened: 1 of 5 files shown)


docs/conf.py
```diff
@@ -89,7 +89,7 @@ def check_if_v046():
 
 # Multiversion settings
 
-smv_tag_whitelist = r'^v0\.4\.6|^v0\.5\.\d+$'
+smv_tag_whitelist = r'^v0\.4\.6|^v0\.5\.\d|^v0\.6\.\d+$'
```
adrinjalali, line 92: should this be a bit more general not to include every single release?

adrinjalali (approved):
otherwise LGTM

docs/conf.py
riedgar-ms: I'm open to suggestions for improving it. The problem is that it must not match anything before `v0.4.6` since the docs were set up very different back then (if they even existed).

README.md
```diff
@@ -19,7 +19,7 @@ Website: https://fairlearn.github.io/
 
 ## Current release
 
-- The current stable release is available at [Fairlearn v0.5.0](https://github.com/fairlearn/fairlearn/tree/release/v0.5.0).
+- The current stable release is available at [Fairlearn v0.5.0](https://github.com/fairlearn/fairlearn/tree/release/v0.6.0).
```
romanlutz, line 22: The string shown will still say v0.5.0 right now.

romanlutz (approved):
Can we also add a new section to CHANGES.md with a new version so that we can add to it?

README.md
riedgar-ms: Arghhh.... Thanks for catching that.

docs/conf.py
adrinjalali: The lazy in me says "ok then" :D

riedgar-ms: I think that the problem will be easier in a few versions time - once `v0.10.0` is our oldest supported version, we can just check for two digits in the second character.

PR #693, opened 2021-02-04
More Python 3.9 builds

riedgar-ms:
Add Python 3.9 to more of the builds we have. Unfortunately, this doesn't work with the 'Nightly Fixed' build.

### DIFF
--- a/devops/nightly-requirements-fixed.yml
+++ b/devops/nightly-requirements-fixed.yml
@@ -25,7 +25,7 @@ jobs:
   parameters:
     platforms:  { Linux: ubuntu-latest, Windows: vs2017-win2016 }
     installationType: 'PipLocal'
-    pyVersions: [3.6, 3.7, 3.8]
+    pyVersions: [3.6, 3.7, 3.8, 3.9]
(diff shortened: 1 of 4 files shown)


romanlutz:
> Add Python 3.9 to more of the builds we have. Unfortunately, this doesn't work with the 'Nightly Fixed' build.

What's the issue with Nightly Fixed?

romanlutz (approved):

riedgar-ms:
> > Add Python 3.9 to more of the builds we have. Unfortunately, this doesn't work with the 'Nightly Fixed' build.
> 
> What's the issue with Nightly Fixed?

The installation of requirements fails. It looks like `pip install` is building things from source, and when it gets to `scikit-learn` that build fails (since `Cython` is not installed). Perhaps our lower-bound version is no longer supported, and hence hasn't had packages build for Python 3.9?

adrinjalali (approved):

adrinjalali:
`scikit-learn` has py3.9 wheels up there, wondering why they fail.

PR #699, opened 2021-02-11
Put requirements file into tarball

### LINKED ISSUE #683: pip install from source fails due to requirements.txt not being included in .tar.gz dist
#### Describe the bug
When installing from src dist, fairlearn fails due to missing requirements.txt file.

#### Steps/Code to Reproduce
```shell
pip download --no-binary :all: -d ./ fairlearn==0.5.0
```

#### Expected Results
```shell
Collecting fairlearn==0.4.6
  Downloading fairlearn-0.4.6.tar.gz (10.5 MB)
     |████████████████████████████████| 10.5 MB 16.6 MB/s
```
#### Actual Results
```shell
Collecting fairlearn==0.5.0
  Using cached fairlearn-0.5.0.tar.gz (12.2 MB)
    ERROR: Command errored out with exit status 1:
     command: /usr/bin/python3.7 -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-download-muxh63mo/fairlearn_d4026bee74b547c9957a0a48e60fe570/setup.py'"'"'; __file__='"'"'/tmp/pip-download-muxh63mo/fairlearn_d4026bee74b547c9957a0a48e60fe570/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base /tmp/pip-pip-egg-info-bfv_4scs
         cwd: /tmp/pip-download-muxh63mo/fairlearn_d4026bee74b547c9957a0a48e60fe570/
    Complete output (5 lines):
      File "<string>", line 1, in <module>
      File "/tmp/pip-download-muxh63mo/fairlearn_d4026bee74b547c9957a0a48e60fe570/setup.py", line 12, in <module>
        with open('requirements.txt') as f:
    FileNotFoundError: [Errno 2] No such file or directory: 'requirements.txt'
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```

#### Screenshots
image

#### Versions
OS: Ubuntu (Windows Subsystem for Linux)
Python: 3..7.17
pip: 20.3.3
fairlearn: 0.5.0

riedgar-ms:
Put the two requirements files into the distribution tarball, so that things such as

```
pip download --no-binary :all: -d ./ fairlearn
```
will work. Closes #683

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -55,6 +55,10 @@ setuptools.setup(
         ('share/jupyter/lab/extensions', [
             'fairlearn/widget/js/'
             'fairlearn_widget/labextension/fairlearn-widget-0.1.0.tgz'
+        ]),
+        ('.', [
+            'requirements.txt',
+            'requirements-customplots.txt'
         ])],
     zip_safe=False,
 )
--- a/setup.py
+++ b/setup.py
@@ -56,8 +56,10 @@ setuptools.setup(
             'fairlearn/widget/js/'
             'fairlearn_widget/labextension/fairlearn-widget-0.1.0.tgz'
         ]),
-        ('.', [
-            'requirements.txt',
+        ('./requirements.txt', [
+            'requirements.txt'
+        ]),
+        ('./requirements-customplots.txt', [
             'requirements-customplots.txt'
         ])],
     zip_safe=False,


romanlutz (approved):
Doesn't seem like this has any drawbacks, right?

riedgar-ms:
> Doesn't seem like this has any drawbacks, right?

I don't think so, but I'll defer to @adrinjalali who has rather more experience in these things.

adrinjalali:
It's not really a common practice to use `requirements.txt` as `install_requires`. 

As [packaging.python.org](https://packaging.python.org/discussions/install-requires-vs-requirements/) nicely put it:

> Whereas install_requires defines the dependencies for a single project, Requirements Files are often used to define the requirements for a complete Python environment.
> 
> Whereas install_requires requirements are minimal, requirements files often contain an exhaustive listing of pinned versions for the purpose of achieving repeatable installations of a complete environment.
> 
> Whereas install_requires requirements are “Abstract”, i.e. not associated with any particular index, requirements files often contain pip options like --index-url or --find-links to make requirements “Concrete”, i.e. associated with a particular index or directory of packages. 1
> 
> Whereas install_requires metadata is automatically analyzed by pip during an install, requirements files are not, and only are used when a user specifically installs them using python -m pip install -r.
> 

Here's a few examples from other packages;
[__numba__](https://github.com/numba/numba/blob/199798e2c849b5e63eeef36972566fda7b84625c/setup.py#L365-L370):
```python
build_requires = ['numpy >={}'.format(min_numpy_build_version)]
install_requires = [
    'llvmlite >={},<{}'.format(min_llvmlite_version, max_llvmlite_version),
    'numpy >={}'.format(min_numpy_run_version),
    'setuptools',
]
```
__scikit-learn__ uses [this file](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/_min_dependencies.py) in [setup.py](https://github.com/scikit-learn/scikit-learn/blob/95fa18f2782c3fcebd7a8388cde85323104f22b0/setup.py#L265) and other places in the same file (and other places in CI):
``` python
                    install_requires=min_deps.tag_to_packages['install'],
```

__imbalanced-learn__ has the code in [setup.py](https://github.com/scikit-learn-contrib/imbalanced-learn/blob/3444430a2a1488de3b96249e363ff4a8aab8a5dd/setup.py#L41-L58):
``` python
INSTALL_REQUIRES = [
    'numpy>=1.13.3',
    'scipy>=0.19.1',
    'scikit-learn>=0.24',
    'joblib>=0.11'
]
EXTRAS_REQUIRE = {
    "tests": ["pytest", "pytest-cov"],
    "docs": [
        "sphinx",
        "sphinx-gallery",
        "sphinx_rtd_theme",
        "sphinxcontrib-bibtex",
        "numpydoc",
        "matplotlib",
        "pandas",
    ],
}
```

This doesn't apply when there are _build_ requirements which are not required for _install_, and for those you can put them in `pyproject.toml` like [scikit-learn's](https://github.com/scikit-learn/scikit-learn/blob/main/pyproject.toml).

riedgar-ms:
This is getting somewhat beyond the narrow issue identified above, however I'm no expert in publishing wheels and the like.

One question for @adrinjalali  based on the above.... it looks like scikit-learn lists a scipy version bound in two places Once in `_min_dependencies.py`:
https://github.com/scikit-learn/scikit-learn/blob/95fa18f2782c3fcebd7a8388cde85323104f22b0/sklearn/_min_dependencies.py#L11
and once in `pyproject.toml`:
https://github.com/scikit-learn/scikit-learn/blob/95fa18f2782c3fcebd7a8388cde85323104f22b0/pyproject.toml#L14

Are those versions allowed to be different? My motivation for the setup in Fairlearn was to ensure that there was exactly one source of truth for each package bound. Hence, `install_requires` reads in the `requirements.txt` - and we have other requirements files to set up the dev environment.

adrinjalali:
build dependencies are not always the same as runtime dependencies @riedgar-ms . Once you build the C code against scipy's ABI, you can then have a different dependency on the python side.

But ideally I agree you should have things in once place, that's why we have the `_min_dependencies.py` file, it's used in the CI as well.

riedgar-ms:
@adrinjalali thanks for pointing that out - I was only thinking of the usual dynamic runtime dependencies on Python code. I completely forgot about the C backing to a bunch of these libraries.

Are you OK with this as a fix for the specific issue raised?

adrinjalali:
I still would rather not use `requirements.txt` for `install_requires`. I think the original issue stems from the fact that we use requirements.txt, rather than a better solution. Why not fix that which would also fix the raised issue?

riedgar-ms:
OK. I'll close this PR.

PR #700, opened 2021-02-16
Remove scheduled builds from old 'master' branch

riedgar-ms:
Attempting to stop a bunch of scheduled builds kicking off on the (recently retired) `master` branch. It appears that since they still have the ADO build definitions referencing `master` the pipelines are still getting triggered. CircleCI may suffer from a similar problem. Adjust the branch name triggers for all of these.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -115,4 +115,4 @@ workflows:
             - build-doc
           filters:
             branches:
-              only: master  # don't deploy to GitHub pages from other branches
+              only: main  # don't deploy to GitHub pages from other branches
--- a/devops/PR-Gate.yml
+++ b/devops/PR-Gate.yml
@@ -3,7 +3,7 @@ variables:
   FreezeFileStem: 'requirements-freeze'
 
 pr:
-- master
+- main
 - release/*
 
 trigger: none # No CI build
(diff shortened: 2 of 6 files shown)


adrinjalali (approved):
@romanlutz missed this one?

romanlutz:
> @romanlutz missed this one?

Odd, somehow didn't show up in the repo string search... thanks @riedgar-ms !!!

romanlutz (approved):

riedgar-ms:
@romanlutz updated the YMLs on the new `main` branch, but left the old ones on `master`. Reading things online, it appears that ADO is checking multiple branches for the pipeline YML, and seeing if it matches.

romanlutz:
> @romanlutz updated the YMLs on the new `main` branch, but left the old ones on `master`. Reading things online, it appears that ADO is checking multiple branches for the pipeline YML, and seeing if it matches.

So the "better" fix would be to delete this branch, wouldn't it?

riedgar-ms:
Once I've straightened out these spurious builds (won't find out if this did the trick until tomorrow), we should investigate ways of 'locking' the old `master` branch.

adrinjalali:
Oh interesting, didn't you use GH's tool to change the default branch? That did remove the `master` branch when I did it for `scikit-learn`, and automatically forwarded all the old links to the new branch. We should also delete all the branches that are there from the time of not creating PRs from your own repos I think.

romanlutz:
Yes I used GH's tool. Somehow the branch seems to still exist. I'm in favor of removing it to avoid confusion.

PR #701, opened 2021-02-18
Add missing MetricFrame tests

riedgar-ms:
The `MetricFrame` smoke tests didn't cover cases with `sample_params` (there are tests of `sample_params` but they check the internal methods, and not the API which users would invoke). Add a couple of tests to rectify this oversight.

### DIFF
--- a/test/unit/metrics/test_metricframe_smoke.py
+++ b/test/unit/metrics/test_metricframe_smoke.py
@@ -8,6 +8,7 @@ import sklearn.metrics as skm
 
 import fairlearn.metrics as metrics
 
+# Bring in some pre-prepared input arrays
 from .data_for_test import y_t, y_p, g_1, g_2, g_3, g_4, s_w
 
 from test.unit.input_convertors import conversions_for_1d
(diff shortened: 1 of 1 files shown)


romanlutz (approved):
Thanks for adding these!!!

test/unit/metrics/test_metricframe_smoke.py
```diff
@@ -226,6 +226,95 @@ def test_1m_1sf_1cf_metric_dict(transform_y_t, transform_y_p):
     assert target_maxs['recall']['m'] == max(recall_m_arr)
 
 
+def test_1m_1_sf_sample_weights():
+
+    def multi_sp(y_t, y_p, p1, p2):
+        assert len(y_t) == len(y_p)
```
adrinjalali, line 232: could you please add a comment about what this function does?

test/unit/metrics/test_metricframe_smoke.py
```diff
@@ -226,6 +226,95 @@ def test_1m_1sf_1cf_metric_dict(transform_y_t, transform_y_p):
     assert target_maxs['recall']['m'] == max(recall_m_arr)
 
 
+def test_1m_1_sf_sample_weights():
```
adrinjalali, line 229: a small comment here wouldn't hurt ;)

test/unit/metrics/test_metricframe_smoke.py
```diff
@@ -226,6 +226,95 @@ def test_1m_1sf_1cf_metric_dict(transform_y_t, transform_y_p):
...
+                                 sensitive_features=g_1,
+                                 sample_params={'p1': param1, 'p2': param2})
+
+    # Sanity check types
+    assert isinstance(target.overall, float)
+    assert isinstance(target.by_group, pd.Series)
+    assert target.by_group.shape == (2,)
+
+    assert target.overall == sum(param2)
+    for g in g_1:
```
adrinjalali, line 253: what's `g_1`? It's imported up there, but it's really not easy to read this code and understand what it does.

adrinjalali (commented):
Thanks @riedgar-ms , same comments apply for the second test :)

test/unit/metrics/test_metricframe_smoke.py
riedgar-ms: Done

test/unit/metrics/test_metricframe_smoke.py
riedgar-ms: Added a note here, and up at the top where it's imported

test/unit/metrics/test_metricframe_smoke.py
riedgar-ms: Done

riedgar-ms:
While there were tests of `sample_params` there were not in the end-to-end flows which users would invoke, but instead called internals.

adrinjalali (approved):

PR #702, opened 2021-02-19
add CNAME file command to website build

romanlutz:
This is now necessary to ensure the CNAME file stays in the fairlearn.github.io repo.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,6 +92,11 @@ jobs:
           # Instead, we want the webpage to just point at index.html
           name: Disable jekyll builds
           command: touch docs/_build/html/.nojekyll
+      
+      - run:
+          # Create CNAME file for fairlearn.org
+          name: Create CNAME file
+          command: echo "fairlearn.org" > docs/_build/html/CNAME
 
       - run:
           name: 'Install gh-pages and configure dependencies'


adrinjalali:
The tests are failing, related?

adrinjalali (approved):
Looking at the log, it seems they fail because one or two dataset IDs are removed from openml O_o. There are also quite a bit of `FutureWarning` and `DeprecationWarning` which we need to fix.

Since the failures are unrelated, LGTM.

riedgar-ms (approved):

romanlutz:
> Looking at the log, it seems they fail because one or two dataset IDs are removed from openml O_o. There are also quite a bit of `FutureWarning` and `DeprecationWarning` which we need to fix.
> 
> Since the failures are unrelated, LGTM.

Where do you get the info about removed datasets from? The "dataset not available" style warnings are not that infrequent, and rerunning has helped in the past. FWIW I can see both those datasets still on the openml website with no indication of removal.

romanlutz:
> > Looking at the log, it seems they fail because one or two dataset IDs are removed from openml O_o. There are also quite a bit of `FutureWarning` and `DeprecationWarning` which we need to fix.
> > Since the failures are unrelated, LGTM.
> 
> Where do you get the info about removed datasets from? The "dataset not available" style warnings are not that infrequent, and rerunning has helped in the past. FWIW I can see both those datasets still on the openml website with no indication of removal.

Turns out I was right! Rerunning fixed it. Completing now.

PR #727, opened 2021-04-06
Fix and Update _create_group_metric_set()

### LINKED ISSUE #729: fairlearn.metrics._group_metric_set._create_group_metric_set fails for regression since y_true contains values other than [0, 1]
#### Describe the bug
I'm trying to run some fairness analysis on some of our models, where we can produce either classification or regression models. With regression, I found that it was failing to calculate metrics as it was expecting y_test to have values of only 0 and 1, which isn't true for our regression dataset. Strangley, even if I sent in a dummy y_test which only contained 0s and 1s, it still failed with the same error.

#### Steps/Code to Reproduce
```
X_test = <real test data for my model>
y_test = <real test data for my model> 
# I tried synthesizing a y_test with only 0 and 1, and it still failed
y_test = pd.DataFrame(0, index=np.arange(164), columns=["ERP"])
y_test['ERP'][0] = 1
y_test
y_pred = model.predict(X_test)
dash_dict = _create_group_metric_set(
                                            y_true=y_test.values,
                                            predictions=y_pred,
                                            sensitive_features=["column_0"],
                                            prediction_type="regression"
    )
```

#### Expected Results
We would expect to get the dashboard dict returned for this regression model instead of getting an exception.

#### Actual Results
  File "fairness.py", line 430, in <module>
    validate_model_ids
  File "fairness.py", line 126, in run_fairness_analysis_single_model
    prediction_type=prediction_type
  File "site-packages\fairlearn\metrics\_group_metric_set.py", line 164, in _create_group_metric_set
    result[_Y_TRUE], prediction, sensitive_features=g[_BIN_VECTOR])
  File "site-packages\fairlearn\metrics\_metric_frame.py", line 151, in __init__
    self._by_group = self._compute_by_group(func_dict, y_t, y_p, sf_list, cf_list)
  File "site-packages\fairlearn\metrics\_metric_frame.py", line 169, in _compute_by_group
    return self._compute_dataframe_from_rows(func_dict, y_true, y_pred, rows)
  File "site-packages\fairlearn\metrics\_metric_frame.py", line 194, in _compute_dataframe_from_rows
    curr_metric = func_dict[func_name].evaluate(y_true, y_pred, mask)
  File "site-packages\fairlearn\metrics\_function_container.py", line 103, in evaluate
    return self.func_(y_true[mask], y_pred[mask], **params)
  File "\site-packages\fairlearn\metrics\_balanced_root_mean_squared_error.py", line 36, in _balanced_root_mean_squared_error
    raise ValueError(_Y_TRUE_NOT_0_1)
ValueError: Only 0 and 1 are allowed in y_true and both must be present

#### Versions

I'm running this on a Linux VM in Azure running pyhton 3.6.8 with latest version of fairlearn from pypi. 

Thanks!

riedgar-ms:
The function `_create_group_metric_set()` was not calculating the correct set of regression metrics, nor was it set up to compute 'probability' metrics (i.e. binary classification where the predictions are probabilities). Rectify these oversights, even though this specific function will soon be removed from Fairlearn itself (it is only used when uploading copies of the dashboard to Azure).

Fixes #729


romanlutz (approved):
Thanks for adding this!

fairlearn/metrics/_group_metric_set.py
```diff
@@ -27,63 +27,71 @@
...
+MEAN_UNDERPREDICTION = "underprediction"
+MISS_RATE = "miss_rate"
+PRECISION_SCORE = "precision_score"
+R2_SCORE = "r2_score"
+RECALL_SCORE = "recall_score"
+ROC_AUC_SCORE = "balanced_accuracy_score"
+ROOT_MEAN_SQUARED_ERROR = "root_mean_squared_error"
+SELECTION_RATE = "selection_rate"
+SPECIFICITY_SCORE = "specificity_score"
+ZERO_ONE_LOSS = "zero_one_loss"
```
adrinjalali, line 63: these changes are not relevant to this PR, right? It's usually easier to review a PR when the scope is limited ;)

adrinjalali (approved):
LGTM, thanks @riedgar-ms

fairlearn/metrics/_group_metric_set.py
riedgar-ms: Strictly speaking, no.

PR #739, opened 2021-04-13
Update main for v0.6.1 Release

riedgar-ms:
Required updates to `main` for the `v0.6.1` release:

- Bump version in `__init__.py`
- Update links on the static landing page

Don't need to update the doc configuration this time (regex is flexible enough)

### DIFF
--- a/fairlearn/__init__.py
+++ b/fairlearn/__init__.py
@@ -9,7 +9,7 @@ import os
 from .show_versions import show_versions  # noqa: F401
 
 __name__ = "fairlearn"
-__version__ = "0.6.1.dev0"
+__version__ = "0.6.2.dev0"
 _base_version = __version__  # To enable the v0.4.6 docs
 
 # Setup logging infrastructure
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -34,4 +34,4 @@ pypandoc
 sphinx
 sphinx-gallery
 git+https://github.com/Holzhaus/sphinx-multiversion.git
-pydata-sphinx-theme>=0.5.0
+pydata-sphinx-theme==0.5.2
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Website: https://fairlearn.github.io/
 
 ## Current release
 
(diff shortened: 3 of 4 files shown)


requirements-dev.txt
```diff
@@ -34,4 +34,4 @@ pypandoc
 sphinx
 sphinx-gallery
 git+https://github.com/Holzhaus/sphinx-multiversion.git
-pydata-sphinx-theme>=0.5.0
\ No newline at end of file
+pydata-sphinx-theme==0.5.2
```
adrinjalali, line 37: ideally you'd fix this on `main`, in a separate PR, and then `cherry-pick` on the release branch.

fairlearn/__init__.py
```diff
@@ -9,7 +9,7 @@
 from .show_versions import show_versions  # noqa: F401
 
 __name__ = "fairlearn"
-__version__ = "0.6.1.dev0"
+__version__ = "0.6.2.dev0"
```
adrinjalali, line 12: This whole PR seems to need to be cherry picked for the release branch, and therefor this change needs to be in a separate PR I think.

requirements-dev.txt
riedgar-ms: This specific fix is now in #740 , which I'll cherry pick to `release/v0.6.1`. Technically this part doesn't need to go into the release branch, since documentation is always built from `main` (with the different versions due to `sphinx-multiversion` doing its thing.

I also plan on opening an issue to update to `pydata-sphinx-theme` v0.6 (how dare they have the same version as our current release! ;-) ).

fairlearn/__init__.py
riedgar-ms: This fix and the changes to `index.html` need to be on `main`, since the docs are only build from `main`. The subsequent PR (#740 ) will get cherry picked, since otherwise the release breaks.

adrinjalali: hmm, ok.

adrinjalali (approved):

fairlearn/__init__.py
riedgar-ms: I'd be open to improved ways of dealing with the versioned docs. I do take your point that this update looks slightly odd.

PR #746, opened 2021-04-16
Remove SECURITY.md file and point to fairlearn-internal mailing list

### LINKED ISSUE #290: SECURITY.md file
It seems the `SECURITY.md` file is from the time where `fairlearn` was under the _microsoft_ org. Now that it's not, does it need to be there?

romanlutz:
closes #290 
related to #433 


adrinjalali:
There are `FutureWarnings` there too. I suggest opening a separate PR to fix those issues first and then getting this one in. We should have crone scheduled CI to test against these new releases of our dependencies. cc @riedgar-ms

fairlearn/reductions/__init__.py
```diff
@@ -18,15 +18,7 @@
 from ._moments import BoundedGroupLoss, SquareLoss, ZeroOneLoss  # noqa: F401
 from ._moments import ClassificationMoment, LossMoment  # noqa: F401
 
-_exponentiated_gradient = [
```
riedgar-ms, line 21: This seems.... unrelated?

romanlutz: Flake8 is requiring this to change. RST902 failure and I can't override it with `noqa` it seems. So I played around with it and this is the only solution I can come up with. This is probably hitting every PR right now (?)

romanlutz:
> There are `FutureWarnings` there too. I suggest opening a separate PR to fix those issues first and then getting this one in. We should have crone scheduled CI to test against these new releases of our dependencies. cc @riedgar-ms

Fair enough. I'll create another PR for the `flake8` fixes. Where would I find the `FutureWarnings`?

fairlearn/reductions/__init__.py
romanlutz: Moving this a different PR

adrinjalali (approved):

MiroDudik (approved):

PR #747, opened 2021-04-16
Fix flake8 warning

romanlutz:
The issue is a newly occurring `flake8` RST902 error based on this file. I couldn't find a different solution than this. Any objections?


riedgar-ms (approved):
I think the previous pattern was a relic from when we were considering some finer grained things here.

romanlutz:
Sadly this PR can't be completed since I managed to create a branch with `.py` at the end which means the `.git` folder has a log file ending in `.py` and `flake8` feels responsible for linting it, which results in errors. Wow! Will create a new branch with identical content and better name. Closing this PR

PR #760, opened 2021-04-30
Add updates to FAQ for multi class classification and multiple/non-binary sensitive features

romanlutz:
While doing this update I also noticed that there was no entry for preprocessing in the corresponding table that I was linking to, so I quickly added that. Obviously, we'll need to add a user guide at some point.

@MiroDudik may want to take a look since he suggested that I create this PR in the first place :-)

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -125,20 +125,19 @@ Which ML libraries does Fairlearn support?
     algorithm can be evaluated using our metrics.
 
 Does Fairlearn support multi-class classification?
-    Fairlearn's assessment capabilities support multi-class classification
-    since the `MetricFrame` does not care about the nature of the ML problem;
-    if the supplied metric function supports multi-class classification,
-    then the `MetricFrame` can use it, producing a result for each of the
-    groups identified by the sensitive feature(s).
-    For example, it is perfectly acceptable to pass
-    `sklearn.metrics.accuracy_score` or `sklearn.metrics.confusion_matrix`
-    as the metric function, and supply multi-class data for y_true and y_pred.
+    On the assessment side, Fairlearn's :class:`MetricFrame` can be used with
+    any metrics for supervised learning, including those for multi-class classification.
+    For example, it is possible to pass
+    :py:func:`sklearn.metrics.accuracy_score` or :py:func:`sklearn.metrics.confusion_matrix`
+    as the metric functions, and supply multi-class data for :code:`y_true` and :code:`y_pred`.
     There are
     `ongoing discussions within the community <https://github.com/fairlearn/fairlearn/issues/752>`_
     to add more extensive support to Fairlearn's assessment capabilities.
     If you have thoughts feel free to add them to the discussion.
-    The existing mitigation techniques do not support multi-class
-    classification.
+    
+    On the mitigation side, our algorithms :class:`ExponentiatedGradient` and :class:`GridSearch`
+    support :ref:`bounded group loss <_bounded_group_loss>` constraints, which are applicable
+    to any supervised learning setting, including multi-class classification.
 
 Does Fairlearn support multiple and non-binary sensitive features?
     Fairlearn's assessment capabilities support sensitive features with more
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -125,13 +125,20 @@ Which ML libraries does Fairlearn support?
     algorithm can be evaluated using our metrics.
 
 Does Fairlearn support multi-class classification?
-    Certain metrics will still work for multi-class classification, e.g.,
-    accuracy score. For others it is more complicated since they rely on
-    binary classes, e.g., true positive rate.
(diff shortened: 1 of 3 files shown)


docs/faq.rst
```diff
@@ -124,10 +124,28 @@ Which ML libraries does Fairlearn support?
     `predict()` methods. Also, any classification or regression
     algorithm can be evaluated using our metrics.
 
+Does Fairlearn support multi-class classification?
```
riedgar-ms, line 127: I think that there should be 'metrics' and 'mitigation' portions to this FAQ.

docs/faq.rst
```diff
@@ -124,10 +124,28 @@ Which ML libraries does Fairlearn support?
...
+    Certain metrics will still work for multi-class classification, e.g.,
+    accuracy score. For others it is more complicated since they rely on
+    binary classes, e.g., true positive rate.
```
riedgar-ms, line 130: How about

> The `MetricFrame` does not care about the nature of the ML problem; if the supplied metric function supports multiclass classification, then the `MetricFrame` can use it, producing a result for each of the groups identified by the sensitive feature(s). For example, it is perfectly acceptable to pass `sklearn.metrics.confusion_matrix` as the metric function, and supply multiclass data for `y_true` and `y_pred`.

docs/faq.rst
```diff
@@ -124,10 +124,28 @@ Which ML libraries does Fairlearn support?
...
+    There are
+    `ongoing discussions within the community <https://github.com/fairlearn/fairlearn/issues/752>`_
+    to add more extensive support to Fairlearn's assessment capabilities.
+    If you have specific thoughts feel free to add them to the discussion.
```
riedgar-ms, line 134: Inviting 'specific thoughts' might make people concerned that their thoughts are insufficiently precise....

docs/faq.rst
```diff
@@ -124,10 +124,28 @@ Which ML libraries does Fairlearn support?
...
+
+Does Fairlearn support multiple and non-binary sensitive features?
+    Fairlearn's assessment capabilities support sensitive features with more
+    than two values as well as multiple sensitive features.
+    Our :ref:`user guide <metrics_with_grouping>` has examples for both of
+    these cases.
+    The mitigation techniques all support mitigation with non-binary and
```
riedgar-ms, line 141: While `GridSearch` can theoretically do this, I believe it runs into dimensional problems quite quickly.

romanlutz: Yes, but I don't want that level of detail in this high-level FAQ. That belongs in the GridSearch API docs or user guide.

docs/faq.rst
romanlutz: I used this, but kept the part about adding thoughts since that is still evolving. Does that work for you?

docs/faq.rst
romanlutz: Indeed.

MiroDudik (approved):

docs/faq.rst
```diff
@@ -124,10 +124,34 @@ Which ML libraries does Fairlearn support?
...
+    any metrics for supervised learning, including those for multi-class classification.
+    For example, it is possible to pass
+    :py:func:`sklearn.metrics.accuracy_score` or :py:func:`sklearn.metrics.confusion_matrix`
+    as the metric functions, and supply multi-class data for :code:`y_true` and :code:`y_pred`.
+    There are
+    `ongoing discussions within the community <https://github.com/fairlearn/fairlearn/issues/752>`_
+    to add more extensive support to Fairlearn's assessment capabilities.
+    If you have thoughts feel free to add them to the discussion.
+    
+    On the mitigation side, our algorithms :class:`ExponentiatedGradient` and :class:`GridSearch`
```
MiroDudik, line 138: We have never tested this, but this should be allowed. I realize that our user guide doesn't say this... we can fix that another time.

riedgar-ms (approved):

MiroDudik:
@romanlutz , @riedgar-ms : for some reason, I don't see the updated FAQ on CircleCI. any idea why?

romanlutz:
> @romanlutz , @riedgar-ms : for some reason, I don't see the updated FAQ on CircleCI, so I'm hesitant to merge. any idea why?

I can't tell either. I re-queued it and will check whether it stays that way.

romanlutz:
> > @romanlutz , @riedgar-ms : for some reason, I don't see the updated FAQ on CircleCI, so I'm hesitant to merge. any idea why?
> 
> I can't tell either. I re-queued it and will check whether it stays that way.

OK, (unsurprisingly) that didn't help. It does seem to check out the branch correctly, though. I just ran the single version sphinx command locally which rendered fine. @riedgar-ms @adrinjalali any ideas?

MiroDudik:
[keeping fingers crossed!]

romanlutz:
@MiroDudik thanks! It does show up now!

image

Issue #728, opened 2021-04-06
ThresholdOptimizer returns different group results when given multiple sensitive features.


LeJit:
#### Describe the bug

The ThresholdOptimizer returns incorrect group performance results when multiple sensitive features are given. If the multiple sensitive feature columns are concatenated into a single column, the results appear more inline with what we intuitively expect.
This bug was found when optimizing for `balanced_accuracy_score` with the fairness constraint of `equalized_odds`.
#### Steps/Code to Reproduce

A script to reproduce the error can be found on the [manandi/scipy_case_study](https://github.com/LeJit/fairlearn/blob/manandi/scipy_case_study/tutorial/scipy/case_study.py) branch.

To reproduce the error, change the `sensitive` variable between the two values specified in lines 23 and 24. 

#### Expected Results

<img width="542" alt="Postprocess_bug" src="https://user-images.githubusercontent.com/1124464/113729775-51cc4f80-96c5-11eb-803f-468249e493e5.png">

When the sensitive features are concatenated together, the `ThresholdOptimizer` yields results that mitigate disparity across sensitive attributes that align with intuition.

#### Actual Results

<img width="542" alt="Postprocess_separate" src="https://user-images.githubusercontent.com/1124464/113729777-51cc4f80-96c5-11eb-8e78-c56a9ab8803c.png">

Within racial groups, we see the ThresholdOptimizer achieves a false positive rate of 0 and false negative rate of 1 for multiple demographic groups. This result is unintuitive and leads to large between-group disparities.

#### Versions

OS: Windows
Python Version: 3.8.5
Fairlearn version: 0.6.0

michaelamoako:
I had a similar issue with other combinations of optimization metric and constraints: 

I think because the constraints are related to parity (and not upperbounds) these types of solutions are **possible**. Commenting to track if progress is made on this, as I have the same question

romanlutz:
I've managed to trace this to `_merge_columns` in `_input_validation.py`. Evidently, if you use `np.apply_along_axis` to merge multiple text labels it cuts them off. Example:
```
>>> np.apply_along_axis(lambda row: ",".join([str(i) for i in row]), axis=1, arr=np.array([["abc", "def"], ["abdddddddddddddddddddddddddddddddd", "danfaenofeinfaoieraj"], ["a", "b"]]))
array(['abc,def', 'abddddd', 'a,b'], dtype='<U7')
```
[...]

PR #763, opened 2021-05-01
Fix _merge_columns when using multiple sensitive features with long names

romanlutz:
Fairlearn supports scenarios with multiple sensitive features (e.g., gender and age, but it can be more than two). Internally, the mitigation techniques currently treat this no different than a single sensitive feature, so there's a column merge step called `_merge_columns()` that happens before applying mitigations.
 
During that step, `numpy.apply_along_axis` has so far been used to merge the string representations of the sensitive features. However, this doesn't return the expected results when the names are long since it cuts them off. For example, @LeJit discovered in #728 that the results from `ThresholdOptimizer` with two sensitive features did not match the results when manually combining the two columns. Since we don't want to restrict users in sensitive feature naming this is not acceptable.

The solution in this PR is to use `pandas.DataFrame.apply` instead of `numpy.apply_along_axis` since it doesn't exhibit the same restrictions. To prevent us from hitting the same issue in the future I've also added a test case which the previous implementation would have failed.

I also suggest we release v0.6.2 once this is in since it might actually affect users (in addition to @LeJit ).

Huge kudos to @LeJit for the excellent repro in #728 . It is a rare pleasure for an engineer to get a simple repro including the specified expectations. :-)


test/unit/metrics/test_metricframe_smoke.py
```diff
@@ -270,7 +269,6 @@ def multi_sp(y_t, y_p, p1, p2):
 
 def test_2m_1sf_sample_weights():
     """Check that sample weights are passed correctly to two metrics."""
-
```
riedgar-ms, line 273: ?

test/unit/postprocessing/test_threshold_optimizer_multiple_sensitive_features.py
```diff
@@ -0,0 +1,124 @@
...
+    # The bug caused the merged names to get cut off, resulting in multiple groups
+    # getting merged internally. To avoid that this test case checks even internal
+    # representations.
+    X = pd.DataFrame([
+        [0, 4], [6, 2], [1, 3], [10, 5], [1, 7], [-2, 1], [3, 10], [14, 5],
+        [1, 3], [1, 5], [1, 7], [-5, 9], [3, 13], [7, 1], [-8, 4], [9, 1]])
+    y = pd.Series([0, 1, 1, 0, 0, 0, 1, 0, 0, 1, 0, 0, 1, 1, 1, 0])
+    a1 = "a"
+    a2 = "a very very very very very very very long group name"
+    a3 = "a group name with commas ,, in , it"
```
riedgar-ms, line 39: Would it be worth having one with a trailing backslash too?

riedgar-ms (changes requested):

test/unit/metrics/test_metricframe_smoke.py
romanlutz: linting issue, now resolved

test/unit/postprocessing/test_threshold_optimizer_multiple_sensitive_features.py
romanlutz: A single one will escape the following characters. What do you want to escape?

riedgar-ms: I meant have the embedded string have the trailing backslash (i.e. put in two), just to make sure that nothing then subsequently used that backslash to escape something.

romanlutz: I added backslashes to the test case. I'm not entirely sure whether you meant something in addition to that. If so, can you provide an example?

riedgar-ms: Looks good

riedgar-ms (approved):

MiroDudik (approved):

PR #766, opened 2021-05-03
Remove FairlearnDashboard and replace it with matplotlib-based plots

romanlutz:
With a slight delay (originally targeted for April) I'm finally removing the `FairlearnDashboard` since a newer version already exists in `raiwidgets`. The documentation is updated to instead use the plots @MiroDudik created with a single line directly from the `MetricFrame`. In the future we want to add more kinds of plots as already mentioned in #758 #666 and #668 . Specifically, the model comparison plots do not yet have a replacement yet.

Note that the "example" added to the `examples` directory is not shown under "Example notebooks" on the webpage, which is intentional since it's technically not a notebook.

This also makes #561 mostly redundant, which I'll close shortly. 
#667 is also directly addressed with this PR as the examples illustrate.

PR #767, opened 2021-05-04
Remove bokeh plot and replace with simple matplotlib plot

romanlutz:
Our doc builds are taking a while, so it seemed like a good idea to replace the custom bokeh plot with a simpler matplotlib plot just like in #766 . Sadly, we are locked into all dependencies we ever had unless we're willing to go into older versions of Fairlearn, remove them from there, and then move the tag. Still, this simpler plotting code seemed like a step in the right direction.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -60,6 +60,7 @@ def check_if_v046():
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
 extensions = [
+    'bokeh.sphinxext.bokeh_plot',
     'sphinx.ext.autodoc',
     'sphinx.ext.doctest',
     'sphinx.ext.extlinks',
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -60,7 +60,6 @@ def check_if_v046():
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
 extensions = [
-    'bokeh.sphinxext.bokeh_plot',
     'sphinx.ext.autodoc',
     'sphinx.ext.doctest',
     'sphinx.ext.extlinks',
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -87,8 +87,9 @@ than $50,000 a year.
     Female    16192
     Name: sex, dtype: int64
 
-.. bokeh-plot:: quickstart_plot.py
-    :source-position: none
+.. figure:: ../auto_examples/images/sphx_glr_plot_quickstart_selection_rate_001.png
+    :target: auto_examples/plot_quickstart_selection_rate.html
+    :align: center
 
 Evaluating fairness-related metrics
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -87,7 +87,7 @@ than $50,000 a year.
     Female    16192
     Name: sex, dtype: int64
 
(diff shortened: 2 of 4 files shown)


adrinjalali:
> Sadly, we are locked into all dependencies we ever had unless we're willing to go into older versions of Fairlearn, remove them from there, and then move the tag.

I'm not sure what you mean here @romanlutz

adrinjalali (approved):

romanlutz:
> > Sadly, we are locked into all dependencies we ever had unless we're willing to go into older versions of Fairlearn, remove them from there, and then move the tag.
> 
> I'm not sure what you mean here @romanlutz

Let me explain what I initially thought, and then my correction to what I wrote:
@riedgar-ms encountered the same issue before with `tempeh`, a package we used in one of the earlier versions of our documentation. Even removing it now isn't quite possible because with sphinx-multiversion we're building all the past versions of the package into the doc, too. In a way, that nice feature comes back to haunt us there because we're basically locked into every dependency we take forever (or rather: as long as we support that version in our docs). Alternatively, one could go back to the past tagged versions, remove that dependency (and replace it or remove it) and then retag, but that's obviously work.

What I realized just now is that `tempeh` isn't in our `requirements-dev.txt` anymore! @riedgar-ms seems to have managed to just put it into our doc build script as an extra dependency. I'll try the same with `bokeh` now and push an update to this, or explain why it didn't work.

romanlutz:
Sadly, I was right with the current version of the PR. I cannot remove `bokeh` completely because the `bokeh-plot` directive is part of `sphinx`'s extensions that are needed to build previous versions, even though it's not part of the current version. That is mildly annoying... I suppose I could go back to previous releases and remove it there also, then re-tag, but that's not as easily doable.

In summary, I don't have further changes to this and would like to merge as is (or rather: someone else should merge it) unless there are concerns. Definitely want to wait to see if @adrinjalali or @riedgar-ms have more input since you've both worked with multiversion in some capacity (and certainly more than me).

adrinjalali:
> Alternatively, one could go back to the past tagged versions, remove that dependency (and replace it or remove it) and then retag

we should definitely NOT do that lol

Regarding sphinx-multiversion, I think we should replace it with a mechanism that builds the different versions completely independently of one another, and that's what we do in sklearn. Each version is completely sandboxed and the dependencies are not leaked into the other ones. We can build the docs for different versions independently, and then put them under different folders on the website, and have those links linking to different versoins.

romanlutz:
@adrinjalali That's interesting. I think the difference is that scikit-learn doesn't use `sphinx-multiversion` and I just somehow assumed it did. I will admit the scikit-learn way of building documentation is scary and looks like it'll take many hours to understand properly.

@riedgar-ms do you think there's a point in filing an issue with `sphinx-multiversion` or is this just fundamentally a part of how it works?

Regardless of the answer to either of these questions, shall I open an issue in our repo about this point in particular (getting multi-version builds done without having to rely on old dependencies) and we discuss it further there? I think this PR has reached its intended scope and these are just discussions beyond that. Wdyt?

riedgar-ms:
The scikit-learn doc build system looking very scary was why I was happy to come across `sphinx-multiversion`. I did ask about these things last summer, and I was told that `sphinx-multiversion` is supposed to build all versions of the documentation with the latest configuration. Which makes things like this inevitable (easy to add dependencies, but take them away, and you have to stop building that doc version); it's also the reason the `conf.py` has special code to detect v0.4.6 being built.

Opening an issue to review the versioned documentation build does seem wise. A related issue would be only building 'new' documentation each time, rather than always building every version (particularly if we're going to build each completely independently).

riedgar-ms (approved):

PR #774, opened 2021-05-07
Release adjustments for v0.6.2

romanlutz:
This PR updates the versions for the release branch. There will be a separate PR for the `main` branch.

PR #775, opened 2021-05-07
Post-v0.6.2 release adjustments

romanlutz:
Other than the adjustments in #774 this puts the version to `0.7.0.dev0` and adds the missing release instructions.

I expect the next release will require a new minor version since I have #766 out which should be completed shortly. We can certainly change that if this assumption turns out to be wrong, though.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 # Changes
 
+### v0.7.0
+
 ### v0.6.2
 
 * Bugfix for `_merge_columns()` when using multiple sensitive features with
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Website: https://fairlearn.github.io/
 
 ## Current release
 
-- The current stable release is available at [Fairlearn v0.6.1](https://github.com/fairlearn/fairlearn/tree/release/v0.6.1).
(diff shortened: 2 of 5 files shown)


riedgar-ms (approved):

adrinjalali (approved):

romanlutz:
I should have been clearer about this, but these sorts of PRs should only be completed after the release is complete. FYI @fairlearn/fairlearn-maintainers 
This killed our webpage :-(

hildeweerts:
@romanlutz could you explain a bit more why this PR shouldn't have been completed? I want to make sure I don't break stuff in the future :D

romanlutz:
@hildeweerts great question! I've tried to explain a bit further in the new release instructions in #794 . The TL;DR is that merging any PR triggers a new doc build that pushes the built webpage to the webpage repo fairlearn/fairlearn.github.io . That's intentional, of course! However, the versioned doc build only picks up already released versions. So by increasing the version in the links of our index.html we effectively pointed to pages that couldn't exist yet since the release wasn't out. Instead, we need to release to pypi first, tag the release, then merge the post-v<x.y.z> adjustments (like this PR) which triggers the build and generates the new webpage including the pages for v<x.y.z>. Does that make sense?

hildeweerts:
Thanks for the explanation @romanlutz! That makes a lot of sense (I can't promise that I will never mistakingly merge something in the future though, lol)

Issue #769, opened 2021-05-05
DOC: Measurement modeling - construct validity


mmadaio:
#### Describe the issue linked to the documentation!
The user guide currently does not include explanations of key concepts from measurement modeling. 

Measurement modeling is a framework from quantitative social sciences, involving a process of measuring theoretical constructs that cannot be observed directly (such as fairness) by making inferences from observable properties.

Because fairness itself is unobservable (and is fundamentally contested), users of Fairlearn need to make decisions about what their theoretical understanding of fairness is, and how they operationalize it via observable properties. At each stage, assumptions are made, which need to be tested. 

Measurement modeling provides a language and a framework for making those assumptions explicit and testing them.

As many users of Fairlearn may not have quantitative social science backgrounds, we need to provide explanations and examples of the process of measurement modeling, including key concepts such as construct validity and construct reliability. This issue is focused on construct validity (to keep the issues smaller and more achievable), but others should include construct reliability.

See [Jacobs and Wallach (2021)](https://arxiv.org/pdf/1912.05511.pdf) for more information.

#### Suggest a potential alternative/fix

Create a glossary explaining key terms from construct validity, with explanations of what they might mean for measuring fairness. Specifically:

- face validity
- content validity
  - contestedness
  - substantive validity
  - structural validity
- convergent validity
- discriminant validity
- predictive validity
- hypothesis validity
- consequential validity

shimst3r:
Measurement and Fairness is still on my desk (and I wouldn't mind reading it again 🥰), do you already have a list of key terms in mind? I'd be interested in collaborating on this issue, but I lack the STS background (at the moment).

romanlutz:
@michaelamoako just created #771 today as well. Would be great if the three of you (@shimst3r , @mmadaio , @michaelamoako ) could outline your ideas here in detail to decide who can do which part :-)
For good measure I'll also tag @hannawallach in case she has any thoughts.

romanlutz:
I just realized #707 also discussed this. @hildeweerts @LisaIbanez @riedgar-ms and @LeJit participated there, too, and may have thoughts, too :-)

michaelamoako:
@romanlutz could you changed linked PR to #787

mmadaio:
> Measurement and Fairness is still on my desk (and I wouldn't mind reading it again 🥰), do you already have a list of key terms in mind? I'd be interested in collaborating on this issue, but I lack the STS background (at the moment).

[...]

PR #787, opened 2021-05-07
Construct Validity Glossary (main) 

michaelamoako:
A writeup summarizing the key terms related to Construct Validity, as educational material. This is to be leveraged by users of Fairlearn, and can be referenced in other materials.

Note: Pull request 771 was incorrectly pointing to master. This points to main.


docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -39,6 +39,79 @@ The Fairlearn package is most applicable to two kinds of harms:
...
+- **Socioeconomic status** is an example of an unobservable theoretical construct. A common measurement model includes annual household income. 
+
+- **Patient benefit** is an example of an unobservable theoretical construct. A common measurement model involves patient care costs.
+
+**Note:** We cite several examples of unobservable theoretical constructs and measurement models for the purpose of explaining the key terms outlined above.  Please reference Measurement and Fairness for more detailed examples.
+ 
+What is Construct Validity? 
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Though the Measurement and Fairness paper explores both Construct reliability and Construct validity, we focus our
```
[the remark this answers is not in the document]
romanlutz: I should add: below in the list we can capitalize the first letter since it's at the start of new bullets, of course.

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -39,6 +39,79 @@ The Fairlearn package is most applicable to two kinds of harms:
...
+
+- **Teacher effectiveness** is an example of an unobservable theoretical construct. Common measurement models include student performance on standardized exams and qualitative feedback for the teacher’s students.
+
+- **Socioeconomic status** is an example of an unobservable theoretical construct. A common measurement model includes annual household income. 
+
+- **Patient benefit** is an example of an unobservable theoretical construct. A common measurement model involves patient care costs.
+
+**Note:** We cite several examples of unobservable theoretical constructs and measurement models for the purpose of explaining the key terms outlined above.  Please reference Measurement and Fairness for more detailed examples.
+ 
+What is Construct Validity?
```
romanlutz, line 78: More of a high-level question: Should this live on this page, or on a sub-page dedicated to this topic? If it stays on this page, is this the place where it should go, or rather at the end of what already exists? @fairlearn/fairlearn-maintainers

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -39,6 +39,79 @@ The Fairlearn package is most applicable to two kinds of harms:
...
+   measurement models for which construct validity has been established? 
+ 
+4. **Discriminant validity** – Do the measurements obtained for the construct of interest correlate with 
+   related constructs as appropriate?  
+
+5. **Predictive validity** – Are the measurements obtained from the measurement model predictive of measurements 
+   of any relevant observable properties or other unobservable theoretical constructs?
+
+6. **Hypothesis validity** – This describes the nature of the hypotheses that could emerge from the measurements 
+   produced by the measurement model, and whether those are “substantively interesting”.
```
romanlutz, line 105: I have seen the sphinx build struggle with these quotes you're using in this line. Can we replace them with the basic ones `"` that you're using in other parts (like in the citation, for example)?

romanlutz (commented):
This PR totally made my day! I'm super excited that we're adding this to the webpage. What a fantastic step forward for the project :-) Thanks a ton, @michaelamoako ! As you can tell, I left a bunch of comments. Let me know if you have questions about them or need help with anything.

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -39,6 +39,79 @@ The Fairlearn package is most applicable to two kinds of harms:
...
+
+- **Sociotechnical context** – A context in which both societal and technical aspects shape who might be harmed by AI systems and how (see fairlearn.ai for more detailed definition).
+
+- **Unobservable theoretical construct** – An idea or concept that is unobservable and cannot be directly measured but must instead be inferred through observable measurements defined in a measurement model. 
+
+- **Measurement model** – The method and approach used to measure the unobservable theoretical construct.
+
+- **Construct reliability** – This is closely related to construct validity (detailed below) and can be thought of as the extent to which the measurements of an unobservable theoretical construct remain the same when measured at different points in time. A lack of construct reliability can either be due to a misalignment between the understanding of the unobservable theoretical construct and the methods being used to measure that construct, or to changes to the construct itself.
+
+- **Construct validity** – This can be thought of as the extent to which the measurement model can be used to produce meaningful and useful results.
```
hildeweerts, line 63: I miss a reference to the construct in this definition. I personally would go for something like "The extent to which the measurement model measures the intended construct." If you really want the meaningful/useful you might add "in way that is meaningful and useful".

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -39,6 +39,79 @@ The Fairlearn package is most applicable to two kinds of harms:
...
 
+Concept Glossary
+----------------------------
+
+The concepts outlined in this glossary are relevant to sociotechnical contexts. 
+
+Construct Validity
+^^^^^^^^^^^^^^^^^^
+
+We explore the concept of construct validity, as described in in Measurement and Fairness [#4]_ , and as it relates to sociotechnical contexts. This in intended to illuminate the concept to readers who are working on solutions in sociotechnical contexts and may also be referenced in explorations that detail how to apply the concept to specific scenarios.
```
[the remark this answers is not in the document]
hildeweerts: I agree, I think it sounds a bit businessy. I think throughout the website we usually refer to something like "AI systems", "ML systems".

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -39,6 +39,79 @@ The Fairlearn package is most applicable to two kinds of harms:
...
+
+Key Terms 
+~~~~~~~~~
+
+- **Sociotechnical context** – A context in which both societal and technical aspects shape who might be harmed by AI systems and how (see fairlearn.ai for more detailed definition).
```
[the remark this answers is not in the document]
hildeweerts: I feel this definition makes it look like there's a specific subsets of systems in which there is a sociotechnical context, which might be a bit misleading. What about something like: "The context surrounding a technical system, including both social aspects (e.g., people, institutions, communities) and technical aspects (e.g., algorithms, technical processes). The sociotechnical context of a system shapes who might benefit or is harmed by AI systems."

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -39,6 +39,79 @@ The Fairlearn package is most applicable to two kinds of harms:
...
+
+- **Unobservable theoretical construct** – An idea or concept that is unobservable and cannot be directly measured but must instead be inferred through observable measurements defined in a measurement model.
```
hildeweerts, line 57: I am not sure whether people unfamiliar with the idea of theoretical constructs would understand this definition. I am also not a big fan of referring to concepts you have not introduced yet. Maybe something like: "An abstract idea or concept that is unobservable and cannot be measured directly." I think it would also help if we add the key term examples **directly** under the associated definition.

hildeweerts, line 57: Would it make sense to first make an entry with "construct"? I think some of our users will be unfamiliar with those. At the very least I would definitely add "observable construct", to distinguish between them. I also think it is important to stress that unobservable theoretical constructs are truly unobservable; it's not due to practical constraints. E.g., in theory we might be able to measure alcohol consumption of drivers, but in practice we can't just add a measurement device in every car and force people to use it.

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -39,6 +39,79 @@ The Fairlearn package is most applicable to two kinds of harms:
...
+Key Terms 
+~~~~~~~~~
+
+- **Sociotechnical context** – A context in which both societal and technical aspects shape who might be harmed by AI systems and how (see fairlearn.ai for more detailed definition).
+
+- **Unobservable theoretical construct** – An idea or concept that is unobservable and cannot be directly measured but must instead be inferred through observable measurements defined in a measurement model. 
+
+- **Measurement model** – The method and approach used to measure the unobservable theoretical construct.
+
+- **Construct reliability** – This is closely related to construct validity (detailed below) and can be thought of as the extent to which the measurements of an unobservable theoretical construct remain the same when measured at different points in time. A lack of construct reliability can either be due to a misalignment between the understanding of the unobservable theoretical construct and the methods being used to measure that construct, or to changes to the construct itself.
```
hildeweerts, line 61: I would remove "This is closely related to construct validity (detailed below) and can be thought of as" from the starting sentence. If you really want to refer to construct validity here, you could still add it to the end: "Construct reliability is closely related to construct validity." Not sure whether it is necessary though, unless we make the relationship explicit (i.e., they're complementary)

docs/user_guide/fairness_in_machine_learning.rst
hildeweerts, line 81: I don't think its necessary to explain in detail why we defer from the paper. Instead, I'd suggest to re-write this paragraph to highlight in what way construct validity is particularly relevant to fairness related harms.

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -39,6 +39,79 @@ The Fairlearn package is most applicable to two kinds of harms:
...
+What is Construct Validity? 
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Though the Measurement and Fairness paper explores both Construct reliability and Construct validity, we focus our 
+exploration below on Construct Validity. We note that both play an important role in understanding fairness in sociotechnical 
+contexts. With that said, Measurement and Fairness offers a fairness-oriented conceptualization of construct validity, that 
+is helpful in thinking about fairness in sociotechnical contexts. We capture the idea in seven key parts that when combined 
+can serve as a framework for analyzing an AI task and attempting to establish construct validity:
+
+1. **Face validity** – On the surface, how plausible does the measurements produced by the measurement model look?
```
hildeweerts, line 87: Please add an example for each component of construct validity.

hildeweerts (commented):
Overall, I feel like the language that is used here is often a bit formal, compared to the rest of our website. As our audience consists of practitioners from various backgrounds I would personally prefer more casual language.

docs/user_guide/fairness_in_machine_learning.rst
michaelamoako: Currently I have: We note that both play an important role in understanding fairness in sociotechnical 
contexts

Any suggestions on what a more detailed explanation of how construct validity is particularly relevant to fairness related harms? 

Having trouble doing this succinctly (since the hope was that the glossary itself would make the relationship clear)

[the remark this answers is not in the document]
michaelamoako: @romanlutz  how about: 
In lists we capitalize the first letter
In titles of sections (i.e: What is Construct Validity) we use camel case 
Within sections we lowercase

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -39,6 +39,79 @@ The Fairlearn package is most applicable to two kinds of harms:
...
+Key Term Examples  - Unobservable theoretical constructs and Measurement models
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- **Fairness** is an example of an unobservable theoretical construct. Several measurement models exist for measuring fairness, including demographic parity. See Fairlearn Metrics for more examples of measurement models for measuring fairness. 
+
+- **Teacher effectiveness** is an example of an unobservable theoretical construct. Common measurement models include student performance on standardized exams and qualitative feedback for the teacher’s students.
+
+- **Socioeconomic status** is an example of an unobservable theoretical construct. A common measurement model includes annual household income. 
+
+- **Patient benefit** is an example of an unobservable theoretical construct. A common measurement model involves patient care costs.
```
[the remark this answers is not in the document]
michaelamoako: That specific example is not referenced in this paper - is the suggestion to plug it in as related work?

If so, how do you see if fitting in (i.e: "See ____ for a related example")?

docs/user_guide/fairness_in_machine_learning.rst
michaelamoako: Which of the following should we add @hildeweerts: 
- construct
- observable construct
- observable theoretical construct 

And how do you suggest we define it? 

Also, how might we make the suggested emphasis in the example above (could you illustrate by taking the current definition and appending a suggestion)?

docs/user_guide/fairness_in_machine_learning.rst
michaelamoako: @hildeweerts how about: 

**This can be thought of as the extent to which the measurements of an unobservable theoretical construct remain the same when measured at different points in time.** A lack of construct reliability can either be due to a misalignment between the understanding of the unobservable theoretical construct and the methods being used to measure that construct, or to changes to the construct itself. **Construct validity and construct reliability as complementary.**

docs/user_guide/fairness_in_machine_learning.rst
michaelamoako: Part of the thinking was that this glossary would be used and referenced by other contributors, who could then utilize the terms above to create examples. Curious to hear other's thoughts on if in addition to the pending educational material that will use the above in specific example, we need to add example here too? 

@mmadaio
@hannawallach  
Any thoughts?

docs/user_guide/fairness_in_machine_learning.rst
michaelamoako: @hannawallach @mmadeo also curious to hear your thoughts on the above suggestion and how we might highlight in what what construct validity is particularly relevant to fairness related harms.

docs/user_guide/fairness_in_machine_learning.rst
michaelamoako: Given the current structure of the examples (where showcase both unobservable theoretical constructs and an measurement models within the same example), how might you suggest restructuring this show that the key term examples appear **directly** under the associated definition. This was my initial structure, but I found that the key term examples were easier to follow if both unobserved tc and measurement model were combined into one example. 

If you think having it directly under would be clearer, could you illustrate how we'd do this

michaelamoako:
> Overall, I feel like the language that is used here is often a bit formal, compared to the rest of our website. As our audience consists of practitioners from various backgrounds I would personally prefer more casual language.

@hildeweerts which part of the writeup were your referring to here? If you could point to specific parts and offer suggestions/replacements that would be helpful!

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -39,6 +39,79 @@ The Fairlearn package is most applicable to two kinds of harms:
...
+- **Measurement model** – The method and approach used to measure the unobservable theoretical construct.
+
+- **Construct reliability** – This is closely related to construct validity (detailed below) and can be thought of as the extent to which the measurements of an unobservable theoretical construct remain the same when measured at different points in time. A lack of construct reliability can either be due to a misalignment between the understanding of the unobservable theoretical construct and the methods being used to measure that construct, or to changes to the construct itself.
+
+- **Construct validity** – This can be thought of as the extent to which the measurement model can be used to produce meaningful and useful results.
+
+Key Term Examples  - Unobservable theoretical constructs and Measurement models
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- **Fairness** is an example of an unobservable theoretical construct. Several measurement models exist for measuring fairness, including demographic parity. See Fairlearn Metrics for more examples of measurement models for measuring fairness.
```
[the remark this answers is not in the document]
michaelamoako: @romanlutz G Here I was referring to the metrics in fairlearn.ai (not another section in this writeup), in api_reference/fairlearn.metrics.html

What I am trying to convey is that the metrics in the Fairlearn package are themselves examples of measurement models or measuring fairness. 

With that clarification, how do you suggest I modify the above?

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -39,6 +39,79 @@ The Fairlearn package is most applicable to two kinds of harms:
   information are extended or withheld. Examples include varying accuracy in
   face recognition, document search, or product recommendation.
 
+Concept Glossary
+----------------------------
+
+The concepts outlined in this glossary are relevant to sociotechnical contexts.
```
mmadaio, line 45: Hmm, I'd say it's less about understanding the sociotechnical context than about something like "relevant to help data scientists test assumptions made when measuring unobservable phenomena such as fairness." Or something along those lines

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -39,6 +39,79 @@ The Fairlearn package is most applicable to two kinds of harms:
...
 
+Concept Glossary
+----------------------------
+
+The concepts outlined in this glossary are relevant to sociotechnical contexts. 
+
+Construct Validity
+^^^^^^^^^^^^^^^^^^
+
+We explore the concept of construct validity, as described in in Measurement and Fairness [#4]_ , and as it relates to sociotechnical contexts. This is intended to illuminate the concept to readers who are trying to build AI systems in sociotechnical contexts and may also be referenced in explorations that detail how to apply the concept to specific scenarios.
```
mmadaio, line 50: I think we can make clear how this glossary relates to fairness in context! Maybe revise to something along the lines of:

"Readers working on measuring fairness are making assumptions and choices about what fairness means for their use case and context. This glossary is intended to provide readers with language and a framework for testing the validity of those assumptions in the sociotechnical context they are working in. Ideally, this glossary may be referenced by extended examples that detail how to apply the concepts to specific scenarios."

docs/user_guide/fairness_in_machine_learning.rst
mmadaio: I'd also be in favor of at least introducing the examples after each one: something like "(e.g, teacher effectiveness, socioeconomic status, or fairness)". You can always call back to those later with the specific measurements, but I think that would help to ground it a bit!

mmadaio: Also agreed that we may want to drop the "defined in a measurement model" since that gets introduced later

mmadaio: Hmm, @hildeweerts, I'm not sure I've seen the term "observable construct" used before. To my understanding, almost by definition, constructs are abstractions. What about something like:

"An abstract idea or concept of interest which is not possible to direct measure, but which must instead be inferred through observable measurements.

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -39,6 +39,79 @@ The Fairlearn package is most applicable to two kinds of harms:
...
+
+Key Terms 
+~~~~~~~~~
+
+- **Sociotechnical context** – The context surrounding a technical system, including both social aspects (e.g., people, institutions, communities) and technical aspects (e.g., algorithms, technical processes). The sociotechnical context of a system shapes who might benefit or is harmed by AI systems.
+
+- **Unobservable theoretical construct** – An idea or concept that is unobservable and cannot be directly measured but must instead be inferred through observable measurements defined in a measurement model. 
+
+- **Measurement model** – The method and approach used to measure the unobservable theoretical construct.
```
mmadaio, line 59: It may be good to add "via inferences from observable properties" or something like that from the paper. Since that connection between the unobservable construct and the observable properties is what they'll be testing for later.

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -39,6 +39,79 @@ The Fairlearn package is most applicable to two kinds of harms:
...
+
+- **Construct reliability** – This is closely related to construct validity (detailed below) and can be thought of as the extent to which the measurements of an unobservable theoretical construct remain the same when measured at different points in time. A lack of construct reliability can either be due to a misalignment between the understanding of the unobservable theoretical construct and the methods being used to measure that construct, or to changes to the construct itself.
+
+- **Construct validity** – This can be thought of as the extent to which the measurement model measures the intended construct in way that is meaningful and useful.
```
mmadaio, line 63: Maybe add here that this has several different aspects, which you'll explain more below?

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -39,6 +39,79 @@ The Fairlearn package is most applicable to two kinds of harms:
...
+
+Key Term Examples  - Unobservable theoretical constructs and Measurement models
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+- **Fairness** is an example of an unobservable theoretical construct. Several measurement models exist for measuring fairness, including demographic parity. See Fairlearn Metrics for more examples of measurement models for measuring fairness.
```
mmadaio, line 68: Hmm, I guess I see the term "measurement model" to refer to the process of moving from abstract ideas to concrete, mathematical measurements. And then fairness metrics like demographic parity are one type of measurement, but the modeling process is what gets us (or the readers) from the abstract idea of fairness to that metric.

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -39,6 +39,79 @@ The Fairlearn package is most applicable to two kinds of harms:
...
+
+- **Fairness** is an example of an unobservable theoretical construct. Several measurement models exist for measuring fairness, including demographic parity. See Fairlearn Metrics for more examples of measurement models for measuring fairness. 
+
+- **Teacher effectiveness** is an example of an unobservable theoretical construct. Common measurement models include student performance on standardized exams and qualitative feedback for the teacher’s students.
+
+- **Socioeconomic status** is an example of an unobservable theoretical construct. A common measurement model includes annual household income. 
+
+- **Patient benefit** is an example of an unobservable theoretical construct. A common measurement model involves patient care costs.
+
+**Note:** We cite several examples of unobservable theoretical constructs and measurement models for the purpose of explaining the key terms outlined above.  Please reference Measurement and Fairness [#4]_ for more detailed examples.
```
mmadaio, line 76: And maybe we can add a plug for contributions here? e.g., "...and other contributors are encouraged to contribute additional examples demonstrating these concepts in more detail."

docs/user_guide/fairness_in_machine_learning.rst
mmadaio: What would be the rationale for moving it outside the user guide? Because there isn't currently functionality to create tests for these, for instance?

I could see an argument to it being in Section 5 on further resources (under a subheader on e.g., "Related Fairness Concepts" or something)

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -39,6 +39,79 @@ The Fairlearn package is most applicable to two kinds of harms:
...
+What is Construct Validity? 
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Though the Measurement and Fairness [#4]_ paper explores both Construct reliability and Construct validity, we focus our 
+exploration below on Construct Validity. We note that both play an important role in understanding fairness in sociotechnical 
+contexts. With that said, Measurement and Fairness [#4]_ offers a fairness-oriented conceptualization of construct validity, that 
+is helpful in thinking about fairness in sociotechnical contexts. We capture the idea in seven key parts that when combined 
+can serve as a framework for analyzing an AI task and attempting to establish construct validity:
+
+1. **Face validity** – On the surface, how plausible do the measurements produced by the measurement model look?
```
mmadaio, line 87: These are great questions! I wonder if we can add another sentence of elaboration for each of the questions about what they mean, maybe using one of the examples you mention above?

But I could also see that being done in a PR from someone else too, pending getting these up on the site!

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -39,6 +39,79 @@ The Fairlearn package is most applicable to two kinds of harms:
...
+
+2. **Content validity** – This has three subcomponents:
+
+   a. **Contestedness** – Is there a single understanding of the unobservable theoretical construct? Or is that understanding contested (and thus context        dependent).
```
mmadaio, line 91: This makes sense to me, but I wonder if it would be clear to a reader why being contested makes it inherently context-dependent? Maybe just a little bit of elaboration would be helpful!

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -39,6 +39,79 @@ The Fairlearn package is most applicable to two kinds of harms:
...
+   of any relevant observable properties or other unobservable theoretical constructs?
+
+6. **Hypothesis validity** – This describes the nature of the hypotheses that could emerge from the measurements 
+   produced by the measurement model, and whether those are “substantively interesting”.
+
+7. **Consequential validity** – Identify and evaluate the consequences and societal impacts of using the 
+   measurements obtained for the measurement model. Framed as questions: how is the world shaped by using the 
+   measurements, and what world do we wish to live in?
+
+**Note:** The order in which the parts above are explored may vary depending on the specific
```
mmadaio, line 111: The order - and the way they're explored too!

mmadaio (commented):
@michaelamoako this is a great start! Huge thank you for moving this forward! See specific comments and suggestions in-line.

docs/user_guide/fairness_in_machine_learning.rst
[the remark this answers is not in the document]
mmadaio: My personal stylistic preference would be for lower case for everything, except the paper title (and then maybe use quotes or a link?).

mmadaio: Agreed @hildeweerts that we probably don't need to say why we differ from the paper, and also since there may be other PRs later addressing construct reliability.

mmadaio: Feel free to use some of the language from the [issue](https://github.com/fairlearn/fairlearn/issues/769), e.g.,:

"Because fairness itself is unobservable (and is fundamentally contested), users of Fairlearn need to make decisions about what their theoretical understanding of fairness is, and how they operationalize that abstract concept via measurements of observable properties. At each stage, assumptions are made, which need to be tested.

Measurement modeling provides a language and a framework for making those assumptions explicit and testing them. Here, we explain some of the concepts that can be helpful in that process."

docs/user_guide/fairness_in_machine_learning.rst
mmadaio: Ah, I had a similar comment here too - I'm sort of of two minds about it. I definitely agree that examples are going to be critical to get these concepts across. But that could also be done in-depth in future PRs. I wonder whether a short (e.g., 1-2 sentence) example using the examples mentioned earlier (e.g., teacher effectiveness, etc) might be a useful starting point? 

I could see an argument to either use fairness as the running example (since... it is Fairlearn after all :), or for _not_ using fairness as the running example, since it might be too close to home for readers, and they'll be thinking about their own context etc, and not the concept itself.

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -39,6 +39,79 @@ The Fairlearn package is most applicable to two kinds of harms:
...
+- **Socioeconomic status** is an example of an unobservable theoretical construct. A common measurement model includes annual household income. 
+
+- **Patient benefit** is an example of an unobservable theoretical construct. A common measurement model involves patient care costs.
+
+**Note:** We cite several examples of unobservable theoretical constructs and measurement models for the purpose of explaining the key terms outlined above.  Please reference Measurement and Fairness [#4]_ for more detailed examples.
+ 
+What is Construct Validity? 
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Though the Measurement and Fairness [#4]_ paper explores both Construct reliability and Construct validity, we focus our
```
hildeweerts, line 81: Did I accidentally delete my previous comment? I thought I had already commented on this but I can't seem to find it. 

I think it's not really necessary to explain why we divert from the paper. This paragraph sounds a bit like we're focussing on construct validity because there is a good list of questions available, whereas I think the actual reason is that in an ML context more fairness related issues arise from construct validity compared to reliability. 

I am also not sure whether we should write out "Measurement and Fairness does X..." but that's more of a stylistic choice which you can feel free to ignore :)

docs/user_guide/fairness_in_machine_learning.rst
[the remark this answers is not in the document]
romanlutz: You can reference anything with sphinx 😄 I think :py:mod:`fairlearn.metrics` should do the trick.

docs/user_guide/fairness_in_machine_learning.rst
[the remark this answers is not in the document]
romanlutz: True. I don't see us bound by the scope of any particular paper, though. The more concrete we can make things the easier it becomes to understand why something is important IMO. 

Your wording with "See ____ for a related example" is fine by me.

docs/user_guide/fairness_in_machine_learning.rst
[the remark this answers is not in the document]
romanlutz: I think we've been capitalizing only the first letter in titles, too. After all, why capitalize "Construct Validity" but not "is"? [We do this in German, but mixing it between titles and remaining text makes little sense to me.]
I believe @MiroDudik came up with this in the first place.

docs/user_guide/fairness_in_machine_learning.rst
romanlutz: No, I meant within the user guide, but in a separate section (or separate subsection of section 1). It should definitely not be outside the user guide.

docs/user_guide/fairness_in_machine_learning.rst
mmadaio: Yup absolutely - I like the language of "observable properties" and "unobservable constructs" from the paper, if that works for others

docs/user_guide/fairness_in_machine_learning.rst
mmadaio: I love this!

docs/user_guide/fairness_in_machine_learning.rst
michaelamoako: Hmm. I see the metrics in fairlearn.metrics as examples of measurements - and perhaps those measurements may come together to form a bigger 'measurement model'. Maybe I can frame it as: 

**Fairness** is an example of an unobservable theoretical construct. Several measurements exist for measuring fairness, including demographic parity. See Fairlearn Metrics for more examples of measurements for measuring fairness. These measurements may come together to form a measurement model, where several measurements are combined to ultimately measure fairness. 

@hildeweerts @mmadiao 
Does this close the gap?

docs/user_guide/fairness_in_machine_learning.rst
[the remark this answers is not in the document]
michaelamoako: Resolving to merge this with the Hilde and Michael thread of comments below

docs/user_guide/fairness_in_machine_learning.rst
michaelamoako: @romanlutz the suggested linking method (:py:mod:fairlearn.metrics) did not work by itself - do I need to define something for this to work?

docs/user_guide/fairness_in_machine_learning.rst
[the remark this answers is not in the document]
michaelamoako: For consistency I will use
:code:`fairlearn.metrics`

docs/user_guide/fairness_in_machine_learning.rst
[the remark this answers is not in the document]
michaelamoako: Will change to AI Systems

docs/user_guide/fairness_in_machine_learning.rst
hildeweerts: I think I should have phrased this better. I too see the metrics in fairlearn as examples of measurements, i.e., simple measurement models of "fairness". But "fairness" is still an umbrella term that covers all different kinds of things. What I was trying to get to is that fairness metrics can be measurements of different underlying theoretical understanding of what it means for an algorithm to be fair (see section 4 of the paper). How about something like:

Fairness is an example of an unobservable theoretical construct - one about which there have been millennia of disagreements. There are different underlying understandings of what it means for an algorithm to be fair (which can be in conflict with each other!). Simple measurement models of these notions of fairness typically come in the form of fairness metrics, such as demographic parity. See :py:mod:fairlearn.metrics for more examples of fairness metrics.

I think it can be phrased better (the sentences are a bit long and maybe "underlying theoretical understanding"
Feel free to improve the phrasing (I feel the sentences are a bit long) . We might @michaelamoako @mmadaio wdyt?

michaelamoako: Hmm - so there's a whole section in Measurement and Fairness expanding on the idea of Fairness as an example of an unobservable theoretical construct. In that context, "highly contested" is the phrase used to capture the idea you're getting at. Maybe we should merge the two: 

Fairness is an example of an unobservable theoretical construct that is highly contested: there are many different understandings of what it means for an algorithm to be fair. Simple measurement models of these notions of fairness typically come in the form of fairness metrics, such as demographic parity. See :py:mod:fairlearn.metrics for more examples of fairness metrics.

docs/user_guide/fairness_in_machine_learning.rst
michaelamoako: Is *formal* an unobservable theoretical construct though @hildeweerts 
👀

hildeweerts: It might even be a highly contested one 👀

docs/user_guide/fairness_in_machine_learning.rst
hildeweerts: Love it!

docs/user_guide/fairness_in_machine_learning.rst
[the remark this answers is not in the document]
michaelamoako: @romanlutz For the paper should it be "Measurement and Fairness" or "Measurement and fairness"

docs/user_guide/fairness_in_machine_learning.rst
michaelamoako: What is the resolution here? 
Same question would apply to the other pending concept definition PRs (i.e: Abstraction Traps, what is the location there)? @romanlutz  @mmadaio

mmadaio: There's a PR open for the abstraction traps for the user guide (for now)!

https://github.com/fairlearn/fairlearn/pull/809/files

docs/user_guide/fairness_in_machine_learning.rst
mmadaio: Good point - "by mapping" maybe? But agreed that that's still a bit vague and yeah examples would help (although fwiw @michaelamoako, future PRs can keep adding to this!)

mmadaio (changes requested):
Thanks for moving this forward so there's a framework for future PRs to iterate on!

Before I approve, can you add the 2 paragraphs @hildeweerts suggested to the beginning to make clear why data scientists should read it?

Thanks!

"In many cases, fairness-related harms can be traced back to the way a real-world problem is translated into a machine learning task. Which target variable do we intend to predict? What features will be included? What (fairness) constraints do we consider? Many of these decisions boil down to what social scientists refer to as measurement: the way we measure (abstract) phenomena.

The concepts outlined in this glossary give an introduction into the language of measurement modeling. This framework can be a useful tool to test the validity of (implicit) assumptions of a problem formulation. In this way, it can help to mitigate fairness-related harms that can arise from mismatches between the formulation and the real-world context of an application."

michaelamoako:
@mmadaio should the two suggested paragraphs still have the reference to the paper: 

In many cases, fairness-related harms can be traced back to the way a real-world problem is translated into a machine learning task. Which target variable do we intend to predict? What features will be included? What (fairness) constraints do we consider? Many of these decisions boil down to what social scientists refer to as measurement: the way we measure (abstract) phenomena.

**The concepts outlined in this glossary give an introduction into the language of measurement modeling - as described in Measurement and Fairness [#4]_ .** This framework can be a useful tool to test the validity of (implicit) assumptions of a problem formulation. In this way, it can help to mitigate fairness-related harms that can arise from mismatches between the formulation and the real-world context of an application.

hildeweerts:
Yes, please add the reference @michaelamoako!

mmadaio (approved):

hildeweerts:
Many thanks for this great contribution @michaelamoako!

Issue #737, opened 2021-04-09
Supporting `count` as a metric


adrinjalali:
It's desired to know the number of data points in each group when working with `MetricFrame`, and that can be achieved by passing a function which returns the length of its input. For instance:

``` python
# report the above scores in a table.
from fairlearn.metrics import MetricFrame
import sklearn.metrics as skm
y_pred = np.asarray(df.y_pred).astype(float)
y_true = np.asarray(df.FraudLabel == "true").astype(int)
multi_metric = MetricFrame({
    'average precision':skm.average_precision_score,
    'roc-auc': skm.roc_auc_score,
    'count': lambda x, y: len(x)},
    y_true, y_pred,
    sensitive_features=df.gender)
multi_metric.by_group
```

We don't necessarily need to change anything in the code or API for this, but we should document it in the user guide and/or the API reference to let people know how it can be done.

romanlutz:
I like this a lot. The `lambda x, y: len(x)` isn't super straightforward if you're not familiar with how to write a metric function. Should we just define the `count` function like we did with `selection_rate` so that people can import it and pass it without having to write the code themselves?

Totally agree on documenting this. I guess the tasks here would be
- [ ] to look at the user guide of `MetricFrame` and add that new `count` metric wherever we have 2 or more metrics passed to the `MetricFrame`, and write a short sentence or two on what this does (i.e., "show the number of people in the corresponding group")
- [ ] to look through notebooks and add `count` whenever we have a `MetricFrame`.

hildeweerts:
I like the convenience of a `count` function, it makes it more beginner friendly I think. Do we need to open a separate issue for that?

adrinjalali:
I think the same issue/pr could do both the doc and the function.

MiroDudik:
To clarify the `count` function proposal. I'm assuming we're just talking about adding `fairlearn.metrics.count`, but not doing anything with the `MetricFrame` API right?

adrinjalali:
> To clarify the `count` function proposal. I'm assuming we're just talking about adding `fairlearn.metrics.count`, but not doing anything with the `MetricFrame` API right?

Correct :)

vamsidesu5:
Hey I can take this issue up!

adrinjalali:
Sure, it's yours @vamsidesu5

stephenrobic:
I will begin working on this issue now. Thanks for your guidance @romanlutz

romanlutz:
[...]

PR #812, opened 2021-05-19
Implement count as a metric

stephenrobic:
This PR seeks to implement `count` as a metric, in order to decipher the number of data points in each group when working with `MetricFrame`.

The following were updated:
- Updated user guide to add new `count` metric wherever more than 1 metric is passed to `MetricFrame`
- Added `count` to notebooks wherever `MetricFrame`.
- Added new `count` function in `fairlearn.metrics.extra_metrics`
- Added `count` to `fairlearn.metrics.group_metric_set` (for consistency, even though deprecated)
- Added unit tests and fixed existing tests pertaining to `count`
  - Fixed expected json results in `unit.metrics.sample_dashboards`)
- Added self to `AUTHORS.md` (and look forward to contributing more)

This is in reference to issue #737


stephenrobic:
:wave: @romanlutz, let me know what you think.

docs/user_guide/assessment.rst
```diff
@@ -123,7 +123,9 @@ metrics simultaneously:
 .. doctest:: assessment_metrics
     :options:  +NORMALIZE_WHITESPACE
 
-    >>> multi_metric = MetricFrame({'precision':skm.precision_score, 'recall':skm.recall_score},
+    >>> multi_metric = MetricFrame({'precision':skm.precision_score,
+                                    'recall':skm.recall_score,
+                                    'count': count},
```
romanlutz, line 128: Don't you need to import this first?

docs/user_guide/assessment.rst
```diff
@@ -138,6 +140,9 @@ metrics simultaneously:
     c                         0.6   0.75
     d                         0.0    0.0
 
+Note that :func:`fairlearn.metrics.count` can be used to show the number of subjects in
```
romanlutz, line 143: Subjects sounds like we're doing an experimental trial of sorts 🙂 How about people / rows? Or simply group size?

romanlutz (commented):
This is great! Thanks for the comprehensive set of changes including all the existing occurrences throughout the user guide! Would you mind adding a few simple unit tests? I'm thinking for an empty group, group of 1, and something slightly longer.

docs/user_guide/assessment.rst
stephenrobic: I think it's imported from `fairlearn.metrics` [here] (https://github.com/fairlearn/fairlearn/blame/8a76289ed54353872b590c790df8de7bd03ba7e3/docs/user_guide/assessment.rst#L7). Is there somewhere else that it needs to be imported?

I built the docs locally, and it seemed to work.

docs/user_guide/assessment.rst
stephenrobic: Yeah, that makes sense. I'll get that changed to group size.

stephenrobic:
Thanks for the speedy feedback :smile: I can definitely play around with adding some unit tests.

docs/user_guide/assessment.rst
romanlutz: That's not an import AFAIK. At least I haven't seen it work that way, but then this wouldn't be the first time I learn something new 😆 Regardless, it would be nice to have it explicitly mentioned where `count` comes from.

fairlearn/metrics/__init__.py
```diff
@@ -46,7 +46,7 @@
     selection_rate,
     _mean_overprediction,
     _mean_underprediction,
-    )
+    count)
```
romanlutz, line 49: I think it would make sense for it to live in the `_extra_metrics.py` file, but in your PR it seems to be in its own file. Can you move it into the file? I'm not entirely sure how it's actually working right now since there should be no `count` in `_extra_metrics`, so what is it actually importing?

romanlutz (commented):
Make sure to also run `flake8` from the repo root to get all linting issues. There's currently only 1 but you never know 😄

fairlearn/metrics/_count.py
```diff
@@ -0,0 +1,32 @@
...
+
+    Parameters
+    ----------
+    y_true : array_like
+        The true labels
+
+    y_pred : array_like
+        The predicted labels (ignored)
+
+    sample_weight : array_like
```
riedgar-ms, line 23: If you're going to ignore it, don't include it.

fairlearn/metrics/_group_metric_set.py
```diff
@@ -44,6 +45,7 @@
 # Issue 269 is about unifying the two sets
 ACCURACY_SCORE = "accuracy_score"
 BALANCED_ROOT_MEAN_SQUARED_ERROR = "balanced_root_mean_squared_error"
+COUNT = "count"
```
riedgar-ms, line 48: It's harmless to add this, but also I don't think it'll be used. This routine is only for the dashboard, and will be removed from Fairlearn in the near future.

romanlutz: And when it's moved won't this be useful to have it included? Otherwise you'll need to manually add it?

fairlearn/metrics/_count.py
stephenrobic: I'll remove it. I was trying to be consistent with this: https://github.com/fairlearn/fairlearn/blob/8a76289ed54353872b590c790df8de7bd03ba7e3/fairlearn/metrics/_selection_rate.py#L20-L27

fairlearn/metrics/__init__.py
stephenrobic: > I think it would make sense for it to live in the `_extra_metrics.py` file, but in your PR it seems to be in its own file. Can you move it into the file? I'm not entirely sure how it's actually working right now since there should be no `count` in `_extra_metrics`, so what is it actually importing?

I was doing something similar to `fairlearn.metrics._selectrion_rate`. The selection rate function is in its own file, but it's imported through `fairlearn.metrics._extra_metrics` here: https://github.com/fairlearn/fairlearn/blob/main/fairlearn/metrics/_extra_metrics.py#L14
Maybe I misunderstood this comment and its answer: https://github.com/fairlearn/fairlearn/issues/737#issuecomment-821566124

romanlutz: True, but you're not importing `count` in `_extra_metrics.py`. Anyway, it's simpler to have it in one file 😄

stephenrobic:
I believe I have fixed everything mentioned here. Thanks for the feedback. How does this look?

docs/user_guide/assessment.rst
romanlutz: Did you want to change this still?

romanlutz (approved):
Looks great! There's the small comment about "subjects" I'd prefer to have changed, but otherwise this looks ready to merge to me. Let's see if the gates pass and @riedgar-ms probably should take another look as well.

docs/user_guide/assessment.rst
```diff
@@ -118,25 +118,32 @@ across groups and also the difference and ratio between the maximum and minimum:
...
+metrics simultaneously (note that :func:`fairlearn.metrics.count` can be used to
+show each group's size):
 
 .. doctest:: assessment_metrics
     :options:  +NORMALIZE_WHITESPACE
 
-    >>> multi_metric = MetricFrame({'precision':skm.precision_score, 'recall':skm.recall_score},
+    >>> from fairlearn.metrics import count
+    >>> multi_metric = MetricFrame({'precision':skm.precision_score,
+                                    'recall':skm.recall_score,
```
romanlutz, line 129: I believe you need to add `...` at the start of continuation lines. Try to run the doc build with `doctest` rather than `html` as the arg to see the doctest output. It should tell you what's wrong. The doc build which builds the webpage (separate from doctest) also failed but that's something we're tackling in #815 and that will probably get merged before your PR, hopefully tomorrow.

stephenrobic:
I got that all fixed. I forgot to take out that line about "subjects", when I added the new one referring to "groups" above it. :facepalm:  I saw that issue #815 earlier when building the docs on my end, so that's great. Thanks for the thorough review!

fairlearn/metrics/_extra_metrics.py
```diff
@@ -215,3 +215,22 @@ def false_negative_rate(y_true,
 def _root_mean_squared_error(y_true, y_pred, **kwargs):
     r"""Calculate the root mean squared error."""
     return skm.mean_squared_error(y_true, y_pred, squared=False, **kwargs)
+
+
+def count(y_true, *args, **kwargs) -> int:
+    r"""Calculate the number of data points in each group when working with `MetricFrame`.
+
+    The ``y_true`` argument is used to make this calculation. For consistency with
+    other metric functions, the ``y_pred`` argument is required, but ignored.
```
riedgar-ms, line 224: This doesn't match the argument list above....

riedgar-ms: Given the pattern shared by other metrics, I'd be inclined to include `y_pred` as the only other argument (my comment above was directed at `sample_weight`). You can check that `y_pred` and `y_true` have the same length before returning.

stephenrobic: Thank you for this feedback. I made the changes.

stephenrobic:
I'm working on fixing the broken tests after adding the `count` function.

stephenrobic:
I have fixed the unit tests, so that `count` is included.

romanlutz:
LGTM! @riedgar-ms feel free to merge if you're happy with it.

riedgar-ms (approved):
Thanks for working on this!

romanlutz:
@stephenrobic last change request 😄 There's a file called CHANGES.md in the repo root directory. Can you add a line to the next release (0.7.0) saying that `count` has been added as a new metric? I'll merge once that's there.

stephenrobic:
@riedgar-ms @romanlutz It was my pleasure to work on this and learn from it. Thank you for guiding me through this!

romanlutz:
Thanks @stephenrobic for this fantastic contribution!

PR #820, opened 2021-05-24
Cost sensitive classification moment

### LINKED ISSUE #804: Support different costs for false positives and false negatives in ExponentiatedGradient

In many binary classification problems, the costs of false positives do not equal to the costs of false negatives, but our `ExponentiatedGradient` only allows classification accuracy as the objective, which assumes that these costs are equal.

We should extend the moment `ErrorRate` to allow for asymmetric costs, and also allow specifying these costs in the constructor to `ExponentiatedGradient`.

I'm thinking that there are two natural API choices. Say, we're considering fairness under demographic parity and the costs for false positives and false negatives are 1.0 and 0.1. Then:

```python
# Current API, which does not allow specifying asymmetric costs for the objective:
expgrad = ExponentiatedGradient(
    learner,
    constraints=DemographicParity())

# API idea (1): Allow specifying the objective via a moment object
expgrad = ExponentiatedGradient(
    learner,
    constraints=DemographicParity(),
    objective=ErrorRate(costs={'fn': 1.0, 'fp': 0.1}))

# API idea (2): Specify the objective costs more directly
expgrad = ExponentiatedGradient(
    learner,
    constraints=DemographicParity(),
    objective_costs={'fn': 1.0, 'fp': 0.1})
```
I slightly prefer API (1), because it allows for extensions into future in case we want to support other objective functions.

The API (2) is a bit simpler and matches #473, but it could be a bit confusing, because `objective_costs` would not be applicable when we are using `ExponentiatedGradient` for regression.

MiroDudik:
Closes #804


fairlearn/reductions/_moments/error_rate.py
```diff
@@ -9,14 +9,57 @@
...
+      c_{FP} P[h(X)=1, Y=0] + c_{FN} P[h(X)=0, Y=1]
+
+    where :math:`c_{FP}` and `c_{FN}` are the costs of false positive
+    and false negative errors respectively. The standard misclassification
+    error corresponds to :math:`c_{FP}=c_{FN}=1.0`.
+
+    Parameters
+    ----------
+    costs : dict
+        The dictionary with keys ``"fp"`` and ``"fn"`` containing the
```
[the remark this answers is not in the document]
MiroDudik: That was the intention, but I'm not sure. Website didn't recompile when I pushed the documentation commit, so I cannot tell whether it worked. I've seen both in the docs and they both seem to render similar. Do you have a preference between the two?

fairlearn/reductions/_moments/error_rate.py
```diff
@@ -9,14 +9,57 @@
...
 
+    def __init__(self, *, costs=None):
+        """Initialize the costs."""
+        super(ErrorRate, self).__init__()
+        if costs is None:
+            self.fp_cost = 1.0
+            self.fn_cost = 1.0
+        elif (type(costs) is dict
+              and costs.keys() == {'fp', 'fn'}
+              and costs['fp'] >= 0.0
```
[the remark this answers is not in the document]
MiroDudik: seems reasonable. updated.

fairlearn/reductions/_moments/error_rate.py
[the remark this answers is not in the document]
romanlutz: I suppose the code one since it's explicit while you need to know additional things for the double backquote(?)
You should definitely try how it renders locally when submitting doc changes 😉

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -31,8 +34,11 @@ class _Lagrangian:
         the estimator to fit in every iteration of :meth:`best_h` using a
         :meth:`fit` method with arguments `X`, `y`, and `sample_weight`
     constraints : fairlearn.reductions.Moment
-        Object describing the parity constraints. This provides the reweighting
+        Object describing the fairness constraints. This provides the reweighting
```
riedgar-ms, line 37: Why this switch?

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -43,11 +49,19 @@ class _Lagrangian:
         (defaults to `sample_weight`)
     """
 
-    def __init__(self, X, y, estimator, constraints, B, opt_lambda=True,
+    def __init__(self, X, y, estimator, constraints, *, B, objective=None, opt_lambda=True,
```
riedgar-ms, line 52: Why not all keyword arguments here?

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -43,11 +49,19 @@ class _Lagrangian:
...
-        self.obj = self.constraints.default_objective()
+        if objective is None:
+            self.obj = self.constraints.default_objective()
+        elif ((isinstance(constraints, ClassificationMoment)
+               and isinstance(objective, ClassificationMoment)) or
+              (isinstance(constraints, LossMoment)
+               and isinstance(objective, LossMoment))):
+            self.obj = objective
+        else:
+            raise ValueError(_MESSAGE_BAD_OBJECTIVE)
```
riedgar-ms, line 64: Might be nice to let the users know what they supplied as the types in the error message.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -11,11 +11,14 @@
...
-from fairlearn.reductions._moments import ClassificationMoment
+from fairlearn.reductions._moments import ClassificationMoment, LossMoment
 
 logger = logging.getLogger(__name__)
 
 
+_MESSAGE_BAD_OBJECTIVE = "Objective needs to be of the same type as constraints."
+
+
 class _Lagrangian:
```
riedgar-ms, line 22: Not for this PR, but there was the little matter of making sure that this is pickleable.... my PR (#662) is waiting on whether it's OK to ditch the callable for an estimator :-)

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -56,11 +59,13 @@ class ExponentiatedGradient(BaseEstimator, MetaEstimatorMixin):
         (defaults to `sample_weight`)
     """
 
-    def __init__(self, estimator, constraints, eps=0.01, max_iter=50, nu=None,
+    def __init__(self, estimator, constraints, objective=None,
```
riedgar-ms, line 62: Since this breaks positional args anyway (after `constraints` at least), do you want to add a `*` here?

fairlearn/reductions/_moments/error_rate.py
```diff
@@ -9,14 +9,58 @@
...
+        costs of false positives and false negatives. If none are provided
+        costs of 1.0 are assumed.
+    """
 
     short_name = "Err"
 
+    def __init__(self, *, costs=None):
+        """Initialize the costs."""
+        super(ErrorRate, self).__init__()
+        if costs is None:
```
riedgar-ms, line 47: This little `if` block might be better as a separate function, for easier unit testing. Plus, it might come in useful elsewhere.

fairlearn/reductions/_moments/error_rate.py
```diff
@@ -9,14 +9,58 @@
...
+            self.fp_cost = 1.0
+            self.fn_cost = 1.0
+        elif (type(costs) is dict
+              and costs.keys() == {'fp', 'fn'}
```
riedgar-ms, line 51: The strings `fp` and `fn` should be made into constants.

fairlearn/reductions/_moments/error_rate.py
```diff
@@ -9,14 +9,58 @@
...
+              and costs['fp'] >= 0.0
+              and costs['fn'] >= 0.0
+              and costs['fp'] + costs['fn'] > 0.0):
+            self.fp_cost = costs['fp']
+            self.fn_cost = costs['fn']
+        else:
+            raise ValueError(_MESSAGE_BAD_COSTS)
+        self.sum_costs = self.fp_cost + self.fn_cost
```
riedgar-ms, line 59: I would suggest making `sum_costs`, `fp_cost` and `fn_cost` into properties rather than fields.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
riedgar-ms: Also, I'm not seeing unit tests for this `if` block.

fairlearn/reductions/_moments/error_rate.py
riedgar-ms: Also, I'm not seeing the unit tests.

riedgar-ms (changes requested):

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
MiroDudik: Our fairness constraints are actually more generic than what I'd call "parity" constraints. For example, bounded group loss is not really a parity constraint. People are confused: https://stackoverflow.com/questions/66910107/can-you-use-fairlearn-for-non-parity-constraints-binned-monotonicity/
Also, we could say "the optimization constraints" vs "the optimization objective". What do you think?

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
MiroDudik: I think that's a good idea. Will do.

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
MiroDudik: I can put `*` before `objective`, but probably not any earlier (at least not in this PR). Sounds good?

fairlearn/reductions/_moments/error_rate.py
MiroDudik: Is this a deal-breaker? We had this conversation elsewhere and made (more or less?) the choice that literal string = constants, e.g., https://github.com/fairlearn/fairlearn/blob/main/fairlearn/postprocessing/_threshold_optimizer.py.

fairlearn/reductions/_moments/error_rate.py
riedgar-ms: Surely the naming convention has nothing to do with whether they're fields or properties?

fairlearn/reductions/_moments/error_rate.py
riedgar-ms: I hadn't noticed those slipping in to `_threshold_optimizer.py`. Strings which pop up more than once really ought to be constants.

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
riedgar-ms: That's what I meant - adding `objective` breaks everything _after_ `constraints` as a positional arg.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
riedgar-ms: That's OK with me. It might be good to put the note about that in the PR description. I just wasn't expecting this to be the first change I saw.

fairlearn/reductions/_moments/error_rate.py
MiroDudik: So maybe I misunderstood. Why do you want them to be properties?

riedgar-ms: Because making them properties gives a place for adding documentation (and type hints)

fairlearn/reductions/_moments/error_rate.py
MiroDudik: Is this required for this PR or should we hold off to a separate conversation re. coding style?

I'm really hesitant to create a constant here. Replacing `'fp'` and `'fn'` by `FP` and `FN` or something of the sort is going to make the code harder to read in my opinion. Right now, the constructor, its documentation and validation are very near-each other in the code.

Besides `ThresholdOptimizer`, we similarly don't create constants for the method argument in `MetricFrame.difference` and there are probably other places in the library.

fairlearn/reductions/_moments/error_rate.py
romanlutz: Perhaps I don't remember it quite right, but I do think that changes the naming convention. Also, the costs are input args and not calculated through calling `fit` so they should not be suffixed with `_`. I like @riedgar-ms's suggestion to make them properties. You could argue that it's different for the sum of costs, but I appreciate the consistency between the three as well.
[I can't shake the feeling that I'm splitting hair here, though, so feel free to proceed either way if I don't respond further.]

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
MiroDudik: [done]

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
MiroDudik: [done on both accounts]

fairlearn/reductions/_moments/error_rate.py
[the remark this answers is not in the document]
MiroDudik: done

fairlearn/reductions/_moments/error_rate.py
riedgar-ms: OK. The point about us not doing constants like that elsewhere is fair.

fairlearn/reductions/_moments/error_rate.py
riedgar-ms: If `sum_costs` is only used in one place, removing it does make sense.

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -39,7 +39,10 @@ class ExponentiatedGradient(BaseEstimator, MetaEstimatorMixin):
         :code:`predict(X)` are either 0 or 1.
         In regression values `y` and predictions are continuous.
     constraints : fairlearn.reductions.Moment
-        The disparity constraints expressed as moments
+        The fairness constraints
```
riedgar-ms, line 42: I'd prefer the docstrings to be expanded rather than trimmed....

riedgar-ms:
Could you also extend the PR description a bit?

romanlutz (approved):
Looks good to me at a glance. @riedgar-ms pointed out a few things that sound reasonable. I don't particularly care about the string constants FWIW. I suggest updating the changelog (in the user guide) and notifying affected PRs once we merge this.

fairlearn/reductions/_moments/error_rate.py
MiroDudik: Added the unit tests for this. Re. pulling out into a separate function--we can maybe hold off until we have the second use case.

docs/user_guide/installation_and_version_guide/v0.7.1.rst
```diff
@@ -18,4 +18,6 @@ v0.7.1
 * :class:`~fairlearn.preprocessing.CorrelationRemover` now exposes
   :code:`n_features_in_` and :code:`feature_names_in_`.
 * Added the ACSIncome dataset and corresponding documentation.
-* Add `sphinxcontrib-bibtex` extension to manage citations in documentation using bibtex.
\ No newline at end of file
+* Add `sphinxcontrib-bibtex` extension to manage citations in documentation using bibtex.
+* Added support for explicitly specifying optimization objective in :class:`~fairlearn.reductions.ExponentiatedGradient`.
+  Added support for cost sensitive classification in :class:`~fairlearn.reductions.ErrorRate`.
```
romanlutz, line 23:
```suggestion
* Added support for explicitly specifying optimization objective in 
  :class:`~fairlearn.reductions.ExponentiatedGradient`.
  Added support for cost sensitive classification in :class:`~fairlearn.reductions.ErrorRate`.
```
max line length 79. Maybe the same in the other line?

docs/user_guide/installation_and_version_guide/v0.7.1.rst
```diff
@@ -18,4 +18,6 @@ v0.7.1
 * :class:`~fairlearn.preprocessing.CorrelationRemover` now exposes
   :code:`n_features_in_` and :code:`feature_names_in_`.
 * Added the ACSIncome dataset and corresponding documentation.
-* Add `sphinxcontrib-bibtex` extension to manage citations in documentation using bibtex.
\ No newline at end of file
+* Add `sphinxcontrib-bibtex` extension to manage citations in documentation using bibtex.
```
romanlutz, line 21: Heads up I'm renaming this to v0.8.0.rst in another PR and depending on who merges first this will create a conflict. Should be trivial to resolve, though.

romanlutz (approved):

PR #823, opened 2021-05-26
Update copyright of docs

hildeweerts:
This PR updates the copyright info in the configuration of our docs, which makes the footer of the website (more) consistent with the footer of the home page.


romanlutz:
Potentially stupid question: some websites say something like 2019-2021 or the oldest year, some say the latest. So I'm wondering whether it matters and how. I haven't found a good explanation yet. 

Regarding the removal of Microsoft I'm still not 100% sure that's possible. The original page was paid for by Microsoft, then replaced by @LeJit. I vaguely remember that Microsoft employees need to apply the copyright wherever they work in open source, but it makes less sense if I think about contributions to, say, pytorch, which certainly won't apply Microsoft copyright. In many ways we're just as removed from Microsoft as Pytorch. Paging @MiroDudik who certainly knows more than me on this topic.

hildeweerts:
> Potentially stupid question: some websites say something like 2019-2021 or the oldest year, some say the latest. So I'm wondering whether it matters and how. I haven't found a good explanation yet.

That's a good question - I've always assumed it was more of a "look how long we've been around!" thing rather than a legal issue. I'd be happy to use 2019 - 2021. I was also thinking it might be better to automate the retrieval of the year so we don't have to update it manually :D

In case it turns out we don't have to copyright Microsoft, we can also think about gradually removing some of the copyright notes that's still in a lot of code files as well.

romanlutz:
Agreed! I think it's also just much simpler and avoids confusion... The project @MiroDudik often points to is VW and they've had a very interesting history if you look at their license:
https://github.com/VowpalWabbit/vowpal_wabbit/blob/master/LICENSE

```
Copyright © Microsoft Corp 2012-2014, Yahoo! Inc. 2007-2012, and many individual contributors.
```
which seems to me like they went with independent governance starting sometime in 2014 OR they haven't updated the file since then (which the git history confirms to be true). Their website simply states that Microsoft is a major contributor, and I've long wanted to have a little page similar to scikit-learn's page that states companies, projects, non-profits, etc. that contribute in some way: https://scikit-learn.org/stable/about.html#funding

After all, credit where credit is due for funding people to do work on Fairlearn! Copyright isn't really that and seems weird to everyone who visits our website or looks at the code. Let's see what we can find out.

hildeweerts:
@MiroDudik do you have thoughts on this?

MiroDudik (approved):

MiroDudik:
I think that for the webpage, I'm happy to go with "Fairlearn contributors".

However, I don't think we should be removing existing copyright from the source code (seems in conflict with MIT license). Newly contributed code could just state "Fairlearn contributors" as a default, with an option for the people to add their name (or possibly name of the company if it's sponsoring their work). The key part is that it's under MIT license.

I'm not sure about adding copyright years, we can ask lawyers.

romanlutz (approved):

romanlutz:
This does not include code changes so the cancelled pipeline can be safely ignored.

PR #851, opened 2021-06-07
TST fix warnings and issues in tests

adrinjalali:
This PR fixes a few `FutureWarning`s and some errors in the tests.

cc @romanlutz @riedgar-ms


riedgar-ms (approved):

adrinjalali:
I found more warnings (my env was a bit old), submitting them soon

test_othermlpackages/test_xgboost.py
```diff
@@ -1,11 +1,12 @@
 # Copyright (c) Microsoft Corporation and Fairlearn contributors.
 # Licensed under the MIT License.
 
+import pytest
 from . import package_test_common as ptc
 
 from fairlearn.reductions import DemographicParity
 
-import xgboost as xgb
+xgb = pytest.importorskip("xgboost")
```
romanlutz, line 9: I'm a little confused as to why this is helping, because the test will fail if we can't import, won't it?

romanlutz (approved):
Thanks for looking into these!

test_othermlpackages/test_xgboost.py
adrinjalali: the test will be skipped, not failed. This raises an exception caught by pytest which leads to it skipping the test rather than failing.

riedgar-ms: It's related to why we have `test_othermlpackages` as a whole other directory.

adrinjalali: That being a separate directory doesn't make tools such as the ones in vscode to ignore them. This makes those tools to actually successfully "discover" the tests.

romanlutz: I see. I think this goes back to our old problem of not wanting to have tests that are skipped without us noticing. Perhaps we were a little paranoid there (?)

If I understand correctly this would allow us to have these tests inside the normal `test` directory, right?

romanlutz: ... regardless of the answer I don't expect that to happen here 😆

adrinjalali: yes, they can go to the normal test suite if you want. AFAIK you can pass `-v` to `pytest` to get a list of skipped tests, if you wanna see what's skipped. Skipping is different than XFAIL though, that I'd understand more if you wanna make sure they pass sometimes?

adrinjalali:
I think this is ready from my side.

adrinjalali:
@romanlutz @riedgar-ms merging this would also reduce the diff size in #854. So please merge if there's nothing I can add here :)

Makefile
```diff
@@ -19,6 +19,9 @@ test-coverage:
 test-unit:
 	$(PYTEST) ./test/unit
 
+test-deprecation:
+	$(PYTEST) -Werror::DeprecationWarning -Werror::FutureWarning ./test/unit
```
riedgar-ms, line 23: Can you add this to the actual build pipelines?

adrinjalali: gonna try, but rusty there, please suggest better ways.

devops/templates/test-run-step-template.yml
```diff
@@ -70,6 +70,14 @@ steps:
   displayName: "Run flake8"
 
 # =================================================
+- ${{ if eq(parameters.testRunType, 'Unit-Deprecation')}}:
+  - script: python -m pytest test/ -Werror::DeprecationWarning -Werror::FutureWarning --ignore=test/install -m "not notebooks" --junitxml=./TEST--TEST.xml -o junit_suite_name="$(Agent.JobName)"
```
riedgar-ms, line 74: I actually meant just adding the `-Werror` arguments to the `Unit` line below. Sorry for the misunderstanding.

Makefile
riedgar-ms: As I said, just add the flags to the `Unit` tests below. No point running all the tests again, just to look for deprecation warnings.

devops/templates/test-run-step-template.yml
adrinjalali: Sure, didn't wanna have it always error. Thought it'd be nice to have the normal tests and just have only one CI run detecting these, for us to see whether they fail because of the warnings or other issues. But could add them here.

riedgar-ms: If your changes above make us 'clean' then I don't see the point in burning the CPU time on a separate pipeline run which effectively just checks for warnings.

If we've a way to go to squelch everything, then maybe a separate pipeline is useful for isolating the issues.

devops/templates/test-run-step-template.yml
```diff
@@ -71,6 +71,7 @@ steps:
 
 # =================================================
 - ${{ if eq(parameters.testRunType, 'Unit')}}:
+  # The imp warning comes from papermill's usage of ansiwrap
```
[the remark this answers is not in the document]
adrinjalali: the config is now in the pyproject.toml actually, so removing this comment

riedgar-ms (approved):

adrinjalali:
Finally green :)

PR #856, opened 2021-06-08
MNT add black config to pyproject.toml

adrinjalali:
This PR adds a config to `pyproject.toml` for people who use black. It doesn't add black to the CI or anything, it's just nice to have it in the repo for the ones who use it.

cc @riedgar-ms @romanlutz

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,2 @@
+[tool.black]
+line-length = 79


riedgar-ms:
Could you add something to the contributor guide about this? It might be a good idea to add `black` as a commit hook eventually, so it would be good to start documenting its use.

riedgar-ms (approved):

adrinjalali:
Finally green! A second review and merge would be lovely.

pyproject.toml
```diff
@@ -0,0 +1,2 @@
+[tool.black]
+line-length = 79
```
romanlutz, line 2: We've been using 99 for code actually since 79 was super constraining and everything went over multiple lines or variable names became unreadable.

adrinjalali: and with 99 it doesn't fit on the monitor to do side by side reviews or to have an interactive interpreter beside the code window :D This doesn't change the flake check, just for the future it'd format the code shorter. Also, we used to have long variable names, but they have become more python like and much shorter, and now it's much easier to fit the code in 80 chars.

MiroDudik: Yeah: https://www.python.org/dev/peps/pep-0008/#maximum-line-length

I think we should keep 79 here; leave the flake8 at 99 for the time being, but try to thoughtfully shorten variable names as we touch the relevant parts of code.

romanlutz: There's definitely a lot to be said for going with defaults (79 in this case), and if it doesn't require us to adjust everything right away I don't mind at all. I'll just adjust from now on 😆

romanlutz (approved):
There was a merge conflict with the other change that added the same `pyproject.toml` file so I resolved it. Whoever sees this first after passing gates should feel free to merge!

PR #829, opened 2021-05-29
Made navigation bar consistent on front page

Gingeh:
#825

PR #860, opened 2021-06-10
Make navigation bar consistent across website

lucioZuccolo:
The navigation bar is currently consistent.


romanlutz:
Hi @lucioZuccolo ! This seems to be the same as #829 and unfortunately has the same problem as described in #829. If you want to fix it as described there feel free to do so.

MiroDudik:
> Hi @lucioZuccolo ! This seems to be the same as #829 and unfortunately has the same problem as described in #829. If you want to fix it as described there feel free to do so.

Also, this is actually not the best first issue since it requires mucking with how sphinx renders the navigation bar (I have removed the "good first issue" tag). Sorry about that. Please let us know if you'd like to continue with this PR.

romanlutz:
Closing this since there's been no response. If you're still interested please reopen and comment. We can sync on specifics if you'd like.

Issue #852, opened 2021-06-07
DOC fix warnings in documentation build


adrinjalali:
There are warnings during doc build which need to be taken care of:

<details>

```
Output from /home/adrin/Projects/fairlearn/fairlearn/examples/plot_new_metrics.py                                                                                                       
/home/adrin/Projects/fairlearn/fairlearn/examples/plot_new_metrics.py:91: SettingWithCopyWarning: 
A value is trying to be set on a copy of a slice from a DataFrame.
Try using .loc[row_indexer,col_indexer] = value instead

See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
  X_raw['race'] = X_raw['race'].map(race_transform).fillna('Other').astype('category')
['Black' 'Other' 'White']
Output from /home/adrin/Projects/fairlearn/fairlearn/examples/plot_new_metrics.py
/home/adrin/Projects/fairlearn/fairlearn/examples/plot_new_metrics.py:125: SettingWithCopyWarning: 
A value is trying to be set on a copy of a slice from a DataFrame.
Try using .loc[row_indexer,col_indexer] = value instead

See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
  A['Credit Score'] = col_credit
/home/adrin/Projects/fairlearn/fairlearn/examples/plot_new_metrics.py:126: SettingWithCopyWarning: 
A value is trying to be set on a copy of a slice from a DataFrame.
Try using .loc[row_indexer,col_indexer] = value instead

See the caveats in the documentation: https://pandas.pydata.org/pandas-docs/stable/user_guide/indexing.html#returning-a-view-versus-a-copy
  A['Loan Size'] = col_loan_size
Output from /home/adrin/Projects/fairlearn/fairlearn/examples/plot_new_metrics.py
/home/adrin/miniforge3/envs/fairlearn/lib/python3.8/site-packages/sklearn/metrics/_classification.py:1495: UndefinedMetricWarning: F-score is ill-defined and being set to 0.0 due to no true nor predicted samples. Use `zero_division` parameter to control this behavior.
  _warn_prf(
[...]

PR #615, opened 2020-10-30
Contributor Guide updates for examples & codeowner definition

romanlutz:
Update: Managed to revive this PR!

The purpose is to 
- update the contributor guide (this is the main thing to review here! All the files under `docs/contributor_guide`), specifically
  - community call information
  - document the example notebook creation process as discussed in the community call several weeks ago
  - development process clarifications
  - remove developer call notes since we haven't been migrating content over from hackmd. Instead this just links to hackmd.
- get the CODEOWNERS file updated so that it actually does the reviewer assignment for PRs. @hildeweerts  and @adrinjalali  please point out which paths you'd like to be added to and I'll make that happen.
- add a MAINTAINERS file consistent with https://github.com/fairlearn/governance/pull/1#discussion_r536371563
- sphinx config updates: add Twitter account & intersphinx matplotlib mapping
- fixed lots of sphinx errors - for some reason they popped up this time and never had before

Once the CODEOWNERS file is updated I'll enable the automation to use it.

Signed-off-by: Roman Lutz <rolutz@microsoft.com>

PR #861, opened 2021-06-10
MNT doc build fixes

romanlutz:
This PR extracts the doc build changes from #615 to make it less controversial and simpler to review.

This is related to #852 but doesn't fix all the errors yet.


hildeweerts (commented):
I am not familiar with the usage of `:class:` and `:obj:` so if these comments don't make sense please ignore. It seems like we use `:class:` in Parameters and `:obj:` in Returns, but I am not really sure why you would use `:class`: there because the input usually is in the form of an object, not a class right?

docs/conf.py
```diff
@@ -72,18 +72,21 @@ def check_if_v046():
     'sphinx_multiversion'
 ]
 
+source_suffix = ['.rst', '.md']
```
adrinjalali, line 75: interesting, are we mixing the two formats? :D

docs/contributor_guide/contributing_code.rst
```diff
@@ -23,8 +23,8 @@ Reductions
 Reduction constructors require a parameter corresponding to an estimator that
 implements the :code:`fit` method with the :code:`sample_weight` argument.
 Parity constraints for reductions are expressed via instances of various
-subclasses of the class :code:`fairlearn.reductions.Moment`. Formally,
-instances of the class :code:`Moment` implement vector-valued random variables
+subclasses of the class :class:`fairlearn.reductions.Moment`. Formally,
+instances of the class :class:`.Moment` implement vector-valued random variables
```
adrinjalali, line 27: as long as you've checked and these work, I'm happy with them :D

docs/faq.rst
```diff
@@ -125,19 +131,24 @@ Which ML libraries does Fairlearn support?
     algorithm can be evaluated using our metrics.
 
 Does Fairlearn support multi-class classification?
-    On the assessment side, Fairlearn's :class:`MetricFrame` can be used with
-    any metrics for supervised learning, including those for multi-class classification.
+    On the assessment side, Fairlearn's :class:`fairlearn.metrics.MetricFrame`
```
adrinjalali, line 134: you can set the `currentmodule` on top of these files to avoid having to repeat `fairlearn` here, and have it as 

```suggestion
    On the assessment side, Fairlearn's :class:`metrics.MetricFrame`
```

docs/user_guide/mitigation.rst
```diff
@@ -15,13 +15,13 @@ Fairlearn contains the following algorithms for mitigating unfairness:
       - binary classification
       - regression
       - supported fairness definitions
-   *  - :code:`fairlearn.` :code:`reductions.` :code:`ExponentiatedGradient`
+   *  - :class:`fairlearn. reductions. ExponentiatedGradient <fairlearn.reductions.ExponentiatedGradient>`
```
adrinjalali, line 18: why the spaces after the dot?

fairlearn/datasets/_fetch_bank_marketing.py
```diff
@@ -33,28 +33,29 @@ def fetch_bank_marketing(*, cache=True, data_home=None,
...
-    cache : boolean, default=True
+    cache : bool, default=True
         Whether to cache downloaded datasets using joblib.
 
-    data_home : optional, default: None
-        Specify another download and cache folder for the datasets. By default
-        all scikit-learn data is stored in '~/.fairlearn-data' subfolders.
+    data_home : str, default=None
+        Specify another download and cache folder for the datasets.
+        By default, all scikit-learn data is stored in '~/.fairlearn-data'
```
adrinjalali, line 41:
```suggestion
        By default, all data is stored in '~/.fairlearn-data'
```

maybe?

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -109,8 +109,10 @@ class ThresholdOptimizer(BaseEstimator, MetaEstimatorMixin):
 
     Parameters
     ----------
-    estimator : estimator object implementing 'predict' and possibly 'fit'
-        An estimator whose output is postprocessed.
+    estimator : object
+        An estimator whose output is postprocessed. The estimator should be
+        implementing a :code:`predict` method. If :code:`prefit=False`, it
+        also needs to implement a :code:`fit` method.
```
adrinjalali, line 115: It's probably easier to say "a scikit-learn compatible estimator". Since there's more to it than fit and predict for it to be something which can then be used in a `Pipeline` or a `GridSearchCV` for instance.

adrinjalali (commented):
otherwise LGTM, thanks @romanlutz

fairlearn/reductions/_moments/utility_parity.py
```diff
@@ -197,8 +198,11 @@ def gamma(self, predictor):
     def bound(self):
         """Return bound vector.
 
-        :return: a vector of bound values corresponding to all constraints
-        :rtype: pandas.Series
+        Returns
+        -------
+        pandas.Series
```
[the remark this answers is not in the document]
romanlutz: So here's the funny thing: in return blocks it renders fine as is without any directive, but elsewhere it doesn't.

docs/contributor_guide/contributing_code.rst
romanlutz: I checked every single one yesterday because I made most of these months ago and didn't trust them 😅
It's just for splitting it over three lines to prevent the table from getting into a weird shape

docs/user_guide/mitigation.rst
romanlutz: It's because the table columns have only very limited space. If I don't have that the other columns become unreadable. This splits the link over three lines (which we've had like that for over a year).

adrinjalali: Do we really want the whole `fairlearn.reductions.ExponentiatedGradient` path? We could have only the `ExponentiatedGradient` displayed and the link would tell the users where the object is:

```suggestion
   *  - :class:`~fairlearn.reductions.ExponentiatedGradient`
```

Or   

```
- :class:~`fairlearn.reductions.ExponentiatedGradient`
```

fairlearn/postprocessing/_threshold_optimizer.py
MiroDudik: I'm fine either way, but I'm not sure whether we require an estimator. With `prefit=True`, I think it just needs to be an object that supports predict.

docs/conf.py
romanlutz: I couldn't quite remember why I did this a few months ago, so I removed the `.md`  portion and rebuilt. Turns out nothing much happens... so I'll remove it.

docs/faq.rst
romanlutz: Done!

docs/user_guide/mitigation.rst
romanlutz: Going with the latter.

fairlearn/postprocessing/_threshold_optimizer.py
romanlutz: I was thinking along the same lines as @MiroDudik here. I think in scikit-learn terms we require
- a predictor if `prefit=True`
- an estimator if `prefit=False`

https://scikit-learn.org/stable/developers/develop.html?highlight=predictor#different-objects

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -251,15 +253,16 @@ def predict(self, X, *, sensitive_features, random_state=None):
...
+        random_state : int or :class:`numpy.random.RandomState` instance, default=None
             Controls random numbers used for randomized predictions. Pass an
             int for reproducible output across multiple function calls.
 
         Returns
         -------
-        Scalar or vector as numpy.ndarray
-            The prediction. If `X` represents the data for a single example
-            the result will be a scalar. Otherwise the result will be a vector
+        numpy.ndarray
```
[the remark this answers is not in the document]
romanlutz: This renders fine without. I believe the "Returns" type is always resolved even without directive.

fairlearn/postprocessing/_threshold_optimizer.py
adrinjalali: I would say  we just say sklearn compatible, and only officially support that.

romanlutz: Replaced as you suggested! This should be future-proof 🙂

romanlutz:
Any other concerns @hildeweerts @MiroDudik ?
I addressed @adrinjalali's remaining comment, so I think this can be merged unless there's anything else 🙂

Issue #872, opened 2021-06-22
DOC: remove duplicates index example notebooks


hildeweerts:
#### Describe the issue linked to the documentation

The lefthand side index of our [Example Notebooks](https://fairlearn.org/main/auto_examples) contains duplicate items. I think this is because some are [in the `toctree` of README.rst](https://github.com/fairlearn/fairlearn/edit/main/examples/README.rst) as well as in the automatically generated examples.

#### Suggest a potential alternative/fix

Remove the items from README.rst and only show automatically generated examples.

@fairlearn/fairlearn-maintainers 

*[if you want to work on to this issue, please let us know in the comments.]*

Fyrebright:
Hello, I have made the change and verified it a local build. I will go ahead and make a PR.

romanlutz:
#873  was merged, so I'm closing this issue.

PR #873, opened 2021-06-22
DOC Remove redundant toctree in gallery header (#872)

Fyrebright:
The left example navbar is now autogenerated by sphinx-gallery so leaving this section  is causing duplicates in the navbar.

### DIFF
--- a/examples/README.rst
+++ b/examples/README.rst
@@ -12,12 +12,3 @@ adding new ones please consult the guide on
 
     The Fairlearn API is still evolving, so if you want to run these
     on your local Fairlearn installation, make sure to match versions.
-
-.. toctree::
-   :maxdepth: 2
-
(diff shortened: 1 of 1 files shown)


romanlutz:
@Fyrebright thanks for the PR! Could you post a screenshot of what it looks like now?

hildeweerts:
I expected the list to disappear from the body so for me this is perfect @Fyrebright! Just to be sure: the build you're showing was created using `make doc` command right, not multi-version (because the version selection is missing)?

If @romanlutz agrees as well I am happy to merge.

hildeweerts (approved):

romanlutz (approved):
Thanks for adding the screenshot! This looks fine. Good to know we don't need to manually maintain the toctree.

PR #876, opened 2021-06-25
Updating the test_tensorflow.py file in the test_othermlpackages directory to use the newer wrappers

### LINKED ISSUE #814: DOC: add example for using tensorflow

We're not super clear about supporting any kind of model for our (currently existing reductions and postprocessing) mitigation techniques. This means people don't know they can use tensorflow with Fairlearn.

We should have a short example that shows how to do that. Perhaps having it as a `plot_tensorflow.py` under `examples` and linking to it from a user guide section makes sense. We've used `literalinclude` for this purpose in parts of the codebase. I am not entirely sure whether including tensorflow into requirements-dev.txt will cause issues, though.

The alternative: Just put a `code-block`, but then the code isn't run during the doc build and we're at risk of this code getting outdated. 

Related: #813

pritishaw1:
This pull request addresses issue #814 and contains a file name "plot_tensorflow.py", under the examples.


examples/plot_tensorflow.py
```diff
@@ -0,0 +1,134 @@
...
+=================================
+An example for tensorflow
+=================================
+"""
+# %%
+
+import pandas as pd
+from sklearn.datasets import fetch_openml
+from sklearn.metrics import accuracy_score, precision_score, recall_score
+import scikeras
```
romanlutz, line 14: This won't quite work under `examples` since it would mean we have to add `scikeras`, `seaborn`, and `tensorflow` to `requirements-dev.txt` which we can't really do. I added another comment in the corresponding issue that may help guide this in the intended direction.

romanlutz (changes requested):

examples/plot_tensorflow.py
pritishaw1: Yes, just to make sure that the dependencies do not change, because as you said that adding these new packages to requirements-dev.txt is not feasible, we can go by adding a code-block for this under mitigation.rst.

I will go with this and commit the new changes soon.

riedgar-ms:
This is something which might be useful in our `test_othermlpackages` directory, though. I used `from tensorflow.keras.wrappers.scikit_learn import KerasClassifier` rather than `scikeras`

pritishaw1:
> This is something which might be useful in our `test_othermlpackages` directory, though. I used `from tensorflow.keras.wrappers.scikit_learn import KerasClassifier` rather than `scikeras`

@riedgar-ms I have used `scikeras` instead of 'from tensorflow.keras.wrappers.scikit_learn import KerasClassifier' because the TensorFlow wrapper utilities are not actively maintained and will be removed in a future release, according to [this](https://github.com/tensorflow/tensorflow/pull/36137#issuecomment-726271760). 

And also scikeras has got some [advantages](https://scikeras.readthedocs.io/en/latest/migration.html) over the TF wrapper.

riedgar-ms:
In that case, would it be possible for you to update the test case there?

pritishaw1:
@riedgar-ms Pardon me if I'm wrong, but from the test case, did you mean adding up comments with reference to the libraries used and the actions performed throughout?

If not, could you please elaborate a little on what you are actually expecting?

riedgar-ms:
I meant updating this file:
https://github.com/fairlearn/fairlearn/blob/main/test_othermlpackages/test_tensorflow.py
and its matching conda YAML, if you are willing. I do not follow TensorFlow closely, and hadn't realised that the wrappers I was using were being deprecated.

pritishaw1:
@riedgar-ms Yeah Sure, I'll be happy to update the existing file for `test_tensorflow.py` in the `test_othermlpackages` directory and also its matching conda YAML.

I'll work it out soon and update the PR, as early as possible.

riedgar-ms:
> @riedgar-ms Yeah Sure, I'll be happy to update the existing file for `test_tensorflow.py` in the `test_othermlpackages` directory and also its matching conda YAML.
> 
> I'll work it out soon and update the PR, as early as possible.

Thanks! That will be really helpful.

pritishaw1:
@riedgar-ms I have updated the existing file for `test_tensorflow.py` along with its matching conda YAML with dependencies on scikeras for using the KerasClassifier wrapper.

Do check it out and let me know if anything needs to be changed.

examples/plot_tensorflow.py
```diff
@@ -0,0 +1,134 @@
...
+
+data = fetch_openml(data_id=1590, as_frame=True)
+X_raw = data.data
+Y = (data.target == '>50K') * 1
+sex = data.data["sex"]
+
+X = X_raw.drop(labels=['sex'], axis=1)
+X = pd.get_dummies(X)
+
+def get_clf(meta, hidden_layer_sizes, dropout):
```
riedgar-ms, line 47: It would be good to have some explanation as to what's going on here, if you want to keep this example separate.

examples/plot_tensorflow.py
```diff
@@ -0,0 +1,134 @@
...
+    'recall': recall_score,
+    'false positive rate': false_positive_rate,
+    'true positive rate': true_positive_rate,
+    'selection rate': selection_rate,
+    'count': lambda Y, y_pred: len(Y),
+    'average precision':skm.average_precision_score,
+    'roc-auc': skm.roc_auc_score,
+}
+
+metric_frame = MetricFrame(metrics,
```
riedgar-ms, line 82: Given that `MetricFrame` only cares about a column of inputs, I'm not really sure that this is really showing TensorFlow being used with `fairlearn`. For that, I think you'd need to show it working with the mitigation algorithms

examples/plot_tensorflow.py
```diff
@@ -0,0 +1,134 @@
...
+# %%
+
+import pandas as pd
+from sklearn.datasets import fetch_openml
+from sklearn.metrics import accuracy_score, precision_score, recall_score
+import scikeras
+import warnings
+from tensorflow import get_logger
+import numpy as np
+from scikeras.wrappers import KerasClassifier
```
riedgar-ms, line 18: Are these actually available in the doc build? I thought that TensorFlow was only installed for the `otherpackages` tests

test_othermlpackages/test_tensorflow.py
```diff
@@ -1,15 +1,19 @@
 # Copyright (c) Microsoft Corporation and Fairlearn contributors.
 # Licensed under the MIT License.
 
+
 import pytest
 from . import package_test_common as ptc
+import scikeras
```
riedgar-ms, line 7: Is this actually used?

riedgar-ms (commented):
Thank you for starting work on the test update!

examples/plot_tensorflow.py
pritishaw1: @riedgar-ms Sure, I will add some comments that would explain the purpose of the code line, in `plot_tensorflow.py`.

I did not do this yet because I am confused if this example would be used for the original directory or not because it has got some libraries that are not aligning with the dependencies which are already there as mentioned in `requirements-dev.txt`. And as per the earlier discussions, I see that the admins do not wish to add dependencies upon the newly used libraries.

examples/plot_tensorflow.py
pritishaw1: Yeah, alright then, I will soon try working it out on the mitigation algorithms.

examples/plot_tensorflow.py
pritishaw1: Actually yes, Tensorflow is not aligning with the dependencies which are already there in `requirements-dev.txt`.

test_othermlpackages/test_tensorflow.py
pritishaw1: Yes, scikeras has been used to import KerasClassifier, which was previously being called with the help of `tf.keras.wrappers.scikit_learn.KerasClassifier`.

examples/plot_tensorflow.py
riedgar-ms: I think the question of example notebooks like that is a bit bigger, since there is the question of having TensorFlow installed (which we don't usually). I would suggest removing this file from the PR for now, and making this PR about updating the `test_tensorflow.py` file in the `test_othermlpackages` directory to use the newer wrappers.

test_othermlpackages/test_tensorflow.py
riedgar-ms: OK. Perhaps you could move this adjacent to the `scikeras.wrappers` import below?

riedgar-ms (commented):
Thanks for your work!

examples/plot_tensorflow.py
pritishaw1: Alright! I will do as suggested and keep this PR only to update the `test_tensorflow.py` file in the `test_othermlpackages` directory.

test_othermlpackages/test_tensorflow.py
pritishaw1: @riedgar-ms Okay. I'll remove this and only keep the import of `scikeras.wrappers` in the updated PR.

romanlutz:
> I think the question of example notebooks like that is a bit bigger, since there is the question of having TensorFlow installed (which we don't usually). I would suggest removing this file from the PR for now, and making this PR about updating the test_tensorflow.py file in the test_othermlpackages directory to use the newer wrappers.

I agree with @riedgar-ms . Plus, we can use `literalinclude` to reference lines from the test case in the docs to show how to do this without actually duplicating code. Doesn't need to happen in this PR, but if you want to try it, you can, @pritishaw1

pritishaw1:
Hi @riedgar-ms, I have updated this PR and kept it only for updating the test_tensorflow.py file with the scikeras wrappers.

Let me know if there's anything else needed for this PR or any other contribution towards the fairlearn library, as a whole. I will be happy to hear back from you and extend help if needed.

Thanks!

riedgar-ms:
Sorry I didn't get back to you sooner @pritishaw1 . You're running into some flake8 errors:
https://dev.azure.com/responsibleai/fairlearn/_build/results?buildId=12891&view=logs&jobId=7f33e5bd-7764-5d8a-ba2e-506e078b9c3f&j=7f33e5bd-7764-5d8a-ba2e-506e078b9c3f&t=b6623cb2-2223-5dfe-fa7b-0ad47bab4310
Basically, you need to move all imports to the top of the file, and remove unused imports.

pritishaw1:
@riedgar-ms I'll look at the error and get back soon.

test_othermlpackages/test_tensorflow.py
```diff
@@ -9,7 +9,8 @@
 tf = pytest.importorskip("tensorflow")
 from tensorflow.keras.layers import Dense # noqa
 from tensorflow.keras.models import Sequential # noqa
-from tensorflow.keras.wrappers.scikit_learn import KerasClassifier # noqa
+import scikeras
```
riedgar-ms, line 12: Is this specific import actually used? `flake8` is still complaining

riedgar-ms: Are you able to see the build reports @pritishaw1 ?

pritishaw1: Yes @riedgar-ms, the highlighted import was not being used so I have removed that in the updated PR.

riedgar-ms:
Thanks for the update. As I hope you can see, there are still failures. First, a `flake8` one:
./test_othermlpackages/test_tensorflow.py:12:1: E402 module level import not at top of file

The pipeline for this test is also failing when building the conda environment. Is there an extra package source which needs to be added to the conda file?

pritishaw1:
Yes, the failures still remain, but I am clueless why is it happening again. All the import files included have been made to use and also for the conda file, all the dependencies are the same as before, except for the scikeras package.

@riedgar-ms  I have a doubt, as fairlearn is not having any dependencies on scikeras, so should it also be imported using `pytest.importorskip`, the way it is done for tensorflow? Maybe that is why it is running into a `flake8` error.

riedgar-ms:
@pritishaw1 it looks like `scikeras` is only available on pip. So the conda file needs to have something like:
```yaml
dependencies:
 - python==3.7
 - # etc.
  - pip:
    - scikeras
```

Now I've looked more closely, for the `flake8` error, you just need to put `# noqa` on the end of the line. The reason is that the `importorskip()` counts as executable code, meaning that the remaining imports aren't at the 'top' of the file.

We use `importorskip()` so that if you accidentally run the tests in an environment without the TensorFlow (or XGBoost, or LightGBM etc.) packages installed, then the test is skipped rather than giving a wall of import exceptions.

@romanlutz we might want to consider:
https://pypi.org/project/flake8-pytest-importorskip

pritishaw1:
@riedgar-ms I have updated the PR, as you have suggested but still one of the checks is failing. Can you please check this?

riedgar-ms:
@pritishaw1 I saw that too. Do you mind if I try editing your PR directly?

pritishaw1:
@riedgar-ms It's absolutely fine. You can go ahead and edit the PR.

pritishaw1:
@riedgar-ms It is absolutely fine, you can go ahead and edit the PR and you can also tell me if it requires an edit on my end.

riedgar-ms:
@pritishaw1 I think I've got it - `scikeras` needed a TensorFlow update, which in turn needed Python 3.8. Not sure why the PR Gate didn't run, though

test_othermlpackages/conda-tensorflow.yaml
```diff
@@ -3,7 +3,9 @@ channels:
 - defaults
 - conda-forge
 dependencies:
-- python==3.7
-- pip>=19.1.1
```
romanlutz, line 7: Why did we remove this lower bound? Seems sensible to have one.

romanlutz (approved):

test_othermlpackages/conda-tensorflow.yaml
riedgar-ms: Restored. It was just one of the things I tried before I discovered the actual issue with TF

riedgar-ms (approved):
Thank you for your patience with this @pritishaw1

pritishaw1:
Thanks, @riedgar-ms, and @romanlutz. It was great collaborating with the team of fairlearn.

PR #877, opened 2021-06-26
FIX add utils module in 0.7.0 release

romanlutz:
#859 added a `utils` directory without `__init__.py` file which made the release pipeline break. This addition fixes it so that we can proceed with release v0.7.0. The same change will be merged into `main` in a separate PR (with a few additional small adjustments).

By pulling in changes from main to get the #878 changes into this PR I used the wrong command (merge instead of cherry-pick) which added PRs #873 #787  #871 and #806 to this PR. Since they all improve the docs this means that our doc page for v0.7.0 will be improved which I consider to be a bonus. Otherwise people won't see those changes until the next release (except if they use the `main` version of the website). If someone has objections please voice them in this thread!

@fairlearn/fairlearn-maintainers please review.

Signed-off-by: Roman Lutz <rolutz@microsoft.com>

PR #878, opened 2021-06-27
FIX create utils module

romanlutz:
This adds an `__init__.py` file just like #877 on the release branch, but additionally moves other utilities (input validation) into the module. The move requires a few import adjustments which are also included.


fairlearn/reductions/_moments/utility_parity.py
```diff
@@ -6,7 +6,8 @@
 
 from .moment import ClassificationMoment
 from .moment import _GROUP_ID, _LABEL, _PREDICTION, _ALL, _EVENT, _SIGN
-from fairlearn._input_validation import _MESSAGE_RATIO_NOT_IN_RANGE, _validate_and_reformat_input
+from fairlearn.utils._input_validation import _MESSAGE_RATIO_NOT_IN_RANGE, \
+    _validate_and_reformat_input
```
adrinjalali, line 10:
```suggestion
from fairlearn.utils._input_validation import (
    _MESSAGE_RATIO_NOT_IN_RANGE,
    _validate_and_reformat_input
)
```

Or even two separate imports :D

fairlearn/utils/__init__.py
```diff
@@ -0,0 +1,4 @@
+# Copyright (c) Microsoft Corporation and Fairlearn contributors.
```
adrinjalali, line 1:
```suggestion
# Copyright (c) Fairlearn contributors.
```

maybe? didn't we change it to this?

test/unit/postprocessing/test_threshold_optimization.py
```diff
@@ -7,7 +7,7 @@
 from copy import deepcopy
 import numpy as np
 import pytest
-from fairlearn._input_validation import \
+from fairlearn.utils._input_validation import \
     (_MESSAGE_Y_NONE,
```
adrinjalali, line 11:
```suggestion
from fairlearn.utils._input_validation import (
    _MESSAGE_Y_NONE,
```

adrinjalali (approved):
Some nits which you can ignore if you don't like the style, otherwise LGTM.

I'd cherry pick this for the branch rather than merging the other PR.

fairlearn/utils/__init__.py
hildeweerts: According to Miro's comment [here](https://github.com/fairlearn/fairlearn/pull/823#issuecomment-857661650)  removing existing copyright from the source code might be in conflict with the MIT license.

adrinjalali: Not if all contributors agree + this is new code with be license

hildeweerts: Good points, in that case I'm very happy to change it to Fairlearn contributors :)

riedgar-ms (approved):

PR #881, opened 2021-07-02
FIX Update documentation in _fetch_boston.py

koaning:
I noticed this issue on scikit-learn https://github.com/scikit-learn/scikit-learn/issues/19657 and figured it makes sense to apply it here too.

### DIFF
--- a/fairlearn/datasets/_fetch_boston.py
+++ b/fairlearn/datasets/_fetch_boston.py
@@ -50,7 +50,7 @@ def fetch_boston(*, cache=True, data_home=None,
     RAD      index of accessibility to radial highways
     TAX      full-value property-tax rate per $10,000
     PTRATIO  pupil-teacher ratio by town
-    B        1000(Bk - 0.63)^2 where Bk is the proportion of black people by town
+    B        1000(Bk - 0.63)^2 where Bk is the proportion of Black people by town
     LSTAT    % lower status of the population
     MEDV     Median value of owner-occupied homes in $1000's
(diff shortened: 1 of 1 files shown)


adrinjalali (approved):
Thanks @koaning

fairlearn/datasets/_fetch_boston.py
```diff
@@ -50,7 +50,7 @@ def fetch_boston(*, cache=True, data_home=None,
     RAD      index of accessibility to radial highways
     TAX      full-value property-tax rate per $10,000
     PTRATIO  pupil-teacher ratio by town
-    B        1000(Bk - 0.63)^2 where Bk is the proportion of blacks by town
+    B        1000(Bk - 0.63)^2 where Bk is the proportion of black people by town
```
romanlutz, line 53: Can we capitalize Black? I forget where I read about this but the first thing that showed up in my search was https://apastyle.apa.org/style-grammar-guidelines/bias-free-language/racial-ethnic-minorities

```suggestion
    B        1000(Bk - 0.63)^2 where Bk is the proportion of Black people by town
```

romanlutz: I'll commit this since you gave "thumbs up" 🙂

romanlutz (approved):

adrinjalali:
I don't think we should be blocking merge if a test fails. Sometimes like here, we know the fail is irrelevant and would merge regardless.

PR #879, opened 2021-06-27
MNT update version references post v0.7.0 release

romanlutz:
This is the usual update for all mentions of the current version according to the release guide.
Additionally, this changes which versions are shown on the website. Instead of using just wildcards we will only publish docs for latest patch version of every minor version (excluding 0.6.0 and 0.6.1 since 0.6.2 exists, for example). This is basically implementing the approach proposed in https://github.com/fairlearn/fairlearn/discussions/659#discussioncomment-766177 since it seemed to resonate with people (or at least @hildeweerts and @riedgar-ms). 

@fairlearn/fairlearn-maintainers

Signed-off-by: Roman Lutz <rolutz@microsoft.com>

PR #882, opened 2021-07-02
MNT Cherry-pick commits from main onto v0.7.0 release branch

romanlutz:
Specifically #878 is what we're looking for to fix the release pipeline. However, the other commits are also useful since they improve the docs.

This branch has only cherry-picks and does not include changes. The target branch already includes the version changes needed to do the release, so I can kick off the release pipeline as soon as this is merged.


romanlutz:
@adrinjalali this should be better than #879 🙂 Thanks for suggesting better practices!

MiroDudik (approved):
This looks good to me (I cannot verify that these are exact cherry picks, but they look sensible).

romanlutz:
> This looks good to me (I cannot verify that these are exact cherry picks, but they look sensible).

Yes, it's a bit tedious... 

Commits on main since the release branch was created (the bottom one in this list is already on release/v0.7.0):
image
[taken from https://github.com/fairlearn/fairlearn/commits/main]

Commits cherry-picked in this PR:
image

Note that the commit IDs on this branch are different because the release branch has 1 extra commit to adjust the versions, so this is to be expected.

adrinjalali (approved):

romanlutz:
The MacOS pipeline is failing with 
```
> assert disparity == pytest.approx(data["disp"], abs=_PRECISION) 
E assert 0.020000000000000018 == 0.019999 ± 1.0e-06 
E + where 0.019999 ± 1.0e-06 = <function approx at 0x106fdbd40>(0.019999, abs=1e-06) 
E + where <function approx at 0x106fdbd40> = pytest.approx 
```
Pytest, scipy, numpy all didn't do a release in the past few days, but pandas did. Not sure how that would have caused this but I'll look into it.

Also, I check the "compare" view that GitHub provides and it seems to either have a bug or these cherry-picks don't work the way I think they do: https://github.com/fairlearn/fairlearn/compare/main...romanlutz:rolutz/cherry_pick_070
Many files are reported as being different even though the contents are the same on the branches. What am I missing? @adrinjalali @riedgar-ms any idea?

MiroDudik:
@romanlutz : the failing test looks actually very suspicious. Given that our "ground truth results" in ` smoke_test_data` seem to be just floors at 6 digits (rather than rounded at 6 digits), we should change `_PRECISION = 1e-6` to `_PRECISION = 1e-5`. since this is failing on a mac, but not on other installations, i'm wondering whether this could be actually related to something extremely low level  (64 bits vs 32 bit arithmetic, or some low-level library implementations stuff), or possibly even hardware related (i seem to remember that some intel processors had some floating point handling bugs).

romanlutz:
FYI I added the cherry-pick of #883 to this PR. This should fix the PR gate and allow me to proceed. Since @adrinjalali and @MiroDudik were closely involved and approved #883 as well I'll merge this once PR gate passes.

PR #889, opened 2021-07-08
Fix comment in postprocessing API documentation #831

### LINKED ISSUE #831: Fix comment in postprocessing API documentation
Hello! Line249 of fairlearn/fairlearn/postprocessing/_tradeoff_curve_utilities.py accumulates 
`false_positives=count[0]
true_positives=count[1]
`
Shouldn't they be true_negatives=count[0] and false_negatives=count[1]? (since you are accumulating in ascending order how many counts are < threshold, thus yhat= 0 which conflicts with definition on lines244 and 245(P[Y_hat=1 | Y=0], P[Y_hat=1 | Y=1])?

I think this is fine if ROC of two classes's ROC curves intersect, but would be problematic when one class needs to be randomized?

bram49:
The comments ascending/increasing should have been descending/decreasing.

Closes #831

### DIFF
--- a/fairlearn/postprocessing/_tradeoff_curve_utilities.py
+++ b/fairlearn/postprocessing/_tradeoff_curve_utilities.py
@@ -68,7 +68,7 @@ def _extend_confusion_matrix(*, true_positives, false_positives, true_negatives,
 
 def _tradeoff_curve(data, sensitive_feature_value, flip=False,
                     x_metric="false_positive_rate", y_metric="true_positive_rate"):
-    """Get a convex hull of achievable tradeoffs between the two provided metrics.
+    """Get a convex hull of achievable trade-offs between the two provided metrics.
 
     The metrics are based on considering all possible thresholds of 'score' column of `data` and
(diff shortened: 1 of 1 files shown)


fairlearn/postprocessing/_tradeoff_curve_utilities.py
```diff
@@ -68,7 +68,7 @@ def _extend_confusion_matrix(*, true_positives, false_positives, true_negatives,
 
 def _tradeoff_curve(data, sensitive_feature_value, flip=False,
                     x_metric="false_positive_rate", y_metric="true_positive_rate"):
-    """Get a convex hull of achievable tradeoffs between the two provided metrics.
+    """Get a convex hull of achievable trade-offs between the two provided metrics.
```
hildeweerts, line 71: I think tradeoff is also accepted in US English, but I'm happy either way :)

hildeweerts (approved):
Thank you for fixing this, @bram49!

PR #896, opened 2021-07-13
FIX Added error handling in MetricFrame

### LINKED ISSUE #800: Stop MetricFrame errors on nonscalar metrics

`MetricFrame` is quite relaxed as to the return values of the metric functions it evaluates. For example, it is perfectly possible to return a confusion matrix or a confidence bound from a metric function.

However such metrics will cause a problem if the user tries calling `group_min()`, `group_max()`, `difference()` or `ratio()` on the resultant `MetricFrame` since those operations are only well defined for scalars.

Currently, we (or rather `pandas`) will throw an exception. This is defensible if only a single metric is being computed by the `MetricFrame` but is likely to lead to a lack of user delight if there are other metrics present for which those operations do make sense. In particular, if a user has constructed a `MetricFrame` consisting of both the `accuracy_score` and `accuracy_score_confidence_interval` then they would reasonably expect `group_min()` to give a useful answer for the `accuracy_score` column.

Catch any exceptions from `group_min()` and `group_max()` and emplace `NaN` in this case.

#### Describe alternatives you've considered, if relevant

Without this, users are forced to split up related metrics into separate `MetricFrame` objects.

#### Additional context

This is important for #235

bram49:
closes #800

Added error argument that can be {'ignore', 'coerce', 'raise'} (copied from [pandas.to_numeric](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.to_numeric.html))
- 'coerce' will except the error and return NaN
- 'ignore' will except the error and return the metric_frame._by_group *
- 'raise' will raise the error

With default values:
- .difference() and .ratio() have default 'ignore'
- .group_max() and .group_min() have default 'raise' **

'*' There might be a better option to return for ignore
'**' Not sure if this should default to 'raise', but thought this would be the safest option to not break dependencies. 

Throws warning message on error
Checks if errors parameter got valid input
Docstrings were added


fairlearn/metrics/_metric_frame.py
```diff
@@ -382,19 +382,41 @@ def group_max(self) -> Union[Any, pd.Series, pd.DataFrame]:
...
+        except ValueError:
+            warnings.warn(f"Metric frame contains non-scalar cells. "
+                          f"Please remove non-scalar columns from your metric frame and try again",
+                          SyntaxWarning)
+            if errors == "raise":
+                raise
+            elif errors == 'coerce':
+                return np.nan
+            else:
+                return self._by_group
```
romanlutz, line 419: I like this a lot. However, I'm not 100% sure I understand when I'd want to have it return `self._by_group`. Can you imagine a situation where that's useful?

romanlutz:
We should probably have test cases for all three options, too.

fairlearn/metrics/_metric_frame.py
hildeweerts: Good question. I *could* imagine that if you prepare a generic pipeline that could deal with metrics for which difference makes sense and metrics for which it doesn't (e.g., a confusion matrix), this could be useful. But that's probably such a niche scenario that we might be better off limiting the options to 'coerce' and 'raise'. 

WDYT?

bram49: I thought it would be useful since the warning message says "some columns contain non-scalar cells", that the user gets to see the columns as output. But I do not see any other usecase then that.

bram49:
> We should probably have test cases for all three options, too.

All right! I will work on that today

fairlearn/metrics/_metric_frame.py
bram49: Shall I change it to only have 'raise' and 'coerce'? And then have 'coerce' as default for .difference() and .ratio()?

fairlearn/metrics/_metric_frame.py
```diff
@@ -382,19 +382,41 @@ def group_max(self) -> Union[Any, pd.Series, pd.DataFrame]:
...
+                result = self._by_group.groupby(level=self.control_levels).max()
+        # Catch ValueError that is thrown by pandas, if not all columns in metric frame are scalar
+        except ValueError:
+            warnings.warn(f"Metric frame contains non-scalar cells. "
+                          f"Please remove non-scalar columns from your metric frame and try again",
+                          SyntaxWarning)
+            if errors == "raise":
+                raise
+            elif errors == 'coerce':
+                return np.nan
```
riedgar-ms, line 416: Don't these `else` clauses need to do different things based on control_levels? Also, how does this behave if some columns in the DataFrame can work with `max()` but others can't?

fairlearn/metrics/_metric_frame.py
riedgar-ms: And (as alluded above) what do you do when there's one column which is a bunch of floats, and another column which consists of confusion matrices?

riedgar-ms (commented):
Very much second the request for tests

test/unit/metrics/test_metricframe_aggregates.py
```diff
@@ -1040,3 +1040,105 @@ def test_2m_1sf_2cf():
         assert ratios_overall[mfn][('m', 'aa')] == min(ratio_overall_m_a)
         assert ratios_overall[mfn][('m', 'ba')] == pytest.approx(min(ratio_overall_m_b),
                                                                  rel=1e-10, abs=1e-16)
+
+
+class Test2m1sf1cfErrorHandlingCM:
+    # Metricframe containing a confusion matrix raises ValueErrors, since cells are non-scalar
+    def _prepare(self):
+        fns = {'confusion_matrix': skm.confusion_matrix}
```
riedgar-ms, line 1048: As mentioned above, there should also be a variant with `fns = { 'recall' : skm.recall_score, 'confusion_matrix': skm.confusion_matrix) }` so that we cover the case where some columns work, but others don't

test/unit/metrics/test_metricframe_aggregates.py
```diff
@@ -1040,3 +1040,105 @@ def test_2m_1sf_2cf():
...
+    def test_min_default(self):
+        # default is 'raise'
+        self._prepare()
+        with pytest.raises(ValueError):
+            self.target.group_min()
+
+    @pytest.mark.parametrize("errors", ["raise", "coerce"])
+    def test_max(self, errors):
+        self._prepare()
+        if errors == "raise":
```
riedgar-ms, line 1086: I'm not a huge fan of turning parts of a test on and off like this - I think it would be better to write out the two test cases, given that the only shared code is `self._prepare()`

test/unit/metrics/test_metricframe_aggregates.py
```diff
@@ -1040,3 +1040,105 @@ def test_2m_1sf_2cf():
...
                                                                  rel=1e-10, abs=1e-16)
+
+
+class Test2m1sf1cfErrorHandlingCM:
+    # Metricframe containing a confusion matrix raises ValueErrors, since cells are non-scalar
+    def _prepare(self):
+        fns = {'confusion_matrix': skm.confusion_matrix}
+        self.target = metrics.MetricFrame(metrics=fns, y_true=y_t, y_pred=y_p,
+                                          sensitive_features=g_2,
+                                          control_features=g_3)
```
riedgar-ms, line 1051: There should also be checks with and without control_features

fairlearn/metrics/_metric_frame.py
bram49: Ah yes! That would be better behavior, that coerce would return all columns that do work with max, min, difference and ratio. Will update this

test/unit/metrics/test_metricframe_aggregates.py
bram49: Yes it made more sense when there was also the ignore option in there. I will make the tests separate.

test/unit/metrics/test_metricframe_aggregates.py
bram49: Can I alter the existing tests to have an extra metric, or do you prefer to both have tests with two metrics and tests with only the CM metric?

bram49: I altered the tests to have an extra metric. Adding tests with only consumer_matrix and errors='coerce', always leads to empty series, so I did leave those tests out.

bram49:
Thank you for the reviews! I implemented the feedback, please let me know what you think.

fairlearn/metrics/_metric_frame.py
romanlutz: @riedgar-ms I'm not entirely sure if you mean that as an argument for or against having this third option 🙂 
If we have a metric for which difference (or ratio, etc.) doesn't make sense, then why are we returning `self._by_group`? We could return `np.nan` perhaps but then that's the `coerce` path.

test/unit/metrics/test_metricframe_aggregates.py
```diff
@@ -1040,3 +1040,195 @@ def test_2m_1sf_2cf():
...
+
+        self.expected = metrics.MetricFrame(metrics={'recall': skm.recall_score}, y_true=y_t, y_pred=y_p,
+                                            sensitive_features=g_2,
+                                            control_features=g_3)
+
+    def test_min_raise(self):
+        self._prepare()
+        with pytest.raises(ValueError):
```
riedgar-ms, line 1059: Can you check the message in the ValueError, just to make sure it's the right error (lots of things will throw a ValueError)

test/unit/metrics/test_metricframe_aggregates.py
```diff
@@ -1040,3 +1040,195 @@ def test_2m_1sf_2cf():
...
+
+class Test2m1sf1cfErrorHandlingCM:
+    # Metricframe containing a confusion matrix raises ValueErrors, since cells are non-scalar
+    def _prepare(self):
+        fns = {'recall': skm.recall_score, 'confusion_matrix': skm.confusion_matrix}
+        self.target = metrics.MetricFrame(metrics=fns, y_true=y_t, y_pred=y_p,
+                                          sensitive_features=g_2,
+                                          control_features=g_3)
+
+        self.expected = metrics.MetricFrame(metrics={'recall': skm.recall_score}, y_true=y_t, y_pred=y_p,
```
riedgar-ms, line 1053: If this is being used how I think it is, then what you're doing is _dropping_ any columns which don't respond well to `min()`, rather than coercing to NaN?

fairlearn/metrics/_metric_frame.py
```diff
@@ -382,19 +382,46 @@ def group_max(self) -> Union[Any, pd.Series, pd.DataFrame]:
...
         """
-        if not self.control_levels:
-            result = pd.Series(index=self._by_group.columns, dtype='object')
-            for m in result.index:
-                max_val = self._by_group[m].max()
-                result[m] = max_val
-        else:
-            result = self._by_group.groupby(level=self.control_levels).max()
+        if errors not in ("raise", "coerce"):
+            raise ValueError("invalid error value specified")
```
riedgar-ms, line 398: Can you list out the valid values in the message? Also, can you make these strings constants at the top of the file

riedgar-ms (commented):
Does the actual behaviour match that specified in the docstrings when `coerce` is set?

test/unit/metrics/test_metricframe_aggregates.py
bram49: Allright will do

test/unit/metrics/test_metricframe_aggregates.py
bram49: Aj yes you are indeed right. The behaviour now is that it will ignore columns that are non-scalar, but it indeed has to output NaN. Will edit this tomorrow.

fairlearn/metrics/_metric_frame.py
bram49: Yes will do

bram49:
Implemented the feedback!
errors='coerce' now outputs nan, for all non-scalar columns. Edited the corresponding test cases and the error messages are now constants at the top of the file.

fairlearn/metrics/_metric_frame.py
```diff
@@ -29,6 +29,9 @@
 _TOO_MANY_FEATURE_DIMS = "Feature array has too many dimensions"
 _SAMPLE_PARAM_KEYS_NOT_IN_FUNC_DICT = \
     "Keys in 'sample_params' do not match those in 'metric'"
+_INVALID_ERRORS_VALUE_ERROR_MESSAGE = "Invalid error value specified"
```
riedgar-ms, line 32: Please specify the valid values - it might be helpful to define another constant along the lines of:
```python
_VALID_ERROR_STRINGS = [ 'raise', 'coerce' ]
```
and reference that in strings where needed (and in the code where the check is made).

fairlearn/metrics/_metric_frame.py
```diff
@@ -382,19 +385,57 @@ def group_max(self) -> Union[Any, pd.Series, pd.DataFrame]:
...
+                warnings.warn(_MF_CONTAINS_NON_SCALAR_WARNING,
+                              SyntaxWarning)
+                if errors == "raise":
+                    raise
+                elif errors == 'coerce':
+                    if not self.control_levels:
+                        # Make all results NaN
+                        result = pd.Series(np.nan, index=self._by_group.columns, dtype='object')
+                        # Fill in the possible columns with their max values
+                        for m in result.index:
```
riedgar-ms, line 422: I'm wondering if rather than the `try:` above (line 404 of the current iteration) it would be better to do
```python
if errors == 'raise':
  # Code as before
elif errors == 'coerce':
  # Rework this 'except' clause
else:
  # raise ValueError("Can't get here")
```
(that last `else` might get objections...).

I think that might be an easier set of logic to follow.

fairlearn/metrics/_metric_frame.py
```diff
@@ -404,7 +445,7 @@ def group_max(self) -> Union[Any, pd.Series, pd.DataFrame]:
         else:
             return result
 
-    def group_min(self) -> Union[Any, pd.Series, pd.DataFrame]:
+    def group_min(self, errors: str = 'raise') -> Union[Any, pd.Series, pd.DataFrame]:
```
riedgar-ms, line 448: We generally prefer to avoid repeating string literals like this - can you turn them into constants?

test/unit/metrics/test_metricframe_aggregates.py
```diff
@@ -1040,3 +1043,463 @@ def test_2m_1sf_2cf():
         assert ratios_overall[mfn][('m', 'aa')] == min(ratio_overall_m_a)
         assert ratios_overall[mfn][('m', 'ba')] == pytest.approx(min(ratio_overall_m_b),
                                                                  rel=1e-10, abs=1e-16)
+
+
+class Test2m1sf1cfErrorHandlingCM:
```
riedgar-ms, line 1048: Can we also have a case which has two metrics which would trigger an exception (even confusion matrix twice, once using `functools.partial` to tweak the output) and one regular metric? That would ensure that the logic handling potential exceptions works when more than one column raises.

fairlearn/metrics/_metric_frame.py
```diff
@@ -382,19 +385,57 @@ def group_max(self) -> Union[Any, pd.Series, pd.DataFrame]:
...
+                              SyntaxWarning)
+                if errors == "raise":
+                    raise
+                elif errors == 'coerce':
+                    if not self.control_levels:
+                        # Make all results NaN
+                        result = pd.Series(np.nan, index=self._by_group.columns, dtype='object')
+                        # Fill in the possible columns with their max values
+                        for m in result.index:
+                            if isinstance(self._by_group[m].values[0], np.float64):
```
riedgar-ms, line 423: @adrinjalali  should this be `numpy.isscalar()`? Although that looks like it might return `True` for strings.

riedgar-ms (commented):
Thanks for the update; sorry for the delay in replying.

fairlearn/metrics/_metric_frame.py
romanlutz: Although I have many times advocated for this sort of change, I have recently learned that many think of this quite differently. I believe in scikit-learn they always duplicate it when it's in the signature. @adrinjalali may have thoughts.

fairlearn/metrics/_metric_frame.py
romanlutz: @riedgar-ms can you comment?

romanlutz:
@riedgar-ms has already pointed out more than I would have discovered in this PR, so I don't really have anything to add atm. I will say that it feels like this is really close to getting merged, and I am very impressed with the thorough tests.

If you'd like you can also add your name to the AUTHORS.md file. Additionally, a short description for CHANGES.md wouldn't hurt (although - heads up - I'm migrating that file in #919 so if I'm faster in merging than you are you'll have to update your PR to make the change to the new location of the file). If this sounds too confusing I'll take care of it when we're ready to merge, no worries!

fairlearn/metrics/_metric_frame.py
bram49: Not sure what the best way is to make this a constant. Should it be something like this?
```
_VALID_ERROR_STRINGS = [ 'raise', 'coerce' ]
_RAISE_ERROR_STRING = 'raise'
_COERCE_ERROR_STRING = 'coerce'
```

test/unit/metrics/test_metricframe_aggregates.py
bram49: I am not familiar with `functools.partial`, could you please elaborate on what it does and how it can be used here? I created tests containing twice the confusion matrix metric.

fairlearn/metrics/_metric_frame.py
bram49: I updated it to np.isscalar(), since this is more readable.

romanlutz (approved):

fairlearn/metrics/_metric_frame.py
```diff
@@ -382,19 +386,51 @@ def group_max(self) -> Union[Any, pd.Series, pd.DataFrame]:
...
+                except ValueError:
+                    raise ValueError(_MF_CONTAINS_NON_SCALAR_ERROR_MESSAGE)
+            elif errors == 'coerce':
+                if not self.control_levels:
+                    # Make all results NaN
+                    result = pd.Series(np.nan, index=self._by_group.columns, dtype='object')
+                    # Fill in the possible columns with their max values
+                    for m in result.index:
+                        if np.isscalar(self._by_group[m].values[0]):
+                            result[m] = self._by_group[m].max()
```
adrinjalali, line 420: you could use the following construct to avoid a for loop:
``` python
values = [self._by_group[m].max()  if  np.isscalar(self._by_group[m].values[0]) else np.nan for m in result.index]
```

fairlearn/metrics/_metric_frame.py
adrinjalali: yeah, I very much prefer having the actual strings here. It makes it much easier to read the code.

test/unit/metrics/test_metricframe_aggregates.py
adrinjalali: https://docs.python.org/3/library/functools.html#functools.partial would give you an idea @bram49 . You can also look for `partial` in our codebase and see how it's used in other places.

adrinjalali (commented):
Do you think we could refactor this code to have a function which would do the operation given the `errors` value, and avoid the copy/pastes of the same patter in this PR?

fairlearn/metrics/_metric_frame.py
riedgar-ms: I was thinking that if you had a mixture of return types where some of them did not have 'less than' well defined, then calling with `raise` would raise an exception, and `coerce` would return the minimums for columns where that was defined, and NAN for columns where a minimum is not defined. Does that answer your question @romanlutz ?

fairlearn/metrics/_metric_frame.py
riedgar-ms: You mean actual strings in the function signature @adrinjalali ?

For other uses @bram49 , I meant roughly that, except you'd define
```python
_VALID_ERROR_STRINGS = [ _RAISE_ERROR_STRING, _COERCE_ERROR_STRING ]
```

test/unit/metrics/test_metricframe_aggregates.py
riedgar-ms: You could do somehing like
```python
norm_conf_matrix = functools.partial(sklearn.metrics.confusion_matrix, normalize='all')
``
as the second metric function.

File: fairlearn/metrics/_metric_frame.py (hunk 11 above)
Comment (adrinjalali) [reply fairlearn/metrics/_metric_frame.py:443]: I meant for almost everywhere. I find `'raise'` much more readable than `_RAISE_ERROR_STRING`

File: CHANGES.md [hunk 15]
```diff
@@ -4,7 +4,8 @@
 
 * Relaxed checks made on `X` in `_validate_and_reformat_input()` since that
   is the concern of the underlying estimator and not Fairlearn
-
+* Added error handling in `MetricFrame`. `group_min`, `group_min`, `difference` and `ratio`
+  now accept `errors` as a parameter, which could either be `raise` or `coerce`.
```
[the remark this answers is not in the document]
romanlutz: Yes, I [pointed this out earlier](https://github.com/fairlearn/fairlearn/pull/896#issuecomment-889654881) before #919 got merged. Now that that has actually happened @bram49 needs to move it over to the corresponding file. I think I need to write a pull request template soon to list all the steps required for a PR...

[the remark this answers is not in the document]
bram49: I removed CHANGES.md and added it to installation_and_version_guide/v0.7.1.rst 👍

fairlearn/metrics/_metric_frame.py
bram49: Thanks for the tip! I now avoided all for loops

adrinjalali (approved):

fairlearn/metrics/_metric_frame.py
```diff
@@ -382,19 +386,46 @@ def group_max(self) -> Union[Any, pd.Series, pd.DataFrame]:
...
-            for m in result.index:
-                max_val = self._by_group[m].max()
-                result[m] = max_val
+            if errors == "raise":
+                try:
+                    mf = self._by_group
+                    max_vals = [mf[m].max() for m in mf.columns]
+                    result = pd.Series(max_vals, index=self._by_group.columns, dtype='object')
+                except ValueError:
+                    raise ValueError(_MF_CONTAINS_NON_SCALAR_ERROR_MESSAGE)
```
riedgar-ms, line 411: Can you include the caught exception by doing a `raise ... from ...` here?

fairlearn/metrics/_metric_frame.py
riedgar-ms: @adrinjalali  has more experience of this sort of thing than I do; I'm happy to go with this suggestion.

fairlearn/metrics/_metric_frame.py
```diff
@@ -414,19 +445,46 @@ def group_min(self) -> Union[Any, pd.Series, pd.DataFrame]:
...
         """
+        if errors not in _VALID_ERROR_STRING:
+            raise ValueError(_INVALID_ERRORS_VALUE_ERROR_MESSAGE)
+
         if not self.control_levels:
-            result = pd.Series(index=self._by_group.columns, dtype='object')
-            for m in result.index:
-                min_val = self._by_group[m].min()
-                result[m] = min_val
+            if errors == "raise":
```
riedgar-ms, line 464: This looks awfully similar to the same block in `group_min()`, and is now large enough to make me concerned about copying code. Would it be possible to refactor this logic into a separate method, and have the final max/min aggregation be an argument?

fairlearn/metrics/_metric_frame.py
```diff
@@ -461,24 +520,39 @@ def difference(self,
...
         if method == 'between_groups':
-            subtrahend = self.group_min()
+            subtrahend = self.group_min(errors=errors)
         elif method == 'to_overall':
             subtrahend = self.overall
         else:
             raise ValueError("Unrecognised method '{0}' in difference() call".format(method))
 
-        return (self.by_group - subtrahend).abs().max(level=self.control_levels)
+        try:
```
riedgar-ms, line 542: We still need another `try` block here? Haven't all possible routes to the exception already been raised, or the values replaced by `NaN`?

riedgar-ms (commented):
Thanks for the update!

fairlearn/metrics/_metric_frame.py
riedgar-ms: (I think that's the right syntax @adrinjalali ?)

adrinjalali, line 411: yes, more precisely: 

```suggestion
                except ValueError as e:
                    raise ValueError(_MF_CONTAINS_NON_SCALAR_ERROR_MESSAGE) from e
```

fairlearn/metrics/_metric_frame.py
bram49: Yes `group_min `and `group_max `do exactly the same, only .`min()` changes in `.max()`. 

Shall I put all logic in a separate method with parameter `['min', 'max']` which is called from `group_min()` and `group_max()`? Such that the dependencies of `group_min `and `group_max `don't break?

fairlearn/metrics/_metric_frame.py
bram49: I found a problem with this piece of code. Cause the code should indeed work with only 
` mf = self.by_group.copy()`
`mf = mf.applymap(lambda x: x if np.isscalar(x) else np.nan)`
`return (mf - subtrahend).abs().max(level=self.control_levels)`

However this doesn't work, since `.applymap()` only works for DataFrames, and Metricframe can sometimes be a Series. So I'm now afraid that the code will not work in the case that Metricframe is a Series and contains a confusion_matrix, so I will first work on new test cases.

romanlutz: `MetricFrame.by_group` can be a `Series`? 😮 Is that intended? @riedgar-ms

riedgar-ms: I believe so. In the case where you supply a single metric function, the result is a Series. If you supply a single metric function in a dictionary, it will be a DataFrame with a single column. We had some lengthy debates about this when putting `MetricFrame` together.

fairlearn/metrics/_metric_frame.py
riedgar-ms: Something like that sounds good. But yes, the users of `group_min()` and `group_max()` mustn't break.

fairlearn/metrics/_metric_frame.py
romanlutz: ... but you can have 1-d DataFrames, so what's the benefit of the special case?

riedgar-ms: Because passing a metric function was considered to be different from passing a dictionary containing a single metric function.

bram49: Believe so, the doc says:
```
@property
def by_group(self) -> Union[pd.Series, pd.DataFrame]:
"""
Returns
        -------
        pandas.Series or pandas.DataFrame
            When a callable is supplied to the constructor, the result is
            a :class:`pandas.Series`, indexed by the combinations of subgroups
            in the sensitive and control features.

            When the metric functions were specified with a dictionary (even
            if the dictionary only has a single entry), then the result is
            a :class:`pandas.DataFrame` with columns named after the metric
            functions, and rows indexed by the combinations of subgroups
            in the sensitive and control features.
```

hildeweerts: If I recall correctly the reason for this was that indexing a 1-d DataFrame without a sensible column name (which it doesn't have if you only supply a metric function) becomes a bit weird. For example, in order to access the score for some group called 'a' you'd have to do something like `mf.by_group.loc[0,'a']` instead of `mf.by_group.loc['a']`.

riedgar-ms: Yes, it was something like that - the 'without a sensible column name' particularly coming from things constructed with `functools.partial()`. And @adrinjalali soothed my ruffled type-feathers by pointing out that changing from function-dictionary to a single function would in most languages result in an overloaded call, making the change in return types less jarring.

fairlearn/metrics/_metric_frame.py
bram49: https://github.com/fairlearn/fairlearn/pull/896/commits/4b6cc6016ac2964fe816986da9244a44ff78c7ea
I fixed it in this way, let me know what you think @riedgar-ms

fairlearn/metrics/_metric_frame.py
```diff
@@ -372,29 +379,67 @@ def sensitive_levels(self) -> List[str]:
         """
         return self._sf_names
 
-    def group_max(self) -> Union[Any, pd.Series, pd.DataFrame]:
-        """Return the maximum value of the metric over the sensitive features.
+    def group(self, grouping_function: str, errors: str = 'raise') -> Union[Any, pd.Series, pd.DataFrame]:
```
riedgar-ms, line 382: This should probably be `_group()` and so not public (well, by convention).

riedgar-ms (commented):
That looks like the sort of thing I had in mind, but I'd make it a private method.

fairlearn/metrics/_metric_frame.py
bram49: Ah yes makes sense! It should be a private method.

adrinjalali:
tests failing?

bram49:
Whoops forgot to flake8. Think the PR is ready for merging

riedgar-ms (approved):
Thank you for all your work on this!

PR #903, opened 2021-07-14
MNT remove gitpython, use git through subprocess instead

romanlutz:
The website build currently doesn't show `main`, only the released versions. This is because of a raised exception while parsing `conf.py`. Interestingly, this does not fail the entire build which is why it escaped detection. `gitpython` is struggling to recognize the repository locally, so instead we'll just run a `git` command to get all release versions.

Note that this will only be possible to validate once it's merged since the `main` version is based on the current `main` branch which has the `conf.py` that can't be parsed properly. I've validated locally that we get at least the same output as currently even if it doesn't fix the issue.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -13,10 +13,10 @@
 # add these directories to sys.path here. If the directory is relative to the
 # documentation root, use os.path.abspath to make it absolute, like shown here.
 #
-from git import Repo
 from packaging import version as packaging_version
 import os
 import pandas as pd
+import subprocess
 import sys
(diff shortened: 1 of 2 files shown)


riedgar-ms (approved):
'Can only be tested when checked in' .... I foresee no problems with that.... ;-)

romanlutz:
> Has this fixed the issue @romanlutz ?

No, but #905 did. Sorry, the automation isn't quite panning out. I spent way more hours on this than I had hoped without any success.

Issue #836, opened 2021-06-02
[Website Redesign] Get Started


nessamilan:
https://www.figma.com/file/wt76z0M87RBDDEoyAEpD2g/OPEN-SOURCE-Fairlearn-Redesign?node-id=2535%3A3126

PR #904, opened 2021-07-15
Extend Migration Guide with Installation Guide

### LINKED ISSUE #736: Create a table with methods/modules and corresponding dependencies
We should start having a very clear table, with methods/modules on one side, and dependencies on the other side, for people to have a clear view about them. Currently `postprocessing` has a soft dependency on `matplotlib`. Soon, this will also be true for `metrics` (once the plotting capabilities are merged). Additionally, #466 seems to converge towards `cvxpy` as another soft dependency.

_Originally posted by @adrinjalali in https://github.com/fairlearn/fairlearn/issues/466#issuecomment-815855327_

The table could either live with our installation instructions or on a separate page of our documentation. The installation instructions are currently part of "Getting started", and we may want to avoid adding too many details on that page, hence the suggestion for a "separate page".

shimst3r:
This closes #736

I extend the migration guide with an installation guide that lists Fairlearn's optional dependencies, as well as how to install them.

Optional dependencies have been identified via

```shell
grep --exclude-dir .venv --include="*.py" -Er "importorskip|^\s{4,}.*import" fairlearn
```

Changes to the sidebar:

changes_sidebar

Changes to the table of contents:

changes_toc

The installation guide itself:

installation_guide


romanlutz (approved):
This is excellent!!! Reading through the changes I liked everything, but looking at the pictures there's one thing that jumped out at me: 4.1.1 seems to be "Installation" in the actual page, but "Dependencies" in the parent page (which is completely missing "Installation" in the TOC). I am a little at a loss as to why, but I'll keep looking.

shimst3r:
I *think* it might've been a caching issue on my local machine. Rebuilding it without changing anything led to the following screenshots:

Installation Guide Detail Page:

installation_guide_2

User Guide Overview Page:

user_guide_2

hildeweerts:
I'm very happy with the contents of these pages, thanks @shimst3r!

Looking at the TOC though I am not sure anymore whether the user guide is the best place to put advanced installation instructions. Especially since we will be renaming 'Quickstart' to 'Getting Started', it might make more sense to add this to the (currently non-existent) lefthandside navigation? WDYT?

I am also happy to merge this as is though and move it when we redesign the Getting Started page.

shimst3r:
My gut feeling is merging it as it is and refactor later. This way new users will be able to benefit from the changes sooner. :nerd_face:

romanlutz:
I agree with @shimst3r . This isn't really even included in the designs of #836 because the designed getting started experience only includes very basic installation without these details. We should iterate on this whenever we get a better idea of where it should live.

Issue #792, opened 2021-05-07
DOC: best practices around demographic data - gender


romanlutz:
Our documentation doesn't currently say anything about best practices for 
- collecting demographic data
- working with demographic data

We should probably start a section for this in our user guide. There's likely specific guidance for various kinds of data (e.g., gender, race, age, etc.). This particular task is for gender only. If you'd like to work on others please create a new issue and link it.

Regarding the collection @MiroDudik posted the following on Gitter based on a discussion with Queer in AI folks:

> Gender: (please select the options which are applicable to you)
> [ ] Woman
> [ ] Man
> [ ] Non-binary / Genderqueer / Third gender
> [ ] Genderfluid / Gender non-conforming
> [ ] Questioning
> [ ] Prefer not to say
> [ ] Specify your own (open text box)
> 
> The key aspect is that this is a multi-select. When this kind of data is collected, there should be an explanation that the data is collected to audit for fairness and/or mitigate any fairness issues, and that if you pick "prefer not to say" you will not be included in any fairness audits.

In some cases people don't start at data collection, but rather already have a dataset with certain demographic data that may not conform to such best practices. Our best practices should mention that as well, and what one can and should do in such a case. A very common situation is to find a dataset with a `sex` column only. Additionally, I've seen plenty of dataset where we see columns like `sex` or `gender` but it's unclear whether these were self-reported or inferred/guessed by someone (or a system) which obviously makes a huge difference.

Further sources:
- https://apastyle.apa.org/style-grammar-guidelines/bias-free-language/gender
- https://medium.com/managing-on-the-margins/respectful-collection-of-demographic-data-56de9fcb80e2
- https://www.morgan-klaus.com/gender-guidelines.html

MiroDudik:
Adding the QAI source:
https://sites.google.com/view/queer-in-ai/diversity-guide

hildeweerts:
Regarding race, this paper is a must-read as well IMO: https://arxiv.org/pdf/1912.03593.pdf

mmadaio:
Katta Spiel et al have a great article in ACM Interactions on inclusive ways to collect gender data on surveys:

https://www.researchgate.net/publication/334073420_How_to_do_better_with_gender_on_surveys_a_guide_for_HCI_researchers

Issue #757, opened 2021-04-27
Use the new Azure DevOps organization "Fairlearn"


romanlutz:
https://dev.azure.com/fairlearn/Fairlearn

I created this new org to decouple our CI from Microsoft's internal CI. Also, all Fairlearn maintainers have admin permissions and can approve releases. The task that remains to be done is to point the existing Azure pipelines at this new organization.

@fairlearn/fairlearn-maintainers

romanlutz:
FYI I tried setting this up tonight. The YAML pipeline definitions were easy to link to the new ADO organization. However, it only auto-queues the pipelines on one org, so that's still currently the "old" organization until I can make sure everything works properly in the new one. So obviously I kicked off a run of one of our pipelines to check whether it does indeed work properly. Surprisingly, I encountered the following issue:
```
##[error]No hosted parallelism has been purchased or granted. To request a free parallelism grant, please fill out the following form https://aka.ms/azpipelines-parallelism-request
```
I've filled out that form and it will take up to 3 business days for a response. Stay tuned!

romanlutz:
As of right now I've migrated all pipelines and disabled all but the release pipeline in the old organization. The release pipeline is a bit special due to the required service connections to PyPI and test-PyPI. I've created those for now, but we have to run an actual release to confirm that they work.

If any issues occur please flag them here @fairlearn/fairlearn-maintainers

riedgar-ms:
The approval is controlled by the 'environment' portion of the job:
https://github.com/fairlearn/fairlearn/blob/a68e74ec90ce99f00446b0b8fe53203f61f640b3/devops/pypi-release-new.yml#L169
which point at:
https://dev.azure.com/responsibleai/fairlearn/_environments

romanlutz:
@riedgar-ms Excellent! I found it and set it to the "Fairlearn team" which includes all @fairlearn/fairlearn-maintainers 

![image](https://user-images.githubusercontent.com/10245648/126729162-4797e84c-1076-416c-8541-964ad43768ac.png)


However, I still don't know how to fix the twine upload. Any idea? Is there some additional setup that needs to happen?

riedgar-ms:
What's the problem exactly? That environment setup should mean approvals are required.

romanlutz:
The issue that happened the first time was (copying from above):
[...]

PR #907, opened 2021-07-17
[WIP] Synthetic dataset creation

### LINKED ISSUE #793: ENH Creating a synthetic example dataset
This is based on a Gitter conversation with @adrinjalali @hildeweerts @MiroDudik where we agreed that it would be nice to have a synthetic dataset available for our examples. @adrinjalali suggested the following code using `sklearn`'s `make_classification`:

```
rng = RandomState(seed=42)

X_women, y_women = make_classification(
    n_samples=500,
    n_features=20,
    n_informative=4,
    n_classes=2,
    class_sep=1,
    random_state=rng,
)

X_men, y_men = make_classification(
    n_samples=500,
    n_features=20,
    n_informative=4,
    n_classes=2,
    class_sep=2,
    random_state=rng,
)

X_unspecified, y_unspecified = make_classification(
    n_samples=500,
    n_features=20,
    n_informative=4,
    n_classes=2,
    class_sep=0.5,
    random_state=rng,
)

X = np.r_[X_women, X_men, X_unspecified]
y = np.r_[y_women, y_men, y_unspecified]
gender = np.r_[["Woman"] * 500, ["Man"] * 500, ["Unspecified"] * 500].reshape(
    -1,
)

X_train, X_test, y_train, y_test, gender_train, gender_test = train_test_split(
    X, y, gender, test_size=0.3, random_state=rng
)
```
@MiroDudik suggested extending this to have at least 2 sensitive features and 1 control feature to allow us to use it in basically all our examples. 

@fairlearn/fairlearn-maintainers any objection with putting this in the `fairlearn.datasets` module?

coreysharris:
This PR adds functionality for creating synthetic datasets. The starting point is the snippet contributed by @romanlutz in #793. I've included a simple example notebook for demonstration purposes, but this could be removed as the work here gets closer to finished.

Closes #793.

### DIFF
--- a/examples/plot_synthetic.py
+++ b/examples/plot_synthetic.py
@@ -1,0 +1,65 @@
+# Copyright (c) Microsoft Corporation and Fairlearn contributors.
+# Licensed under the MIT License.
+
+"""
+==========================
+Synthetic dataset example with visualization
+==========================
+"""
+
+# %%
+from fairlearn.datasets import make_synthetic_dataset
+from fairlearn.metrics import (
+    MetricFrame,
+    false_positive_rate,
+    true_positive_rate,
+    selection_rate,
+    count
+)
+import numpy as np
+from sklearn.metrics import accuracy_score, precision_score, recall_score
+from sklearn.model_selection import train_test_split
+from sklearn.tree import DecisionTreeClassifier
+
+
+rng = np.random.RandomState(seed=42)
+X, y, gender = make_synthetic_dataset(rng=rng)
+
+X_train, X_test, y_train, y_test, gender_train, gender_test = train_test_split(
+    X, y, gender, test_size=0.3, random_state=rng
+)
+
+classifier = DecisionTreeClassifier(min_samples_leaf=10, max_depth=4)
+classifier.fit(X, y)
+y_pred = classifier.predict(X)
+
+# Analyze metrics using MetricFrame
+metrics = {
+    'accuracy': accuracy_score,
+    'precision': precision_score,
+    'recall': recall_score,
+    'false positive rate': false_positive_rate,
+    'true positive rate': true_positive_rate,
+    'selection rate': selection_rate,
+    'count': count}
+metric_frame = MetricFrame(metrics=metrics,
+                           y_true=y,
+                           y_pred=y_pred,
+                           sensitive_features=gender)
+metric_frame.by_group.plot.bar(
+    subplots=True,
+    layout=[3, 3],
+    legend=False,
+    figsize=[12, 8],
+    title="Show all metrics",
+)
+
+# Customize plots with kind
+metric_frame.by_group.plot(
+    kind="pie",
+    subplots=True,
+    layout=[3, 3],
+    legend=False,
+    figsize=[12, 8],
+    title="Show all metrics in pie",
+)
--- a/fairlearn/datasets/__init__.py
+++ b/fairlearn/datasets/__init__.py
@@ -8,9 +8,11 @@
 from ._fetch_adult import fetch_adult
 from ._fetch_boston import fetch_boston
 from ._fetch_bank_marketing import fetch_bank_marketing
+from .synthetic import make_synthetic_dataset
 
 __all__ = [
     "fetch_adult",
     "fetch_boston",
     "fetch_bank_marketing",
+    "make_synthetic_dataset",
 ]
--- a/fairlearn/datasets/synthetic.py
+++ b/fairlearn/datasets/synthetic.py
@@ -4,8 +4,23 @@ import numpy as np
 from sklearn.datasets import make_classification
 
 
-def make_synthetic_dataset(rng=None):
+def make_synthetic_dataset(seed=12345):
     """Create a synthetic dataset.
+
+    Parameters
+    ----------
+    seed : int, default=12345
+        The random number generator seed to use.
+
+    Returns
+    -------
+    (X, y, gender) : tuple of numpy.ndarray
+        X : ndarray
+            The generated samples.
+        y : ndarray
+            Labels for the binary classification.
+        gender : ndarray
+            The sensitive feature label.
     """
     if rng is None:
         rng = np.random.RandomState(seed=42)
--- a/fairlearn/datasets/synthetic.py
+++ b/fairlearn/datasets/synthetic.py
@@ -1,0 +1,52 @@
+# Copyright (c) Microsoft Corporation and Fairlearn contributors.
+# Licensed under the MIT License.
+import numpy as np
+from sklearn.datasets import make_classification
+
+
+def make_synthetic_dataset(rng=None):
+    """Create a synthetic dataset.
+    """
+    if rng is None:
+        rng = np.random.RandomState(seed=42)
+
+    classification_kwargs = {
+        'n_features': 20,
+        'n_informative': 4,
+        'n_classes': 2,
+        'random_state': rng,
+    }
+
+    X_women, y_women = make_classification(
+        n_samples=500,
+        class_sep=1,
+        **classification_kwargs,
+    )
+
+    X_men, y_men = make_classification(
+        n_samples=500,
+        class_sep=2,
+        **classification_kwargs,
+    )
+
+    X_other, y_other = make_classification(
+        n_samples=500,
+        class_sep=0.5,
+        **classification_kwargs,
+    )
+
+    X_unspecified, y_unspecified = make_classification(
+        n_samples=500,
+        class_sep=0.5,
+        **classification_kwargs,
+    )
+
+    X = np.r_[X_women, X_men, X_other, X_unspecified]
+    y = np.r_[y_women, y_men, y_other, y_unspecified]
+    gender = np.r_[
+        ["Woman"] * 500,
+        ["Man"] * 500,
+        ["Other"] * 500,
+        ["Unspecified"] * 500
+    ].reshape(-1)
+    return X, y, gender


fairlearn/datasets/synthetic.py
```diff
@@ -0,0 +1,67 @@
...
+        'n_informative': 4,
+        'n_classes': 2,
+        'random_state': rng,
+    }
+
+    X_women, y_women = make_classification(
+        n_samples=500,
+        class_sep=1,
+        **classification_kwargs,
+    )
```
coreysharris, line 38: This can be cleaned up a lot. I'm avoiding early abstractions since I wasn't sure what parameters will be exposed.

examples/plot_synthetic.py
```diff
@@ -0,0 +1,65 @@
+# Copyright (c) Microsoft Corporation and Fairlearn contributors.
+# Licensed under the MIT License.
+
+"""
+==========================
```
romanlutz, line 5: The = signs need to be at least as long as the header AFAIK

examples/plot_synthetic.py
```diff
@@ -0,0 +1,65 @@
...
+                           sensitive_features=gender)
+metric_frame.by_group.plot.bar(
+    subplots=True,
+    layout=[3, 3],
+    legend=False,
+    figsize=[12, 8],
+    title="Show all metrics",
+)
+
+# Customize plots with kind
```
romanlutz, line 57: I think the bar plot is sufficient. Was there anything in particular you want to show with the pie?

fairlearn/datasets/synthetic.py
```diff
@@ -0,0 +1,67 @@
...
+
+    X = np.r_[X_women, X_men, X_other, X_unspecified]
+    y = np.r_[y_women, y_men, y_other, y_unspecified]
+    gender = np.r_[
+        ["Woman"] * 500,
+        ["Man"] * 500,
+        ["Other"] * 500,
+        ["Unspecified"] * 500
+    ].reshape(-1)
+    return X, y, gender
```
coreysharris, line 66: This is specific to the gender example, but we could have a more general function which just produces a certain number of sensitive features. Is that desirable?

fairlearn/datasets/synthetic.py
```diff
@@ -0,0 +1,66 @@
+# Copyright (c) Microsoft Corporation and Fairlearn contributors.
+# Licensed under the MIT License.
+import numpy as np
+from sklearn.datasets import make_classification
+
+
+def make_synthetic_dataset(seed=12345):
```
romanlutz, line 7: Is it normal to specify a seed as default? I would have assumed that it uses something like system time by default.

fairlearn/datasets/__init__.py
```diff
@@ -8,9 +8,11 @@
 from ._fetch_adult import fetch_adult
 from ._fetch_boston import fetch_boston
 from ._fetch_bank_marketing import fetch_bank_marketing
+from .synthetic import make_synthetic_dataset
```
coreysharris, line 11: I can move this to `_synthetic`.

fairlearn/datasets/synthetic.py
```diff
@@ -0,0 +1,66 @@
...
+    -------
+    (X, y, gender) : tuple of numpy.ndarray
+        X : ndarray
+            The generated samples.
+        y : ndarray
+            Labels for the binary classification.
+        gender : ndarray
+            The sensitive feature label.
+    """
+    rng = np.random.RandomState(seed=seed)
```
romanlutz, line 25: How would this not be None?

romanlutz: Did you mean to check the seed here?

fairlearn/datasets/synthetic.py
```diff
@@ -0,0 +1,66 @@
...
+
+    classification_kwargs = {
+        'n_features': 20,
+        'n_informative': 4,
```
romanlutz, line 29: Can these be input args with these defaults?

romanlutz (commented):
Can you add some tests, too?

fairlearn/datasets/synthetic.py
romanlutz: Fair enough! I do think we should start simple and then extend as needed.

fairlearn/datasets/synthetic.py
romanlutz: Definitely desirable. I'm wondering whether we should start with a 1-D version, though, and then add more in a follow-up PR.

fairlearn/datasets/synthetic.py
coreysharris: I'm not sure what's more common, but I think using `seed=None` here is quite reasonable. I'll make the change.

fairlearn/datasets/synthetic.py
coreysharris: Sure!

fairlearn/datasets/synthetic.py
```diff
@@ -0,0 +1,66 @@
...
+        **classification_kwargs,
+    )
+
+    X_other, y_other = make_classification(
+        n_samples=500,
+        class_sep=0.5,
+        **classification_kwargs,
+    )
+
+    X_unspecified, y_unspecified = make_classification(
```
romanlutz, line 52: #792 specifies the categories one should ideally provide to users. I'm not sure whether this needs to conform to that, but it's definitely something we should keep on our radar for a follow-up PR if not for this one. 
A related point that occurred to me was that the group sizes should probably be configurable. My initial thought was to have a dictionary where for each sensitive feature (gender, race, nationality, etc.) you can specify the group sizes ("male": 300, "female": 400, etc.) but then it's unclear how to map that onto individuals so that it still makes sense (e.g., should races be distributed equally among the genders?), so it's probably best to let the user specify that (?).
I do think this goes beyond the initial PR for sure, but if you have thoughts on how this could be achieved in a sensible way we should get the conversation started 🙂

fairlearn/datasets/synthetic.py
romanlutz: Actually, I stand corrected. https://scikit-learn.org/stable/modules/generated/sklearn.datasets.make_classification.html?highlight=make_classification#sklearn.datasets.make_classification
We should probably do what sklearn does and use `random_state`. We can probably reuse their `check_random_state` code from https://github.com/scikit-learn/scikit-learn/blob/2beed55847ee70d363bdbfe14ee4401438fba057/sklearn/datasets/_samples_generator.py#L159 since it's part of `sklearn.utils` (which we already heavily rely on)

fairlearn/datasets/__init__.py
romanlutz: Good point! This is mostly to show everything at the module level (`from fairlearn.datasets import make_synthetic_dataset` in the users' code) rather than having people import `fairlearn.datasets.synthetic`.

examples/plot_synthetic.py
coreysharris: Not really. It was just easy enough to include. I'll remove it.

examples/plot_synthetic.py
```diff
@@ -0,0 +1,65 @@
+# Copyright (c) Microsoft Corporation and Fairlearn contributors.
```
romanlutz, line 1: You can leave the Microsoft part out and just make it
```
# Copyright (c) Fairlearn contributors.
```
That's basically just an add-on we used in older files before the project became independent 🙂 This applies to all files, of course

examples/plot_synthetic.py
```diff
@@ -0,0 +1,65 @@
...
+
+rng = np.random.RandomState(seed=42)
+X, y, gender = make_synthetic_dataset(rng=rng)
+
+X_train, X_test, y_train, y_test, gender_train, gender_test = train_test_split(
+    X, y, gender, test_size=0.3, random_state=rng
+)
+
+classifier = DecisionTreeClassifier(min_samples_leaf=10, max_depth=4)
+classifier.fit(X, y)
```
romanlutz, line 33: We should probably use `X_train, y_train` here and predict using `X_test`?
And then evaluate the metrics on `y_test` and `gender_test`.

romanlutz:
Ah before I forget: Could you add a line to CHANGES.md describing your change (in the 0.7.1 section since that'll be the next release). Additionally, feel free to add your name to AUTHORS.md if you like.

fairlearn/datasets/synthetic.py
coreysharris: Sure. Happy to keep this PR small.

examples/plot_synthetic.py
coreysharris: Indeed! I didn't think carefully about this at all when I copied things from the existing notebook. It's fixed now.

CHANGES.md
```diff
@@ -2,6 +2,8 @@
 
 ### v0.7.1
 
+* Added new `make_synthetic_dataset()` in `fairlearn.datasets` to allow for
+  creation of a dataset with user-selected numhber of features.
```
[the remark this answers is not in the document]
coreysharris: Fixed, thanks!

coreysharris:
I've cleaned things up to hopefully clarify how we could expand into more interesting waters. The code that defines the gender feature is now quite small and could easily be separated into another function.

fairlearn/datasets/synthetic.py
coreysharris: Just saw this comment... I didn't know about `check_random_state`. Very useful! I've incorporated it now.

fairlearn/datasets/_synthetic.py
```diff
@@ -0,0 +1,74 @@
...
+        classes = {
+            'Man': {'n_samples': 500, 'class_sep': 1},
+            'Other': {'n_samples': 500, 'class_sep': 0.5},
+            'Unspecified': {'n_samples': 500, 'class_sep': 0.5},
+            'Woman': {'n_samples': 500, 'class_sep': 2},
+        }
+
+    Xs, ys, genders = [], [], []
+
+    for label, kwargs in classes.items():
```
romanlutz, line 63: `kwargs` is typically used for the `kwargs` passed into the method (in the signature), so perhaps something more descriptive like `class_config` would be preferable? Actually, I think `group_config` might work better since the term class is often used for the target (`y`).

test/unit/datasets/test_datasets.py
```diff
@@ -38,3 +40,26 @@ def test_dataset_as_X_y(self, as_frame, fetch_function):
...
+    def test_synthetic_datasets(self):
+        """Ensure that dataset creation is deterministic."""
+        rng = np.random.RandomState(12345)
+        X, y, gender = make_synthetic_dataset(random_state=rng)
+
+        X_train, _, y_train, _, gender_train, _ = train_test_split(
+            X, y, gender, test_size=0.3, random_state=rng
+        )
+
+        assert np.sum(X_train[0] < 0) == 8
```
romanlutz, line 53: Are these two lines just random aggregate stats you found? If so, perhaps asserting the dimensions and mean of the features and targets is useful?

test/unit/datasets/test_datasets.py
```diff
@@ -38,3 +40,26 @@ def test_dataset_as_X_y(self, as_frame, fetch_function):
...
+        assert np.sum(y_train) == 699
+
+        counts = {'Man': 0, 'Woman': 0, 'Other': 0, 'Unspecified': 0}
+        for k in gender_train:
+            counts[k] += 1
+        expected_counts = {
```
romanlutz, line 59: This is great! However, can we also check for the number of samples with 0/1? I would hope that there's some significant difference between at least 2 of the groups because that's kind of the point 🙂 
Taking a step back, I don't think we need the train/test split here, because we're not training anything. We're just asserting properties of the utility function you created while assuming that scikit-learn tests `train_test_split` thoroughly. Wdyt?

fairlearn/datasets/_synthetic.py
```diff
@@ -0,0 +1,74 @@
...
+
+def make_synthetic_dataset(classes=None, n_features=20, n_informative=4, random_state=None):
+    """Create a synthetic dataset with a single sensitive feature: 'gender'.
+
+    Parameters
+    ----------
+    classes : dict, default=None
+        A dict whose keys are the labels for the gender feature, and whose set
+        the `n_samples` and `class_sep` parameters for `make_classification`.
+        If `None`, the following default dict is used::
```
romanlutz, line 16: `defaultdict` is actually a utility that's part of `collections` (and quite useful!), so perhaps I'd rephrase to
```
If `None`, the following dictionary is used by default:
```
The double colon is probably there by accident, right? (or is it special syntax for ReST?)

fairlearn/datasets/_synthetic.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def make_synthetic_dataset(classes=None, n_features=20, n_informative=4, random_state=None):
+    """Create a synthetic dataset with a single sensitive feature: 'gender'.
+
+    Parameters
+    ----------
+    classes : dict, default=None
+        A dict whose keys are the labels for the gender feature, and whose set
+        the `n_samples` and `class_sep` parameters for `make_classification`.
```
romanlutz, line 15:
```suggestion
        the :code:`n_samples` and :code:`class_sep` parameters for :py:func:`sklearn.utils.make_classification`.
```

fairlearn/datasets/_synthetic.py
```diff
@@ -0,0 +1,74 @@
...
+            }
+
+    n_features : int, default=20
+        The total number of features.
+
+    n_informative : int, default=4
+        The number of informative features.
+
+    random_state : int or RandomState instance, default=None
+        The random number generator seed or RandomState to use.
```
romanlutz, line 32:
```suggestion
        The random number generator seed or :class:`numpy.random.RandomState` to use.
```
Note that this is only needed in this line for the description, since the line above (the type annotation) doesn't require the ReST directive.

fairlearn/datasets/_synthetic.py
```diff
@@ -0,0 +1,74 @@
...
+
+    Returns
+    -------
+    (X, y, gender) : tuple of numpy.ndarray
+        X : ndarray
```
romanlutz, line 37:
```suggestion
        X : numpy.ndarray
```

fairlearn/datasets/_synthetic.py
```diff
@@ -0,0 +1,74 @@
...
+            The generated samples.
+        y : ndarray
```
romanlutz, line 39:
```suggestion
        y : numpy.ndarray
```

test/unit/datasets/test_datasets.py
```diff
@@ -38,3 +40,26 @@ def test_dataset_as_X_y(self, as_frame, fetch_function):
         assert isinstance(X, pd.DataFrame if as_frame else np.ndarray)
         assert y is not None
         assert isinstance(y, pd.Series if as_frame else np.ndarray)
+
+    def test_synthetic_datasets(self):
```
romanlutz, line 44: Can we add the following test cases as well?
- pass in custom configuration with a different number of samples than the default and assert that the sum of the `n_samples` values adds up to the total size
- pass in a different number of features and assert the resulting array dimension

Bonus: try out invalid args
- `n_features=0`
- empty dict for `classes`

fairlearn/datasets/_synthetic.py
```diff
@@ -0,0 +1,74 @@
...
+from sklearn.datasets import make_classification
+from sklearn.utils import check_random_state
+
+
+def make_synthetic_dataset(classes=None, n_features=20, n_informative=4, random_state=None):
+    """Create a synthetic dataset with a single sensitive feature: 'gender'.
+
+    Parameters
+    ----------
+    classes : dict, default=None
```
romanlutz, line 13: I'm not super happy with the arg name `classes` since that's typically used for target classes (as in multi-class classification, for example). Groups is the term I'd prefer, but it's also a config dictionary for the groups, so perhaps `group_config`? I strongly suspect we'll expand this in the future to work for multiple sensitive features, so this arg will likely evolve.

This is also focused on binary classification right now (which is perfectly fine!), so perhaps we should include that in the name (similar to `sklearn`'s `make_classification`)?

@adrinjalali @hildeweerts @riedgar-ms @MiroDudik any thoughts on naming and API?

romanlutz (commented):
@coreysharris this is fantastic and I've pointed out everything I possibly could since we're getting very close to merging this.

Do not worry about the currently failing (single) PR-Gate pipeline. That's a manifestation of the issue I described in https://github.com/fairlearn/fairlearn/issues/757#issuecomment-881224903 and won't be a blocker for your PR.

fairlearn/datasets/_synthetic.py
coreysharris: Fixed!

fairlearn/datasets/_synthetic.py
coreysharris: I've reworded as you suggested. The double colon is the ReST syntax for making the next paragraph a literal block.

coreysharris: BTW, I don't really like that I've duplicated this code into the docstring, but I'm hoping things can be sufficiently cleaned up in the next PR so that this will become unnecessary.

fairlearn/datasets/_synthetic.py
coreysharris: Fixed, thanks!

fairlearn/datasets/_synthetic.py
coreysharris: Fixed!

fairlearn/datasets/_synthetic.py
coreysharris: Fixed!

fairlearn/datasets/_synthetic.py
coreysharris: Good point. I've gone with `group_config` here, and replaced the `classes` kwarg with `feature_config`.

test/unit/datasets/test_datasets.py
coreysharris: I had included the `train_test_split` because I wanted to test the interaction between the two RNGs. Now that it's shared, there's no benefit, so I've removed it.

> I would hope that there's some significant difference between at least 2 of the groups because that's kind of the point

Of course :) Right now, because the feature is randomly assigned to the samples, we're getting pretty even distribution. I'll add in a utility function to help assign things unfairly.

test/unit/datasets/test_datasets.py
coreysharris: As it stands, such invalid args result in a `ValueError`. Is that acceptable, or should we do some kind of sanitization for the user? I'm specifically thinking about `classes` (now called `feature_config`). It's not very user-friendly to ask for this dict in the first place, and I hope we can replace it with something cleaner in the follow-up, but for now I wonder if we should try to ensure that the dict has entries of the form `'label': {'n_samples': n_samples, 'class_sep': class_sep}`.

test/unit/datasets/test_datasets.py
coreysharris: I've made a simple attempt at this in 0d2cda7. Let me know what you think.

fairlearn/datasets/_synthetic.py
coreysharris: I've replaced `classes` with `feature_config`. See https://github.com/fairlearn/fairlearn/pull/907#discussion_r671835815 .

I actually originally called it `make_classification`, rather than `make_synthetic_dataset`! I replaced it because I wasn't sure about overloading such a common function from scikit-learn. What about `make_sensitive_classification` or `make_gendered_classification`?

romanlutz: It's certainly not an unreasonable thought. I'll wait for others to chime in (at least @adrinjalali may have thoughts), but `make_sensitive_classification` sounds like a name that'll work even with more sensitive features (even though that may not be part of the initial PR).

test/unit/datasets/test_datasets.py
romanlutz: That's fine 🙂 We have plenty of tests for invalid inputs where we assert the correct errors/exceptions are raised. Typically comes out as
```
with pytest.raises(ValueError) as error:
    assert str(error.value) == expected_msg
```

I completely agree on the dict not being ideal. I started thinking a bit about having a well-defined config object to which you can add groups. Perhaps something that would work like this:
```
sensitive_feature_config = SensitiveFeatureConfig(names=["gender", "age"])
sensitive_feature_config.add_group(gender="male", "age"="0-20", n_samples=123, class_sep=1.1, ...)
sensitive_feature_config.add_group(gender="female", "age"="0-20", n_samples=321, class_sep=0.6, ...)
sensitive_feature_config.add_group(gender="non-binary", "age"="0-20", n_samples=231, class_sep=1.7, ...)
...
sensitive_feature_config.add_group(gender="male", "age"="80+", n_samples=103, class_sep=1.05, ...)
sensitive_feature_config.add_group(gender="female", "age"="80+", n_samples=183, class_sep=1.4, ...)
sensitive_feature_config.add_group(gender="non-binary", "age"="80+", n_samples=123, class_sep=0.2, ...)
...
```
[I would allow all sorts of args to be passed through to `make_classification` via `kwargs`.]

Again, I think this goes quite a bit beyond the basic version you have in this PR, so I wouldn't expect this right away, and I feel like we should get some more input on this from others such as @adrinjalali @MiroDudik @riedgar-ms @hildeweerts

fairlearn/datasets/_synthetic.py
romanlutz: Oh neat, I didn't know about that one! 🙂 

I agree 100% on the duplication (and the future removal), but having good small examples in the API docs is also a priority.

fairlearn/datasets/_synthetic.py
```diff
@@ -0,0 +1,82 @@
...
+        genders.append([label] * group_config['n_samples'])
+        Xs.append(X)
+        ys.append(y)
+
+    X = np.concatenate(Xs)
+    y = np.concatenate(ys)
+    gender = np.concatenate(genders)
+
+    # Now generate disparities by randomly flipping classes
+    bias = {group_name: rng.uniform(0.4, 0.6) for group_name in feature_config}
```
romanlutz, line 75: I see what you're doing here and it's a nice and simple way of achieving this, although it doesn't necessarily guarantee that we'll have any disparities, which might mean that users have to try a few times with different seeds. We could alternatively make this `bias` a part of the input args. I've added another comment where I suggest using a class for setting the config, so you can think of this as another arg to the config. I do think this could be another part of the follow-up PR, though.

test/unit/datasets/test_datasets.py
romanlutz: Good idea! I added a comment separately, and I'm resolving this one. https://github.com/fairlearn/fairlearn/pull/907#discussion_r671991344

fairlearn/datasets/_synthetic.py
```diff
@@ -0,0 +1,82 @@
...
+
+    Parameters
+    ----------
+    feature_config : dict, default=None
+        A dict whose keys are the labels for the gender feature, and whose set
+        the :code:`n_samples` and :code:`class_sep` parameters for :code:`make_classification`.
+        If `None`, the following dictionary is used by default::
+
+            {
+                'Man': {'n_samples': 500, 'class_sep': 1},
```
riedgar-ms, line 19: Are these the keys we want to use?

fairlearn/datasets/_synthetic.py
```diff
@@ -0,0 +1,82 @@
...
+
+    X = np.concatenate(Xs)
+    y = np.concatenate(ys)
+    gender = np.concatenate(genders)
+
+    # Now generate disparities by randomly flipping classes
+    bias = {group_name: rng.uniform(0.4, 0.6) for group_name in feature_config}
+    for idx in range(y.size):
+        # 1/4 chance we'll overwrite this value
+        if rng.uniform(0, 1) > 0.25:
```
riedgar-ms, line 78: Should that chance be a parameter as well?

fairlearn/datasets/_synthetic.py
```diff
@@ -0,0 +1,82 @@
...
+                'Other': {'n_samples': 500, 'class_sep': 0.5},
+                'Unspecified': {'n_samples': 500, 'class_sep': 0.5},
+                'Woman': {'n_samples': 500, 'class_sep': 2},
+            }
+
+    n_features : int, default=20
+        The total number of features.
+
+    n_informative : int, default=4
+        The number of informative features.
```
riedgar-ms, line 29: A little more explanation might be nice here, even if it's mainly just passing through to SciKit. I'm guessing this means that four of the features will have some effect on the prediction, but the other 16 will be completely random?

fairlearn/datasets/_synthetic.py
```diff
@@ -0,0 +1,82 @@
+# Copyright (c) Fairlearn contributors.
+# Licensed under the MIT License.
+import numpy as np
+from sklearn.datasets import make_classification
+from sklearn.utils import check_random_state
+
+
+def make_synthetic_dataset(feature_config=None, n_features=20, n_informative=4, random_state=None):
```
riedgar-ms, line 8: Can this API reasonably be extended to multiple sensitive features in the future? If not, I'd suggest tweaking the name to reflect that it can only generate for a single sensitive feature.

fairlearn/datasets/_synthetic.py
```diff
@@ -0,0 +1,82 @@
...
+    """Create a synthetic dataset with a single sensitive feature: 'gender'.
```
riedgar-ms, line 9: This should explain how the 'biases' I see in the code down below are applied.

riedgar-ms (commented):
Thank you for contributing!

fairlearn/datasets/_synthetic.py
romanlutz: Already linked it in another comment🙂

fairlearn/datasets/_synthetic.py
romanlutz: Yes and it will be. I wrote sort of a proposal for what that might look like in another comment.

fairlearn/datasets/_synthetic.py
```diff
@@ -0,0 +1,82 @@
...
+            Labels for the binary classification.
+        gender : numpy.ndarray
+            The sensitive feature label.
+    """
+    rng = check_random_state(random_state)
+
+    classification_kwargs = {
+        'n_features': n_features,
+        'n_informative': n_informative,
+        'n_classes': 2,
```
LeJit, line 49: I feel like we should specify somewhere (either in function name or docstring) that this function produces a binary classification dataset.  Are there any plans to incorporate multi-class synthetic datasets in the future?

romanlutz: I suspect we'll have a separate function for multiclass, similar to what scikit-learn does. I'm all in favor of clarity, of course, so including "binary" in the name sounds very reasonable to me.

fairlearn/datasets/_synthetic.py
coreysharris: I was thinking about how to explain in this in light of the bias that we add by flipping classes, but the problem is that degrades these informative features. Rather, we should leave the labels alone, and rearrange the sensitive feature assignments.

This is actually much easier, because `make_classification` accepts a `weights` kwarg that lets you set the proportion of positive labels. Update incoming...

fairlearn/datasets/_synthetic.py
coreysharris: I've renamed the function to `make_sensitive_classification` and expanded the docstring.

fairlearn/datasets/_synthetic.py
coreysharris: I've replaced this with a much better solution. See https://github.com/fairlearn/fairlearn/pull/907#discussion_r673120236

fairlearn/datasets/_synthetic.py
```diff
@@ -0,0 +1,82 @@
...
+from sklearn.datasets import make_classification
+from sklearn.utils import check_random_state
+
+
+def make_synthetic_dataset(feature_config=None, n_features=20, n_informative=4, random_state=None):
+    """Create a synthetic dataset with a single sensitive feature: 'gender'.
+
+    Parameters
+    ----------
+    feature_config : dict, default=None
```
adrinjalali, line 16: I'm not sure if we need to give users this much flexibility. It's making the API quite obscure.

fairlearn/datasets/_synthetic.py
```diff
@@ -0,0 +1,82 @@
...
+
+    classification_kwargs = {
+        'n_features': n_features,
+        'n_informative': n_informative,
+        'n_classes': 2,
+        'random_state': rng,
+    }
+
+    if feature_config is None:
+        feature_config = {
```
adrinjalali, line 57: do we want these to be gender? Or is it enough to have them as `A, B, C, ...`? It doesn't matter what those categories are, does it?

fairlearn/datasets/_synthetic.py
adrinjalali: do we need this? Different ways of calling `make_classification` is making that bias already.

test/unit/datasets/test_datasets.py
coreysharris: This seems like a nice way to handle multiple features. I hadn't thought of dealing directly with group intersections in the config.

adrinjalali (commented):
Thanks @coreysharris. I'm not quite happy with the API yet. We need to fix that first I think.

fairlearn/datasets/_synthetic.py
coreysharris: I've expanded the docstring to say something about this.

fairlearn/datasets/_synthetic.py
romanlutz: I think it should be configurable (as mentioned in another comment).
You'd be surprised how often I've explained something with A, B, C, D and people didn't get it until I said "well, imagine that these are gender or age groups. I'm fine with it being something generic in the first version, though. Moving to a configurable option in a future PR is also in line with what I wrote in the other comment.

test/unit/datasets/test_datasets.py
adrinjalali: I like @romanlutz 's proposal, except that I'd have an object for each sensitive feature, and not have the Cartesian product written in the code.

I think it's worth having the API as something we like in the first version. I rather not have to deprecate later since we already know we don't like the API as is. I'm happy to have a `SensitiveFeatureGenerator` object, but the API needs to be a bit less verbose than what's proposed by Roman. I'm not sure if we have to give users this much control over how the data is generated. If they really need this much control, they can write their own generator. This function is supposed to generate some dummy data for demonstration purposes. It's not supposed to generate data similar to the real world issues we have.

coreysharris: I've tried something like this here: coreysharris/fairlearn#1. I opened it as a PR against this branch so as not to muddy the waters here if people don't like it. I think the API is much cleaner, and it makes it easier to extend functionality later.

One note: although it _looks_ like it now supports multiple features, that's not the case. The implementation needs to be reworked, but I haven't gotten that far.

adrinjalali:
An alternative is also not to have any configuration for the features allowed in the API (for now), and generate like 3 sensitive features, two categorical and one numerical, and later figure the rest out.

test/unit/datasets/test_datasets.py
romanlutz: Somehow I missed this, sorry for the delay!
I **love** it! Do you want to pull that change into your branch so that we can review it as part of this PR?

romanlutz:
FYI @coreysharris CHANGES.md migrated to `docs/user_guide/installation_and_version_guide/` and the corresponding file for you would be `v0.7.1.rst`. Sorry for the trouble!

adrinjalali:
@coreysharris should we pick this up? It's totally find if you want us to continue the PR

coreysharris:
Eek! Sorry, I thought I was still waiting on feedback. I'll clean things up and bring in the changes from my other branch.

test/unit/datasets/test_datasets.py
coreysharris: Ok, great! In order to support multiple features, the API has to change slightly. It looks more and more like your suggestion above 😛

coreysharris:
@romanlutz @adrinjalali @riedgar-ms @LeJit This is now working for multiple sensitive features.
It is in much more rough shape than the previous versions were, because I want to make sure you like this before I smooth out the sharp edges.

Here is the MetricFrame generated by the example notebook:
Untitled

docs/user_guide/installation_and_version_guide/v0.7.1.rst
```diff
@@ -9,6 +9,8 @@ v0.7.1
 * Relaxed checks made on :code:`X` in :code:`_validate_and_reformat_input()`
   since that is the concern of the underlying estimator and not Fairlearn.
 * Add support for Python 3.9
+* Added new `SensitiveDatasetMaker` in `fairlearn.datasets` to allow
+  for creation of a dataset with user-selected number of features.
```
romanlutz, line 13:
```suggestion
* Added new `SensitiveDatasetMaker` in `fairlearn.datasets` to allow
  for creation of a synthetic dataset with user-selected number of features.
```

examples/plot_synthetic.py
```diff
@@ -0,0 +1,67 @@
...
+
+rng = np.random.RandomState(42)
+
+gender_feature = SensitiveFeature('Gender', ['Man', 'Other', 'Unspecified', 'Woman'])
+age_feature = SensitiveFeature('Age', ['<21', '21-64', '>64'])
+
+dataset = SensitiveDatasetMaker(sensitive_features=[gender_feature, age_feature], random_state=rng)
+
+# need a convenience method here
+dataset.configured_groups[('Man', '<21')].classification_kwargs['n_samples'] = 100
```
romanlutz, line 35: When reading this example I'm not entirely sure whether the other groups were forgotten or whether it will "just work" by having all the other groups with random sizes. Can you add a little bit of description (as @adrinjalali suggested in another comment)?

fairlearn/datasets/_synthetic.py
```diff
@@ -0,0 +1,119 @@
...
+                      for group_name in feature.groups]
+            all_groups.append(groups)
+        group_dicts = []
+        for group in product(*all_groups):
+            group_dicts.append(OrderedDict(
+                [(feature_name, group_name) for feature_name, group_name in group]
+            ))
+        return group_dicts
+
+    def init_configured_groups(self):
```
romanlutz, line 75: Is this meant to be called by users or only internally? If it's only internal then please prefix with `_`.

fairlearn/datasets/_synthetic.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def __repr__(self):
+        return f'{self.name}[{",".join(self.groups)}]'
+
+
+class SensitiveDatasetMaker:
+
+    def __init__(self, sensitive_features=None, random_state=None):
+        self.rng = check_random_state(random_state)
+        self.sensitive_features = sensitive_features or []
```
romanlutz, line 54: I'm surprised `sensitive_features` is optional. What's the use case for having `sensitive_features=[]`?

fairlearn/datasets/_synthetic.py
```diff
@@ -0,0 +1,119 @@
...
+            X, y = make_classification(**group_config)
+            Xs.append(X)
+            ys.append(y)
+
+            for feature_name, feature_list in sensitive_features.items():
+                feature_list.extend(
+                    [group_dict[feature_name]] * group_config['n_samples']
+                )
+
+        X = np.concatenate(Xs)
```
romanlutz, line 113: Do we want to shuffle these? Otherwise the groups are separated. Ideally you'd hope the user considers that possibility, of course, but since we already have a random number generator here I'm wondering whether it's best to shuffle the X and y (the same way, of course). Wdyt?

test/unit/datasets/test_datasets.py
```diff
@@ -38,3 +41,49 @@ def test_dataset_as_X_y(self, as_frame, fetch_function):
...
+        counts = [127, 370, 324, 210]
+        for i in range(4):
+            assert np.sum(y[i*500:(i+1)*500]) == counts[i]
+
+        assert np.isclose(np.mean(X), 0.0097618)
+        assert X.shape == (2500, 20)
+        assert y.shape == (2500,)
+        assert gender.shape == (2500,)
+
+    def test_custom_make_sensitive_classification(self):
```
romanlutz, line 64: Would be awesome to have a test case with at least 2 sensitive features, too. In the current setup you seem to allow 0 as well, but if that's still changing we don't need a test since it would be obsolete. If you see 0 as a legitimate use case we'll need a test, too, of course.

fairlearn/datasets/_synthetic.py
```diff
@@ -0,0 +1,119 @@
...
+            group = tuple(group_dict.values())
+            self.configured_groups[group] = SensitiveFeatureGroupConfig(group_dict)
+
+    def make_sensitive_classification(self, **kwargs):
+        classification_kwargs = {
+            'n_features': 20,
+            'n_informative': 4,
+            'n_classes': 2,
+            'n_samples': 50,
+            'random_state': self.rng,
```
romanlutz, line 89: These seem like rather important default values that should be highlighted in the docstring of this method.

examples/plot_synthetic.py
romanlutz: After re-reading the code I realize now that all groups will have 300 rows except man<21 which will have 100. Originally, I had assumed that 300 is the total size and since man<21 gets 100 rows the remaining 200 are divided up into the other groups at random (or perhaps equally). This misunderstanding of mine shows that we'll need to be a bit clearer in explaining the functionality. I can see a few ways:
- rename the `n_samples` param because it's really `n_samples_per_group` OR change functionality (e.g., making that the total number of samples, but that's probably harder to achieve since the other way is already implemented)
- add a comment here and a nice docstring in the class explaining what's happening under the hood
- explain what happens to the groups that aren't manually configured
The bar plot below should show the group sizes, so that's great!

fairlearn/datasets/_synthetic.py
```diff
@@ -0,0 +1,119 @@
...
+
+    def __init__(self, group_dict, **classification_kwargs):
+        """ `group_dict` is a dict assigning to a group in each sensitive
+        feature category. E.g.,::
+            {
+                'gender': 'Other',
+                'age': '50-60',
+            }
+        `classification_kwargs` is the set of kwargs that will be passed on to
+        make_classification.
```
romanlutz, line 22: It may be preferable to have a description of the class, then an example and the parameters section.

examples/plot_synthetic.py
romanlutz, line 35: don't we need to skew the parameter passed to `make_classification`? Otherwise the classes will be indistinguishable for the model.

romanlutz (commented):
I like where this is going! Thanks a ton for making all the requested changes, @coreysharris !

@fairlearn/fairlearn-maintainers you should really take a look at this and provide feedback now if you have thoughts 🙂

examples/plot_synthetic.py
hildeweerts: I was also confused by this and I like @romanlutz's suggested solution!

hildeweerts (changes requested):
Many of the docstrings are incomplete or missing. Could you please make sure to add one to each class and function?

It is also not 100% clear to me what specific form of unfairness/bias is simulated in the dataset (which I expect will be much clearer once you've updated the docstrings and added more detailed comments to the example). Am I correct in assuming it includes (1) a different data distribution for each sensitive group, (2) option to generate different number of samples per sensitive group, but *not* (3) option to have different base rate (i.e., proportion of positives) for each sensitive group?

Thanks!

fairlearn/datasets/_synthetic.py
coreysharris: The idea was that one could instantiate the `SensitiveDatasetMaker` and add sensitive features later, then manually call `init_configured_groups`, but I think this is too complicated. I've removed this option.

fairlearn/datasets/_synthetic.py
coreysharris: See below. Fixed!

examples/plot_synthetic.py
coreysharris: Changed to `n_samples_per_group`. 

> OR change functionality (e.g., making that the total number of samples, but that's probably harder to achieve since the other way is already implemented)

I did implement this in the previous iteration, but I've removed it here, just to reduce complexity. If you think it would be useful it wouldn't be hard to bring back in.

docs/user_guide/installation_and_version_guide/v0.7.1.rst
coreysharris: Fixed, thanks!

fairlearn/datasets/_synthetic.py
coreysharris: I agree that shuffling is preferable. Fixed!

examples/plot_synthetic.py
```diff
@@ -0,0 +1,67 @@
...
+)
+import numpy as np
+from sklearn.metrics import accuracy_score, precision_score, recall_score
+from sklearn.model_selection import train_test_split
+from sklearn.tree import DecisionTreeClassifier
+
+rng = np.random.RandomState(42)
+
+gender_feature = SensitiveFeature('Gender', ['Man', 'Other', 'Unspecified', 'Woman'])
+age_feature = SensitiveFeature('Age', ['<21', '21-64', '>64'])
```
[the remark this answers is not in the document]
coreysharris: I've replaced these with more arbitrary values 😄

coreysharris:
> Am I correct in assuming it includes (1) a different data distribution for each sensitive group, (2) option to generate different number of samples per sensitive group, but not (3) option to have different base rate (i.e., proportion of positives) for each sensitive group?

Better documentation to come, but the short answer is that all three are included. (3) is randomized for each group, but can be overwritten by the user.

examples/plot_synthetic.py
romanlutz: I'm fine with `n_samples_per_group`. Would love to hear other @fairlearn/fairlearn-maintainers opinions, though!

romanlutz:
@coreysharris I'm checking in on a few open PRs to see if the author is still interested in continuing. If not, someone else can pick it up from here and continue. Of course, if you are still interested it's best that you continue even if it's not right away (no hurry!). Just let us know! If there's no response by 12/28 (two weeks from today) I'll assume others can pick it up.

PR #909, opened 2021-07-19
MNT add section for maintainers with avatars

romanlutz:
Adding a section to the "About us" page similar to scikit-learn's core-dev page. It auto-updates if you change your profile pic on GitHub. To make sure the single version doc build doesn't break the latest version is always checked in, too, so we need to update the script if we add or remove maintainers (which shouldn't happen too often).

image


hildeweerts:
oh no... 

image

hildeweerts (approved):

docs/about/maintainers.rst
```diff
@@ -0,0 +1,32 @@
+.. raw :: html
+
+    <!-- Generated by generate_maintainers_table.py -->
+    <div class="maintainers-container">
+    <style>
+      img.avatar {border-radius: 10px;}
+    </style>
+    <div>
+    <a href='https://github.com/adrinjalali'><img src='https://github.com/adrinjalali.png' class='avatar' /></a> <br />
```
riedgar-ms, line 9: Is this what's generated by the script below?

romanlutz: See PR description. Without that file you can't run the basic doc build (without multiversion). Also, I don't expect it to change every other week.

adrinjalali: I'm also a bit confused as why the single doc build would fail w/o this file. We need to fix our whole doc build :see_no_evil:

romanlutz: Well tbh I didn't try but if you have a missing file it should at least complain, shouldn't it? In any case it's rerun during multiversion build.

I consider the single version doc build a feature rather than a bug. I learned a lot about scikit-learn's doc build the past week and see how it works better in some ways, but it can't help you switch between versions of the same page (only to the start page of another version).

romanlutz: Correction: it doesn't fail, just prints the following and doesn't add anything.
```
fairlearn\docs\about\index.rst:152: WARNING: Problems with "include" directive path:
InputError: [Errno 2] No such file or directory: 'docs/about/maintainers.rst'.
```
image

romanlutz: I removed it in the latest commit.

PR #912, opened 2021-07-21
DOC fix generation script path and add generated file

romanlutz:
The doc build in CircleCI currently fails due to a script that it can't find. This PR fixes the corresponding path.

However, the generated file does not show up in the resulting website because we check out the `main` version (and other released versions) of the repo for the doc build. They don't have the generated file so it's not shown on the website. To fix it I propose to check in that file and update it using the script whenever needed. This really shouldn't happen all that often so I don't see a huge problem with that. @adrinjalali and @riedgar-ms objected to that before, though, so I'd like to hear what they have to say.

I don't really see another solution in the short term (unless someone is willing to spend a week on moving us over to something that doesn't use sphinx-multiversion).


adrinjalali (commented):
As an alternative, we can have it that the script is only run manually, and the file is directly used in the docs. We can add it to our release docs that the release manager should make sure the maintainers file is up to date by running the script. And remove running the script step from the doc build.

romanlutz:
> As an alternative, we can have it that the script is only run manually, and the file is directly used in the docs. We can add it to our release docs that the release manager should make sure the maintainers file is up to date by running the script. And remove running the script step from the doc build.

Done!

adrinjalali (approved):

PR #913, opened 2021-07-22
FEAT add python 3.9 to setup.py

romanlutz:
This PR formally adds Python 3.9 to setup.py and several of our pipelines. As it turns out we were running on Python 3.9 in some pipelines even before. I had intended for this PR to check whether the pipelines will succeed and then to take whatever steps necessary to make them work with Python 3.9.


adrinjalali (approved):
I'm surprised this wasn't there already.

riedgar-ms (approved):
Great if this is working in all builds :-)

romanlutz:
> I'm surprised this wasn't there already.

You're not alone 😀

Issue #665, opened 2020-12-31
Replace `get_dummies` in examples with `Pipeline`


romanlutz:
This issue is based on the following comment:
could we use a `Pipeline` instead of `get_dummies`? See #649

_Originally posted by @adrinjalali in https://github.com/fairlearn/fairlearn/pull/614#discussion_r547870456_

Using a `Pipeline` with all the preprocessing steps is cleaner and a better practice. However, we currently can't pass a `Pipeline` to mitigation techniques since they run `check_X_y` internally. If there are preprocessing steps, they are not executed before this check and we get a failure.

# Goals
- use `Pipeline` in examples, tests, etc. (we have several places with `get_dummies` which are almost identical
- fix mitigation technique behavior to work with `Pipeline`s that include preprocessing steps
- stretch goal (perhaps out of scope for the initial PR): make it possible to pass our mitigation techniques to `Pipeline`s.

romanlutz:
As far as I remember we wanted to get our mitigation techniques to be passed into `Pipeline`s from sklearn. That's not possible right now since we require `sensitive_features` which currently can't be passed to the `Pipeline`. I believe @adrinjalali mentioned that passing things like `sensitive_features` might be possible at some point. I found the following SLEP which I *think* refers to this https://scikit-learn-enhancement-proposals.readthedocs.io/en/latest/slep006/proposal.html
@adrinjalali if you have any other updates or thoughts to share please add 🙂

adrinjalali:
What we can do, is to pass a `Pipeline` to our mitigation techniques, which includes the preprocessing part. There's a bug in validating the input in our side which prevents us from doing so, but otherwise it would work.

romanlutz:
I just tried it out locally. It actually just works! I'll send out a (few?) PR(s) asap to create a notebook that shows this and then reference it everywhere else. This will also be super useful in #614 .

romanlutz:
Update: I created #916 but I can't seem to extract the code WITH output from the examples to show in the new user guide for #614  which is a shame because it would mean that I have to duplicate it after all. I thought I might be able to use the same example and just `literalinclude` all over the docs. @adrinjalali do you know of a fix for that?

Additionally, `Pipeline` seems to fail `check_is_fitted` so I check for the `vars(pipeline)` where end in `_` but don't start with `__` (which is what `check_is_fitted` does) and it's indeed `[]` so it's supposed to fail. Does `check_is_fitted` not apply to `pipeline`?
[...]

PR #916, opened 2021-07-25
DOC add an example with sklearn.pipeline.Pipeline 

romanlutz:
Adds an example that uses `sklearn.pipeline.Pipeline` to show that Fairlearn works with estimators that are made up of various steps. This is especially useful since we previously weren't compatible with `Pipeline`.

Update: refactored the following out into #918. ~~Additionally, I'm renaming `interpolation_dict` to `threshold_interpolation_` since we never liked that name to begin with (see #614) and it's really the information about how to interpolate between thresholds to get your predictions. If people dislike this name I can undo this, though. Since `InterpolatedThresholder` isn't a class that's actually visible in our API docs I don't think this constitutes a breaking change. However, we may want to consider making it public by explicitly mentioning it in the postprocessing module's `__init__.py`.~~

This is also related to #665 since it represents a first step towards replacing usage of `get_dummies`.
Furthermore, #614 will need to be updated following this PR.


examples/plot_mitigation_pipeline.py
```diff
@@ -0,0 +1,102 @@
...
+from fairlearn.postprocessing import ThresholdOptimizer, plot_threshold_optimizer
+from fairlearn.reductions import ExponentiatedGradient, DemographicParity
+from sklearn.compose import ColumnTransformer
+from sklearn.impute import SimpleImputer
+from sklearn.linear_model import LogisticRegression
+from sklearn.model_selection import train_test_split
+from sklearn.preprocessing import StandardScaler, OneHotEncoder
+from sklearn.compose import make_column_selector as selector
+from sklearn.pipeline import Pipeline
+
```
adrinjalali, line 28: I'd add a section here:

```suggestion
# %%
```

examples/plot_mitigation_pipeline.py
```diff
@@ -0,0 +1,102 @@
...
+(X_train, X_test, y_train, y_test, A_train, A_test) = train_test_split(
+    X_raw, y, A, test_size=0.3, random_state=12345, stratify=y)
+
+X_train = X_train.reset_index(drop=True)
+X_test = X_test.reset_index(drop=True)
+y_train = y_train.reset_index(drop=True)
+y_test = y_test.reset_index(drop=True)
+A_train = A_train.reset_index(drop=True)
+A_test = A_test.reset_index(drop=True)
+
```
adrinjalali, line 43: same here (with a one sentence saying what the sections do)

fairlearn/postprocessing/_interpolated_thresholder.py
```diff
@@ -78,10 +77,10 @@ class InterpolatedThresholder(BaseEstimator, MetaEstimatorMixin):
        [Online]. Available: https://arxiv.org/abs/1610.02413.
     """
 
-    def __init__(self, estimator, interpolation_dict, prefit=False,
+    def __init__(self, estimator, threshold_interpolation, prefit=False,
                  predict_method='deprecated'):
         self.estimator = estimator
-        self.interpolation_dict = interpolation_dict
+        self.threshold_interpolation_ = threshold_interpolation
```
adrinjalali, line 83: this looks like a meta-estimator to me, so I'd rather it follow estimator API and have this as `threshold_interpolation` instead of the one with a trailing underscore.

adrinjalali (approved):
I'm happy with the PR, but I think this is actually two separate PRs?

adrinjalali:
Also, to me this is a DOC PR rather than a FEAT

fairlearn/postprocessing/_interpolated_thresholder.py
romanlutz: It is indeed a meta-estimator. Changed it in the latest commit (which goes to a separate PR), but I have to say I wasn't aware of different rules for meta-estimators vs estimators. If you happen to have a reference I'd love to educate myself 🙂

romanlutz:
Refactored the `InterpolatedThresholder` changes out into #918. Let's merge #918 first and then this one (which includes the same changes and builds on them).

fairlearn/postprocessing/_interpolated_thresholder.py
adrinjalali: I meant "it's an estimator because it's a meta-estimator." All meta-estimators are estimators, and there isn't much to it other than: 

https://github.com/scikit-learn/scikit-learn/blob/45bb9ab916351c95a6d15a5cb06469036a0707cf/sklearn/base.py#L810-L813

adrinjalali: In more advanced cases, there are things such as _estimator tag delegation_ and passing `fit_params` and `score_params` to the underlying estimator, etc, but that can come next.

adrinjalali (approved):
LGTM, good to merge once the other PR is merged.

romanlutz:
Looks like the other PR may have a hard(er) time so I'll just remove those changes from here and will update the other one accordingly. That way we can merge this one asap.

examples/plot_mitigation_pipeline.py
```diff
@@ -0,0 +1,113 @@
...
+    prefit=False)
+threshold_optimizer.fit(X_train, y_train, sensitive_features=A_train)
+print(threshold_optimizer.predict(X_test, sensitive_features=A_test))
+print(json.dumps(
+    threshold_optimizer.interpolated_thresholder_.interpolation_dict,
+    default=str,
+    indent=4))
+plot_threshold_optimizer(threshold_optimizer)
+
+# %%
```
riedgar-ms, line 101: No GridSearch?

romanlutz: I didn't really want to show each and every mitigation technique, but felt like it makes sense to have one of postprocessing and reductions each. I expect we'll have, say, 20 mitigation techniques at some point, and we can't have all of them everywhere. Does that make sense?

riedgar-ms: OK

riedgar-ms (approved):

romanlutz:
Looks like CircleCI refuses to schedule a build for this pipeline. I've even tried adding another commit, but it didn't trigger it. To prove that it does indeed work I've run this locally. The changes are all on a single page, so it should be easy to check the screenshots:

image

image

image

The warning about the pipeline will be fixed once scikit-learn provides a way to properly check for whether a pipeline is fitted (currently in PR).

Since these look just like they should I'll go ahead and merge this since we have the necessary approvals and this has been sitting around for a few days.

romanlutz:
Voila: https://fairlearn.org/main/auto_examples/plot_mitigation_pipeline.html#sphx-glr-auto-examples-plot-mitigation-pipeline-py

adrinjalali:
Thanks @romanlutz

PR #917, opened 2021-07-25
MNT remove erroneously added deps.txt file

romanlutz:
This is a mistakenly added file from #913 from analyzing dependencies.


romanlutz:
This PR is theoretically blocked on the broken build-webpage pipeline but we could also just merge it since #912 will address that issue and this is guaranteed to be unrelated.

adrinjalali (approved):
yep, that sneaked in, sneaky Roman :D

PR #918, opened 2021-07-25
FEAT export InterpolatedThresholder as a public object and update its API (e.g., rename an argument)

romanlutz:
This PR does the following:

- Exposes `InterpolatedThresholder` and `ThresholdOperation` as public objects in `fairlearn.postprocessing` (adds them to `__all__`).
- **Breaking**: reshapes `InterpolatedThresholder`'s per-group interpolation data. The old `interpolation_dict` constructor argument (a `dict[group -> sklearn.utils.Bunch]`) is replaced by `threshold_interpolation`, a single `pandas.DataFrame` indexed by sensitive feature value with columns `p0`, `operation0`, `p1`, `operation1` (plus `p_ignore`, `prediction_constant` for equalized odds).
  - Before: `it.interpolation_dict[group].p0`
  - After: `it.threshold_interpolation.loc[group, "p0"]`
  - Nested-dict view still available via `it.threshold_interpolation.to_dict("index")`.
- Updates the docstring, `versionchanged:: 0.15` note, doctests in `docs/user_guide/mitigation/postprocessing.rst`, and the `plot_mitigation_pipeline.py` example to match the new shape.
- Adds a `Breaking changes` entry in `docs/user_guide/installation_and_version_guide/v0.15.0.rst` with a before/after snippet.
- Refactored out of #916 to minimize the scope of each PR.

Addresses review feedback from @MiroDudik on flattening the deep dict-of-`Bunch` structure before making the class public.


adrinjalali (approved):

fairlearn/postprocessing/_interpolated_thresholder.py
```diff
@@ -21,14 +21,13 @@ class InterpolatedThresholder(BaseEstimator, MetaEstimatorMixin):
 
     At prediction time, the predictor takes as input both standard and sensitive features.
     Based on the values of sensitive features, it then applies a randomized thresholding
-    transformation according to the provided `interpolation_dict`.
+    transformation according to the provided `threshold_interpolation`.
```
MiroDudik, line 24: why this change? if you don't like the original, we should go for something more concise. maybe something like `threshold_info` or `thresholding_info` or even `thresholds` (but that is a bit confusing, because it sounds like the structure would be simpler).

fairlearn/postprocessing/_interpolated_thresholder.py
```diff
@@ -21,14 +21,13 @@ class InterpolatedThresholder(BaseEstimator, MetaEstimatorMixin):
...
 
     Parameters
     ----------
     estimator :
         base estimator
-
-    interpolation_dict : dict
+    threshold_interpolation : dict
```
MiroDudik, line 30: See my comment above re. naming. By the way, do we like the structure of this argument? The main reason why I've originally excluded this estimator from the module was that I wasn't sure we all liked this deep structure and I didn't want to commit to it... I usually prefer "shallower" input arguments.

fairlearn/postprocessing/__init__.py
```diff
@@ -7,10 +7,12 @@
...
 """
 
+from ._interpolated_thresholder import InterpolatedThresholder  # noqa: F401
 from ._threshold_optimizer import ThresholdOptimizer  # noqa: F401
 from ._plotting import plot_threshold_optimizer  # noqa: F401
 
 __all__ = [
-    "ThresholdOptimizer",
-    "plot_threshold_optimizer"
+    "InterpolatedThresholder",
```
MiroDudik, line 15: You'll also want to export `ThresholdOperation`. That's another structured object, where we should review whether we like its current API enough to make it external.

MiroDudik (changes requested):
Since we are taking a private object and making it public, we should review its API to make sure we like.

fairlearn/postprocessing/_interpolated_thresholder.py
romanlutz: Info isn't a great name. Everything is "info", yet we don't call "width" variables "width_info". It's a stylistic improvement as well as a description improvement: interpolation and threshold are the two important things worth mentioning IMO.
If you don't think it's an improvement I don't mind scrapping it, though (not the hill I choose to die on). It was originally part of an older PR about documenting post processing (which I'm actively working on) and got refactored out into its own change for simplicity.

romanlutz:
> Since we are taking a private object and making it public, we should review its API to make sure we like.

Adding to the list: tests!

fairlearn/postprocessing/_interpolated_thresholder.py
MiroDudik: How about `threshold_weights` since basically each value of a threshold gets a certain probabilistic weight?

romanlutz: I don't dislike it, but that's just p0 and p1 🙂 

It's really not a bad name though...

TamaraAtanasoska:
this is something started really long ago, would it be still nice to expose `InterpolatedThresholder`? From what I gather above there are a few things missing:
- tests
- naming conventions 
- decision about if the API is good enough in some cases 

Have maybe some opinions changed about this @MiroDudik and @romanlutz? We can also discuss this at the meeting as it seems it might need some discussion as naming and API stuff do :)

romanlutz:
I know some people felt strongly about this at some point 🤣 Personally, I'm happy with it as is. The naming improvement is truly a small one if at all. Happy to abandon if that's preferable.

TamaraAtanasoska:
@romanlutz great :) are you interested in seeing it through sometime soon when time permits? I think practically it is the tests only missing the rest is opinion/discussion based.

fairlearn/postprocessing/_interpolated_thresholder.py
romanlutz: Revisiting this after merging main. I ended up flattening the structure in this PR as well (see the thread below), so I wanted to pick a name that fit the flatter shape too. Went with `threshold_interpolation`: it's now a `pandas.DataFrame` indexed by sensitive feature value with columns `p0`, `operation0`, `p1`, `operation1` (and `p_ignore`, `prediction_constant` for equalized odds), so `threshold_interpolation` reads as "the interpolation table, one row per group". Not attached to the name though — happy to switch to `interpolation_rules` or similar if you'd prefer. Added a `versionchanged:: 0.15` note, a `Breaking changes` entry in `v0.15.0.rst` with a before/after snippet, and updated the docstring, doctests, and example to match.

fairlearn/postprocessing/_interpolated_thresholder.py
romanlutz: On reflection, you're right — flattening the structure before making this public is the right call, and I didn't want to ship a deep structure just to redo the API a release later. Reshaped in this PR: `threshold_interpolation` is now a `pandas.DataFrame` indexed by sensitive feature value, with columns `p0`, `operation0`, `p1`, `operation1`, plus `p_ignore` and `prediction_constant` when the equalized-odds constraint is used (rather than a `dict[group -> sklearn.utils.Bunch]`). The DP vs EO shape difference is now a column-presence check instead of a per-`Bunch` attribute lookup, and users get a compact tabular repr by default, with `.to_dict('index')` available for the nested-dict view (the doctests use that for stable output). Called out as a breaking change in `v0.15.0.rst` with a before/after snippet.

fairlearn/postprocessing/__init__.py
romanlutz: Done — `ThresholdOperation` is now exported from `fairlearn.postprocessing` and listed in `__all__` alongside `InterpolatedThresholder`. Its own API is tiny (`threshold` and `operator`) so I don't think it needs reshaping the way the interpolation dict did; happy to revisit in a follow-up if you disagree.

Issue #738, opened 2021-04-09
Move changelog to the main documentation


riedgar-ms:
#### Describe the issue linked to the documentation

The `CHANGES.md` file is only accessible in the repo, and not part of the main website (fairlearn.org)

#### Suggest a potential alternative/fix

Move (and convert to ReST) the file so that the file appears on fairlearn.org. Ideally it would be kept prominent in the repo itself, though (which might complicate things a little)

romanlutz:
We've had this discussion a while ago, although I'm having trouble tracking down where exactly. What I remember from that time:
- CHANGES.md is meant for ALL changes to the package (excluding, for example, fixing a build pipeline since that's not going on PyPI) made, regardless of how small. This is mainly for developers, not users.
- The website should have something similar, but reduced to just the most important changes for users. Example in scikit-learn: https://scikit-learn.org/stable/auto_examples/release_highlights/plot_release_highlights_0_24_0.html
- [upgrading instructions](https://fairlearn.org/v0.6.0/user_guide/migrating_versions/index.html): this is somewhat related and we've neglected that again in the last few releases

I'm definitely not saying that we need to stick with that system, but I do see the point of the separation into ALL changes and a well-described set of important changes to users. It's basically serving a different audience (at least in my mental model). Wdyt? @fairlearn/fairlearn-maintainers

hildeweerts:
I like the idea of release highlights a lot (with the added bonus that it's very tweet-able #fairlearn)!

Upgrading instructions sound like something that we probably should be doing (although they also don't sound very fun to be doing, lol). Ideally we don't add too many breaking changes which will hopefully make future upgrading instructions very minimal.

adrinjalali:
Upgrading instructions could go into the release highlights, to keep it simple if we want.

The changelog is something like the one you have here:
https://scikit-learn.org/stable/whats_new/v0.24.html

It does include everything which changes something for the users, but it's still different from the release highlights, and it doesn't have to include CI commits for instance.

The idea is that the changelog is updated at the same time as the PRs usually, if the PR touches on something that users should know about. But not all users care about all those changes, hence the existence of the release highlights.

romanlutz:
I've started taking a stab at this. My upcoming PR moves the previously called CHANGES.md to the website (it's a lot harder than it sounds! 🤣 ). From there we can add a line per PR, and then distill that into highlights (basically a kind of TL;DR section I'm thinking...).

Open questions:
[...]

PR #919, opened 2021-07-26
DOC move changelog to the website

romanlutz:
This is a first step towards #738.
The contents are not changed other than formatting & typo fixes. For versions where a migration guide existed the migration guide was appended to the version changelog.
This PR also removes numbering from the user guide since section names like "4.2.1 v0.4.6" are completely unreadable. If you have objections please voice them!  
Finally, the release instructions needed to be updated slightly to reflect this new changelog location.


adrinjalali:
Awesome work @romanlutz , I'd be even happy with having all _older_ files to be together rather than separate, but you've done it all already.

I think installation guide is not of the same _level_ of importance compared to change log. I'm happy if the change log is a bit harder to reach compared to installation guide.

romanlutz:
> Awesome work @romanlutz , I'd be even happy with having all _older_ files to be together rather than separate, but you've done it all already.
> 
> 
> 
> I think installation guide is not of the same _level_ of importance compared to change log. I'm happy if the change log is a bit harder to reach compared to installation guide.

This is just a first pass. In the future I hope to add
- list of PRs that got merged similar to sklearn (should be updated in every PR)
- release highlights (sort of what we have right now, but with more details)
- credit to the release manager (that is so far mostly @riedgar-ms 🙂 )
- potentially credit to people who participated in filing and discussing issues when they get closed
- anything else people want to include

So that means some extra content per release and lots of these files will be a bit more substantial than they are now and would warrant their own file(s).

Is there anything you want me to do regarding the second part (re installation guide)? I agree 100%. @shimst3r added this entire section recently 🥳

riedgar-ms (approved):

romanlutz:
@adrinjalali if this works for you as a first version please merge. I'll be iterating on this in follow-up PRs as mentioned earlier.

romanlutz:
Yes! This is what it looks like right now: https://fairlearn.org/main/user_guide/installation_and_version_guide/v0.7.0.html

There's definitely something off in the way the list on the bottom left is rendered but it doesn't really affect anything much
image

PR #925, opened 2021-08-04
DOC small doc formatting fixes

romanlutz:
Moving some more tiny changes out of #614 to minimize the scope of changes there.


adrinjalali (approved):
Not sure why you have so many commits here, but this LGTM.

romanlutz:
It's because it came out of #614 and I just removed other changes rather than starting afresh and including these changes. Shouldn't make a difference once it's squashed, though.

PR #929, opened 2021-08-05
DOC Add versionadded and versionchanged to API reference

matthew-so:
Just a test to ensure that I am properly adding versionadded/versionchanged documentation. Are my current additions correct, given the functions' edit history?


fairlearn/datasets/_fetch_adult.py
```diff
@@ -25,6 +25,9 @@ def fetch_adult(*, cache=True, data_home=None,
     Prediction task is to determine whether a person makes over $50,000 a
     year.
 
+    ..versionadded: 0.5.0
+    ..versionchanged: 0.7.1
```
adrinjalali, line 29: version changed should have a comment explaining what's changed.

romanlutz: In this particular case nothing changed other than the documentation, so I think we can remove that line and keep the `versionadded` one only.

romanlutz: Come to think of it, the same applies to the other two in this module.

matthew-so:
Sorry for the delay; classes just started. I'll get it done as soon as possible.

romanlutz (approved):

romanlutz:
@fairlearn/fairlearn-maintainers can we merge this?

hildeweerts:
Did someone check whether this works locally? It seems that other libraries (e.g., scikit-learn) usually use something like
`.. versionadded:: 0.5.0` rather than `..versionadded: 0.5.0`, see also [sphinx docs](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionadded).

fairlearn/datasets/_fetch_adult.py
```diff
@@ -25,6 +25,8 @@ def fetch_adult(*, cache=True, data_home=None,
     Prediction task is to determine whether a person makes over $50,000 a
     year.
 
+    ..versionadded: 0.5.0
```
romanlutz, line 28:
```suggestion
    .. versionadded: 0.5.0
```

fairlearn/datasets/_fetch_bank_marketing.py
```diff
@@ -30,6 +30,7 @@ def fetch_bank_marketing(*, cache=True, data_home=None,
     The classification goal is to predict if the client will subscribe a
     term deposit (variable y).
 
+    ..versionadded: 0.5.0
```
romanlutz, line 33:
```suggestion
    .. versionadded: 0.5.0
```

fairlearn/datasets/_fetch_boston.py
```diff
@@ -55,6 +55,7 @@ def fetch_boston(*, cache=True, data_home=None,
     MEDV     Median value of owner-occupied homes in $1000's
     =======  ======================================================================
 
+    ..versionadded: 0.5.0
```
romanlutz, line 58:
```suggestion
    .. versionadded: 0.5.0
```

romanlutz (commented):
Thanks for the suggestion, @hildeweerts ! Better to be safe and add the whitespace.

hildeweerts:
> Thanks for the suggestion, @hildeweerts ! Better to be safe and add the whitespace.

It's not just whitespace but also a double colon!

romanlutz:
> > Thanks for the suggestion, @hildeweerts ! Better to be safe and add the whitespace.
> 
> It's not just whitespace but also a double colon!

Ooops. Updated!

romanlutz:
@hildeweerts Thanks again for the suggestions!!! Do you want to merge this?

PR #934, opened 2021-08-11
Removed py3.9 dependency from nightly-requirements

### LINKED ISSUE #915: MNT Bump requirements for compatibility with python 3.9
Some of our min requirements don't exist in Python 3.9, e.g., scikit-learn seems to have adopted python 3.9 in version 0.24, whereas our nightly "requirements-fixed" build runs with our minimum requirement of 0.22.1 and fails accordingly.

Should we bump our min requirements to

```
numpy>=1.20.0
pandas>=1.2.0
scikit-learn>=0.24.0
scipy>=1.6.0
```
from currently
```
numpy>=1.17.2
pandas>=0.25.1
scikit-learn>=0.22.1
scipy>=1.4.1
```
(and probably several others in requirements-dev.txt as well as `matplotlib>=3.4.0` for requirements-customplots.txt)?

I'm generally in favor of moving forward rather than staying on old versions, but I know there are people who are forced into old versions of some packages that might not be able to use our latest version as a consequence.

@fairlearn/fairlearn-maintainers Wdyt?

sushmitaS16:
closes #915 

Python version 3.9 dependency is removed from [nightly-requirements-fixed.yml](https://github.com/fairlearn/fairlearn/blob/main/devops/nightly-requirements-fixed.yml)

### DIFF
--- a/devops/nightly-requirements-fixed.yml
+++ b/devops/nightly-requirements-fixed.yml
@@ -1,6 +1,6 @@
 # Nightly build pipeline with requirements pinned to lower bounds (if any)
 # This does not run on MacOS due to issues with the pinned requirements
-# There are also issues with running on Python 3.9
+# There are also issues with running on Python 3.9, hence dependency removed (for now)
 
 variables:
   FreezeArtifactStem: 'freeze'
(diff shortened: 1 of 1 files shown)


romanlutz (approved):
Thank you @sushmitaS16 !!!

I can't kick off a test run since our Azure DevOps instance only takes branches from our repo, but not forks. Still, it succeeded on those remaining three python versions as recently as last night: https://dev.azure.com/fairlearn/Fairlearn/_build/results?buildId=208&view=results

So this looks fine to me and can be merged as far as I am concerned. @fairlearn/fairlearn-maintainers

adrinjalali (approved):

Issue #940, opened 2021-09-02
Bug when using custom Lagrange multiplier grids


AndreFCruz:
#### Describe the bug

When passing a custom set of Lagrangian multipliers to the `GridSearch` reduction, via the `grid` kwarg, the loss computation for model selection throws a `KeyError`.

The multipliers at `self.gammas_`, which is a pandas.DataFrame, are being inserted using the name of the column as index.
See [line 162](https://github.com/fairlearn/fairlearn/blob/main/fairlearn/reductions/_grid_search/grid_search.py#L162) for the values of `i` and [line 194](https://github.com/fairlearn/fairlearn/blob/main/fairlearn/reductions/_grid_search/grid_search.py#L194) for how the values are inserted.

Afterwards, in [line 201](https://github.com/fairlearn/fairlearn/blob/main/fairlearn/reductions/_grid_search/grid_search.py#L201), these values are accessed assuming the indices are ordered integer values.
This works when using the default GridGenerator because the column names are indeed ordered integer indices, but not when providing a custom DF.

#### Steps/Code to Reproduce

Running a GridSearch reduction with a custom grid represented by a DF whose column names are not ordered integer indices.

#### Versions
This happens on all versions as far as I can tell, including the most recent on the `main` branch.

romanlutz:
@riedgar-ms should we close this since #941 was merged?

AndreFCruz:
Yep, closed :slightly_smiling_face:

PR #941, opened 2021-09-02
Fixed bug when using custom Lagrange multipliers #940

AndreFCruz:
Fixes issue #940 

Please see the issue for more details.

### DIFF
--- a/docs/user_guide/installation_and_version_guide/v0.7.1.rst
+++ b/docs/user_guide/installation_and_version_guide/v0.7.1.rst
@@ -12,6 +12,6 @@ v0.7.1
 * Added error handling in :code:`MetricFrame`. Methods :code:`group_max`, :code:`group_min`,
   :code:`difference` and :code:`ratio` now accept :code:`errors` as a parameter,
   which could either be :code:`raise` or :code:`coerce`.
-* Fixed a bug whereby passing a custom :code: `grid` object to a :code: `GridSearch`
-  reduction would result in a :code: `KeyError` if the column names were not ordered
+* Fixed a bug whereby passing a custom :code:`grid` object to a :code:`GridSearch`
+  reduction would result in a :code:`KeyError` if the column names were not ordered
   integers.
--- a/docs/user_guide/installation_and_version_guide/v0.7.1.rst
+++ b/docs/user_guide/installation_and_version_guide/v0.7.1.rst
@@ -12,3 +12,6 @@ v0.7.1
 * Added error handling in :code:`MetricFrame`. Methods :code:`group_max`, :code:`group_min`,
   :code:`difference` and :code:`ratio` now accept :code:`errors` as a parameter,
   which could either be :code:`raise` or :code:`coerce`.
+* Fixed a bug whereby passing a custom :code: `grid` object to a :code: `GridSearch`
+  reduction would result in a :code: `KeyError` if the column names were not ordered
+  integers.
--- a/fairlearn/reductions/_grid_search/grid_search.py
+++ b/fairlearn/reductions/_grid_search/grid_search.py
@@ -198,7 +198,7 @@ class GridSearch(BaseEstimator, MetaEstimatorMixin):
         if self.selection_rule == TRADEOFF_OPTIMIZATION:
             def loss_fct(i):
                 return self.objective_weight * self.objectives_[i] + \
-                    self.constraint_weight * self.gammas_[i].max()
+                    self.constraint_weight * self.gammas_[grid.columns[i]].max()
             losses = [loss_fct(i) for i in range(len(self.objectives_))]
             self.best_idx_ = losses.index(min(losses))
         else:
--- a/test/unit/reductions/grid_search/test_grid_search_arguments.py
+++ b/test/unit/reductions/grid_search/test_grid_search_arguments.py
@@ -3,7 +3,6 @@
 
 import logging
 import numpy as np
-from numpy.lib.function_base import disp
 import pandas as pd
 import pytest
 from sklearn.linear_model import LogisticRegression, LinearRegression
@@ -247,7 +246,8 @@ class ArgumentTests:
 
         disparity_moment.load_data(X, y, sensitive_features=A)
 
-        grid = _GridGenerator(grid_size, grid_limit,
+        grid = _GridGenerator(
+            grid_size, grid_limit,
             disparity_moment.pos_basis, disparity_moment.neg_basis,
             disparity_moment.neg_basis_present, False, grid_offset).grid
 
@@ -267,7 +267,6 @@ class ArgumentTests:
             transformY(y),
             sensitive_features=transformA(A))
 
-
     # ----------------------------
 
     @pytest.mark.parametrize("transformA", candidate_A_transforms)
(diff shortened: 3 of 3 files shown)


riedgar-ms (changes requested):
Please can you include a test of this change.

AndreFCruz:
I added a test checking that the GridSearch fit now runs successfully when using custom grids. Can you check it?

test/unit/reductions/grid_search/test_grid_search_arguments.py
```diff
@@ -231,6 +232,40 @@ def test_grid_size_warning_up_to_5_sensitive_feature_group(self, transformX, tra
...
             assert len(log_records) == 0
 
+    @pytest.mark.parametrize("transformX", candidate_X_transforms)
+    @pytest.mark.parametrize("transformY", candidate_Y_transforms)
+    @pytest.mark.parametrize("transformA", candidate_A_transforms)
+    def test_custom_grid(self, transformX, transformY, transformA):
+
+        grid_size = 10
+        grid_limit = 2.0
+        grid_offset = 0.1
```
riedgar-ms, line 242: I'm guessing that the `grid_offset` is the key? If so, can you add a comment to that effect

test/unit/reductions/grid_search/test_grid_search_arguments.py
```diff
@@ -231,6 +232,40 @@ def test_grid_size_warning_up_to_5_sensitive_feature_group(self, transformX, tra
...
+        X, y, A = _quick_data(False)
+
+        disparity_moment.load_data(X, y, sensitive_features=A)
+
+        grid = _GridGenerator(grid_size, grid_limit,
+            disparity_moment.pos_basis, disparity_moment.neg_basis,
+            disparity_moment.neg_basis_present, False, grid_offset).grid
+
+        indices = [7, 3, 4]
+        grid = grid.iloc[:, indices]
```
riedgar-ms, line 254: Or is it this bit which matters? An extra comment of two in the test code would be appreciated.

riedgar-ms:
Tagging @MiroDudik who wrote the `GridGenerator`

romanlutz:
A change description in the corresponding file for v0.7.1 under docs/user_guide in the version directory would also be appreciated 🙂

test/unit/reductions/grid_search/test_grid_search_arguments.py
AndreFCruz: Yes, this is the bit that matters, which makes a grid DataFrame whose column names are not ordered integers. I'll add a comment.

test/unit/reductions/grid_search/test_grid_search_arguments.py
AndreFCruz: Nope, these values don't matter, I'm just using the defaults.

AndreFCruz:
Just added the changelog comment and clarification comments to the test.

docs/user_guide/installation_and_version_guide/v0.7.1.rst
```diff
@@ -12,3 +12,6 @@ v0.7.1
 * Added error handling in :code:`MetricFrame`. Methods :code:`group_max`, :code:`group_min`,
   :code:`difference` and :code:`ratio` now accept :code:`errors` as a parameter,
   which could either be :code:`raise` or :code:`coerce`.
+* Fixed a bug whereby passing a custom :code: `grid` object to a :code: `GridSearch`
+  reduction would result in a :code: `KeyError` if the column names were not ordered
+  integers.
```
romanlutz, line 17:
```suggestion
* Fixed a bug whereby passing a custom :code:`grid` object to a :code:`GridSearch`
  reduction would result in a :code:`KeyError` if the column names were not ordered
  integers.
```

best to remove the whitespace to avoid confusion sphinx

riedgar-ms:
There are a few `flake8` errors. Are you able to see:
https://dev.azure.com/responsibleai/fairlearn/_build/results?buildId=12924&view=logs&jobId=7f33e5bd-7764-5d8a-ba2e-506e078b9c3f&j=7f33e5bd-7764-5d8a-ba2e-506e078b9c3f&t=b6623cb2-2223-5dfe-fa7b-0ad47bab4310

AndreFCruz:
my bad, the pipeline passes now

fairlearn/reductions/_grid_search/grid_search.py
```diff
@@ -198,7 +198,7 @@ def predict_fct(X): return current_estimator.predict(X)
         if self.selection_rule == TRADEOFF_OPTIMIZATION:
             def loss_fct(i):
                 return self.objective_weight * self.objectives_[i] + \
-                    self.constraint_weight * self.gammas_[i].max()
+                    self.constraint_weight * self.gammas_[grid.columns[i]].max()
```
romanlutz, line 201: I'm honestly a bit confused here. Above it says `for i in grid.columns` which should iterate over the column names, right? So how does `grid.columns[i]` fix this?

`grid` doesn't have a specified input type in the API reference so I'm a bit worried we're breaking this in case people pass a `np.ndarray`. Perhaps I'm misinterpreting this but I think so far we assumed `grid` to be a list of columns. I don't quite know for sure since I didn't write the original code and re-reading it isn't making things super obvious either. @riedgar-ms  @MiroDudik

AndreFCruz: Line 162 says `for i in grid.columns`, but the `i` on line 201 refers to the argument of the `loss_fct` which comes from the list comprehension on line 202 `for i in range(len(self.objectives_))`

test/unit/reductions/grid_search/test_grid_search_arguments.py
```diff
@@ -231,13 +231,50 @@ def test_grid_size_warning_up_to_5_sensitive_feature_group(self, transformX, tra
...
+            estimator=LogisticRegression(solver='liblinear'),
+            constraints=EqualizedOdds(),
+            grid=grid,
+        )
+
+        # Check that fit runs successfully with the custom grid
+        gs.fit(
+            transformX(X),
+            transformY(y),
+            sensitive_features=transformA(A))
```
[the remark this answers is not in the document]
AndreFCruz: Yep that does make sense. I added it.

riedgar-ms (approved):
I'm OK with this but @MiroDudik should review as well.

MiroDudik (approved):
Thanks for catching this. The bugfix looks fine to me and I'm fine merging this in.

But this bug has uncovered a potentially nasty issue, where we change the indexing between input (based on `grid.columns`) and the output, meaning: `predictors_`, `objectives_`, `oracle_execution_times_`, and `best_idx_`, which is integral. I imagine that much confusion can ensue in cases where grid.columns are integers that are not of the form 0,1,...,n-1.

riedgar-ms:
> But this bug has uncovered a potentially nasty issue, where we change the indexing between input (based on `grid.columns`) and the output, meaning: `predictors_`, `objectives_`, `oracle_execution_times_`, and `best_idx_`, which is integral. I imagine that much confusion can ensue in cases where grid.columns are integers that are not of the form 0,1,...,n-1.

Have you opened an issue on that @MiroDudik ?

PR #942, opened 2021-09-06
FEAT Introduce StreamingMetricFrame to accumulate batches

### LINKED ISSUE #655: Support fort streaming metrics

Dref360:
Introduce StreamingMetricFrame to accumulate batches and enable batch evaluation.

This version focuses on maintainability and is much easier to work with than the previous one. Most checks are now done by MetricFrame.

Fixes #655 

### Notes

In the previous PR (#670), we mentioned that converting everything to pandas would be valuable.
I think this would be useful, but this approach gets complicated with nested dict found in `sample_params`. I am open to propositions on how to fix that and maybe get rid of `StreamingMetricFrame._concat_batches`.

### DIFF
--- a/fairlearn/metrics/_streaming_metric_frame.py
+++ b/fairlearn/metrics/_streaming_metric_frame.py
@@ -107,11 +107,13 @@ class StreamingMetricFrame:
             control features are present.
         """
         check_consistent_length(y_true, y_pred)
+        check_consistent_length(y_true, sensitive_features)
         self._y_true = self._concat_if_not_none(self._y_true, y_true)
         self._y_pred = self._concat_if_not_none(self._y_pred, y_pred)
         self._sensitive_features = self._concat_if_not_none(self._sensitive_features,
                                                             sensitive_features)
         if control_features is not None:
+            check_consistent_length(y_true, control_features)
             self._control_features = self._concat_if_not_none(self._control_features,
                                                               control_features)
         elif self._control_features is not None:
@@ -121,7 +123,7 @@ class StreamingMetricFrame:
             self._sample_params = self._concat_if_not_none(self._sample_params,
                                                            sample_params)
 
-    def get_metric(self) -> MetricFrame:
+    def get_metric_frame(self) -> MetricFrame:
         """Get the MetricFrame computed on the accumulated values."""
         if self._y_true is None:
             raise ValueError(_EMPTY_BATCHES_ERR)
--- a/fairlearn/metrics/_streaming_metric_frame.py
+++ b/fairlearn/metrics/_streaming_metric_frame.py
@@ -53,11 +53,15 @@ class StreamingMetricFrame:
         self._sample_params = None
 
     def reset(self):
-        """Reset accumulators: y_true, y_pred, sensitive_features and control_features."""
+        """Reset accumulators.
+
+        Reset y_true, y_pred, sensitive_features, control_features and sample_params.
+        """
         self._y_true = None
         self._y_pred = None
         self._sensitive_features = None
         self._control_features = None
+        self._sample_params = None
 
     def add_data(self, y_true, y_pred, sensitive_features, control_features=None,
                  sample_params: Optional[Union[Dict[str, Any], Dict[str, Dict[str, Any]]]] = None):
--- a/fairlearn/metrics/_streaming_metric_frame.py
+++ b/fairlearn/metrics/_streaming_metric_frame.py
@@ -1,0 +1,178 @@
+# Copyright (c) Microsoft Corporation and Fairlearn contributors.
+# Licensed under the MIT License.
+
+from typing import Callable, Dict, Optional, Any, Union, List, TypeVar
+
+import numpy as np
+import pandas as pd
+from sklearn.utils import check_consistent_length
+
+from fairlearn.metrics import MetricFrame
+
+_EMPTY_BATCHES_ERR = "No data to process, please add batches with `add_data`."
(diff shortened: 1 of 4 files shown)


fairlearn/metrics/_streaming_metric_frame.py
```diff
@@ -0,0 +1,182 @@
...
+            self._control_features = self._concat_if_not_none(self._control_features,
+                                                              control_features)
+        elif self._control_features is not None:
+            raise ValueError(_CF_BAD_STATE_ERR)
+
+        if sample_params is not None:
+            self._sample_params = self._concat_if_not_none(self._sample_params,
+                                                           sample_params)
+
+    def get_metric(self) -> MetricFrame:
```
riedgar-ms, line 124: Perhaps `evaluate_metrics()` ? @MiroDudik @hildeweerts @adrinjalali ?

fairlearn/metrics/_streaming_metric_frame.py
```diff
@@ -0,0 +1,182 @@
...
+            a nested dictionary, with the first set of string keys identifying the
+            metric function name, with the values being the string-to-array-like dictionaries.
+
+            **Note** the types returned by members of the class vary based on whether
+            control features are present.
+        """
+        check_consistent_length(y_true, y_pred)
+        self._y_true = self._concat_if_not_none(self._y_true, y_true)
+        self._y_pred = self._concat_if_not_none(self._y_pred, y_pred)
+        self._sensitive_features = self._concat_if_not_none(self._sensitive_features,
```
riedgar-ms, line 112: Should sensitive features be checked for consistent length?

fairlearn/metrics/_streaming_metric_frame.py
```diff
@@ -0,0 +1,182 @@
...
+                                                            sensitive_features)
+        if control_features is not None:
+            self._control_features = self._concat_if_not_none(self._control_features,
+                                                              control_features)
+        elif self._control_features is not None:
+            raise ValueError(_CF_BAD_STATE_ERR)
+
+        if sample_params is not None:
+            self._sample_params = self._concat_if_not_none(self._sample_params,
+                                                           sample_params)
```
riedgar-ms, line 122: Shouldn't this have an else clause like the control features? And a consistent length check?

fairlearn/metrics/_streaming_metric_frame.py
```diff
@@ -0,0 +1,182 @@
...
+            **Note** the types returned by members of the class vary based on whether
+            control features are present.
+        """
+        check_consistent_length(y_true, y_pred)
+        self._y_true = self._concat_if_not_none(self._y_true, y_true)
+        self._y_pred = self._concat_if_not_none(self._y_pred, y_pred)
+        self._sensitive_features = self._concat_if_not_none(self._sensitive_features,
+                                                            sensitive_features)
+        if control_features is not None:
+            self._control_features = self._concat_if_not_none(self._control_features,
```
riedgar-ms, line 115: Consistent length?

riedgar-ms (commented):
Thank you for trying again at this!

fairlearn/metrics/_streaming_metric_frame.py
adrinjalali: that's a good question. Somehow I was thinking of this being sth that I could add data to, and then call `.evaluate()` which either returns nothing or `self`, and it would have all the features like `MetrifFrame`. If we don't want to go down that path, I think `get_metricframe` would be what this does?

hildeweerts:
At a first glance it seems like @Dref360 has processed al requested changes from @riedgar-ms. Is there anything else that's blocking this PR from getting merged?

Tagging @fairlearn/fairlearn-maintainers

fairlearn/metrics/_streaming_metric_frame.py
```diff
@@ -0,0 +1,184 @@
...
+            alongside the ``y_true`` and ``y_pred`` arrays, and passed to the metric function.
+            If there are multiple metric functions (passed as a dictionary), then this is
+            a nested dictionary, with the first set of string keys identifying the
+            metric function name, with the values being the string-to-array-like dictionaries.
+
+            **Note** the types returned by members of the class vary based on whether
+            control features are present.
+        """
+        check_consistent_length(y_true, y_pred)
+        check_consistent_length(y_true, sensitive_features)
```
adrinjalali, line 110: you could merge these calls and pass the three objects

adrinjalali (commented):
Otherwise LGTM, except that I'm not happy that we don't have codecov and I can't check if tests cover the new code. I guess I should set it up.

Issue #944, opened 2021-09-07
DOC add example code for the metrics package


Thomasq99:
The API section of the website does not contain code examples.  Adding code examples would help new users to get acquainted with Fairlearn.  

This is a sub issue for #894.

romanlutz:
Indeed! Would you like to do this @Thomasq99 ?

Thomasq99:
@romanlutz Already on it!  @hildeweerts told me to make a new issue and she was going to assign me to the issue, but she has not gotten around to it yet.

romanlutz:
Fantastic! Let us know if you need any help/questions answered.

hildeweerts:
Thank you for opening this issue @Thomasq99! Feel free to get started with a single class (e.g., `MetricFrame`) for your first PR.

Thomasq99:
I wrote an example for the MetricFrame class utilizing all methods of the class.

![image](https://user-images.githubusercontent.com/38749496/132654884-e2c533f9-5f7b-47f7-981c-54b191fa632f.png)

I was thinking about making an example per separate function and class. For example, the above example for the MetricFrame class, an example for fairlearn.metrics.count, and so on. Is this indeed what you want? 
If not, please let me know!

PR #946, opened 2021-09-09
DOC: example MetricFrame class #944

Thomasq99:
Added an example for the MetricFrame class showcasing all methods.

related to #944

### DIFF
--- a/fairlearn/metrics/_metric_frame.py
+++ b/fairlearn/metrics/_metric_frame.py
@@ -203,9 +203,9 @@ class MetricFrame:
     ...      y_true=y_true,
     ...      y_pred=y_pred,
     ...      sensitive_features=sex)
-    >>> selection_rates.by_group
+    >>> selection_rates.by_group # doctest: +NORMALIZE_WHITESPACE
                         accuracy selection rate
-    sensitive_feature_0                        
+    sensitive_feature_0
     Female                   0.8            0.8
     Male                     0.6            0.4
     >>> selection_rates.difference()
--- a/fairlearn/metrics/_metric_frame.py
+++ b/fairlearn/metrics/_metric_frame.py
@@ -258,11 +258,7 @@ class MetricFrame:
     group_min       0.6            0.4
     group_max       0.8            0.8
 
-    You can also easily plot all of the metrics using DataFrame plotting capabilities
-
-    >>> mf2.by_group.plot.bar(subplots=True, layout= [1,3], figsize=(12, 4),
-    ...                  legend=False, rot=-45, position=1.5); # doctest: +SKIP
-
+    You can find more information about plotting grouped metrics in the following section: :ref:`plot`
     """
 
     # The deprecation decorator does two things:
--- a/fairlearn/metrics/_metric_frame.py
+++ b/fairlearn/metrics/_metric_frame.py
@@ -202,7 +202,7 @@ class MetricFrame:
     >>> selection_rates.by_group
     sensitive_feature_0
     Female    0.8
-    Male      0.6
+    Male      0.4
     Name: selection_rate, dtype: object
     >>> selection_rates.difference()
     0.4
--- a/fairlearn/metrics/_metric_frame.py
+++ b/fairlearn/metrics/_metric_frame.py
@@ -205,7 +205,7 @@ class MetricFrame:
     ...      sensitive_features=sex)
     >>> selection_rates.by_group
                         accuracy selection rate
-    sensitive_feature_0
+    sensitive_feature_0                        
     Female                   0.8            0.8
     Male                     0.6            0.4
     >>> selection_rates.difference()
--- a/fairlearn/metrics/_metric_frame.py
+++ b/fairlearn/metrics/_metric_frame.py
@@ -95,7 +95,7 @@ def _deprecate_metric_frame_init(new_metric_frame_init):
 
 
 class MetricFrame:
-    """Collection of disaggregated metric values.
+    r"""Collection of disaggregated metric values.
 
     This data structure stores and manipulates disaggregated values for any number of underlying
     metrics. At least one sensitive feature must be supplied, which is used
--- a/fairlearn/metrics/_metric_frame.py
+++ b/fairlearn/metrics/_metric_frame.py
@@ -193,25 +193,25 @@ class MetricFrame:
     Examples
     --------
     >>> from fairlearn.metrics import MetricFrame, selection_rate
-    >>> y_true = [0,1,1,1,0,0,1,1]
-    >>> sex = ['Female']*4 + ['Male']*4
+    >>> y_true = [0,1,1,1,1,0,0,1,1,1]
+    >>> sex = ['Female']*5 + ['Male']*5
     >>> selection_rates = MetricFrame(metrics=selection_rate,
     ...                       y_true=y_true,
     ...                       y_pred=y_true,
     ...                       sensitive_features=sex)
     >>> selection_rates.by_group
     sensitive_feature_0
-    Female    0.75
-    Male       0.5
+    Female    0.8
+    Male      0.6
     Name: selection_rate, dtype: object
     >>> selection_rates.difference()
-    0.25
+    0.4
     >>> selection_rates.group_max()
-    0.75
+    0.8
     >>> selection_rates.group_min()
-    0.5
+    0.4
     >>> selection_rates.ratio()
-    0.6666666666666666
+    0.5
     """
 
     # The deprecation decorator does two things:
--- a/fairlearn/metrics/_metric_frame.py
+++ b/fairlearn/metrics/_metric_frame.py
@@ -193,28 +193,25 @@ class MetricFrame:
     Examples
     --------
     >>> from fairlearn.metrics import MetricFrame, selection_rate
-    >>> from fairlearn.datasets import fetch_adult
-    >>> data = fetch_adult(as_frame=True)
-    >>> X = data.data
-    >>> y_true = (data.target == '>50K') * 1
-    >>> sex = X['sex']
+    >>> y_true = [0,1,1,1,0,0,1,1]
+    >>> sex = ['Female']*4 + ['Male']*4
     >>> selection_rates = MetricFrame(metrics=selection_rate,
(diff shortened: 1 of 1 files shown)


sawravchy (approved):
LGTM 👍

fairlearn/metrics/_metric_frame.py
```diff
@@ -189,6 +189,32 @@ class MetricFrame:
 
         .. deprecated:: 0.7.0
             `metric` will be removed in version 0.10.0, use `metrics` instead.
+
+    Examples
+    --------
+    >>> from fairlearn.metrics import MetricFrame, selection_rate
```
romanlutz, line 195: Is this code actually executed? As in, if you change a number will that break the documentation build? If so, that's exactly what I'd want here :-)

Thomasq99: Sadly not, wihtout explicitly writing the output in the docstring there would be no output shown. However, I am not sure if it is possible to have code executed since it is written in the docstring of the class to be shown in the API reference automatically. I looked at the source code of numpy, pandas and scipy and this is how they did it.

fairlearn/metrics/_metric_frame.py
```diff
@@ -189,6 +189,32 @@ class MetricFrame:
...
+    >>> from fairlearn.datasets import fetch_adult
+    >>> data = fetch_adult(as_frame=True)
```
riedgar-ms, line 197: Rather than yet another use of the Adult Census dataset, you could set up a few short arrays like the tests to, with values which are easy to check by inspection.

Thomasq99: I agree that would be more useful. I will change it tomorrow.

fairlearn/metrics/_metric_frame.py
romanlutz: That is perfect! Thanks for checking!

fairlearn/metrics/_metric_frame.py
```diff
@@ -189,6 +189,29 @@ class MetricFrame:
...
+    Male       0.5
+    Name: selection_rate, dtype: object
+    >>> selection_rates.difference()
+    0.25
+    >>> selection_rates.group_max()
+    0.75
+    >>> selection_rates.group_min()
+    0.5
+    >>> selection_rates.ratio()
+    0.6666666666666666
```
riedgar-ms, line 214: If you want to avoid the recurring decimal (and if you look in the other docs, something like `0.666...` should be a neater alternative in case Python ever changes its default output format), then you could extend the array to 10 items, which should allow the ratio to be a finite decimal.

riedgar-ms (approved):
Small suggestion to make the output slightly neater, but I'm happy to merge as-is.

fairlearn/metrics/_metric_frame.py
Thomasq99: Should be fixed now!

fairlearn/metrics/_metric_frame.py
```diff
@@ -189,6 +189,29 @@ class MetricFrame:
...
+
+    Examples
+    --------
+    >>> from fairlearn.metrics import MetricFrame, selection_rate
+    >>> y_true = [0,1,1,1,1,0,0,0,1,1]
+    >>> sex = ['Female']*5 + ['Male']*5
+    >>> selection_rates = MetricFrame(metrics=selection_rate,
+    ...                       y_true=y_true,
+    ...                       y_pred=y_true,
+    ...                       sensitive_features=sex)
```
romanlutz, line 201:
```suggestion
    >>> selection_rates = MetricFrame(
    ...      metrics=selection_rate,
    ...      y_true=y_true,
    ...      y_pred=y_true,
    ...      sensitive_features=sex)
```

At the risk of getting into really nitpicky territory... this indentation looks off. You can adjust it yourself, too, or this would be my suggestion.

romanlutz (approved):
Just one nitpicky comment, otherwise this looks good to merge IMO

fairlearn/metrics/_metric_frame.py
```diff
@@ -189,6 +189,30 @@ class MetricFrame:
 
         .. deprecated:: 0.7.0
             `metric` will be removed in version 0.10.0, use `metrics` instead.
+
+    Examples
+    --------
+    >>> from fairlearn.metrics import MetricFrame, selection_rate
+    >>> y_true = [0,1,1,1,1,0,0,0,1,1]
```
hildeweerts, line 196: I think most of our library focusses on evaluating an ML model rather than the data, wouldn't it make more sense to use `y_pred` here instead of `y_true`?

@romanlutz @Thomasq99

Thomasq99: Should I then use a different `y_true` and `y_pred` for this case? In principle the `y_true` does not matter for the selection rate.  However, it feels odd to set `y_true=y_pred`.

hildeweerts: My preferred solution would be to use `y_true=None` and `y_pred=y_pred`. This will also hopefully prevent wrong usage of `selection_rate` by people who copy paste examples! :)

Thomasq99: It seems to me that this is not possible. Both `y_true` and `y_pred` need to be of the same shape. Therefore I propose the following solution: `y_true = [None]*10` and `y_pred=y_pred`. For now I have implemented it this way.

romanlutz: That is a reasonable concern... Maybe we should use a different metric to avoid confusion? Let's say `accuracy_score`? Then we can provide `y_true` and `y_pred`. Or both `accuracy_score` AND `selection_rate` to show how to pass multiple?

riedgar-ms (approved):
Thanks!

fairlearn/metrics/_metric_frame.py
hildeweerts: Oh I like the suggestion to add both!

Thomasq99: I did that in my new commit but I get the following test fail:
image

I can't figure out what is wrong. I copy the output from my IDE to the example directly and it seems to be exactly the same but still something is wrong. Can someone help me out?

riedgar-ms: In our other docs, we have a `WHITESPACE` directive, which might help.

Thomasq99: I am not sure what that means, is this something I need to download on my IDE?

Thomasq99: I think that is specifically for writing documentation. I am writing code in the docstring so I am not sure whether that will work in code. Let me know if it does though!

Thomasq99:
Currently my problem is the following: 
To write a code example which is added to the API referenceI need to add code to the docstring of the class/function the example is about. In this example you need to write both code and output to get it shown on the API reference.

The output of the `sensitive_features.by_group`  a functionality  I want to showcase is a DataFrame. When this is printed, indentation (including "trailing" white spaces) is used to get the layout of the dataframe nice. I need to add this output exactly to the docstring. However, when I add the indentation required to have it have the same format as the printed dataframe I get an error that I have trailing white spaces when running the tests. Subsequently I get an error that the indentation is not correct (I think the "trailing" white spaces are automatically removed).

When I remove the "trailing" white spaces myself the indentation is not correct since it does not adhere to the exact output of the printed dataframe. Therefore, it throws an error.

Both when I add correct indentation and when I remove this indentation I get an error. It seems to me there is no way to add this code example without the tests throwing an error. Does someone have a suggestion to what to do? I am quite lost. 

Erro when correct indentation is removed for reference:

image

image

fairlearn/metrics/_metric_frame.py
```diff
@@ -189,6 +189,41 @@ class MetricFrame:
 
         .. deprecated:: 0.7.0
             `metric` will be removed in version 0.10.0, use `metrics` instead.
+
+    Examples
```
MiroDudik, line 193: I think that it is really important to have good docstring examples for `MetricFrame`. What would you think about essentially just copy-pasting the section "Fairness assessment with MetricFrame" from `https://github.com/fairlearn/talks/blob/main/2021_scipy_tutorial/fairness-in-AI-systems-instructors.ipynb` with a small inline-constructed dataset? This would be longer than our usual docstrings, but not too long by pandas standards.

fairlearn/metrics/_metric_frame.py
Thomasq99: I misunderstood, but with some help I got it, seems to be working now!

fairlearn/metrics/_metric_frame.py
Thomasq99: I added the code from the tutorial to the examples. Is this okay?

hildeweerts:
@Thomasq99 could you add a screenshot of what the example looks like locally?

@MiroDudik could you check whether Thomas's latest changes reflect what you wanted?

Thomasq99:
image
image

It looks like this in the webpage

romanlutz:
@Thomasq99 I really like this! The plotting part at the end is not showing because that kind of code can't render plots (I think), but I'm also tempted to just link to the user guide section that explains plotting in detail. After all, this isn't functionality we provide but rather pandas.DataFrame functionality that's convenient.

Is there a macro for linking elsewhere? If not, a simple textual note with a link will do.

@MiroDudik I'm happy with this as is. Is this missing anything compared to what you were thinking of?

Thomasq99:
For now I have changed it to the following:
image
Is this what you mean? @romanlutz

romanlutz:
> For now I have changed it to the following:
> 
> image
> 
> Is this what you mean? @romanlutz 

Yes, except that it shouldn't be an absolute link, but rather the link directive using

```
:ref:`plot`
```

Thomasq99:
> > For now I have changed it to the following:
> > image
> > Is this what you mean? @romanlutz
> 
> Yes, except that it shouldn't be an absolute link, but rather the link directive using
> 
> ```
> :ref:`plot`
> ```

It looks like this now:
image

romanlutz (approved):

PR #967, opened 2021-09-30
ENH make CorrelationRemover expose n_features_in_ and feature_names_in_

### LINKED ISSUE #964: TST: CorrelationRemover is failing estimator checks
#### Describe the bug

`CorrelationRemover` is failing in our unit tests to check for estimator API compatibility: 
```
TypeError: fit() got an unexpected keyword argument 'sample_weight'
```

In a recent scikit-learn commit the following change occurred: https://github.com/scikit-learn/scikit-learn/commit/cbfbd863804da1005eda2fe8ee34ebe6380f9b3a# 
The important part there is https://github.com/scikit-learn/scikit-learn/commit/cbfbd863804da1005eda2fe8ee34ebe6380f9b3a#diff-49f814863816a0e8d199ba647472450ecc59a4375171db23cb1ca36063d6dc25R876-R888 that is, it doesn't check whether `sample_weight` is an argument on `fit` anymore. This seems to be the root cause.

This is failing tests so we are blocked from merging PRs until it's fixed.

#### Steps/Code to Reproduce
`python -m pytest test/unit/preprocessing`

#### Versions
```
System:
    python: 3.8.10 (default, Jun  2 2021, 10:49:15)  [GCC 9.4.0]
executable: /usr/bin/python
   machine: Linux-5.11.0-34-generic-x86_64-with-glibc2.29

Python dependencies:
    Cython: None
matplotlib: 3.4.3
     numpy: 1.21.2
    pandas: 1.3.3
       pip: 20.0.2
     scipy: 1.7.1
setuptools: 45.2.0
   sklearn: 1.0
    tempeh: None

```

@adrinjalali do you have a good intuition on what the fix here should be? Since that technique doesn't yet support `sample_weight` (and we have no plans of adding that anytime soon) it seems wrong to add the argument just to satisfy this check. Wdyt?
FYI @koaning who added `CorrelationRemover`

adrinjalali:
fixes #964

Makes `CorrelationRemover` pass sklearn common tests, and include 2 tests which are going to fail as `xfail`. In the process, the estimator now exposes `n_features_in_` and `feature_names_in_` via calling `_validate_data`.

It's also formatting the two files with `black`.

### DIFF
--- a/fairlearn/preprocessing/_correlation_remover.py
+++ b/fairlearn/preprocessing/_correlation_remover.py
@@ -71,13 +71,16 @@ class CorrelationRemover(BaseEstimator, TransformerMixin):
         if isinstance(X, pd.DataFrame):
             self.lookup_ = {c: i for i, c in enumerate(X.columns)}
             return X.values
+        # correctly handle a 1d input
+        if len(X.shape) == 1:
+            return {0: 0}
         self.lookup_ = {i: i for i in range(X.shape[1])}
         return X
 
     def fit(self, X, y=None):
         """Learn the projection required to make the dataset uncorrelated with sensitive columns."""  # noqa: E501
         self._create_lookup(X)
-        X = check_array(X, estimator=self)
+        X = self._validate_data(X)
         X_use, X_sensitive = self._split_X(X)
         self.sensitive_mean_ = X_sensitive.mean()
         X_s_center = X_sensitive - self.sensitive_mean_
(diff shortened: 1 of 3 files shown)


fairlearn/preprocessing/_correlation_remover.py
```diff
@@ -100,3 +105,17 @@ def transform(self, X):
         X_use = np.atleast_2d(X_use)
         X_filtered = np.atleast_2d(X_filtered)
         return self.alpha * X_filtered + (1 - self.alpha) * X_use
+
+    def _more_tags(self):
+        return {
+            "_xfail_checks": {
```
riedgar-ms, line 111: This looks like test things in the product code? I'm sure this is the right thing to do, but can you explain what's going on? Is `_more_tags()` a magic name for something in pytest?

adrinjalali: no it has nothing to do with pytest. Estimator tags are things which can be set on an estimator _instance_, which can define its behavior in certain ways, including when the test suite runs tests on it.

A bit more about tags here: https://scikit-learn.org/stable/developers/develop.html?highlight=estimator%20tags#estimator-tags

and this particular tag is used here in the test suite: https://github.com/scikit-learn/scikit-learn/blob/844b4be24d20fc42cc13b957374c718956a0db39/sklearn/utils/estimator_checks.py#L502

We are working on making the interface for these tags a bit better than what it is.

riedgar-ms: Thanks for the explanation.

riedgar-ms (approved):

romanlutz (approved):

Issue #785, opened 2021-05-07
ENH Add "adversarial debiasing" to mitigations


romanlutz:
The paper [Mitigating Unwanted Biases with Adversarial Learning](https://arxiv.org/abs/1801.07593) by Zhang, Lemoine, and Mitchell presents an "adversarial debiasing" technique. The goal of this task is to implement it.

AIF360 has an [implementation](https://aif360.readthedocs.io/en/v0.2.3/modules/inprocessing.html#aif360.algorithms.inprocessing.AdversarialDebiasing) of this technique, but it uses `tensorflow`. Fairlearn shouldn't depend on `tensorflow` (or `pytorch`, for that matter), but instead have a soft dependency on the package of choice. @adrinjalali has previously noted that Fairlearn should prefer `pytorch` over `tensorflow`, which we should take into account for this.

[Note: The name "adversarial debiasing" is in conflict with [our policy](https://fairlearn.org/main/contributor_guide/how_to_talk_about_fairness.html) around the words "bias"/"debiasing". What else could we call it? Adversarial fairness? @fairlearn/fairlearn-maintainers ]

Completing this item requires:
- code for the technique in `fairlearn.adversarial` (@fairlearn/fairlearn-maintainers  is that a suitable module name? Other suggestions?)
- unit tests in `test.unit.adversarial`
- descriptive API reference (directly in the docstring)
- ideally a short user guide in docs.user_guide.mitigation.rst

A fully fledged example notebook is not required.

To claim this task please respond below. Of course, you can also ask questions!

adrinjalali:
I'd be a bit concerned with our policy of not using the word "bias" or "debias" completely. These words are used in the literature and folks use them to find resources and implementations. I completely agree that we can't completely debias a dataset, or a model, but we do remote some sort of a bias when we use these methods. We can still use these words, but be very clear about what kind of bias it is that the method is removing, instead of saying this method debiases the model/data.

romanlutz:
Exactly, but "Adversarial Debiasing" isn't a clear name since it misses exactly that point. In fact, it might give the impression that it's a fix for all biases.

Thanks @riedgar-ms for the title fix!

hildeweerts:
Within the paper itself they even acknowledge that debiasing might be problematic:*"“debias” may not be quite the right word, as all bias is not necessarily removed"*. 

[...]

PR #973, opened 2021-10-06
ENH Add "adversarial debiasing"

SeanMcCarren:
Solves issue #785 and implements [Adversarial Debiasing paper](https://dl.acm.org/doi/pdf/10.1145/3278721.3278779)

Updated 24 March.

# UPDATE
Should have responded to all feedback now

# Old
- resolved issues mentioned in #1009 
- Every experiment from the original paper ([Adversarial Debiasing paper](https://dl.acm.org/doi/pdf/10.1145/3278721.3278779)) is reproduced on [this repo](https://github.com/SeanMcCarren/adversarial-debiasing)! Not exact replicas, but close. See repo for discussion of results as well.
### To summarize important design choices:
- Follow scikit-learn guidelines to create this estimator
  * `fit(X, y, sensitive_feature)`, `predict(X)`, `decision_function(X)`. 
  * we handle preprocessing of `y` and `sensitive_features`, but the user needs to preprocess `X`. (the models are NN's, so we require numeric inputs everywhere)
- Base class is general and mostly handles API, `BackendEngine` provides PyTorch/TensorFlow-specific code.
- Allow many kwargs, in order to serve many use cases. Especially note:
  * `y` and `sensitive_features` can be from arbitrary distributions. We try to infer the distribution of this data and use this to choose appropriate preprocessors (`y_transform` and `a_transform`), loss functions (`predictor_loss` and `adversary_loss`), and the decision function (`predictor_function`). Currently, we only infer whether `y` or `sensitive_features` is univariate binomial, univariate multinomial, or multivariate normal. If we can infer such a distribution, we know precisely what to choose as aforementioned kwargs. Otherwise, the user must supply these kwargs explicitly.
  * `predictor_optimizer` and `adversary_optimizer` are kwargs, because in practice we see many different optimizers used.
  * `callbacks`. I found supporting callback functions is particularly useful (as is done in skorch for instance)

### DIFF
--- a/fairlearn/adversarial/_preprocessor.py
+++ b/fairlearn/adversarial/_preprocessor.py
@@ -6,6 +6,7 @@ from ._constants import (
     _TYPE_CHECK_ERROR,
     _ARG_ERROR_MESSAGE,
     _TYPE_UNKNOWN_ERROR,
+    _INVALID_OHE,
 )
 from sklearn.utils import check_array
 from sklearn.preprocessing import OneHotEncoder
@@ -14,7 +15,7 @@ from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn.utils.multiclass import type_of_target
 from numpy import all as np_all
 from numpy import sum as np_sum
-from numpy import unique
+from numpy import unique, isin
 
 # FIXME: memoize type_of_target. It is quite expensive and called repeatedly.
 
@@ -223,6 +224,9 @@ def _get_type(data, assumption):
         # or not? because it could be multiple categorical... who knows!
         if not np_all(np_sum(data, axis=1) == 1):
             raise ValueError(_TYPE_CHECK_ERROR.format("category"))
+        u = unique(data)
+        if len(u) != 2 or (not isin([0,1], u).all()):
+            raise ValueError(_INVALID_OHE)
 
     # Match inferred with dist_assumption
     if inferred == "binary" and assumption in [
--- a/fairlearn/adversarial/_constants.py
+++ b/fairlearn/adversarial/_constants.py
@@ -12,6 +12,7 @@ _INVALID_DATA = (
     "Data of '{}' can not be interpreted. Only continuous data or array "
     + "with only 0s and 1s."
 )
+_INVALID_OHE = "One-hot-encoded matrix does not contain precisely 0 and 1"
 _PREDICTION_FUNCTION_AMBIGUOUS = (
     "Can not interpret prediction_function. Please provide a callable as "
     + "key-word argument that maps soft-probabilities "


hildeweerts (commented):
I have embarrassingly little experience with pytorch/tensofrlow so I've mostly added a few nitpicks re. naming and such.

romanlutz:
This is perhaps naive for reasons I haven't quite thought through yet, but how about reading the targets and deciding if it's binary, multiclass, or regression based on that?

ExponentiatedGradient does this (without multiclass) AFAIK without dedicated input variable.

hildeweerts:
For binary/multiclass there's a whole bunch of [utils in scikit-learn](https://github.com/scikit-learn/scikit-learn/blob/844b4be24d20fc42cc13b957374c718956a0db39/sklearn/utils/multiclass.py) that may be of help.

For regression a separate `AdversarialMitigationRegressor` class seems more intuitive to me (this is also a pattern that is common in scikit-learn, e.g., `RandomForestClassifier` versus `RandomForestRegressor`).

I bet @adrinjalali has thoughts on this as well.

hildeweerts:
> Or is there a lot of use for also supporting multiclass features, and letting the users map various groups of columns of sensitive_features?

Tutorials like to pretend that everything is binary, but in practice there's hardly any sensitive feature that can truly be considered binary. So my first reaction would be to do things the other way around: assume *none* of the features are one-hot encoded and do one-hot encoding internally for multicategorical features (if necessary?)

To distinguish categorical / continuous features I could imagine an argument `infer_type` (or whatever we want to call it) that's either `'auto'` (automatically infer type of sensitive features) or a dict `{ 'colname1' : 'continuous', 'colname2' : 'categorical' }`. 

The independence assumption should be described clearly in the documentation btw, because sensitive features may be statistically related even if they are not one-hot-encoded.

SeanMcCarren:
@adrinjalali Hi :) I have some questions regarding testing.
1) CI build doctest failing because in `examples/plot_adversarial_mitigation.py` there is no pytorch or tensorflow. How do we ensure that for that notebook there is torch or tensorflow installed? We also discussed this in the last comment, but I do not know how to solve this in the way that we'd like.

2) In `test_othermlpackages/test_*.py` you write `importorskip`, but I wonder why? Shouldn't the test fail if we can not import? Just curious, as I added some test cases here myself too and I want them to run.

adrinjalali:
> 1) CI build doctest failing because in `examples/plot_adversarial_mitigation.py` there is no pytorch or tensorflow. How do we ensure that for that notebook there is torch or tensorflow installed? We also discussed this in the last comment, but I do not know how to solve this in the way that we'd like.

If we want those examples to be rendered in our documentation, then we do need to install them on the CI where they're rendered. I'd rather do only pytorch though, at least for a rendered version of the docs. Alternatively, we can just not have a notebook, and only have an `.rst` file explaining things and showing example code, which is not run, so we won't need the dependency.

> 2) In `test_othermlpackages/test_*.py` you write `importorskip`, but I wonder why? Shouldn't the test fail if we can not import? Just curious, as I added some test cases here myself too and I want them to run.

The tests which have dependencies which are not always installed, should pass if the dependency is not installed. We should make sure there is at least one CI where those dependencies are installed so that those tests are run at least once.

> 3) Lastly, a question regarding documentation, you said earlier that you'd like an extensive user guide. What is the precise location of this? In `user_guide/mitigation.rst`? I ask because there are some empty sections here still for Exponentiated Gradient and Grid search.

It would be a standalone `.rst` file under `docs/user_guide`. We're in the process of adding more material there, and yours would be one of them.

SeanMcCarren:
> 3.

Alright! I'll put a guide in the User guide.

> 1. If we want those examples to be rendered in our documentation, then we do need to install them on the CI where they're rendered. I'd rather do only pytorch though, at least for a rendered version of the docs. Alternatively, we can just not have a notebook, and only have an `.rst` file explaining things and showing example code, which is not run, so we won't need the dependency.

So, I currently have a notebook set up and it would be cool if it was rendered, but I'm not sure how to install pytorch properly on a CI then. But yes the altnerative is also fine, I can just translate the notebook to a rst in that case

examples/plot_adversarial_mitigation.py
```diff
@@ -0,0 +1,366 @@
...
+X_train, X_test, Y_train, Y_test, Z_train, Z_test = \
+        train_test_split(X, y, sensitive_feature,
+                         test_size=0.2,
+                         random_state=12345,
+                         stratify=y)
+
+# %%
+# Now, we can use :class:`fairlearn.adversarial.AdversarialFairnessClassifier` to train on the
+# UCI Adult dataset. As our predictor and adversary models, we use for
+# simplicity the fairlearn built-in constructors for fully connected neural
```
hildeweerts, line 100:
```suggestion
# simplicity the default constructors for fully connected neural
```

examples/plot_adversarial_mitigation.py
```diff
@@ -0,0 +1,366 @@
...
+# networks with sigmoid activations. We initialize neural network constructors
```
hildeweerts, line 101:
```suggestion
# networks with sigmoid activations implemented in Fairlearn. We initialize neural network constructors
```

examples/plot_adversarial_mitigation.py
```diff
@@ -0,0 +1,366 @@
...
+# Intuitively, you should imagine that there are "very easy" local minima that the
+# models may converge to. For instance, if the predictor always outputs class=0,
+# then the adversary's objective is much easier, namely it only has to output the
+# most correlated sensitive feature to class=0. Such minima are very easily reached
+# unfortunately. #TODO this description is very bad at the moment, working on
+# understanding it:).
+#
+# So, to finetune our model and the training thereof, we start by defining our
+# neural networks in the way we'd like them and in the way we can easily tweak them.
+# I'll be using PyTorch, but the same can be achieved using Tensorflow!
```
hildeweerts, line 188:
```suggestion
# We will be using PyTorch, but the same can be achieved using Tensorflow!
```

examples/plot_adversarial_mitigation.py
```diff
@@ -0,0 +1,366 @@
...
+    progress_updates=None,
+    callback_fn=callback_fn)
+
+
+# %%
+# Then, we more carefully improve this by changing the model. That is, we do not
+# change the predictor/adversary model parameters, but we do change the learning
+# parameters. In particular, we set very small learning rates and alpha a bit higher.
+# As mentioned earlier, training adversarially is tricky, and ideally
+# you'd train this in a controlled way.
```
hildeweerts, line 330: With "controlled" do you mean systematically go over different settings or something else?

examples/plot_adversarial_mitigation.py
```diff
@@ -0,0 +1,366 @@
...
+    sensitive_features=Z,
+    epochs=epochs,
+    batch_size=batch_size,
+    shuffle=True,
+    progress_updates=None,
+    callback_fn=callback_fn)
+# %%
+# We take a look at the results. Notice we achieve a much lower demographic parity
+# difference than in Exercise 1! This does come at the cost of some accuracy, but
+# such a tradeof is to be expected.
```
hildeweerts, line 354: We could add a sentence on why it is expected.

hildeweerts (commented):
A few notes on the example notebook - it's in a really good shape already!

examples/plot_adversarial_mitigation.py
SeanMcCarren: ah yea i could be more clear, I meant "in a slow and careful way" because training can quickly diverge and you want to avoid this.

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,979 @@
...
+    """The interface of a mixin class."""
+
+    def _evaluate(self, X: ndarray) -> ndarray:
+        """
+        Evaluate the model given input `X`.
+
+        Feed 2d `numpy.ndarray` through model and receive output as
+        2d `numpy.ndarray`.
+        """
+        pass
```
adrinjalali, line 695: according to many, and I agree with them, we shouldn't have any method which has only `pass` as its body. @riedgar-ms may disagree.

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,979 @@
...
+            self.cuda = False
+        elif (cuda):
+            if (not self.torch):
+                raise ValueError("Cuda is PyTorch-only")
+            if not torch.cuda.is_available():
+                raise ValueError("Cuda is not available")
+            self.cuda = True
+            self.device = torch.device("cuda:0")
+
+    def _extend_instance(self, cls_):
```
adrinjalali, line 663: instead of this, we can have a `BackendEngine` kinda thing, which implements the methods, and we instantiate one in `fit`, and use its methods, instead of adding them here to `self`. I think it'd be much cleaner.

test/install/test_no_ml.py
```diff
@@ -0,0 +1,20 @@
...
+
+import pytest
+
+from fairlearn.adversarial import AdversarialFairness
+from fairlearn.adversarial._constants import _IMPORT_ERROR_MESSAGE
+
+
+@pytest.mark.parametrize("library", [None, 'torch', 'tensorflow'])
+def test_no_library(library):
+    expected_library = "torch or tensorflow" if library is None else library
```
adrinjalali, line 12: in a test, you can use `importorskip` from pytest.

test/unit/adversarial/__init__.py
```diff
@@ -0,0 +1,2 @@
+# Copyright (c) Fairlearn contributors.
+# Licensed under the MIT License.
```
adrinjalali, line 2: why do we need this file?

test_othermlpackages/package_test_common.py
```diff
@@ -119,3 +121,59 @@ def run_thresholdoptimizer_classification(estimator):
                                                       mitigated_predictions,
                                                       sensitive_features=A_test)
     assert dp_diff_mitigated <= dp_diff_unmitigated
+
+
+def run_AdversarialFairness_classification(estimator):
```
adrinjalali, line 126: I'm happy if we put all adverserial related tests in their own dedicated files.

setup.cfg
```diff
@@ -35,5 +35,6 @@ per-file-ignores =
     scripts/*.py:D100,D101,D102,D103,D104,D107,D205,D400,D401,RST201,RST205,RST301
     test/*.py:D104
     test/*/*.py:D100,D101,D102,D103,D104,D105,D107
+    test/unit/adversarial/*.py:A003,D100,D101,D102,D103,D104,D105,D106,D107
```
adrinjalali, line 38: why? :D

SeanMcCarren:
> Thanks a bunch for the work @SeanMcCarren. This is quite nice work! Enjoyed going through it.

Thanks so much for the helpful feedback!!!!! :)

> I didn't spend too much time on the bits very specific to pytorch and tf, but I'd be curious if we could use a scikeras or a skorch model instead.

I like this suggestion, but I'm not sure if this will make the API any nicer. Will let you know!

docs/user_guide/mitigation.rst
```diff
@@ -649,6 +658,203 @@ Exponentiated Gradient
...
+~~~~~~
+
+One can implement the predictor and adversarial neural networks as
+a `torch.nn.Module` (using PyTorch) or as a `tensorflow.keras.Model` (using TensorFlow).
+This implementation has a soft dependency on either PyTorch or TensorFlow, and the user
+needs to have installed either one of the two soft dependencies. It is not possible to
+mix these dependencies, so a PyTorch predictor with TensorFlow loss functions are not
+possible.
+
+It is very important to define the neural network models with no activation function
```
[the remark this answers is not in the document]
SeanMcCarren: ah yea thats the wording I was looking for, perfect

docs/user_guide/mitigation.rst
```diff
@@ -649,6 +658,203 @@ Exponentiated Gradient
...
+    >>> from sklearn.compose import make_column_transformer, make_column_selector
+    >>> from sklearn.preprocessing import OneHotEncoder, StandardScaler
+    >>>
+    >>> ct = make_column_transformer(
+    ...     (StandardScaler(),
+    ...      make_column_selector(dtype_include='number')),
+    ...     (OneHotEncoder(drop='if_binary', sparse=False),
+    ...      make_column_selector(dtype_include="category")))
+    >>>
+    >>> X_transformed = ct.fit_transform(X)
```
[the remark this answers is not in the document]
SeanMcCarren: What do you mean by leaking data? That we are losing the original mapping and not automatically using the inverse transform?

Maybe a related question which I can't easily find an answer to online, if we add this model to the pipeline, how can we pass `sensitive_features`? Is the sklearn Pipeline able to transform this data too and pass it to our model?

docs/user_guide/mitigation.rst
```diff
@@ -649,6 +658,203 @@ Exponentiated Gradient
...
+        adversary_model=[6, 6],
+        predictor_loss=torch.nn.BCEWithLogitsLoss(),
+        adversary_loss=torch.nn.MSELoss(),
+        predictor_function=lambda pred: (pred >= 0.).astype(float)
+    )
+
+*Note that the PyTorch and TensorFlow modules switch the order of arguments in their loss functions, so be extra careful when defining custom loss functions so that you adhere to the module-specific argument order*
+
+We do, however, offer some shortcuts for binary, categorical, and continuous data. Namely, instead of passing a function, you can pass any of the keywords defined in `fairlearn.adversarial.Keyword` to signal a certain distribution type and use a certain PyTorch or TensorFlow specific loss function. For instance, the above example is equivalent to passing the keywords :code:`predictor_loss='binary'`, :code:`adversary_loss='continuous'` and :code:`predictor_function='binary'`. The default argument for any of these three keywords is :code:`'auto'`, which automatically infers the distribution type. *Note: distribution types are validated on the first batch of data that the model sees, so a* :code:`'binary'` *assumption on non-binary data will raise an error.*
```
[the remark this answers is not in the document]
SeanMcCarren: Maybe not? Generally I don't like magic values like 'binary' and such in code, but I'm open to suggestions or changes

examples/plot_adversarial_mitigation.py
```diff
@@ -0,0 +1,366 @@
...
+from fairlearn.adversarial import AdversarialFairnessClassifier, \
+    AdversarialFairness
+from pandas import Series
+from numpy import number, random, mean
+from sklearn.compose import make_column_transformer, make_column_selector
+from sklearn.preprocessing import OneHotEncoder, StandardScaler
+from sklearn.datasets import fetch_openml
+from sklearn.model_selection import train_test_split
+import torch
+random.seed(123)
```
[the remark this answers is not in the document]
SeanMcCarren: Sure, but I wonder why? Is it because this rng may be used in more places at the same time?

examples/plot_adversarial_mitigation.py
```diff
@@ -0,0 +1,366 @@
...
+torch.manual_seed(123)
+
+X, y = fetch_openml(data_id=1590, as_frame=True, return_X_y=True)
+
+# Remove rows with NaNs. In many cases, dropping rows that contain missing values is not the best approach for dealing with missing values,
+# but for this example we ignore that.
+non_NaN_rows = ~X.isna().any(axis=1)
```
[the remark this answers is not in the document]
SeanMcCarren: Should we? I was not sure, because imputing binary with most_frequent is also not very clean

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,979 @@
...
+                                            Z[batch_slice])
+                predictor_losses.append(LP)
+                adversary_losses.append(LA)
+            if shuffle and epoch != epochs - 1:
+                X, Y, Z = self._shuffle(X, Y, Z)
+            if callable(callback_fn):
+                callback_fn()
+
+    def _shuffle(self, X, Y, Z):
+        """Shuffle the rows of X, Y, Z."""
```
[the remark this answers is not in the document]
SeanMcCarren: PyTorch mixin overwrites this functionality because sklearn doesn't handle PyTorch tensors I think

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,979 @@
...
+        Y_pred = self._evaluate(X)
+
+        assert Y_pred.ndim == 2
+
+        Y_pred = self.predictor_function(Y_pred)
+
+        return Y_pred
+
+    def _check_type(self, Y, choice, data_name):
+        """Identify user query :code:`choice`."""
```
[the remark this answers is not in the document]
SeanMcCarren: @adrinjalali Oh `check_classification_targets` may be perfect! Thanks! Earlier, you said that the `Keyword` class was not necessary, and I think the possible responses from `check_classification_targets` is a nice replacement, but do you know of a util in sklearn where the possible classes like `binary` and `multiclass-multioutput` are stored? I can imagine sklearn may change these names at some point, and then our code will break too...

setup.cfg
SeanMcCarren: Hihi yeah.... Lazy? In my defense most of these also appear in the other modules haha :)

test/install/test_no_ml.py
SeanMcCarren: yeah no but this is to test the error message if none are installed. Quite silly though.

SeanMcCarren:
@adrinjalali Thanks again for the great feedback! It was really helpful, and with that in mind I took a good look again at the API and at https://scikit-learn.org/stable/developers/develop.html and it raised some more questions that I can't really answer. 
- How can we use a pipeline if we also need to pass in `sensitive_features`?
- So I know about how fit should work with only being persistent iff `warm_start=True`, but what about `partial_fit`

docs/user_guide/mitigation.rst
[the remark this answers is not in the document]
SeanMcCarren: So to clarify, softmax is okay, but discrete predictions are not okay

docs/user_guide/mitigation.rst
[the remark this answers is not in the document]
hildeweerts: I think what @adrinjalali refers to here is that you need to make sure that all pre-processing steps are based *solely* on the training data. If you fit `StandardScaler` on all data the mean and variance are computed based on both training and test data, meaning that you ["leak" information](https://en.wikipedia.org/wiki/Leakage_(machine_learning)) about the test data into the pre-processed training data.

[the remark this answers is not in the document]
SeanMcCarren: That makes sense, thanks!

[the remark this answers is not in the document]
SeanMcCarren: But still the question of how you want to use the pipeline remains, I can't find a sklearn-style way to also pass sensitive features, except possibly indicating that the sensitive features are some columns of X, which is done in ThresholdOptimizer I think. That is not really possible if you preprocess such that the column layout changes (such as when doing a one hot encoding).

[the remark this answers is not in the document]
adrinjalali: you can pass things to the estimators' `fit` through pipeline's `fit` as: `pipeline.fit(X, y, estimator_name__param_name=value)`

docs/user_guide/mitigation.rst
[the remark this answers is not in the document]
adrinjalali: but thos _magic values_ are the way things are done in scikit-learn. As an example, you can check [here](https://github.com/scikit-learn/scikit-learn/blob/2330c4c4652074609b614c193ac90d131a68dac5/sklearn/linear_model/_logistic.py#L469)

examples/plot_adversarial_mitigation.py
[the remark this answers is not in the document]
adrinjalali: It's just a better practice to use a local rng rather than the global one. Since setting the global seed can also change the generated values in other unrelated files if they also use the global rng.

examples/plot_adversarial_mitigation.py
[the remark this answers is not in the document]
adrinjalali: If you're not happy with a simple strategy available on `SimpleImputer`, you can also stack a `MissingIndicator` to them to have the information that the info here was missing, or by using `add_indicator=True` to `SimpleImputer`, or you could have a much fancier way of imputing them by `IterativeImputer` if you think that makes more sense.

fairlearn/adversarial/_adversarial_mitigation.py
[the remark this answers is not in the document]
adrinjalali: But I don't think having those mixins the way they are now makes sense anyway.

fairlearn/adversarial/_adversarial_mitigation.py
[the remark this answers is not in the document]
adrinjalali: No, we don't store them in a union or a class or anything, they're just constant values, and if we rename such a constant, it'll go through a deprecation period.

setup.cfg
adrinjalali: Just run `black` on your code.

adrinjalali:
You can pass extra arrays to pipeline's steps by `pipeline.fit(X, y, step_name__param_name=value)`, but it's true that transforming a third array is not supported in sklearn. For those cases, you may want to write a meta-estimator which accepts an array with the sensitive attributes attached to it, internally decouple them, do your transformation, and return them again together.

SeanMcCarren:
> For those cases, you may want to write a meta-estimator which accepts an array with the sensitive attributes attached to it, internally decouple them, do your transformation, and return them again together.
@adrinjalali 
1) Do you think this may be a separate issue? I think other fair estimators may benefit from such a pipeline as well
2) I am not 100% certain how you envision this meta-estimator: is a pipeline the estimator we pass, or AdversarialFairness?
3) More concretely, is this a solution to your description? a `ParallelPipeline` which takes a pipeline, and `fit(*args, **kwargs)` transforms every arg or kwarg using a clone of pipeline (without the final estimator), and combines every arg, kwarg to pass to the estimator at once. So you could call ParallelPipeline.fit(A, something=B) which clones the pipeline for A and B, fits them separately, and finally calls estimator.fit(A, something=B) again.

adrinjalali:
I guess before we get into that discussion, the question is whether we need this feature in the first place. We don't need a pipeline where the sensitive attributes are augmented in this PR, do we?

SeanMcCarren:
I'm afraid we do, but not 100% certain because we could also change the loss. But in essence, we need to compute the loss between the sensitive features and the adversary's predictions.

adrinjalali:
But we do that inside the estimator, it's not like the sensitive feature needs to be transformed before it reaches our estimator, and since this is a predictor, we don't need to send the sensitive features to the next step in the pipeline.

adrinjalali:
I may be mistaken though, could you point me to the place I could check how we deal with it?

SeanMcCarren:
Oh sorry maybe I misunderstood you. You are right, you don't need it transformed before the estimator. 

But, you do need to also transform the sensitive feature at some point in time. The adversarial neural network needs to predict the (distribution of) the sensitive feature, and I don't see how we can compute the loss between this prediction and the actual sensitive feature(s) if it is not preprocessed beforehand. 

You are right in that we *can* do the preprocessing in the predictor, we could just apply some standard transforms like binary to 0/1s, category to one-hot-encoding, standardize numeric columns. 

But then I have some concerns
1) If we preprocess the sensitive features and the sensitive are multiple features, then the adversarial will predict multiple outputs, and it is weird to define a loss function for this. Ideally, if the user wants to use this model with multiple sensitive features, I'd let it up to the user to define a weird loss function to go with it. But how do we tell the user *beforehand* which sensitive features map to which dimensions in the adversary's prediction vector? For instance, with one-hot encoding, there is no longer a one-to-one mapping between sensitive_features and preprocessed sensitive_features.
2) We discussed earlier in this PR (@hildeweerts ) about preprocessing for X and y too, so that the user doesn't have to preprocess anything. But this may be weird?
3) It seems nice to be able to preprocess X and sensitive_features in the same way because sensitive_features could be a subset of X.
4) If we want to add a partial_fit method that also needs to accept every possible class of sensitive_features (particularly nasty if there are multiple sensitive features?)

I'm curious to hear what you think about this, you did kind of open my eyes to incorporating a preprocessor for the sensitive_features in the estimator directly, but I see no way of implementing this yet.

test/unit/adversarial/__init__.py
SeanMcCarren: It is also on all other subdirs of /test/unit, but i don't know

fairlearn/adversarial/__init__.py
```diff
@@ -0,0 +1,8 @@
+# Copyright (c) Microsoft Corporation and Fairlearn contributors.
+# Licensed under the MIT License.
+
+"""Adversarial techniques to help mitigate fairness disparities."""
```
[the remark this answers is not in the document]
SeanMcCarren: I don't like it either, but flake8 is telling me to

test_othermlpackages/package_test_common.py
SeanMcCarren: It is currently very nice how test_pytorch and test_tensorflow are called from CI, would you like me to move it to somewhere that is triggered from test_pytorch and test_tensorflow, or should the CI be changed?

SeanMcCarren:
@hildeweerts @adrinjalali @romanlutz 

For those interested, this PR is up for review again :)

romanlutz:
I don't have the time to review this in detail right now, but I do want to point out that you may want to consider splitting parts of this PR into separate PRs. This can significantly speed up the overall work as I've experienced many times. For more on this phenomenon see [my tweet](https://twitter.com/romanlutz13/status/1451289545627086849?s=20) 
#LutzsLaw

SeanMcCarren:
> I don't have the time to review this in detail right now, but I do want to point out that you may want to consider splitting parts of this PR into separate PRs. This can significantly speed up the overall work as I've experienced many times. For more on this phenomenon see [my tweet](https://twitter.com/romanlutz13/status/1451289545627086849?s=20) #LutzsLaw

Haha I like this yes, and want to say that I have thought about this before, but find it difficult to find meaningful "splits" of this PR. One split could be pytorch/tensorflow separately, but the module-specific code is only a tiny part so I don't know... Another split could be code/user guide, but I don't know, the user guide is quite small still.

docs/user_guide/mitigation.rst
```diff
@@ -45,6 +45,15 @@ Fairlearn contains the following algorithms for mitigating unfairness:
       - ✔
       - ✔
       - ✘
+   *  - :class:`~fairlearn.adversarial.AdversarialFairness`
+      - Inprocessing algorithm based on the paper *Mitigating Unwanted Biases*
+        *with Adversarial Learning* [#4]_. This algorithm uses a predictor and an adversarial neural network, defined either as a `PyTorch module
```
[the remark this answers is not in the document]
romanlutz: I don't think you can actually wrap lines while writing links, but maybe I'm wrong... Just reopening to confirm.

fairlearn/adversarial/toying_around.py
```diff
@@ -0,0 +1,174 @@
...
+#         LP = torch.nn.BCELoss()(Y_hat, y[batch_slice])
+#         LP.backward()  # Check what this does at some point in time
+
+#         predictor_optimizer.step()
+
+
+#     if True and epoch != epochs - 1:
+#         x, y, z = model.backendEngine_.shuffle(x, y, z)
+
+# evaluate(model.predict(X))
```
romanlutz, line 170: Is this meant to be removed?

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,748 @@
...
+from sklearn.utils.validation import (
+    check_is_fitted,
+    check_random_state,
+    check_array,
+)
+from sklearn.exceptions import NotFittedError
+from numpy import zeros, argmax, arange
+
+
+class AdversarialFairness(BaseEstimator):
```
romanlutz, line 34: This name seems (to me) like it's missing a part, like `estimator` or `mitigator`.

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,748 @@
...
+    r"""Train complex predictors while mitigating biases in PyTorch or Tensorflow.
```
romanlutz, line 35:
```suggestion
    r"""Train complex predictors while mitigating unfairness in PyTorch or Tensorflow.
```
Would that work? Bias is such a vague word that I try to avoid it wherever possible. Unfairness isn't that much clearer, but at least it eliminates some of the confusion from ambiguous definitions that bias doesn't. The authors of that paper use the word bias, of course, which makes this a bit trickier....
It would be nice to have a paragraph in the API reference on what types of biases/unfairness this can mitigate.

romanlutz: Nevermind on the last sentence. I see the `constraints` description contains this already.

romanlutz (commented):
I'll take back my comment on splitting this PR in multiple chunks. Half the lines are from tests, so those can be skipped in early reviews which we should all be doing right now. The remaining lines are all essential for a PR as we've defined on the community calls (API reference, user guide). Nevertheless, this is not a small PR, so it may take a little while to dig through all of this. You've done a fantastic job in presenting us with a PR that makes sense and can be reviewed, with plenty of comments to help reviewers. Thanks for your patience!
@fairlearn/fairlearn-maintainers may want to have (another) look at this PR.

fairlearn/adversarial/_adversarial_mitigation.py
SeanMcCarren: Yes! Excellent point. AdversarialFairnessEstimator or AdversarialMitigator or so, what do you like?

hildeweerts: Given `AdversarialFairnessClassifier` and `AdversarialFairnessRegressor` I think `AdversarialFairnessEstimator` makes the most sense, but I'm happy either way.

riedgar-ms:
Gentle ping on this, since we added it to the v.8 milestone @hildeweerts @romanlutz @adrinjalali

examples/plot_adversarial_mitigation.py
```diff
@@ -0,0 +1,367 @@
...
+#
+# In short, the authors take the classic supervised learning setting in which
+# a predictor neural network is trained, and extend it with an adversarial
+# network that aims to predict the sensitive feature given these
+# predictions and possibly also given the true target. Then, they train the
+# predictor not only to minimize its own loss, but also minimize the predictive
+# ability of the adversarial. If this model converges properly,
+# the adversary will attain a loss equal to the entropy, so the adversary
+# can not
+# predict the sensitive features from the predictions.
```
adrinjalali, line 22:
```suggestion
# can not predict the sensitive features from the predictions.
```

examples/plot_adversarial_mitigation.py
```diff
@@ -0,0 +1,367 @@
...
+# objective that we choose for this example is demographic parity, so we also
+# set :code:`objective = "demographic_parity"`. We generally follow sklearn API,
+# but in this case we require some extra kwargs. In particular, we should
+# specify the number of epochs, batch size, whether to shuffle the rows of data
+# after every epoch, and optionally after how many seconds to show a progress
+# update.
+
+
+mitigator = AdversarialFairnessClassifier(
+    backend="tensorflow",
```
adrinjalali, line 131: I thought we wanna do a pytorch example.

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,822 @@
...
+                raise ValueError(
+                    _KWARG_ERROR_MESSAGE.format(kwname, "a non-negative number")
+                )
+
+        # Positive or -1 parameters
+        for kw, kwname in (
+            (self.batch_size, "batch_size"),
+            (self.epochs, "epochs"),
+            (self.max_iter, "max_iter"),
+        ):
```
adrinjalali, line 322: you might wanna check `check_scalar` from sklearn (https://scikit-learn.org/dev/modules/generated/sklearn.utils.check_scalar.html?highlight=check_scalar#sklearn.utils.check_scalar)

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,822 @@
...
+            ):
+                raise ValueError(
+                    _KWARG_ERROR_MESSAGE.format(
+                        kwname,
+                        "a sklearn Transformer (subclass TransformerMixin)",
+                    )
+                )
+
+        self.y_transform_ = self.y_transform.fit(Y)
+        self.z_transform_ = self.z_transform.fit(Z)
```
adrinjalali, line 402: if we don't want to expose these as _public_ to the user (which I think we don't need to), I'd do them as `self._y_transformer` instead

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,822 @@
...
+
+        if self.cuda and not isinstance(self.cuda, str):
+            raise ValueError(
+                _KWARG_ERROR_MESSAGE.format(
+                    "cuda",
+                    'None, or the device to use (such as "cuda:0")',
+                )
+            )
+
+        self.random_state_ = check_random_state(self.random_state)
```
adrinjalali, line 412: do we need to store this in the instance? It could be used just where it's needed, can't it?

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,822 @@
...
+        self.step_ = 0
+        for epoch in range(epochs):
+            for batch in range(batches):
+                if self.progress_updates:
+                    if (time() - last_update_time) > self.progress_updates:
+                        last_update_time = time()
+                        progress = (epoch / epochs) + (
+                            batch / (batches * epochs)
+                        )
+                        print(
```
adrinjalali, line 488: can we use logger instead?

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,822 @@
...
+
+        if (not is_fitted) or (reinitialize):
+            self.__setup(X, Y, Z)
+
+        if not self.skip_validation:
+            Y = self.y_transform_.transform(Y)
+            Z = self.z_transform_.transform(Z)
+
+        # Check for equal number of samples
+        if not (X.shape[0] == Y.shape[0] and X.shape[0] == Z.shape[0]):
```
adrinjalali, line 636: https://scikit-learn.org/stable/modules/generated/sklearn.utils.check_consistent_length.html

fairlearn/adversarial/_preprocessor.py
```diff
@@ -0,0 +1,203 @@
...
+from pandas import Series, DataFrame
+from sklearn.base import BaseEstimator, TransformerMixin
+from sklearn.utils.multiclass import type_of_target
+from numpy import all as np_all
+from numpy import sum as np_sum
+
+
+class FloatTransformer(BaseEstimator, TransformerMixin):
+    """
+    Transformer that maps dataframes to numpy arrays of floats.
```
adrinjalali, line 21: could you better explain here what it does and why it's needed?

adrinjalali (commented):
can't check in detail, but I checked the API and it looks quite nice now, nice work :)

adrinjalali:
We're hoping to have this in the next release (hopefully by end of March), so heads up :)

docs/user_guide/adversarial.rst
```diff
@@ -0,0 +1,281 @@
...
+trains the predictor with gradient:
+
+.. math::
+    \nabla_W L_P - \text{proj}_{\nabla_W L_A} \nabla_W L_P - \alpha \nabla_W L_A
+
+If this model converges properly,
+the adversary will attain a loss equal to the entropy, so the adversary
+can not
+predict the sensitive features from the predictions.
+Moreover, this model can be trained for either *demographic parity* or
```
riedgar-ms, line 29: Side question: can this be extended to just do True Positive Rate Parity or False Positive Rate Parity? I suppose what I'm getting at is that I suspect those are just the _implemented_ constraints?

docs/user_guide/adversarial.rst
```diff
@@ -0,0 +1,281 @@
...
+do recommend you create your own neural networks beforehand,
+as you can tweak these to your
+liking.
+
+.. _data_preprocessing:
+
+Data Preprocessing
+~~~~~~~~~~~~~~~~~~
+
+In contrast to other mitigation techniques, we require you to
```
riedgar-ms, line 110: Stylistic note.... I'm seeing "the user" rather than "you" in other places.

examples/plot_adversarial_mitigation.py
riedgar-ms: I'm not so keen on prediciting things from the predictions. Should it be 'recover the sensitive features?'

examples/plot_adversarial_mitigation.py
```diff
@@ -0,0 +1,367 @@
...
+# %%
+# Example 2: Finetuning
+# =====================
+# Adversarial Learning is inherently difficult because of various issues,
+# such as mode collapse, divergence, and diminishing gradients.
+# In particular, mode collapse seems a real problem on this dataset: the
+# predictor and adversary trap themselves in a local minimum by favoring one
+# class (mode). Problems with diverging parameters may also occur, which
+# may be an indication of a bad choice of hyperparameters, such as a
+# learning rate that is too large. This problems one may encounter are
```
riedgar-ms, line 180: 'This problems one may encounter' ?

examples/plot_adversarial_mitigation.py
```diff
@@ -0,0 +1,367 @@
...
+# that finetune your estimator,
+# Example 2 will demonstrate some problem-specific
+# techniques we can use such as using dynamic hyperparameters,
+# validation, and early stopping to improve this method.
+
+# %%
+# Example 2: Finetuning
+# =====================
+# Adversarial Learning is inherently difficult because of various issues,
+# such as mode collapse, divergence, and diminishing gradients.
```
riedgar-ms, line 175: One line definitions for 'mode collapse' etc. would be helpful.

examples/plot_adversarial_mitigation.py
```diff
@@ -0,0 +1,367 @@
...
+# class (mode). Problems with diverging parameters may also occur, which
+# may be an indication of a bad choice of hyperparameters, such as a
+# learning rate that is too large. This problems one may encounter are
+# of course case specific, but general good practices when training
+# such models are: train slowly, ensuring the
+# losses remain balanced, and keep track of validation accuracies.
+# Additionally, we found that single hidden layer neural
+# networks work best for this use case.
+
+# In this example, we show some of these good practices in practice.
```
riedgar-ms, line 187: 'practices in practice' ?

examples/plot_adversarial_mitigation.py
```diff
@@ -0,0 +1,367 @@
...
+            and accuracy > 0.8
+            and selection_rate > 0.01
+            and selection_rate < 0.99
+        ):
+            return True
+
+
+# %%
+# Then, the instance itself. Notice that we do not explicitely define loss
+# functions, because the model is able to infer this on its own in this example.
```
riedgar-ms, line 286: 'infer this'  does not match 'loss functions'

fairlearn/adversarial/__init__.py
```diff
@@ -0,0 +1,23 @@
...
+
+from ._adversarial_mitigation import (
+    AdversarialFairness,
+    AdversarialFairnessClassifier,
+    AdversarialFairnessRegressor,
+)
+
+# from ._backend_engine import BackendEngine
```
riedgar-ms, line 12: If these comment lines are obsolete, please can you remove them (and below)

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,822 @@
...
+    `"Mitigating Unwanted Biases with Adversarial Learning"
+    <https://dl.acm.org/doi/pdf/10.1145/3278721.3278779>`_ [1]_.
+    This algorithm takes as input two neural network
+    models, a predictor and an adversarial, defined either as a `PyTorch module
+    <https://pytorch.org/docs/stable/generated/torch.nn.Module.html>`_ or
+    `Tensorflow2 model
+    <https://www.tensorflow.org/api_docs/python/tf/keras/Model>`_. You train this
+    predictor using an API that is similar to estimators in `sklearn`.
+
+    As per [1]_, the neural network models :code:`predictor_model` and
```
riedgar-ms, line 48: The use of 'may' in this paragraph is confusing me. I think that it's saying that the two models have to have outputs which are class probabilities, rather than a single prediction?

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,822 @@
...
+    :code:`adversary_model` may not have a discrete prediction at the end of the
+    model. Instead, even if we are dealing with discrete predictions, we may output
+    the sigmoidal or soft-max, but we may not output discrete integer predictions.
+
+    The distribution types of :code:`y` and :code:`sensitive_features`
+    are set by their preprocessor :code:`y_transform` and :code:`z_transform`
```
riedgar-ms, line 54: `z` for sensitive features? Isn't `A` more common in the literature?

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,822 @@
...
+        the number of nodes :math:`k_i` (if :math:`k_i` is integer) or a keyword
+        that indicates an activation function (if :math:`k_i` is a string) or
+        a layer or activation function instance directly (if :math:`k_i` is
+        callable). The default parameter is :code:`[]`, which indicates
+        a neural network without any hidden layers.
+        However, the number of nodes in the input
+        and output layer are automatically inferred from data, and the final
+        activation function (such as softmax for categorical
+        predictors) are inferred from data.
+        If :code:`backend` is specified, we cannot pass a model
```
riedgar-ms, line 88: This 'we cannot pass' does not match the 'You can also pass' in the previous parameter block.

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,822 @@
...
+        FloatTransformer that maps strictly binary or categorical tables
+        to binary or
+        one-hot encodings, and it maps strictly continuous-valued (possible 2d)
+        to itself.
+
+    learning_rate : float, default = 0.001
+        A small number greater than zero to set as initial learning rate
+
+    alpha : float, default = 1.0
+        A small number :math:`\alpha` as specified in the paper.
```
riedgar-ms, line 175: Could you give a sentence about what this controls?

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,822 @@
...
+        The preprocessor to use on the :code:`sensitive_features`.
+        The given transformer *must* map data
+        to a 2d ndarray containing only floats. Per default, we use a
+        FloatTransformer that maps strictly binary or categorical tables
+        to binary or
+        one-hot encodings, and it maps strictly continuous-valued (possible 2d)
+        to itself.
+
+    learning_rate : float, default = 0.001
+        A small number greater than zero to set as initial learning rate
```
riedgar-ms, line 172: Should the parameter be called `initial_learning_rate` then?

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,822 @@
...
+        A small number :math:`\alpha` as specified in the paper.
+
+    epochs : int, default = 1
+        Number of epochs to train for.
+
+    batch_size : int, default = -1
+        Batch size. For no batching, set this to -1.
+
+    shuffle : bool, default = False
+        Iff true, shuffle the data after every iteration.
```
riedgar-ms, line 184: `Iff` typo?

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,822 @@
...
+        elif self.constraints == "equalized_odds":
+            self.pass_y_ = True
+        else:
+            raise ValueError(
+                _KWARG_ERROR_MESSAGE.format(
+                    "constraints", "'demographic_parity' or 'equalized_odds'"
+                )
+            )
+
+        # Non-negative parameters
```
riedgar-ms, line 306: Would the pure validation checks here (as opposed to those which also set internals, such as the constraints check above) be better placed in a separate method?

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,822 @@
...
+
+        # Initialize backend
+        # here, losses and optimizers are also set up.
+        self.backendEngine_ = self.backend_(self, X, Y, Z)
+
+        # Sklearn-parameters
+        self.n_features_in_ = X.shape[1]
+        self.n_features_out_ = self.y_transform_.n_features_in_
+
+    def fit(self, X, y, *, sensitive_features=None):
```
riedgar-ms, line 424: What will happen if nothing is passed for `sensitive_features` ?

fairlearn/adversarial/_adversarial_mitigation.py
riedgar-ms: Agree on using logger.

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,822 @@
...
+
+                self.step_ += 1
+
+                # Purposefully first stop and then handle callbacks
+                if self.max_iter != -1 and self.step_ >= self.max_iter:
+                    return self
+
+                if self.callbacks:
+                    stop = False
+                    if isinstance(self.callbacks, list):
```
riedgar-ms, line 522: Small thought, rather than have this logic here, why not 'normalise' the callbacks so that even if the use passes a single callback function, `self.callbacks` will always be a list?

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,822 @@
...
+        if not self.skip_validation:
+            X = check_X(X)
+
+        try:  # TODO check this
+            check_is_fitted(self)
+            is_fitted = True
+        except NotFittedError:
+            is_fitted = False
+
+        if Z is None:
```
riedgar-ms, line 624: Don't we use `A` for sensitive features?

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,822 @@
...
+            pass
+        elif isinstance(self.predictor_function_, str):
+            kw = self.predictor_function_
+            if kw == "binary":
+                self.predictor_function_ = lambda pred: (pred >= 0.5).astype(
+                    float
+                )
+            elif kw == "category":
+
+                def loss(pred):
```
riedgar-ms, line 765: Whence is this coming?

fairlearn/adversarial/_preprocessor.py
riedgar-ms: The description is making it sound like an sklearn pipeline

riedgar-ms (commented):
I've not been through everything in detail yet, but I wanted to leave some comments. Some of them should be fairly easy to address.

docs/user_guide/adversarial.rst
SeanMcCarren: More constraints can be implemented, such as equality of opportunity, but for TPR or FPR I am not quite sure what objective the adversary should then have. Perhaps it is possible.

docs/user_guide/adversarial.rst
SeanMcCarren: Oh that is good to know, I'll adjust this, thanks

fairlearn/adversarial/_adversarial_mitigation.py
SeanMcCarren: I sometimes overuse if and only if haha :)

fairlearn/adversarial/_adversarial_mitigation.py
SeanMcCarren: it depends on what the user does with the optimizers, decaying learning rate is not the default (though it would be nice!). About the optimizers, I'm not entirely a fan of my own work here, as initializing the optimizers is quite messy. But despite trying I did not see a more elegant solution that provides flexibility to the user. Maybe you have more experience with this?

fairlearn/adversarial/_adversarial_mitigation.py
SeanMcCarren: I do agree, but I saw many sklearn methods do it like this

fairlearn/adversarial/_adversarial_mitigation.py
SeanMcCarren: Yes, I mistakenly used the paper's conventions

fairlearn/adversarial/_adversarial_mitigation.py
SeanMcCarren: I'm not sure what you mean?

fairlearn/adversarial/_adversarial_mitigation.py
SeanMcCarren: Yes! Changed this thanks

fairlearn/adversarial/_adversarial_mitigation.py
SeanMcCarren: That is a good question, I should check these cases.

fairlearn/adversarial/_adversarial_mitigation.py
SeanMcCarren: That's an awfully scary function for such a simple task :)

fairlearn/adversarial/_adversarial_mitigation.py
SeanMcCarren: I agree they needn't be public. This then overrules the sklearn convention of appending a `_` to validated parameters?

fairlearn/adversarial/_adversarial_mitigation.py
SeanMcCarren: Perhaps I interpreted this in the wrong way, but I was following this convention https://scikit-learn.org/stable/developers/develop.html#random-numbers

examples/plot_adversarial_mitigation.py
SeanMcCarren: yea we did say this indeed

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,822 @@
...
+        """
+        check_is_fitted(self)
+        X = check_X(X)
+
+        Y_pred = self.backendEngine_.evaluate(X)
+        return Y_pred
+
+    def predict(self, X):
+        """
+        Compute discrete predictions for given test data.
```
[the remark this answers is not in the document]
SeanMcCarren: good catch

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,822 @@
...
+
+        Important note is that we follow call `__setup` from here, because the
+        setup procedure requires the validated data. If `reintialize` is True,
+        then always call `__setup`.
+        """
+        if not self.skip_validation:
+            X = check_X(X)
+
+        try:  # TODO check this
+            check_is_fitted(self)
```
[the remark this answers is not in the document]
SeanMcCarren: Oh I was looking for something like this, it felt slow!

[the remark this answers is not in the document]
SeanMcCarren: Haha you made this :D

fairlearn/adversarial/_preprocessor.py
```diff
@@ -0,0 +1,203 @@
...
+# hash = lambda d : id(d) + sum(d.shape) + d.size + id(d.dtype)
+# memoize_store = {}
+def _infer_type(data):
+    """Memoize type_of_target without prohibiting data garbage-disposal."""
+    # memoized = memoize_store.get(hash(data), None)
+    # if memoized is not None:
+    #     print("hit")
+    #     return memoized
+    # print("miss")
+    inferred = type_of_target(data)
```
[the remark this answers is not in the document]
SeanMcCarren: I would like to memoize this expensive function somehow, but I'll add this as a comment and remove the function for now

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,822 @@
...
+        # Or all other cases, a mismatch between model model (and installation)
+        raise ValueError(
+            _KWARG_ERROR_MESSAGE.format(
+                "predictor_model and adversary_model",
+                "a list, torch.nn.Module, or tensorflow.keras.Model. Also, "
+                + "make sure to have installed the corresponding backend",
+            )
+        )
+
+    def _set_predictor_function(self):
```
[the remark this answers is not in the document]
SeanMcCarren: I like this, but even for classifier there are still two options for predictor function (mapping sigmoid to 0/1, or mapping to vector to argmax). Also, how I currently intended it is for the main class AdversarialFairness to be useable in any case, and that stuff like predictor function is inferred (instead of implicitly specifying it through the choice of AdversarialFairness{Classifier,Regressor})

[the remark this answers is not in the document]
SeanMcCarren: Curious what you think. I guess a lot of design choices regarding adversarial fairness could (and maybe should) have been in favor of simpler API and taking away some choices from the user.

fairlearn/adversarial/_adversarial_mitigation.py
SeanMcCarren: Even here, in case we are certain that everything is a numpy array?

SeanMcCarren:
Thanks for the feedback @adrinjalali @riedgar-ms. I closed the comments that were resolved.

Also, I moved the notebook to the test_othermlpackages, but that CI is not triggered. Did you disable that CI @riedgar-ms ? I forget.

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,841 @@
...
+        predictors) are inferred from data.
+        If :code:`backend` is specified, you cannot pass a model
+        that uses a different backend.
+
+    adversary_model : list, torch.nn.Module, tensorflow.keras.Model
+        The adversary model to train. Defined similarly as predictor_model.
+        Must be the same type as the
+        :code:`predictor_model`.
+
+    predictor_loss : str, callable, default = 'auto'
```
MiroDudik, line 103: I don't love the current keyword choices for `predictor_loss`, `adversary_loss`, because they seem to refer to the type of the target, rather than the loss. If we go that route, we should consider using something similar to sklearn's [target types](https://scikit-learn.org/stable/modules/generated/sklearn.utils.multiclass.type_of_target.html#sklearn.utils.multiclass.type_of_target).

Alternatively, we could use keywords that describe the loss, like `"square_loss"`, `"logistic_loss"`... but let me think a bit more about this.

MiroDudik (commented):
I'm not fully done yet with my pass. I'm focusing on API and documentation.

fairlearn/adversarial/_adversarial_mitigation.py
SeanMcCarren: You raise an excellent point. Ideally, we'd also provide something like `Y_distribution_type` and `A_distribution_type`. I've thought about this before, but I can't remember why I let go of the idea. We should not get rid of the loss parameters though, and we would have to be sure that the inferred distribution type of the preprocessor agrees.

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,841 @@
...
+    alpha : float, default = 1.0
+        A small number :math:`\alpha` as specified in the paper. It
+        is the factor that balances the training towards predicting :math:`Y`
+        (choose :math:`\alpha` closer to zero) or increasing fairness
+        (choose larger :math:`\alpha`).
+
+    epochs : int, default = 1
+        Number of epochs to train for.
+
+    batch_size : int, default = -1
```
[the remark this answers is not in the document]
SeanMcCarren: One element per batch? I think that is a bit uncommon, but I can see why that is better than everything in one batch.

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,841 @@
...
+
+    cuda : str, default = None
+        A string to indicate which device to use when training. For instance,
+        set :code:`cuda='cuda:0'` to train on the first GPU. Only for PyTorch
+        backend.
+
+    warm_start : bool, default = False
+        Normally, when set to False, a call to fit triggers reinitialization,
+        which discards the models and intializes them again. Setting to
+        True triggers reuse of these models. Note: if pre-initialized models
```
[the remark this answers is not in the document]
SeanMcCarren: Yes, how you write it is more clear. :)

fairlearn/adversarial/_preprocessor.py
SeanMcCarren: I changed it a bit, is there anything I'm missing still perhaps?

fairlearn/adversarial/_adversarial_mitigation.py
riedgar-ms: Can you explain the origin of this loss function? Why is it defined deeply embedded here, rather than somewhere it can be unit tested?

fairlearn/adversarial/_adversarial_mitigation.py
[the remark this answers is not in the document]
MiroDudik: I was just thinking about what should be a sensible default behavior. If I understand correctly, with epoch=1 and batch_size=-1, there would be only one update, which doesn't seem like a good idea.

There's usually an interplay between batch_size and learning_rate, see, e.g., the plots in [Revisiting small batch training for deep neural networks](https://arxiv.org/pdf/1804.07612.pdf). My sense is that choosing a small value of batch_size is generally a good choice for optimization as long as the learning rate is also appropriately small. However, to leverage parallel execution, it can be beneficial to increase the batch_size and also appropriately adjust learning_rate.

fairlearn/adversarial/_adversarial_mitigation.py
MiroDudik: Actually, another option would be to make `AdversarialFairnessEstimator` a private class and only export `AdversarialFairnessClassifier` and `AdversarialFairnessRegressor`. I can see arguments either way. Do folks have any thoughts on this?

docs/user_guide/mitigation.rst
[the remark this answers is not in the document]
adrinjalali: I agree with links, But this is normal text.

fairlearn/adversarial/_adversarial_mitigation.py
adrinjalali: but it makes your code shorter and error messages more consistent.

fairlearn/adversarial/_adversarial_mitigation.py
adrinjalali: sklearn's convention is that you `attr_` is a fitted public attribute, and `_attr` is a private fitted attribute.

fairlearn/adversarial/_adversarial_mitigation.py
adrinjalali: it's not about data conversion, it's about not writing the same code multiple times and having consistent error messages.

fairlearn/adversarial/_adversarial_mitigation.py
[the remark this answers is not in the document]
adrinjalali: I generally agree with you, but I think for the user to choose a regressor or a classifier is not something we should hide from them.

fairlearn/adversarial/_preprocessor.py
adrinjalali: Reading the description, my question is then why not use a columntransformer?

test_othermlpackages/package_test_common.py
adrinjalali: we can change the CI.

fairlearn/adversarial/_adversarial_mitigation.py
MiroDudik: I had some further thoughts on this.

I think that the very basic question is whether we want to represent binary classification problems via networks with a single output or two outputs... so that should be decided upfront.

With that said, what do you think about the following tweaks of the current API:
* base class (currently `AdversarialFairness`)
   * allow specifying `y_transform`, `sf_transform`, `predictor_loss`, `adversary_loss`, `predictor_function`
   * they can take values `None` (no transformation; this might still mean casting ints as a floats?), `'auto'` (default), a callable, and additionally:
     * for `predictor/adversary_loss`, we support `'logistic_loss'`, `'square_loss'`
       * if needed, we could also distinguish `'multinomial_logistic_loss'` (which acts on one-hot encoding), but this could be inferred from the number of outputs of the network
     * for `y/sf_transform`, we support `'one_hot_encoder'`
     * for `predictor_function`, we support `'argmax'` (for one-hot representation) and `'threshold'` (for one-output representation of binary classifiers), with an additional argument `threshold_value` with the default value 0
  
* `AdversarialFairnessClassifier`
  * allow specifying `sf_transform`, `adversary_loss`
  * fill in `y_transform`, `predictor_loss`, `adversary_loss`, `predictor_function` so as to support (multinomial) logistic regression; I wouldn't allow overriding these (if that's what you want to do, you can just call the base class)
  * besides `predict` and `decision_function`, we should consider supporting `predict_proba` and `predict_log_proba`

* `AdversarialFairnessRegressor`
  * allow specifying `sf_transform`, `predictor_loss`, `adversary_loss`
  * fill in `y_transform`, `predictor_function` as `None`

And one more idea--not sure how much I like it, but something that occurred to me. We could change:
* `predictor_loss` -> `y_loss`
* `adversary_loss` -> `sf_loss`

SeanMcCarren: I like your thoughts! I really appreciate you taking the time! 

I just want to reiterate that we (it was a suggestion from adrin or hilde I think) made the design choice that as much is automatically inferred from data as possible, so that means what kind of transform/loss/predictor function to use. Looking at this now, I am not so sure why I included `y/sf_transform` as parameters, because without this `FloatTransformer` class thingy (`FloatTransform` is the default `y/sf_transform`) we lose this automatic inference. The nice thing is that if the user has already applied a one-hot-encoding, the FloatTransform class will still infer and tell AdversarialFairness that the data is categorical, which is required to infer what loss function to use. I agree that the current API needs changing, but I do not see yet how we can resolve this nicely.

> * for `y/sf_transform`, we support `'one_hot_encoder'`

I think I like this keyword-style better than how it is currently done, but I'm still on the fence. Currently, you can achieve this `sf_transform='one_hot_encoder'` with `sf_transform=FloatEncoder('categorical')`, so what you suggest is definitely cleaner. What do you think should happen if the user provides a custom transform? Should we then require the user to (1) also pass loss functions, because we can no longer infer them from FloatEncoder? Or (2) should we do the inferring outside of FloatEncoder? or (3) should we not even let the user pass a custom transform? I find this a difficult choice because all options feel bad in some sense. I've actually switched implementations from (2) to (1) in the past, but I am now tending back towards (2) and using the keywords you suggested. I'd love to hear your thoughts on this. @adrinjalali you were quite involved in this design choice regarding preprocessing in the past, so perhaps you can help us here. 

> * for `predictor/adversary_loss`, we support `'logistic_loss'`, `'square_loss'`

Currently we accept `'binary'`, `'category'`, `'continuous'`, I chose those because they are descriptive of the distribution that you assume, but I understand that you favor more precise names such as `'logistic_loss'`, `'square_loss'`, or `'argmax'`? I might agree actually.

> * fill in `y_transform`, `predictor_loss`, `adversary_loss`, `predictor_function` so as to support (multinomial) logistic regression;

Then we'd still need to infer whether it is categorical or binary I'd say, hence I'd still love to know the distribution type somehow (rather through inferring it from the data than through a keyword parameter)

> And one more idea--not sure how much I like it, but something that occurred to me.

I like this!


> * besides `predict` and `decision_function`, we should consider supporting `predict_proba` and `predict_log_proba`

Yeah sure!

MiroDudik: Let me try to answer your questions--but also let me know if I left something unanswered!

**What to do about `'auto'` loss functions and `predictor_function` with custom `y/sf_transform`.**
* The behavior should be the same as what we do when the transform is `None`. I'm not sure what you do currently, but a sensible option would be an automatic inference among:
  1. univariate {0.0,1.0} (-> univariate logistic model with sign-based prediction function),
  2. one-hot-encoded categorical (-> multinomial logistic with argmax prediction function),
  3. continuous univariate or vectors (-> square loss/L2 norm with identity prediction function),
  4. otherwise: throw an exception

**Loss versus link function ("distribution assumption").**
* I think by "distribution assumption" you mean the [link function](https://en.wikipedia.org/wiki/Generalized_linear_model#Link_function) that is used in generalized linear models. I'd be in favor of keeping the idea of link (which is basically our `prediction_function`) separate from loss function. For example, it would be awkward to specify Huber loss (for robust regression) using the language of link functions.

**Categorical vs. binary encoding in classification.**
* For this one, I'm not sure that we have any other option other than inferring from the provided `y`. My idea would be to support the binary vs. multiclass classification similarly to sklearn's [LogisticRegression](https://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html).

SeanMcCarren: Thanks, this was really clarifying!

> What to do about `'auto'` loss functions and `predictor_function` with custom `y/sf_transform`.

Oh so doing this inference if the transform is None, and otherwise just do the transform and infer then. That sounds good. Basically decoupling this inference from the transform. Okay! Apart from that, I make the same inferences (the i. through iv.). (I initially chose to be a bit more general and accept mixed column types (so for instance binary and continuous columns), but we decided there was no use for this.)

> Loss versus link function ("distribution assumption").

I must say I am learning a lot here, I was not familiar with these terms. I understand a reason for keeping these things seperate now, thanks!

SeanMcCarren: @MiroDudik Hey, I finally got around to this, was a bit more work than I anticipated. What do you think? Internally, I still use "binary", "category", "continuous", because I feel this may be relevant to know later, but the API changed and accepts the better more concrete terms now.

As for the `predict_(log_)proba`, I am not sure how to interpret this. In case of a single output neuron $\hat y$, should we unfold it into a vector with separate probabilities for $\hat y = 0$ and $\hat y = 1$? And what if there are more weird cases? For now, we do have `decision_function`, which might give enough flexibility?

fairlearn/adversarial/_preprocessor.py
SeanMcCarren: @adrinjalali my apologies I missed your response. A columntransformer such as in the example notebook would have worked fine too yes, and I had that at some point in time. But I decided I don't want to, because I don't want to deal with defining loss functions for mixed type columns. Say you pass two categorical columns for `y`. Then, how should we define a loss function for this? We have thought about multi-output loss functions, but decided it was too strange. Maybe a good option would have been to do the columntransform nonetheless, and then just raise an error if the user does not supply a loss function with it. That should be do-able besides some minor detail I won't mention. With more thought, there may be something better, but I think we should see if there is desire for this first.

fairlearn/adversarial/_adversarial_mitigation.py
SeanMcCarren: @riedgar-ms sorry, i missed this message somehow. It seems trivial, but I would understand if you'd like this extracted and unit tested

fairlearn/adversarial/_adversarial_mitigation.py
SeanMcCarren: ah okey that makes sense!

fairlearn/adversarial/_adversarial_mitigation.py
SeanMcCarren: Rather not, because we are purposefully exposing the losses and other API to make it applicable to more difficult use cases that do not fall under classification or regression, such as multi-output/multitask classification

fairlearn/adversarial/_adversarial_mitigation.py
SeanMcCarren: oh sorry, I see now that I wrote "initial" in the sentence below

fairlearn/adversarial/_adversarial_mitigation.py
SeanMcCarren: @adrinjalali What do you think?

fairlearn/adversarial/_adversarial_mitigation.py
SeanMcCarren: Was raising a warning, but changed it to an Exception

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,841 @@
...
+        (choose larger :math:`\alpha`).
+
+    epochs : int, default = 1
+        Number of epochs to train for.
+
+    batch_size : int, default = -1
+        Batch size. For no batching, set this to -1.
+
+    shuffle : bool, default = False
+        When true, shuffle the data after every iteration.
```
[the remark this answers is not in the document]
SeanMcCarren: @MiroDudik Do you like the `shuffle_start` parameter, that if true shuffle at the start?

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,899 @@
...
+        self.shuffle = shuffle
+        self.shuffle_start = shuffle_start
+        self.progress_updates = progress_updates
+        self.skip_validation = skip_validation
+        self.callbacks = callbacks
+        self.cuda = cuda
+        self.warm_start = warm_start
+        self.random_state = random_state
+
+    def __setup(self, X, Y, A):
```
adrinjalali, line 308:
```suggestion
    def _setup(self, X, Y, A):
```

the canonical way to indicate a method is private is a single `_`.

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -0,0 +1,899 @@
...
+            raise ValueError(
+                _KWARG_ERROR_MESSAGE.format(
+                    "constraints", "'demographic_parity' or 'equalized_odds'"
+                )
+            )
+
+        # Numbers
+        check_scalar(self.threshold_value, "threshold_value", (int, float))
+
+        # Non-negative numbers
```
adrinjalali, line 336: I'm agnostic on whether these should be a separate method or not. Either way is fine with me.

fairlearn/adversarial/_preprocessor.py
adrinjalali: What I meant was to have a `ColumnTransformer` that does exactly what this transformer does. The output would be an array of floats, and then nothing would be different. You'd apply your ColumnTransformer internally on your `y`.

fairlearn/adversarial/_adversarial_mitigation.py
MiroDudik: I thought about all of this some more, and I'm still finding parts of the constructor API unnecessarily complicated--and there are various dependencies spread across multiple parameters. In particular, I'd like to suggest an alternative to what's currently handled by: `predictor_loss, adversary_loss, predictor_function, y_transform, sf_transform`

**AdversarialMitigationClassifier**
* can handle binary classification and multi-class classification
* for encoding of `y`, we follow sklearn`s [target types](https://scikit-learn.org/stable/glossary.html#glossary-target-types), which are described [here](https://scikit-learn.org/stable/modules/multiclass.html)
* _Binary classification_
  * `y` has shape `(n_samples,)` and it contains two values typically {0,1}; if other values are provided they are transformed into {0,1} during fit and predict
  * the neural net is outputting a single scalar, corresponding to the logit of P(Y=1|X), and is trained by minimizing logistic loss
 * _Multiclass classification_
   * `y` has either shape `(n_samples,n_classes)` with 0/1 values implementing one-hot-encoding, or it has shape `(n_samples,)` and contains 3 or more distinct values; in the latter case, it is transformed into one-hot-encoding during fit and predict
   * the neural net is outputting a vector of `n_classes` scalars, corresponding to the linear scores of multinomial logistic model of P(Y|X), and is trained by minimizing log loss (this is called CrossEntropyLoss in PyTorch)
* _Adversarial model_
   * sensible defaults around fitting sensitive features are a bit more tricky, but I think that we should begin by covering the common case of binary or categorical sensitive features:
     * `sensitive_features` is shape `(n_samples,)` or `(n_samples,n_sensitive_features)` where every column has two distinct values -> columns are transformed into {0,1}, and the training loss is the sum of logistic losses (i.e., treated as the sum of binary problems)
     * `sensitive_features` is shape `(n_samples,)` or `(n_samples,n_sensitive_features)` where at least one column has three or more values -> columns are transformed using one-hot-encoding, and the training loss is the sum of log losses (i.e., treated as the sum of multiclass problems)
* **This means that we could get rid of the parameters `predictor_loss, adversary_loss, predictor_function, y_transform, sf_transform`.**
  * If we want to expose more generality (say continuous sensitive features), we could do that later.

**AdversarialMitigationRegressor**
* `y` has shape `(n_samples,)` and contains floats, the loss is square loss
* sensitive features and adversarial model are treated the same way as in AdversarialMitigationClassifier.
* again, this means we could get rid of `predictor_loss, adversary_loss, predictor_function, y_transform, sf_transform`, and add more generality later.

**AdversarialMitigation**
* I suggest to make this private; in that case I don't care too much about the API.

fairlearn/adversarial/_adversarial_mitigation.py
MiroDudik: See my comment above -> if we specialize to logistic models, then `predict_proba` and `predict_log_proba` are obvious, right? These would be only provided for AdversarialMitigationClassifier.

adrinjalali:
Meeting summary: have separate PR from this one with the minimal implementation, as small as we're happy to have it released, and have the main base class private at least for now.

riedgar-ms:
Is this still active, or should it be closed in favour of #1079 ?

Issue #825, opened 2021-05-27
Make navigation bar consistent across website


hildeweerts:
#### Describe the issue linked to the documentation

The navigation bar is currently inconsistent between the homepage and the remainder of the website:

* The order of items differs
* *Contributor guide* on home page refers to #contribute rather than https://fairlearn.org/v0.6.2/contributor_guide/index.html
* *Getting Started* vs *Quickstart*
* *API docs* versus *API Reference*
* *About* vs *About us*

#### Suggest a potential alternative/fix

Make the navigation bars consistent.

@MiroDudik @LeJit

@romanlutz we might split this up into multiple smaller issues in case we're still looking for good first issues for the SciPy sprint.

MiroDudik:
This issue is generating a lot of confusion. It's definitely not a good first issue. The solution is to:
* Make the navbar in the documentation pages look just like the navbar on the landing page. The landing page is the one that has been vetted.

romanlutz:
Reordering the items and renaming them in some cases seems like a very good first issue if you ask me. The only confusion arose from which part should be made consistent with which other part. Now it's just a fix in a single rst file.

MiroDudik:
Ok--I've updated my comment above. Besides reordering, some section titles may need to be updated.

romanlutz:
Let's keep this issue open since the associated PR(s) weren't merged as they preceded @MiroDudik's more specific instructions (on fixing the bar in all the other pages to make it consistent with the landing page, rather than the other way round). This is up for grabs for anyone who'd like to try! Just comment below.

romanlutz:
Some more information: the top navbar seems to be configurable as described in https://pydata-sphinx-theme.readthedocs.io/en/latest/user_guide/sections.html#the-navbar-items

The default templates are at https://github.com/pydata/pydata-sphinx-theme/tree/master/pydata_sphinx_theme/_templates

PR #976, opened 2021-10-08
DOC make headers consistent across website

romanlutz:
Adjusting section titles for the headers to be consistent on the landing page and the other pages. This does not fix the layout differences or font differences, though, so we may want to keep the associated issue open.

related to #825


hildeweerts:
We should probably also adjust the ordering of the sphinx generated docs, I think that should be done [in this file](https://github.com/fairlearn/fairlearn/blob/romanlutz/website_nav/docs/index.rst).

Issue #924, opened 2021-08-02
DOC Fix CITATION.cff formatting


romanlutz:
It looks like I formatted it the wrong way when adding the file in #921 and hence it doesn't render properly

![image](https://user-images.githubusercontent.com/10245648/127935204-8543b8a3-b334-4724-9420-9397be4a0a3b.png)

https://github.com/fairlearn/fairlearn/blob/main/CITATION.cff

hildeweerts:
I've been playing around a bit with this in a test repo: https://github.com/hildeweerts/testcitation/blob/main/CITATION.cff.

There was a minor issue with missing quotation marks in the article title. But the main issue with the current file seems to be that it is required to have authors and a version. It turns out that it doesn't actually allow you to directly cite the references; only the metadata provided in the 'main' part of the citation file:

<img width="340" alt="image" src="https://user-images.githubusercontent.com/24417440/127983618-36afebc3-426a-4411-81b5-13689ae7df3c.png">

Despite requiring the 'version', the rendering is a bit inconsistent; in APA you get the version, but in BibTex not. This is not entirely unsurprising as `misc` doesn't have a `version` field, but even manually adding a `note="version 0.7.0"` field doesn't show in the BibTex on GitHub:

APA: 
````
Lisa M., Bot H. (2021). Fairlearn (version 0.7.0).
````

BibTex: 
```@misc{Lisa_Fairlearn_2021,
author = {Lisa, Mona and Bot, Hew},
month = {7},
title = {Fairlearn},
url = {https://github.com/fairlearn/fairlearn},
year = {2021}
}
```

Although I'm not a fan of the inconsistency, I do think it makes some sense to cite the package rather than the paper, given that the current version of Fairlearn is substantially different from the version for which the white paper was written. But I'm not sure who should be included as 'authors' of the package in that case.

hildeweerts:
Although I really like your idea in terms of inclusivity, intuitively it doesn't feel right to me that something like a typo fix would get equal attribution to somebody who wrote an entire section in the user guide. But like you already said, it's very tricky to decide "what's enough". We could be very rigorous and only add maintainers plus maybe a few other people. On a more practical note, many people are very sloppy with citations and will simply cite whatever they find first, so I doubt that they will carefully select the one for the version they are using, resulting in messy attribution after all...

[...]

PR #979, opened 2021-10-13
MNT Remove broken CITATION.cff file

romanlutz:
As discussed in #924 with @hildeweerts we agree that just citing the existing paper leaves a lot to be desired (including the fact that it excludes lots of people who contributed after we changed the governance). The current file doesn't even render properly because the formatting is broken. For that reason, I'm removing it until we find a solution we're happier with. Let's leave #924 open as a place to discuss options.


romanlutz:
It is on the about page, but at least that one isn't broken. This one is.

PR #980, opened 2021-10-13
MNT deprecate python 3.6

### LINKED ISSUE #966: Remove python 3.6 support
In #963 I discovered that scikit-learn isn't supporting py3.6 anymore. This issue is there to discuss dropping this in Fairlearn, too. Additionally, as @adrinjalali mentions in that issue:

> re: py36 support, that one's long gone, and even py37 support is already removed from the main branches of numpy and scipy AFAIK, and we'll be removing it too for the next release.

So should we also make these changes? I'm definitely in favor. (If you're also in favor and have nothing to add, simply upvote. If you're against or you have something to add please comment!)

Indeed, numpy dropped 3.7 already https://github.com/numpy/numpy/blob/main/setup.py
The full deprecation schedule is at https://numpy.org/neps/nep-0029-deprecation_policy.html and shows that 3.7 support is dropped in December 2021. I guess that means that they won't fix issues anymore, but dropping it from newer versions happens even before that (?)

romanlutz:
Closes #966

~~#265 was about a failing test due to non-deterministic behavior of a solver which we won't need anymore once python 3.6 is deprecated. In newer versions this behaves deterministically.~~ Update: still not resolved in 3.7, so this will stay.

### DIFF
--- a/fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
+++ b/fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
@@ -22,6 +22,12 @@ class ExponentiatedGradient(BaseEstimator, MetaEstimatorMixin):
     The exponentiated gradient algorithm is described in detail by
     `Agarwal et al. (2018) <https://arxiv.org/abs/1803.02453>`_.
 
+    .. versionchanged:: 0.3.0
+        Was a function before, not a class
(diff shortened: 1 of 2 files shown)


adrinjalali:
I also don't understand, we actually do these changes in the middle of two releases in scikit-learn.

PR #981, opened 2021-10-13
MNT add support for python 3.10 and remove 3.7

romanlutz:
Trying this out with this PR. If there are issues in the CI I'll address them accordingly.


romanlutz:
Update: spicy, numpy, matplotlib, and pandas all support 3.10 now. Scikit-learn is still on 3.9, probably until the next release (see [this comment](https://github.com/scikit-learn/scikit-learn/issues/21401#issuecomment-975524954)), so once that's there I can retry.

docs/user_guide/installation_and_version_guide/v0.7.1.rst
```diff
@@ -8,7 +8,7 @@ v0.7.1
 
 * Relaxed checks made on :code:`X` in :code:`_validate_and_reformat_input()`
   since that is the concern of the underlying estimator and not Fairlearn.
-* Add support for Python 3.9
+* Add support for Python 3.9 and 3.10, deprecate support for Python 3.6
```
adrinjalali, line 11: should go to version 0.8, and "remove" instead of "deprecate".

adrinjalali (commented):
What's left here @romanlutz ?

adrinjalali:
There has been support for 3.10 from sklearn for a while now ;)

romanlutz:
I'll try to update this and get it ready to merge!

docs/user_guide/installation_and_version_guide/v0.7.1.rst
romanlutz: Will do!

romanlutz:
> There has been support for 3.10 from sklearn for a while now ;)

https://pypi.org/project/scikit-learn/ still shows 3.9 as the highest version:
image

adrinjalali:
> https://pypi.org/project/scikit-learn/ still shows 3.9 as the highest version:

That's a lie 😁 you have the 310 wheels here: https://pypi.org/project/scikit-learn/#files
We need to fix the `setup.py` metadata on sklearn side, thanks for pointing it out.

romanlutz:
Thanks @adrinjalali!

For some reason the unit test pipeline seems to think it should use Python 3.1 rather than 3.10. I checked all occurrences of versions in the templates and we definitely don't specify 3.1 anywhere. I'm a bit stumped... any ideas? @riedgar-ms maybe?

image

riedgar-ms:
@romanlutz perhaps put `3.10` in quotes? It looks like it's being interpreted as a float, not a string?

romanlutz:
> @romanlutz perhaps put `3.10` in quotes? It looks like it's being interpreted as a float, not a string?

🤣 OMG I can't believe this... Thanks I'll try that.

adrinjalali:
yep, numbers are interpreted as numbers unless put in quotation marks 😁

romanlutz:
It worked this time! @adrinjalali and @riedgar-ms if you feel like taking a quick look at the changes that would be great! I expect the pipelines will succeed in the next 30-60 minutes so feel free to merge if there's already another approval 🙂 Thanks!

devops/pypi-release-new.yml
```diff
@@ -46,7 +46,7 @@ stages:
     parameters:
       platforms:  { Linux: ubuntu-latest, Windows: windows-latest }
       installationType: 'PipLocal'
-      pyVersions: [3.7, 3.8, 3.9]
+      pyVersions: [3.7, 3.8, 3.9, "3.10"]
```
riedgar-ms, line 49: Nitpick.... quote all of these? And/or perhaps add a comment about `# 3.10 not 3.1` ?

devops/templates/test-run-step-template.yml
```diff
@@ -17,8 +17,8 @@ parameters:
 - name: installationType
   type: string
 - name: pythonVersion
-  type: number
```
riedgar-ms, line 20: You can probably blame me, then....

riedgar-ms (approved):
You might want to change all the Python version to strings, but that's not a huge issue.

devops/templates/test-run-step-template.yml
```diff
@@ -17,8 +17,8 @@ parameters:
...
-  default: 3.7
+  type: string
+  default: "3.7"
```
adrinjalali, line 21: 3.9 as default maybe? or even 3.10?

adrinjalali (approved):
LGTM. Note that 3.7 support is removed in our dependencies already.

devops/templates/test-run-step-template.yml
romanlutz: Good point. Will adjust this and check for others!

romanlutz: #966 is the only issue where this was discussed to my knowledge. You made the IMO correct point that we should remove 3.7 if the release is not in 2021 (which it isn't), so I'll remove all occurrences of 3.7 (and in pipelines increase to something higher).

romanlutz: @fairlearn/fairlearn-maintainers FYI

romanlutz: Would love to kick off the nightly builds to validate that it's working. [On that note, the "fixed" one is currently broken it seems.]

adrinjalali:
I'm not sure, but it seems it's coming from pytorch and not us. For that we can disable the error by ignoring certain deprecation warnings. In sklearn for instance we do `-Wignore:tostring:DeprecationWarning` since that's coming from our dependencies and not us.

adrinjalali (approved):

Issue #948, opened 2021-09-13
DOC Add versionadded and versionchanged to API reference (fairlearn.reductions)


bramreinders97:
Add versionadded and versionchanged to API reference of fairlearn.reductions. Sub-issue of #855.

bramreinders97:
Hi, after diving in to all the changes made in different versions, I am wondering what level of detail is expected in the versionchanged part of this issue.

To give an example of my doubts, consider the changed mentioned in the version guide related to ExponentGradient: 

v0.4.5:  
![image](https://user-images.githubusercontent.com/50204600/133609095-df9a7f0f-e7d1-4eb6-ba61-c49a2bb2215c.png)

v0.4.6:  
![image](https://user-images.githubusercontent.com/50204600/133609142-ad3b2a04-7852-4dd8-9a77-d753ae33950b.png)
v0.5.0:     
![image](https://user-images.githubusercontent.com/50204600/133609178-0bd9e991-73a5-434c-9ebe-747320404163.png)
![image](https://user-images.githubusercontent.com/50204600/133609187-2680bd01-f03b-4180-a2a2-701b5368031c.png)
![image](https://user-images.githubusercontent.com/50204600/133609201-300d95e8-b57a-4263-9daf-27ae3d152753.png)
![image](https://user-images.githubusercontent.com/50204600/133609211-d8bba913-7c36-4724-83da-28f18bec86cf.png)


So my question is, should all these changes be included, and if not, how do I know which to include?

romanlutz:
That's a great question! I'm actually not entirely sure. @fairlearn/fairlearn-maintainers wdyt? Have you tried out whether it's technically possible to add more than one `versionchanged` tag? If it's not possible then I suppose it has to be the changes in the latest version in which it was changed.

Whenever it is an individual argument you can add it just for that one, e.g., `sample_weight_name` was added in v0.5.0. `max_iter`, `eta0`, and `run_linprog_step` were also added in v0.5.0.

bramreinders97:
Turns out it is possible to include multiple `versionchanged` tags. So I'll try to add all mentioned changes.

adrinjalali:
I usually apply the tags only when a parameter changes behavior, or a major change in behavior is happening in a class. It's not supposed to mirror the changelog in any way IMO.

romanlutz:
I agree. We've had some changes in the datasets module which @adrinjalali and I recommended omitting from `versionchanged` since only small changes or doc changes happened.

[...]

PR #983, opened 2021-10-17
DOC Add versionadded and versionchanged to API reference (fairlearn.reductions._moments) 

bramreinders97:
Related to issue #948.

I added versionadd and versionchanged tags for moments. When checking all my changes on the site, I noticed that `ConditionalLossMoment` and `MeanLoss` do not show up on the docs page. I guess this is because these classes on their own are not very relevant for the user? 

However, especially for `ConditionalLossMoment`, I think some useful information might be lost? To illustrate: in bounded_group_loss.py the following comments are present: 

image

, but this info is nowere on the fairlearn.reductions docs page, so a user will never see the explanation of `upper_bound` for example. Is this supposed to be the case or is this a mistake?

### DIFF
--- a/docs/user_guide/installation_and_version_guide/v0.5.0.rst
+++ b/docs/user_guide/installation_and_version_guide/v0.5.0.rst
@@ -56,7 +56,7 @@ v0.5.0
     For a comprehensive overview of available constraints refer to the new
     :ref:`user guide on fairness constraints for reductions methods <reductions>`.
 
-* Added :code:`True` to provide the
+* Added :code:`TrueNegativeRateParity` to provide the
   opposite constraint of :class:`~fairlearn.reductions.TruePositiveRateParity`
   to be used with reductions techniques.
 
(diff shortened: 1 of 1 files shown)


romanlutz:
I'll briefly summarize what I understand from the discussion here. @bramreinders97 observed that significant parts of the documentation are only on the parent classes `ConditionalLossMoment` and `MeanLoss`. The child classes that users actually work with don't duplicate most of this important documentation. Since only the child classes are exported in the module (see [here](https://github.com/fairlearn/fairlearn/blob/5d14271343951156c3b98a516da4e089d655bd59/fairlearn/reductions/__init__.py#L21)) the documentation isn't presented on the webpage's API reference. @adrinjalali suggested exporting those classes (`ConditionalLossMoment` and `MeanLoss`) as well.

The only parts I'm not clear on and I think @MiroDudik may want to chime in here: Both `ConditionalLossMoment` and `MeanLoss` are meant as abstract parent classes AFAIK, so should we do anything to prevent users from inadvertently using them directly rather than their child classes?

Other classes further up the hierarchy, such as `ClassificationMoment` or `Moment` are also shown in the API reference, so I think it should apply to these two, too, and also to `LossMoment`.
@fairlearn/fairlearn-maintainers wdyt?

hildeweerts:
I agree with @romanlutz and @adrinjalali and would love to hear @MiroDudik's thoughts as well.

adrinjalali:
If we're just worried about the docs, I'm used to having the docs copy/pasted to all child classes. Not ideal, but it works. And the shared long documentation goes into user guide instead of API docs.

PR #985, opened 2021-10-22
Improve MetricFrame performance

riedgar-ms:
Rework the innards of `MetricFrame` to use `DataFrame.groupby()` to improve performance. The public API does not change, although there is a slight change to the return - the resultant `DataFrame`s will have `dtype=float64` etc. when appropriate, rather than just being `object`.

The `FunctionContainer` class has been replaced by `AnnotatedMetricFunction`.  In addition, the `GroupFeature` class has been simplified, removing its dependence on `LabelEncoder`. These changes have obsoleted quite a few tests.


riedgar-ms:
@adrinjalali could you expand on your concerns about using `DataFrame`? Also, is there some fast way of doing the masking with numpy, which can match `DataFrame.groupby()`? That's the performance bottleneck.

adrinjalali:
I don't think we have to use `pd.DataFrame` to improve the performance here. Based on the answers [here](https://stackoverflow.com/questions/8623047/group-by-max-or-min-in-a-numpy-array), you can sort the array first, and then do your aggregations on the parts of data you need. Doing min/max groupby I see them being quite fast. It may not be as fast as pandas, but it's much faster than what we have I think.

For instance, to get `min`, you could do

``` python
def group_min(groups, data):
    # sort with major key groups, minor key data
    order = np.lexsort((data, groups))
    groups = groups[order] # this is only needed if groups is unsorted
    data = data[order]
    # construct an index which marks borders between groups
    index = np.empty(len(groups), 'bool')
    index[0] = True
    index[1:] = groups[1:] != groups[:-1]
    return data[index]
```

You could do a similar thing for other aggregations.

I personally don't have the data in a pandas DataFrame to feed to `MetricFrame`, with this change, I'd need to first construct a dataframe, just to feed it to the `MetricFrame`.

riedgar-ms:
@adrinjalali , this current change would not require the input be in `DataFrame` format - although it would narrow the inputs deemed suitable to 'must be assignable to `DataFrame` column' - this precludes passing a 2D array for `y_true`.

riedgar-ms:
Just noting that @adrinjalali also commented on the first pass at this:
https://github.com/fairlearn/fairlearn/pull/975#discussion_r745923571
For my part, I think that we do want to support metrics which don't take `y_true` or `y_pred`. Even just simple like `selection_rate` doesn't want both, so you wind up writing functions which completely ignore one of their arguments.

riedgar-ms:
@adrinjalali and @MiroDudik , I've adjusted the logic to 'crack' a 2D ndarray across multiple columns of a `DataFrame`. All the tests now pass (although I have had to change one test from `np.array_equal` to `np.allclose`)

fairlearn/metrics/_annotated_metric_function.py
```diff
@@ -0,0 +1,50 @@
+# Copyright (c) Microsoft Corporation and Fairlearn contributors.
```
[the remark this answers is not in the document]
adrinjalali: The new policy is that for new files, we don't need microsoft there, but for legal reasons, we are not removing microsoft from the existing ones.

fairlearn/metrics/_annotated_metric_function.py
```diff
@@ -0,0 +1,60 @@
...
+            self._args = args
+        self._kwargs = dict()
+        if kwargs is not None:
+            self._kwargs = kwargs
+
+    @property
+    def func(self) -> Callable:
+        """Return the wrapped function."""
+        return self._func
+
```
adrinjalali, line 29: could we not have properties for these and just store them in public attributes instead of private ones? It makes the code much smaller and simpler.

fairlearn/metrics/_annotated_metric_function.py
```diff
@@ -0,0 +1,60 @@
...
+    def args(self) -> List[str]:
+        """Return the list of positional arguments."""
+        return self._args
+
+    @property
+    def kwargs(self) -> Dict[str, str]:
+        """Return the mapping from column names to kwarg names."""
+        return self._kwargs
+
+    def invoke(self, df: pd.DataFrame, split_columns: Dict[str, List[str]]):
```
adrinjalali, line 40: do you want to have `invoke`, or make them a callable with `__call__` and have them be called as usual?

fairlearn/metrics/_metric_frame.py
```diff
@@ -39,6 +41,18 @@
     "Please remove non-scalar columns from your metric frame or use parameter errors='coerce'."
 
 
+def apply_to_dataframe(
```
adrinjalali, line 44: please add complete docstrings to methods and functions, makes the review much easier, since then I'd know what the method/function's supposed to do, and I can check the implementation against that :)

adrinjalali (commented):
Thanks @riedgar-ms

fairlearn/metrics/_annotated_metric_function.py
riedgar-ms: _sobs_
Removed.

fairlearn/metrics/_annotated_metric_function.py
riedgar-ms: On the whole, I prefer an `invoke()` method, although I'm not going to die on that particular hill.

fairlearn/metrics/_metric_frame.py
riedgar-ms: I've been adding some. The documentation is far from final though, since the code isn't finalised yet....

fairlearn/metrics/_annotated_metric_function.py
adrinjalali: The reason I say that, is that in the place where we check what metric needs what, we can simply check

``` python
params = defaults
if hasattr(metric, "kwargs"):
    # extend params
value = metric(**params)
```
we don't need to have a separate syntax for callables and these objects there then.

adrinjalali:
It'd be nice to apply `black` to the files you're touching, then you won't need to deal with flake8 anymore :)

riedgar-ms:
@adrinjalali 

> It'd be nice to apply `black` to the files you're touching, then you won't need to deal with flake8 anymore :)

I've just done that to a bunch of files.... I'm a little reluctance to run `black` on `MetricFrame` itself, since that could make it harder to see what's changed. I'd prefer to do that in a separate PR. Or at least wait for signoffs on this one, and do it as a final commit prior to merge.

riedgar-ms:
OK, I've done some tinkering with `GroupFeature`, although not to the point of removing it.

I'm not a huge fan of the `_all_data` and `_split_columns` private members. Together with the `_accrete_array()` method, they make unit testing things rather hard (compounded by how much work `MetricFrame` does in its constructor). However, the alternative would be introducing a new class to hold the two of them.... and I'm not sure that would be viewed as better.

Thoughts @adrinjalali @MiroDudik @romanlutz @hildeweerts ?

fairlearn/metrics/_annotated_metric_function.py
```diff
@@ -0,0 +1,92 @@
...
+logger = logging.getLogger(__name__)
+
+_DEFAULT_NAME = "metric"
+
+_METRIC_FUNCTION_NONE = "Found 'None' instead of metric function"
+_METRIC_FUNCTION_NOT_CALLABLE = "Object passed as metric function not callable"
+
+
+class AnnotatedMetricFunction:
+    """Wrapper for functions, to give them args and kwargs."""
```
MiroDudik, line 18: This doesn't seem quite the right description. My sense is that the goal of this wrapper is so that it can be applied to a `DataFrame`. So, if anything, the class should be probably called `DataFrameMetric` or something like that?

fairlearn/metrics/_metric_frame.py
```diff
@@ -783,20 +845,25 @@ def _process_features(self, base_name, features, sample_array) -> List[GroupFeat
 
         return result
 
-    def _mask_from_tuple(self, index_tuple, feature_list) -> np.ndarray:
-        """Generate a mask for the ``y_true``, ``y_pred`` and ``sample_params`` arrays.
+    def _accrete_array(
```
MiroDudik, line 848: Mhmm... this seems to be a complicated mechanism for this. I had in mind more something along the following lines (i.e., splitting into rows rather than columns):

```python
y_true=np.array([[0, 1], [1, 0], [0, 1]])
y_pred=np.array([[0, 1], [0, 1], [1, 0]])
sf = ['a','a','b']

df = pd.DataFrame({"sf": sf, "y_true": list(y_true), "y_pred": list(y_pred)})
df_accuracy_score = lambda df: skm.accuracy_score(list(df['y_true']), list(df['y_pred']))
by_group = df.groupby('sf').apply(df_metric)
```

fairlearn/metrics/_annotated_metric_function.py
```diff
@@ -0,0 +1,92 @@
...
+        positionally, while those
+        in :code:`self.kw_argument_mapping` are supplied as keyword arguments.
+
+        There are two subtleties.
+        Firstly, the keyword arguments expected by the function might not
+        have the same names as the columns in the supplied DataFrame.
+        This is the reason :code:`self.kw_argument_mapping` is a dictionary, rather than
+        just a list of column names.
+
+        The second issue is coping with when users have passed in a 2D array as
```
MiroDudik, line 65: I think we should be able to avoid this subtlety (see my comment below re. `_accrete_array`).

MiroDudik (commented):
I was hoping that we'd be able to reduce the complexity / number of lines in MetricFrame more, but for the time being I am fine moving forward with the current implementation.

I've left a couple of questions / comments for your consideration. None of them are must-haves though.

fairlearn/metrics/_metric_frame.py
riedgar-ms: I've incorporated this, since it slightly simplifies the handling of multi-dimensional columns. I just hope that the extra type change doesn't have some other subtle side effect :-/

fairlearn/metrics/_annotated_metric_function.py
riedgar-ms: I'm a little concerned about all the type changes, but this does allow me to ditch `_accrete_array()` which is a plus.

fairlearn/metrics/_annotated_metric_function.py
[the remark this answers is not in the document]
riedgar-ms: `MetricForDataFrame` or `FunctionForDataFrame` ?

test/unit/postprocessing/test_threshold_optimizer_multiple_sensitive_features.py
```diff
@@ -19,7 +19,7 @@
     "false_positive_rate": false_positive_rate,
     "positive_count": lambda true, pred: np.sum(true),
     "false_negative_rate": false_negative_rate,
-    "negative_count": lambda true, pred: np.sum(1-true),
+    "negative_count": lambda true, pred: np.sum(1-np.asarray(true)),
```
riedgar-ms, line 22: So.... we have breaking changes @MiroDudik @adrinjalali @romanlutz

fairlearn/metrics/_group_feature.py
```diff
@@ -65,27 +63,22 @@ def __init__(self,
...
-        return self.get_mask_for_class_index(idx)
-
-    def get_mask_for_class_index(self,
-                                 target_class_index: int) -> np.ndarray:
-        """Fetch a mask array for the given class index."""
-        return (self._encoded == target_class_index)
-
     @property
-    def name(self) -> str:
+    def name_(self) -> str:
```
adrinjalali, line 72: why not store the values in these attributes and then remove the 3 properties?

adrinjalali (commented):
Thanks for the work @riedgar-ms .

I'm struggling to find maybe a place in the user guide to see the changes that affect the users' experience, which would also be related to your question on the breaking changes.

Would you mind telling me where I should look to see those changes?

fairlearn/metrics/_group_feature.py
riedgar-ms: I can do that - I thought this was the SciKit-Learn convention, though? Or have I misunderstood it....

adrinjalali: our convention there is not to have many properties. If the user is to have access to a `name_` attribute, it's stored under `obj.name_` w/o any property to hide the _private_ attribute. We just store it publicly in the first place.

riedgar-ms: OK. Changed to match this

riedgar-ms:
@adrinjalali , I think I've managed to work around the breaking change (it wasn't documented yet, since I didn't want to have more stuff to undo; I should perhaps change that policy). It does involve some typewashing, though.

test/unit/postprocessing/test_threshold_optimizer_multiple_sensitive_features.py
riedgar-ms: I have a workaround....

riedgar-ms:
@fairlearn/fairlearn-maintainers green builds at last....

riedgar-ms:
So some performance test code, as requested by @adrinjalali :

```python

import numpy as np
import pandas as pd
import sklearn.metrics as skm
import time

import fairlearn.metrics as metrics


def group_gen(x, runs, groups):
    # For producing arrays such as ['a', 'a', 'b', 'b', 'a', 'a', ... ]
    return groups[(x // runs) % len(groups)]


def array_gen(groups, run_length, num_samples):
    vals = [group_gen(x, run_length, groups) for x in range(num_samples)]
    return np.asarray(vals)


def run_one_test(num_samples: int, rand_seed: int) -> float:
    rng = np.random.default_rng(seed=rand_seed)

    metric_dict = {
        "recall": skm.recall_score,
        "prec": skm.precision_score,
        "acc": skm.accuracy_score,
    }

    # Generate 'true' and 'predicted' values
    y_true = rng.integers(low=0, high=1, endpoint=True, size=num_samples)
    y_pred = rng.integers(low=0, high=1, endpoint=True, size=num_samples)

    # Generate the conditional and sensitive features
    # Note that we have quite a few prime numbers here
    cf_1_groups = ["aa", "aaa", "aaaa"]
    cf_2_groups = ["p", "qq"]
    sf_1_groups = ["A", "B", "C", "D", "E"]
    sf_2_groups = ["H", "L", "M", "P", "Q", "R", "S"]
    cf_1_arr = array_gen(cf_1_groups, 11, num_samples)
    cf_2_arr = array_gen(cf_2_groups, 13, num_samples)
    sf_1_arr = array_gen(sf_1_groups, 17, num_samples)
    sf_2_arr = array_gen(sf_2_groups, 19, num_samples)

    # Sanity check that all groups appear
    assert len(np.unique(cf_1_arr)) == len(cf_1_groups)
    assert len(np.unique(sf_1_arr)) == len(sf_1_groups)
    assert len(np.unique(cf_2_arr)) == len(cf_2_groups)
    assert len(np.unique(sf_2_arr)) == len(sf_2_groups)

    # Make some inputs
    sf_frame = pd.DataFrame(
        np.stack([sf_1_arr, sf_2_arr], axis=1), columns=["SF0", "SF1"]
    )
    cf_array = np.stack([cf_1_arr, cf_2_arr], axis=1)

    # Create the target object
    start = time.time()
    target = metrics.MetricFrame(
        metrics=metric_dict,
        y_true=y_true,
        y_pred=y_pred,
        sensitive_features=sf_frame,
        control_features=cf_array,
    )
    end = time.time()
    elapsed = end - start
    print(f"MetricFrame evaluation took {elapsed} secs")
    return elapsed


print("Running MetricFrame performance test")

num_tests = 20
num_elements = 100000

warm_time = run_one_test(int(num_elements / 10), int(time.time()))
print(f"Warmup complete in {warm_time} secs")

run_times = []
for _ in range(num_tests):
    nxt_time = run_one_test(num_elements, int(time.time()))
    run_times.append(nxt_time)

print("Tests complete")
mean_time = np.mean(run_times)
std_dev = np.std(run_times)
print(f"Time = {mean_time}+/-{std_dev} secs")
```
I can check this in, if it would be helpful to keep around. Problem is a performance test isn't really pass/fail.

riedgar-ms:
Running on Python 3.7. First installing `fairlearn` from PyPI:
```
Time = 163.8702350616455+/-13.086753780340425 secs
```
and with the new improved innards:
```
Time = 1.2914242148399353+/-0.025541574801125556 secs
```

riedgar-ms:
@adrinjalali , I've gone back and improved some of the documentation. However, the changes are smaller now that the user-facing API is being preserved.

fairlearn/metrics/_metric_frame.py
```diff
@@ -272,26 +290,39 @@ def __init__(self,
                  y_true,
                  y_pred,
                  sensitive_features,
-                 control_features: Optional = None,
+                 control_features=None,
                  sample_params: Optional[Union[Dict[str, Any], Dict[str, Dict[str, Any]]]] = None):
         """Read a placeholder comment."""
         check_consistent_length(y_true, y_pred)
+
+        all_data = pd.DataFrame(dtype='object')
```
adrinjalali, line 298: why is this needed?

fairlearn/metrics/_metric_frame.py
```diff
@@ -272,26 +290,39 @@ def __init__(self,
...
+
         y_t = _convert_to_ndarray_and_squeeze(y_true)
         y_p = _convert_to_ndarray_and_squeeze(y_pred)
 
-        func_dict = self._process_functions(metrics, sample_params)
+        all_data['y_true'] = list(y_t)
```
adrinjalali, line 303: why do we convert things to an ndarray and then a list here?

fairlearn/metrics/_metric_frame.py
```diff
@@ -303,53 +334,39 @@ def __init__(self,
...
             # Prepend the conditional features, so they are 'higher'
-            rows = copy.deepcopy(cf_list) + rows
+            grouping_features = copy.deepcopy(cf_list) + grouping_features
 
-        return self._compute_dataframe_from_rows(func_dict, y_true, y_pred, rows)
+        temp = all_data.groupby([x.name_ for x in grouping_features]).apply(
+            apply_to_dataframe,
+            metric_functions=annotated_funcs)
+        if len(grouping_features) > 1:
+            all_indices = pd.MultiIndex.from_product([x.classes_ for x in grouping_features],
```
adrinjalali, line 364: I wouldn't mind if you apply `black` on this file :D

fairlearn/metrics/_annotated_metric_function.py
```diff
@@ -0,0 +1,107 @@
...
+            raise ValueError(_METRIC_FUNCTION_NONE)
+        if not callable(func):
+            raise ValueError(_METRIC_FUNCTION_NOT_CALLABLE)
+        self.func = func
+
+        if name is None:
+            if hasattr(func, "__name__"):
+                self.name = func.__name__
+            else:
+                logger.warning("Supplied 'func' had no __name__ attribute")
```
adrinjalali, line 63: this is not tested (hard to find what's not tested, this is why I love codecov)

test/unit/metrics/test_metricframe_comprehensive.py
```diff
@@ -54,9 +55,13 @@ def test_3m_2sf_2cf():
     cf_array = np.stack([cf_1_arr, cf_2_arr], axis=1)
 
     # Create the target object
+    start = time.time()
     target = metrics.MetricFrame(metrics=metric_dict, y_true=y_true, y_pred=y_pred,
                                  sensitive_features=sf_frame,
                                  control_features=cf_array)
+    end = time.time()
+    elapsed = end - start
+    print(f"MetricFrame evaluation took {elapsed} secs")
```
adrinjalali, line 64: why are we printing in a test?

adrinjalali (commented):
Otherwise I think it looks quite good to me!

fairlearn/metrics/_metric_frame.py
riedgar-ms: It was to make the object exist as a `DataFrame`. I have changed how it's constructed.

fairlearn/metrics/_metric_frame.py
riedgar-ms: Done

test/unit/metrics/test_metricframe_comprehensive.py
riedgar-ms: My mistake. This was leftover from when I was using this test for performance testing. Have removed

fairlearn/metrics/_annotated_metric_function.py
```diff
@@ -0,0 +1,107 @@
...
+
+        self.func = func
+        self.postional_argument_names = ["y_true", "y_pred"]
+        if postional_argument_names is not None:
+            self.postional_argument_names = postional_argument_names
+        self.kw_argument_mapping = dict()
+        if kw_argument_mapping is not None:
+            self.kw_argument_mapping = kw_argument_mapping
+
+    def invoke(self, df: pd.DataFrame):
```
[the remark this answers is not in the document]
riedgar-ms: Changed.

fairlearn/metrics/_metric_frame.py
riedgar-ms: Because if you try assigning a 2D array to a single pandas column, it complains. However, it doesn't object to a list-of-lists.

fairlearn/metrics/_annotated_metric_function.py
riedgar-ms: Do you mean that I don't check for the log message? This branch of the `if` statement is covered by `test_constructor_no_name()`

fairlearn/metrics/_annotated_metric_function.py
```diff
@@ -0,0 +1,107 @@
...
+
+        The second issue is coping with when users have passed in a 2D array as
+        a named argument (especially, `y_true` or `y_pred`).
+        For this reason, we perform some extra list-washing, to make sure the
+        expected types are passed to the underlying metric function.
+        """
+        args = []
+        for arg_name in self.postional_argument_names:
+            # Need to convert to list first in case we have 2D arrays
+            args.append(np.asarray(list(df[arg_name])))
```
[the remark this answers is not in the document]
riedgar-ms: This is the complement to your query below. In order to get multi-dimensional arrays into a single DataFrame column, I convert to a nested list. So in that case `df[arg_name]` will return a series of lists, the 'list' makes that into a list-of-lists which `np.asarray()` converts to a proper multi-D array.

fairlearn/metrics/_annotated_metric_function.py
riedgar-ms: Ping @adrinjalali I'm not sure which part you mean isn't tested.

riedgar-ms:
@adrinjalali please could you take another look?

riedgar-ms:
Ping.... @adrinjalali , have I adequately addressed your concerns? I would really like to get this checked in.

fairlearn/metrics/_annotated_metric_function.py
adrinjalali: Yes, the test doesn't check that this warning is raised (I guess that means the test actually raises a warning, which it ideally shouldn't).

adrinjalali (commented):
Otherwise this looks good to me. I think it's a positive change. I'm still not super confident with the code since I find it hard to follow in some parts, but this is making things much cleaner than before.

riedgar-ms:
@adrinjalali I've adjusted the test so that the warning is checked (although I had to go through `warnings.warn()` rather than `logging`). Is there somewhere in particular you feel it would be helpful to have more documentation?

test/unit/metrics/test_annotated_metric_function.py
```diff
@@ -15,7 +16,7 @@ def test_constructor_unnamed():
     assert len(fc.kw_argument_mapping) == 0
 
 
-def test_constructor_no_name():
+def test_constructor_no_name(recwarn):
```
adrinjalali, line 18: I didn't know about `recwarn`, it's neat.

adrinjalali (approved):
I'm happy to have this merged as is, and do improvements in future PRs.

test/unit/metrics/test_annotated_metric_function.py
riedgar-ms: Neither did I until yesterday :-)

MiroDudik (approved):

Issue #958, opened 2021-09-20
ENH Add Reject Object Classification and Discrimination-Aware Ensemble based on Kamiran,Karim,Zhang


bramreinders97:
The paper [Decision Theory for Discrimination-Aware Classification](https://ieeexplore.ieee.org/document/6413831) by Kamiran, Karim and Zhang presents two solutions for discrimination-aware classification that neither require data modification nor classifier tweaking; Reject Option based Classification (ROC) and Discrimination-Aware Ensemble (DAE) . The goal of this task is to implement them. 

Completing this item requires:

- code for the technique in ?? (@fairlearn/fairlearn-maintainers I'm not entirely sure where this should go?)
- unit tests in `test.unit.??`
- descriptive API reference (directly in the docstring)
- ideally a short user guide in docs.user_guide.mitigation.rst

A fully fledged example notebook is not required.

PS. I think I covered all the stuff I'm required to put in here, if not please let me know :).

bramreinders97:
For some reason my @fairlearn/fairlearn-maintainers tag did not work, so I'll do it like this:  @adrinjalali  @hildeweerts  @MiroDudik @mmadaio @riedgar-ms @romanlutz

bramreinders97:
I would like to work on this issue :)

hildeweerts:
Thank you for opening this issue @bramreinders97! I personally think this technique would be a good addition to Fairlearn (but I am 100% biased because I would like to get rid of AIF360 in my teaching). I think it would be sufficient to stick to Reject Option based Classification for now.

@bramreinders97: could you perhaps write a short summary of the technique such that @fairlearn/fairlearn-maintainers can get an idea of how the method works?

romanlutz:
I have no objection. But let's not use ROC as an acronym here. It will 100% get confused with the more commonly used receiver operating characteristic 🙂

hildeweerts:
100% agree re. naming, @romanlutz!

bramreinders97:
> @bramreinders97: could you perhaps write a short summary of the technique such that @fairlearn/fairlearn-maintainers can get an idea of how the method works?
>

Reject Option Classification is a technique based on the intuitive hypothesis that discriminatory decisions are often made close to the decision boundary because of the decision maker’s bias. 

To tackle this problem, the authors focus on a critical region around the decision boundary. This critical region contains all instances for which: 

[...]

PR #997, opened 2021-11-25
Adding Thresholder and RejectOptionClassifier 

bramreinders97:
For issue #958 

So far I have the first version for `Thresholder`, and the first version of a unit test for `Thresholder`. 

To those files, I for sure want to add: 
- A test for prefit = True, and some more options for `predict_method`
- A way to deal with multiple sensitive features, when we have decided on how to do so (see discussion in issue)

Furthermore, I will start working on RejectOptionClassifier when everything for `Thresholder` is done.

### DIFF
--- a/fairlearn/postprocessing/__init__.py
+++ b/fairlearn/postprocessing/__init__.py
@@ -7,10 +7,12 @@ The predictor's output is adjusted to fulfill specified parity constraints. The 
 learn how to adjust the predictor's output from the training data.
 """
 
+from ._thresholder import Thresholder
 from ._threshold_optimizer import ThresholdOptimizer  # noqa: F401
 from ._plotting import plot_threshold_optimizer  # noqa: F401
 
 __all__ = [
+    "Thresholder"
     "ThresholdOptimizer",
-    "plot_threshold_optimizer"
+    "plot_threshold_optimizer",
 ]
--- a/fairlearn/postprocessing/_thresholder.py
+++ b/fairlearn/postprocessing/_thresholder.py
@@ -1,0 +1,105 @@
+# Copyright (c) Microsoft Corporation and Fairlearn contributors.
+# Licensed under the MIT License.
+
+import numpy as np
+from sklearn import clone
+from sklearn.base import BaseEstimator, MetaEstimatorMixin
+from sklearn.exceptions import NotFittedError
+from sklearn.utils.validation import check_is_fitted
+from warnings import warn
+
+
+from postprocessing._threshold_operation import ThresholdOperation
+
+# try to get it to work now
+from utils._common import _get_soft_predictions
+from utils._input_validation import _validate_and_reformat_input
+from postprocessing._constants import (
+    BASE_ESTIMATOR_NONE_ERROR_MESSAGE,
+    BASE_ESTIMATOR_NOT_FITTED_WARNING)
+
+
+# how it should be
+# from ..utils._common import _get_soft_predictions
+# from ..utils._input_validation import _validate_and_reformat_input
+# from ._constants import (
+#     BASE_ESTIMATOR_NONE_ERROR_MESSAGE,
+#     BASE_ESTIMATOR_NOT_FITTED_WARNING)
+
+
+class Thresholder(BaseEstimator, MetaEstimatorMixin):
+    r"""Create my own description here
+
+    Parameters
+
+    References
+    ----------
+
+    """
+
+    def __init__(self, estimator, threshold_dict, prefit=False,
+                 predict_method='deprecated'):
+        self.estimator = estimator
+        self.threshold_dict = threshold_dict
+        self.prefit = prefit
+        self.predict_method = predict_method
+
+    def fit(self, X, y, **kwargs):
+        """Fit the estimator.
+
+        If `prefit` is set to `True` then the base estimator is kept as is.
+        Otherwise it is fitted from the provided arguments.
+        """
+        if self.estimator is None:
+            raise ValueError(BASE_ESTIMATOR_NONE_ERROR_MESSAGE)
+
+        if self.predict_method == "deprecated":
+            warn(
+                "'predict_method' default value is changed from 'predict' to "
+                "'auto'. Explicitly pass `predict_method='predict' to "
+                "replicate the old behavior, or pass `predict_method='auto' "
+                "or other valid values to silence this warning.",
+                FutureWarning,
+            )
+            self._predict_method = "predict"
+        else:
+            self._predict_method = self.predict_method
+
+        if not self.prefit:
+            self.estimator_ = clone(self.estimator).fit(X, y, **kwargs)
+        else:
+            try:
+                check_is_fitted(self.estimator)
+            except NotFittedError:
+                warn(BASE_ESTIMATOR_NOT_FITTED_WARNING.format(type(self).__name__))
+            self.estimator_ = self.estimator
+        return self
+
+    def predict(self, X, *, sensitive_features):
+        """
+        Predict stuff, write better explanation later"""
+
+        check_is_fitted(self)
+
+        # get soft predictions
+        base_predictions = np.array(
+            _get_soft_predictions(self.estimator_, X, self._predict_method)
+        )
+
+        # validate and reformat input
+        _, base_predictions_vector, sensitive_feature_vector, _ = _validate_and_reformat_input(
+            X, y=base_predictions, sensitive_features=sensitive_features, expect_y=True,
+            enforce_binary_labels=False)
+
+        final_predictions = 0.0*base_predictions_vector
+
+        for a, threshold in self.threshold_dict.items():
+
+            operation = ThresholdOperation('>', threshold)
+
+            thresholded_predictions = 1.0 * operation(base_predictions_vector)
+
+            final_predictions[sensitive_feature_vector == a] = \
+                thresholded_predictions[sensitive_feature_vector == a]
+
+        return final_predictions
--- a/test/unit/postprocessing/test_thresholder.py
+++ b/test/unit/postprocessing/test_thresholder.py
@@ -1,0 +1,58 @@
+# Copyright (c) Microsoft Corporation and Fairlearn contributors.
+# Licensed under the MIT License.
+
+import numpy as np
+import pandas as pd
+
+from sklearn.linear_model import LogisticRegression
+from fairlearn.postprocessing import Thresholder
+
+
+def test_thresholder():
+    X = pd.DataFrame([
+        [0, 4], [6, 2], [1, 3], [10, 5], [1, 7], [-2, 1], [3, 10], [14, 5],
+        [1, 3], [1, 5], [1, 7], [-5, 9], [3, 13], [7, 1], [-8, 4], [9, 1]])
+    y = pd.Series([0, 1, 1, 0, 0, 0, 1, 0, 0, 1, 0, 0, 1, 1, 1, 0])
+    a1 = "A"
+    a2 = "B"
+    a3 = "C"
+    a4 = "D"
+    A = pd.DataFrame([[a1, a3, a1 + a3], [a1, a3, a1 + a3],
+                      [a2, a3, a2 + a3], [a2, a3, a2 + a3], [a2, a3, a2 + a3], [a2, a3, a2 + a3],
+                      [a2, a4, a2 + a4], [a2, a4, a2 + a4], [a2, a4, a2 + a4], [a2, a4, a2 + a4],
+                      [a2, a4, a2 + a4], [a2, a4, a2 + a4], [a2, a4, a2 + a4], [a2, a4, a2 + a4],
+                      [a1, a4, a1 + a4], [a1, a4, a1 + a4]],
+                     columns=['SF1', 'SF2', 'SF1+2'])
+
+    estimator = LogisticRegression()
+    estimator.fit(X, y)
+
+    # finish when there are answers about how to deal with multiple sensitive features
+    # thresholder_multi = Thresholder(estimator=estimator, threshold_dict=//todo,
+    #                                 prefit=True,
+    #                                 predict_method='predict_proba')
+
+    threshold_dict_combined = {'AC': .36, 'BC': .43, 'BD': .4, 'AD': .465}
+    thresholder_combined = Thresholder(estimator=estimator,
+                                       threshold_dict=threshold_dict_combined,
+                                       prefit=True,
+                                       predict_method='predict_proba')
+
+    X_test = pd.concat([
+        pd.DataFrame([[5, 4], [7, 2], [0, 3], [1, 2], [-2, 9], [1, 1], [0, 5], [-3, 3]]),
+        X])
+    A_test = pd.concat([
+        pd.DataFrame([[a1, a3, a1 + a3], [a1, a3, a1 + a3],
+                      [a2, a3, a2 + a3], [a2, a3, a2 + a3],
+                      [a2, a4, a2 + a4], [a2, a4, a2 + a4],
+                      [a1, a4, a1 + a4], [a1, a4, a1 + a4]],
+                     columns=['SF1', 'SF2', 'SF1+2']),
+        A])
+
+    expected_y = pd.Series([1, 0, 1, 0, 1, 0, 0, 1, 1, 1, 0, 0,
+                           1, 0, 1, 0, 1, 1, 1, 1, 1, 0, 1, 0])
+
+    thresholder_combined.fit(X, y)
+    outputted_y = thresholder_combined.predict(X_test, sensitive_features=A_test.loc[:, 'SF1+2'])
+
+    assert (np.array_equal(outputted_y, expected_y))


bramreinders97:
Note that in _thresholder.py, I changed from relative to absolute imports. The reason for this is that I had many problems with the relative imports, and this fixed it for me. When we'll get this merged I'll change it back to relative imports.

bramreinders97:
I'm wondering if I should include more tests? I feel like the main thing to test is if the output is as expected, which is the case.

fairlearn/postprocessing/__init__.py
```diff
@@ -1,16 +1,18 @@
...
+# Copyright (c) Microsoft Corporation and Fairlearn contributors.
+# Licensed under the MIT License.
+
+"""This module contains methods which operate on a predictor, rather than an estimator.
+
+The predictor's output is adjusted to fulfill specified parity constraints. The postprocessors
+learn how to adjust the predictor's output from the training data.
+"""
+
+from ._thresholder import Thresholder
```
romanlutz, line 10: I believe you'll find `flake8` issues. Just run `flake8` in the repo root directory on your machine and it should point out the issues like this one (needs `  # noqa: F401`)

fairlearn/postprocessing/_interpolated_thresholder.py
```diff
@@ -9,8 +9,14 @@
...
 
-from ..utils._common import _get_soft_predictions
-from ..utils._input_validation import _validate_and_reformat_input
+# for me to get it to work now
+from utils._common import _get_soft_predictions
+from utils._input_validation import _validate_and_reformat_input
+
+# how it should be
+# from ..utils._common import _get_soft_predictions
+# from ..utils._input_validation import _validate_and_reformat_input
```
romanlutz, line 18: Anything you want help with here, or do you know how to work with this? Feel free to ask!

fairlearn/postprocessing/_thresholder.py
```diff
@@ -0,0 +1,105 @@
+# Copyright (c) Microsoft Corporation and Fairlearn contributors.
```
romanlutz, line 1: In new files we can remove the Microsoft part and just write Fairlearn contributors. 🙂 I'm sure that's a bit confusing, sorry for that!

fairlearn/postprocessing/_thresholder.py
```diff
@@ -0,0 +1,105 @@
...
+        )
+
+        # validate and reformat input
+        _, base_predictions_vector, sensitive_feature_vector, _ = _validate_and_reformat_input(
+            X, y=base_predictions, sensitive_features=sensitive_features, expect_y=True,
+            enforce_binary_labels=False)
+
+        final_predictions = 0.0*base_predictions_vector
+
+        for a, threshold in self.threshold_dict.items():
```
romanlutz, line 96: I think this must be based off of an early version of Fairlearn where we used `a` for sensitive feature. Perhaps `sensitive_feature` or `sf` would be preferable here.

fairlearn/postprocessing/_thresholder.py
```diff
@@ -0,0 +1,105 @@
...
+
+            operation = ThresholdOperation('>', threshold)
```
romanlutz, line 98: So we don't allow < thresholds? Wouldn't the threshold dict have the full specification of the threshold operation? Maybe I'm misunderstanding the input type of `threshold_dict`, though, which is entirely possible given how long it's been since I worked on it. If you're not sure yourself I'll be happy to dig in a bit more, just lmk!

test/unit/postprocessing/test_thresholder.py
```diff
@@ -0,0 +1,86 @@
+# Copyright (c) Microsoft Corporation and Fairlearn contributors.
```
romanlutz, line 1: Same here: Microsoft not needed

romanlutz (commented):
I like your plan as outlined in the description to resolve the high-level questions in the issue. This should be a good way to get started, and that way the changes stay relatively small. I'll respond on the issue as well. Thanks for this contribution!!!

fairlearn/postprocessing/_interpolated_thresholder.py
bramreinders97: I was/am struggling with getting these relative imports to work for some reason, however at some point I figured I just change it to absolute imports as for writing the actual code it does not really matter how these things are imported. If you have a clear idea what might be going wrong here (I get this error: from ..utils._common import _get_soft_predictions
ImportError: attempted relative import beyond top-level package), I'm all ears, but as I said I can manage with these absolute imports if not.

fairlearn/postprocessing/_thresholder.py
bramreinders97: Consider the following example `threshold_dict`: {'A' : .6}

Now, for group 'A' we predict 1 if the output > .6, and 0 otherwise. It seems a bit redundant (and prone to confusion) to also allow threshold operations going the other way to be included in the dict.

bramreinders97:
There is a strange problem when testing `Thresholder` with a simple Pipeline (see code below/in one of the last commits). When testing in my local IDE (visual studio code using python 3.9.9), I get BASE_ESTIMATOR_NOT_FITTED_WARNING, while if I perform the same tests in google colab (using python 3.7.12), there are no problems. I guess this is probably down to the python version? Is this something we want to look into? And if so, what kind of things should I be looking into?

image

bramreinders97:
The idea behind the `validate_and_reformat` function is the following:

The keys of `threshold_dict` represent the values of the sensitive feature column(s). If there is just one sensitive feature, these values are from just 1 column, and should therefore have the same dtype. If there are multiple sensitive features, I plan to have the user input the combinations in the `threshold_dict` like this: { ('SF1', 'SF2') : threshold }, so all keys should be tuples, hence have the same dtype. 

In the last commit I have included two warnings, one to warn the user if the situation explained above is not the case, and one to warn the user about providing sensitive features at prediction time. The `warn()` function also asks for a `Type[Warning]`, however I’m not sure what type of warning I should be providing. Do you have any suggestions for this? @hildeweerts @romanlutz @adrinjalali @mmadaio @MiroDudik @riedgar-ms

fairlearn/postprocessing/_interpolated_thresholder.py
SeanMcCarren: I think your syntax is correct, I was having similar issues at some point too! Perhaps you are running this script directly? If so then that file is not part of the fairlearn module but the file is __main__ itself instead and hence .. is above the top-level package (itself).

Otherwise, you might find an answer [here](https://stackoverflow.com/questions/14132789/relative-imports-for-the-billionth-time/14132912#14132912)

fairlearn/postprocessing/_thresholder.py
```diff
@@ -0,0 +1,241 @@
...
+
+import numpy as np
+from sklearn import clone
+from sklearn.base import BaseEstimator, MetaEstimatorMixin
+from sklearn.exceptions import NotFittedError
+from sklearn.utils.validation import check_is_fitted
+from warnings import warn
+
+
+from postprocessing._threshold_operation import ThresholdOperation
```
romanlutz, line 12:
```suggestion
from ._threshold_operation import ThresholdOperation

```
at least locally that's what makes this all work for me

fairlearn/postprocessing/_interpolated_thresholder.py
romanlutz: I think @SeanMcCarren is right. I removed your import changes using the `..` and made one other import change (removing `postprocessing` in one import as mentioned in another comment) and everything worked fine. Perhaps try `python -m purest test/unit/postprocessing` from the repo root directory to confirm.

fairlearn/postprocessing/_thresholder.py
romanlutz: I see. I think what confused me is that I used the term `threshold_dict` elsewhere in the past as a dict mapping the sensitive feature group names to `ThresholdOperation` objects, not just to floating point numbers. In my case, I can make this richer in the sense that I get to specify not just the float, but also the operator, so that's where my confusion stems from.

Taking a step back, I'm still not convinced we should give up on such a capability (thresholds in both directions). What do others think @fairlearn/fairlearn-maintainers ?

romanlutz:
> The idea behind the `validate_and_reformat` function is the following:
> 
> The keys of `threshold_dict` represent the values of the sensitive feature column(s). If there is just one sensitive feature, these values are from just 1 column, and should therefore have the same dtype. If there are multiple sensitive features, I plan to have the user input the combinations in the `threshold_dict` like this: { ('SF1', 'SF2') : threshold }, so all keys should be tuples, hence have the same dtype.
> 

Instead of using tuples of strings as dict keys I'd suggest utilizing something similar to the [existing functionality that maps such tuples to a single string](https://github.com/fairlearn/fairlearn/blob/5d14271343951156c3b98a516da4e089d655bd59/fairlearn/utils/_input_validation.py#L114) with the difference that you're mapping just the sensitive features names rather than entire arrays of them, so it's probably somewhat less convoluted that in the function I pointed to above.

romanlutz:
> There is a strange problem when testing `Thresholder` with a simple Pipeline (see code below/in one of the last commits). When testing in my local IDE (visual studio code using python 3.9.9), I get BASE_ESTIMATOR_NOT_FITTED_WARNING, while if I perform the same tests in google colab (using python 3.7.12), there are no problems. I guess this is probably down to the python version? Is this something we want to look into? And if so, what kind of things should I be looking into?

Where do I find this code? This is a screenshot so I couldn't copy. The tests all ran fine (after the import adjustments mentioned in other comments) for me locally using python 3.8.5 on MacOS.

bramreinders97:
> Instead of using tuples of strings as dict keys I'd suggest utilizing something similar to the existing functionality that maps such tuples to a single string with the difference that you're mapping just the sensitive features names rather than entire arrays of them, so it's probably somewhat less convoluted that in the function I pointed to above.

I am struggling to see the difference between what you're proposing and [what is already implemented](https://github.com/bramreinders97/fairlearn/blob/5b8a6bc3aab731acbf3afbfad17425f9e9ea7a50/fairlearn/postprocessing/_thresholder.py#L100) in `reformat_threshold_dict_keys()`. The main reason I proposed using tuples as dict keys is to have a simple and clear way for the user to specify subgroups (if there are multiple sensitive features). Internally this tuple is changed to a single string.

fairlearn/postprocessing/_interpolated_thresholder.py
bramreinders97: It seems to work now, thx :)!

fairlearn/postprocessing/_thresholder.py
hildeweerts: I don't have a very strong opinion on this. I think defaulting`'>'` suffices for most use cases and simplifies the API. I'm not opposed to having the capability, provided it does not add too much complexity for how the user needs to supply the thresholds.

What would you suggest the API to look like if we go for both directions, @romanlutz? E.g., should the user supply a dict of `ThresholdOperation` objects, a dict of tuples `(float, operator)`, a dict of floats and a separate argument for `operator`, or perhaps something else entirely?

bramreinders97:
> I am struggling to see the difference between what you're proposing and what is already implemented in reformat_threshold_dict_keys(). The main reason I proposed using tuples as dict keys is to have a simple and clear way for the user to specify subgroups (if there are multiple sensitive features). Internally this tuple is changed to a single string.

To clarify, the tuple structure is what the user inputs as keys into the `threshold_dict`, to make it clearer for the user. Internally, this tuple structure is transformed into a single string.

bramreinders97:
In an offline discussion with @hildeweerts this morning, we decided that we'd add a keyword argument `default_threshold` to which we compare groups that are not mentioned in `threshold_dict`. This seems nicer for the user as it seems like a realistic situation that you'd only want to change the thresholds of a subset of all groups in the data, instead of manually specifying them all.

bramreinders97:
In the last commit I've added a first version of the user guide for `Thresholder`. Please let me know if you miss something/anything is unclear/other feedback @adrinjalali @romanlutz @hildeweerts.

There are some things that I have doubts about and would like some feedback on:

- Throughout the guide, I use a class I wrote for myself which produces plots that help gain insight into the data. (The class is called `PlotPredictionInfoClassifier`, and the instance of it I use is called `plotter`.  However, since the class contains quite a lot of code, and is not directly relevant for the explanation of the workings of `Thresholder`, I did not include the code in the user guide. This does mean that the reader will presumably not understand where this `plotter` is coming from and why it does what it does. I realize that this is probably not how it should be, but I do feel that these plots can help in understanding how `Thresholder` works. In conclusion: I'm not really sure what direction to take here, so I would appreciate some input :)
- I am not sure where exactly in the docs I should include the guide? In the index.rst file, I placed it between mitigation and datasets, but there was no real logic behind this decision.
- Whenever I start a code block in the user guide, I start it off by typing:
image
I just copied this from another .rst file, but am not sure if this is correct?

docs/user_guide/thresholder.rst
```diff
@@ -0,0 +1,560 @@
...
+    >>> X_train, X_test, Y_train, Y_test, A_train, A_test, df_train, df_test = train_test_split(
+    ...        X,
+    ...        Y,
+    ...        A,
+    ...        df,
+    ...        test_size=0.50,
+    ...        stratify=Y,
+    ...        random_state=445)
+
+    >>> #Resample datset such that training data has same number of positive and negative samples
```
romanlutz, line 99:
```suggestion
    >>> # Resample dataset such that training data has same number of positive and negative samples
```

docs/user_guide/thresholder.rst
```diff
@@ -0,0 +1,560 @@
...
+    >>>   X_train = X_train.loc[balanced_ids, :]
+    >>>   Y_train = Y_train.loc[balanced_ids]
+    >>>   A_train = A_train.loc[balanced_ids]
+    
+    >>>   return X_train, Y_train, A_train
+
+    >>> X_train, Y_train, A_train = resample_dataset(X_train, Y_train, A_train)
+   
+
+Classifier: 'Neutral' predictions
```
romanlutz, line 116: By "neutral", do you mean a uniform threshold for every group? Neutral is somewhat ambiguous...

docs/user_guide/thresholder.rst
```diff
@@ -0,0 +1,560 @@
...
+classification algorithms). 
+
+Considering the nature of the problem at hand, important aspects of our model
+to investigate are:
+
+- How many patients are recommended for care (selection rate = what % of predictions are positive) 
+- Reducing false negative rate (we don't want people that need extra care not receiving it)
+
+.. doctest:: thresholder
+    :options:  +NORMALIZE_WHITESPACE
```
romanlutz, line 131: Re: your question:
Yes, this is how you can start code cells that you want executed and where you want the output to be checked. The whitespace option is needed to avoid weird issues with formatting when you output tables. You can remove it and see what happens. It might start complaining... Certainly fine to add it!

docs/user_guide/thresholder.rst
```diff
@@ -0,0 +1,560 @@
...
+    >>> plotter.plot_proba_distr(Y_pred_proba_clf)
+
+.. figure:: ../auto_examples/images/user_guide_thresholder_proba_distr.png
+    :align: center
+
+
+Two ways of specifying positive thresholds
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Say, you want to change the threshold of a specific group to 0.4 (all instances
+with a probability to get 1 > 0.4 will be predicted as 1). There are two ways to
```
romanlutz, line 175:
```suggestion
Say, you want to change the threshold of a specific group to 0.4 (all scores
(or probabilities) above 0.4 will be predicted as 1). There are two ways to
```

docs/user_guide/thresholder.rst
```diff
@@ -0,0 +1,560 @@
...
+Changing this threshold for groups 'Asian' and 'Other' shows the impact a
+change in threshold of just 0.1 can have on the false negative rates and the
+amount of positive predictions.
+
+.. doctest:: thresholder
+    :options:  +NORMALIZE_WHITESPACE
+
+    >>> from fairlearn.postprocessing import Thresholder
+
+    >>> threshold_dict = {'Asian': .4, 'Other': ('>' , .4)}
```
romanlutz, line 190: Aren't these both `.4`?

Also, can we add a leading 0?

docs/user_guide/thresholder.rst
```diff
@@ -0,0 +1,560 @@
...
+    :align: center
+
+
+Two ways of specifying positive thresholds
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Say, you want to change the threshold of a specific group to 0.4 (all instances
+with a probability to get 1 > 0.4 will be predicted as 1). There are two ways to
+specify this threshold in the :code:`threshold_dict`: 
+
+    - 0.4
```
romanlutz, line 178: I'm conflicted as to whether we should allow this...

docs/user_guide/thresholder.rst
```diff
@@ -0,0 +1,560 @@
...
+    >>> plotter.plot_positive_predictions(Y_pred_thresholded)
+
+.. figure:: ../auto_examples/images/user_guide_thresholder_two_ways_pos_threshold.png
+    :align: center
+
+Specifying a negative threshold
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+It is also possible to specify a negative threshold. This means that all
+instances with a probability to get 1 < the specified threshold will be
+predicted as 1. This is done by switching the sign in the threshold. Doing
```
romanlutz, line 221:
```suggestion
It is also possible to specify a negative threshold. This means that all
instances with a score (or probability) less than the specified threshold will be
predicted as 1. This is done by switching the sign in the threshold. Doing
```

docs/user_guide/thresholder.rst
```diff
@@ -0,0 +1,560 @@
...
+this for 'Asian' and 'Other' shows that this flips the false negative rate and
+percentage of positive predictions compared to the previous example. Note that
+we have to take a threshold just above 0.4 to make sure all the predictions
+are switched, as there might be instances with a probability of exaclty 0.4. 
+
+.. doctest:: thresholder
+    :options:  +NORMALIZE_WHITESPACE
+
+    >>> threshold_dict = {'Asian': ('<' , .400001), 'Other': ('<' , .400001)}
```
romanlutz, line 230: These are the same. How come this does anything different at all?

romanlutz (commented):
Didn't quite get all the way through, but here are some thoughts:

romanlutz:
> * Throughout the guide, I use a class I wrote for myself which produces plots that help gain insight into the data. (The class is called `PlotPredictionInfoClassifier`, and the instance of it I use is called `plotter`.  However, since the class contains quite a lot of code, and is not directly relevant for the explanation of the workings of `Thresholder`, I did not include the code in the user guide. This does mean that the reader will presumably not understand where this `plotter` is coming from and why it does what it does. I realize that this is probably not how it should be, but I do feel that these plots can help in understanding how `Thresholder` works. In conclusion: I'm not really sure what direction to take here, so I would appreciate some input :)

You can check the existing plotting code of threshold optimizer. Does it relate to that in some way? If not, feel free to add another file or add to the plotting file. In the past we felt like it's best to have a function rather than a class, but perhaps there's a good reason to have a class. I'd have to learn more to be able to recommend something 🙂  FWIW I am all in favor of plotting to help users understand what's happening!

docs/user_guide/thresholder.rst
```diff
@@ -0,0 +1,560 @@
...
+as 1, and a sample with probability to get 1 of 0.4 will be classified as 0.
+If this threshold would be changed to 0.7, both samples would be classified
+as 0.
+
+
+Why use it
+~~~~~~~~~~~
+:class:`Thresholder` is an easy to interpret tool, that can help explore different
+trade-offs. Where an algorithm like 
+:class:`fairlearn.postprocessing.ThresholdOptimizer` “blindly” satisfies
```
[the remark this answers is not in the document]
bramreinders97: When building the docs locally, I don’t experience any problems with this. Does this mean it is okay?

docs/user_guide/thresholder.rst
bramreinders97: Yes, that’s what I mean (the standard threshold of 0.5). I changed it to default now, to emphasize that this are predictions without using `Thresholder`. This does not really focus on the fact that there is now a uniform threshold, but I hope the text after the header makes this clear enough?

docs/user_guide/index.rst
```diff
@@ -9,6 +9,7 @@ User Guide
    fairness_in_machine_learning
    assessment
    mitigation
+   thresholder
```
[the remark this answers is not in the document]
bramreinders97: I have moved it to mitigation.rst, in the Postprocessing part. I did not see `ThresholdOptimizer` there (because it is still in progress I guess?), so could not place it in front of that. Also, I added a small description in the table at the top of the page. I added a check mark for regression of which I’m not entirely sure: `Thresholder` can take a regressor as input, but the output will always be binary, so not sure if this should be a check mark or a cross?

docs/user_guide/thresholder.rst
```diff
@@ -0,0 +1,560 @@
...
+^^^^^^^^
+The dataset used in all examples below, unless specified otherwise, is a
+clincial dataset of hospital re-admissions over a ten-year period (1998-2008)
+for diabetic patients across 130 different hospitals in the US. Each record
+represents the hospital admission records for a patient diagnosed with
+diabetes whose stay lasted one to fourteen days. We would like to develop a
+classification model, which decides whether the patients should be suggested
+to their primary care physicians for an enrollment into the high-risk care
+management program. The positive prediction will mean recommendation into the
+care program.
```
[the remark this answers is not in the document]
bramreinders97: Out of these two options I would prefer the former, mainly because this means that I don’t have to wait for #907, and this can be merged faster. 
To check if I understand correctly, if we would go for this option I should:
-	Move the short description I have here to the dataset user guide
-	Create a separate PR for someone else to complete the description in the dataset user guide
-	Refer to the dataset user guide instead of explaining the dataset in ‘my’ user guide

I should still include the codeblock where I import the dataset and do the train_test_split and stuff in ‘my’ user guide I assume, in order to have the reader understand where all variables later on in the guide are coming from?

romanlutz:
All of what you explained makes sense to me @bramreinders97. I still prefer the explicit way of specifying the threshold just because it leaves no doubt as to what is meant, but perhaps I'm alone on this one 🙂 Please ping me and others when you want another round of reviews or feedback!

bramreinders97:
>To check if I understand correctly, if we would go for this option I should:
> - Move the short description I have here to the dataset user guide
> - Create a separate PR for someone else to complete the description in the dataset user guide
> - Refer to the dataset user guide instead of explaining the dataset in ‘my’ user guide

I have completed the first and last bullet point. However, for the second one, I'm unsure if I should create a PR or an issue?

> I still prefer the explicit way of specifying the threshold just because it leaves no doubt as to what is meant, but perhaps I'm alone on this one.

Before changing this/leaving it like this I would like to know how others feel about this? @hildeweerts @adrinjalali @mmadaio @MiroDudik @riedgar-ms

Furthermore, I would like some feedback [on this problem](https://github.com/fairlearn/fairlearn/pull/997#discussion_r786919383) :).

I plan to start working on `RejectOptionClassifier` now, therefore it would be nice if everything related to `Thresholder` can be merged. Do you think this is 'merge-ready' (apart from the things I mentioned above)? @romanlutz @hildeweerts @adrinjalali @mmadaio @MiroDudik @riedgar-ms

docs/user_guide/index.rst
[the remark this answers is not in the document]
romanlutz: I agree, regression probably doesn't make sense. 

The ThresholdOptimizer PR is still in progress and I won't get to it for at least a few weeks. I'm basically blocked on a few other issues (like the synthetic data PR) which I'd like to rely on for that PR. I suppose I could do the same as you and use this dataset instead. But by all means go ahead and put your user guide into the postprocessing section! That's perfectly fine. I'll adjust around it. I think yours is more foundational anyway.

docs/user_guide/mitigation.rst
```diff
@@ -81,6 +87,560 @@ Postprocessing
...
+~~~~~~~~~~~~
+
+What it does
+^^^^^^^^^^^^
+
+:class:`fairlearn.postprocessing.Thresholder` is a post processing technique
+which outputs a binary prediction based on group-specific thresholds. The
+specified thresholds can work in two directions (positive/negative), and can
+be applied to the output of the :code:`predict()` function a regressor,
+:code:`predict_proba()[:,1]` of a classifier, or on
```
romanlutz, line 100: That `[:,1]` only applies to a binary classifier, right?

fairlearn/postprocessing/__init__.py
```diff
@@ -1,16 +1,22 @@
...
+from ._plotting import plot_threshold_optimizer, plot_histograms_per_group, \
+     plot_positive_predictions, plot_proba_distr  # noqa: F401
+
+__all__ = [
+    "Thresholder",
+    "ThresholdOptimizer",
+    "plot_threshold_optimizer",
+    "plot_histograms_per_group",
+    "plot_positive_predictions",
+    "plot_proba_distr"
```
romanlutz, line 21: nit: if we're spelling out "predictions" and "histograms" then "distribution" shouldn't be an issue either 🤣 One could make a case for "hist" and "pos_preds", of course. I'm always a fan of fully spelled out names, but then I'm not like everyone else in this regard... @hildeweerts and other please do feel free to voice your opinions 🙂

fairlearn/postprocessing/_plotting.py
```diff
@@ -104,3 +108,269 @@ def plot_threshold_optimizer(threshold_optimizer, ax=None, show_plot=True):
 
     if show_plot:
         plt.show()
+
+
+"""Utilities to plot information about the output of an estimator"""
+
+
+def _check_A(A):
```
romanlutz, line 116: The whole `A` terminology is based on some papers and I don't think we should propagate that any further than we've already done. We try to use `sensitive_features` all throughout the code. If you see anything suggesting otherwise feel free to file an issue or fix it yourself with a PR.

The reason is just that "A" isn't terribly obvious. I can see why researchers choose it in papers, but in code you don't always have the luxury of explaining what every letter means. It's not as pervasive as X and y, either.

fairlearn/postprocessing/_plotting.py
```diff
@@ -104,3 +108,269 @@ def plot_threshold_optimizer(threshold_optimizer, ax=None, show_plot=True):
...
+        Sensitive features to identify groups by. If there are multiple
+        sensitive features, these are combined into one string
+
+    all_sf : numpy.ndarray
+        List with all sensitive feature (combinations) observed in A
+
+    samples_each_sf : pandas.Series
+        The number of samples observed of each sensitive feature (combination)
+
+    nmbr_test_samples : int
```
romanlutz, line 136:
```suggestion
    n_test_samples : int
```
Wdyt?

fairlearn/postprocessing/_plotting.py
```diff
@@ -104,3 +108,269 @@ def plot_threshold_optimizer(threshold_optimizer, ax=None, show_plot=True):
...
+        Sensitive features to identify groups by
+
+    Returns
+    -------
+    A : pandas.Series
+        Sensitive features to identify groups by. If there are multiple
+        sensitive features, these are combined into one string
+
+    all_sf : numpy.ndarray
+        List with all sensitive feature (combinations) observed in A
```
romanlutz, line 131: wait, list or ndarray? List sounds reasonable. ndarray would confuse me a bit

romanlutz: oh it's because of `np.unique`.... just add `.tolist()` and it's a list 😉

romanlutz (commented):
This is amazing! I need to read the user guide section in detail, but this is WAY more comprehensive than I would have dared to dream 🙂 

Don't forget to add an entry to https://fairlearn.org/main/user_guide/installation_and_version_guide/v0.7.1.html

fairlearn/postprocessing/_plotting.py
bramreinders97: makes sense to me, i'll change it

docs/user_guide/mitigation.rst
bramreinders97: True, I wrote it like this because if the `predict_method` specified by the user is `predict_proba`, the proba that is used internally is the one obtained by calling `predict_proba()[:,1]` (by calling `_get_soft_predictions()`).

bramreinders97:
>Don't forget to add an entry to https://fairlearn.org/main/user_guide/installation_and_version_guide/v0.7.1.html

I've added two bullet points, is that all that you meant I should do for this?

fairlearn/postprocessing/_plotting.py
```diff
@@ -104,3 +108,269 @@ def plot_threshold_optimizer(threshold_optimizer, ax=None, show_plot=True):
...
+    statistics_dict = {}
+
+    for sf in sf_to_plot:
+        statistics_dict[sf] = deepcopy(value)
+
+    return statistics_dict
+
+
+def plot_proba_distr(sensitive_features, Y_pred_proba, specified_sf_to_plot=None):
+    """Create a stacked barplot showing the distribution of probabilities.
```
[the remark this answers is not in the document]
bramreinders97: Do you mean here in the discussion, or somewhere on the docs page?

If you meant the former:

`plot_positive_predictions(sensitive_features,Y_pred)`


output


`plot_proba_distribution(sensitive_features,Y_pred_proba)`


output


`plot_histograms_per_group(sensitive_features,Y_pred_decision_func)`


output

adrinjalali:
Just a note that I'd be happy if we have `NORMALIZE_WHITESPACE` and `ELLIPSIS` globally set.

bramreinders97:
I have added a first version of `RejectOptionClassifier` to the postprocessing package, and included a user guide in the mitigation section. Feedback is very welcome :)

Some questions/remarks I have:
-	I feel like the text “The predictor’s output is adjusted to fulfill specified parity constraints. The postprocessors learn how to adjust the predictor’s output from the training data” at the top of the API Docs page for fairlearn.postprocessing was written with only `ThresholdOptimizer` in mind, and is not correct anymore now that `Thresholder` and `RejectOptionClassifier` joined the party. Wdyt?
-	I’m in doubt about the exact definition of `theta`. I feel like the definition provided in the paper is not very intuitive. To illustrate the two options, take as an example a critical region that lies between 0.4 and 0.6. (So each instance with a probability to get 1 between 0.4 and 0.6 is inside the critical region)
The paper defines this as follows: The critical region contains all instances for which :  `theta` (with 0.5 < `theta` < 1). 
On the other hand, I feel it's much more intuitive if the critical region is specified as containing all instances in the region [0.5 - `theta`, 0.5 + `theta`], with (0 < `theta` < 0.5). Wdyt? (For now, I have included both options, until we decide what to do).
- I have omitted a parameter specifying the `predict_method`, because the only thing we are interested in is the probability to get 1, with 0.5 "in the middle" (removing the possibility of specifying `decision_function`)

hildeweerts:
This PR is now ready for review! Tagging @fairlearn/fairlearn-maintainers

docs/user_guide/mitigation.rst
```diff
@@ -30,6 +30,19 @@ Fairlearn contains the following algorithms for mitigating unfairness:
       - ✔
       - ✔
       - DP, EO, TPRP, FPRP, ERP, BGL
+   *  - :class:`~fairlearn.postprocessing.Thresholder`
+      - Postprocessing algorithm that outputs a binary prediction and allows
+        the user to manually specify group-specific thresholds.
+      - ✔
+      - ✘
+      - ✘
```
MiroDudik, line 38:
```suggestion
      - N/A
```

fairlearn/postprocessing/_interpolated_thresholder.py
MiroDudik: Are we reverting this change or leaving as is?

docs/user_guide/mitigation.rst
```diff
@@ -30,6 +30,19 @@ Fairlearn contains the following algorithms for mitigating unfairness:
...
+   *  - :class:`~fairlearn.postprocessing.RejectOptionClassifier`
+      - Postprocessing algorithm based on the paper *Decision Theory for
+        Discrimination-aware Classification* [#4]_. This algorithm 
+        alters decisions made with low certainty in favor of the deprived group.
+      - ✔
+      - ✘
+      - ✘
```
MiroDudik, line 45: I think that the algorithm is designed to help with demographic parity.

docs/user_guide/mitigation.rst
```diff
@@ -30,6 +30,19 @@ Fairlearn contains the following algorithms for mitigating unfairness:
...
+   *  - :class:`~fairlearn.postprocessing.Thresholder`
+      - Postprocessing algorithm that outputs a binary prediction and allows
+        the user to manually specify group-specific thresholds.
+      - ✔
+      - ✘
+      - ✘
+   *  - :class:`~fairlearn.postprocessing.RejectOptionClassifier`
+      - Postprocessing algorithm based on the paper *Decision Theory for
+        Discrimination-aware Classification* [#4]_. This algorithm 
+        alters decisions made with low certainty in favor of the deprived group.
```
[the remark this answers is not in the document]
romanlutz: I agree. I've read this in a few papers. Without a concrete application context in which we elaborate more this term feels odd to me.

fairlearn/postprocessing/_reject_option_classifier.py
```diff
@@ -0,0 +1,272 @@
...
+    r"""A classifier that produces group specific classifications inside a critical region.
+
+    :code:`RejectOptionClassifier` looks at the certainty by which binary classification
+    decisions are made. For decisions made with high certainty (= with a posterior
+    probability close to 0 or 1), the result of the classification remains the same.
+    For decisions with low certainty (= with a probability to get 1 close to 0.5), the
+    result is disregarded and instances from the favorable group receive the undesirable label,
+    whilst instances from the deprived group receive the desirable label. The exact range of
+    decisions considered to have low certainty is specified by the critical region.
+
```
MiroDudik, line 30: The original paper has the terminology that I don't think is best, and also `theta` parameter is confusing. So how about this:
* `deprived_group` -> `group_to_upselect`
* `favorable_group` -> `group_to_downselect`
* `desired_label` -> `selection_label`
Instead of `theta`, I would suggest something like:
* `critical_width` -- this would mean that the critical region is [0.5-cw/2, 0.5+cw/2]

fairlearn/postprocessing/_interpolated_thresholder.py
bramreinders97: we revert the change (so go with relative import)

bramreinders97:
tagging @hildeweerts

docs/user_guide/mitigation.rst
```diff
@@ -81,6 +94,707 @@ Postprocessing
...
+How to use it (and deal with multiple sensitive features)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Using :class:`RejectOptionClassifier` requires manually specifying the critical region, the
+group to (up/down) select, and the selection label.
+
+As mentioned above, :class:`RejectOptionClassifier` assumes that the sensitive feature is binary.
+Of course, it can happen that there are multiple sensitive features, or that the sensitive feature
+is not binary (just like in our previous example). In this case it is easiest to create a new
+sensitive feature column which is binary. To see how this, and using
+:class:`RejectOptionClassifier` itself, works, consider the same `classifier <https://
```
hildeweerts, line 692:
```suggestion
:class:`RejectOptionClassifier` itself, works, consider the same :class:`sklearn.ensemble.RandomForestClassifier`
```

docs/user_guide/mitigation.rst
```diff
@@ -651,7 +1365,23 @@ Grid Search
 
 .. _references:
 
+   .. [#1] Agarwal, Dudik, Wu `"Fair Regression: Quantitative Definitions and
```
hildeweerts, line 1368: We have recently moved to BibTex! Please add the references in the `ref.bib` file and cite in the text accordingly. See also #1056 and #1069

hildeweerts (commented):
This review only focuses on the user guide for now, I will try to review the code later this week!

Small comment: is  "examples/filename.png" supposed to be part of this PR?

hildeweerts (commented):
I was just going through some of the plotting functionality. I think it might be better to create a separate PR for these, to ensure that the thresholding functionality is not blocked by plotting and vice versa.

It also seems like we should be able to reproduce at least some of the proposed plotting functionality with a lot less code using `MetricFrame.by_group.plot()`. It might be worthwhile to look into that @bramreinders97.

Tagging @romanlutz who probably has thoughts on plotting!

adrinjalali:
@bramreinders97 would you be able to work on the suggestions here?

bramreinders97:
The smaller changes in the user guide/names of variables I can fix in one or two weeks time I think. For the changes related to changing the code functionality / looking into possible changes for the code functionality I'm not sure if I can get to that before the summer (begin/mid july). Is that okay? @adrinjalali

adrinjalali:
Good to know @bramreinders97 , thanks for letting us know. I'll remove the milestone for now then.

bramreinders97:
When is the deadline to do have it included in the milestone? @adrinjalali

adrinjalali:
Not sure, with the pace we're going, I'd say June.

docs/user_guide/installation_and_version_guide/v0.7.1.rst
```diff
@@ -0,0 +1,25 @@
+v0.7.1
```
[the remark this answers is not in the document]
bramreinders97: I added this here because of [this comment](https://github.com/fairlearn/fairlearn/pull/997#pullrequestreview-878163541). Considering the recent changes about including/not including this PR in v.8 milestones, maybe it should be changed to v0.8.0.rst or removed completely? Wdyt @hildeweerts @romanlutz @adrinjalali

docs/user_guide/datasets/boston_housing_data.rst
```diff
@@ -447,3 +446,16 @@ you pause about using it in the future.
   
   .. [#11] Kinmberlé Crenshaw, Mapping the margins: Intersectionality, identity politics, and violence against women of color, 
       Stanford Law Review, 1993, 43(6), 1241-1299.
+
+.. _hospital_readmissions_dataset:
```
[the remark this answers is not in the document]
bramreinders97: At first I briefly introduced the dataset in the user guide, but moved it as I did because @romanlutz said  
>I'm not opposed to either one of these, but I'm not a fan of introducing the dataset in between the mitigation subsections.

in [this comment](https://github.com/fairlearn/fairlearn/pull/997#discussion_r786936129).

I agree that it does not make a lot of sense to have this dataset described in the boston housing dataset file. Shall I create a new file *hospital_readmissions_dataset.rst* , and create a separate PR with the purpose of providing information about this dataset similar to the way this is done in *dataset_x.rst*, for someone else to finish (as also suggested by roman in the comment just mentioned)?

docs/user_guide/mitigation.rst
```diff
@@ -81,6 +94,707 @@ Postprocessing
...
 .. currentmodule:: fairlearn.postprocessing
 
+What is a threshold?
+~~~~~~~~~~~~~~~~~~~~~
+Many postprocessing algorithms make decisions based on thresholds. But what
+is a threshold? 
+
+A threshold is a number which represents the border between positive and
+negative predictions. For most classification algorithms, the default
+threshold is 0.5, which means that a sample with a probability to get 1
```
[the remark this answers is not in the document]
bramreinders97: I think changing this to confidence score is more confusing for readers learning about all this, as the word probability is used in the majority of places when referring to the output of `predict_proba`.

docs/user_guide/mitigation.rst
```diff
@@ -81,6 +94,707 @@ Postprocessing
...
+
+The group-specific thresholds are specified in :code:`threshold_dict`. This is
+a dictionary where the keys are the sensitive feature value of the sub-group,
+and the values corresond to the threshold. Below, we will show all different
+kind of ways to specify thresholds.
+
+Furthermore, we will plot some information about the obtained predictions,
+in order to help gain an insight in the way :class:`Thresholder` works. Throughout
+this example, unless specified otherwise, we will be using the
+:ref:`clinical hospital readmissions dataset<hospital_readmissions_dataset>`.
```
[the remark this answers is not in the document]
bramreinders97: Not resolved yet, but discussion about this going on [here](https://github.com/fairlearn/fairlearn/pull/997#discussion_r848101413).

docs/user_guide/mitigation.rst
```diff
@@ -81,6 +94,707 @@ Postprocessing
...
+
+    >>> from sklearn.ensemble import RandomForestClassifier
+    >>> from fairlearn.metrics import MetricFrame, false_negative_rate
+    >>> from fairlearn.postprocessing import plot_positive_predictions, plot_proba_distribution
+
+    >>> classifier = RandomForestClassifier(random_state=1).fit(X_train,Y_train) 
+    >>> Y_pred_proba_clf = classifier.predict_proba(X_test)[:, 1]
+    >>> Y_pred_clf = classifier.predict(X_test)
+
+    >>> MetricFrame(metrics=false_negative_rate,
```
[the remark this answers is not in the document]
bramreinders97: Selection rate is shown using the plotting functionality, so I thought would be redundant to also include in the metric frame.

docs/user_guide/mitigation.rst
```diff
@@ -81,6 +94,707 @@ Postprocessing
...
+    :align: center
+
+Dealing with the Prefit parameter
+**********************************
+In all examples above, the :code:`prefit` parameter was set to :code:`True`.
+This parameter indicates whether or not fit has been called on the supplied
+estimator. If :code:`Prefit==False`, the :code:`Thresholder.fit()` function
+will fit the estimator. For the final predictions it does not
+matter whether or the estimator is fit beforehand or in the 
+:code:`Thresholder.fit()` call:
```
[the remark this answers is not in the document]
bramreinders97: Yeah I wasn’t sure how detailed I should go into the different parameters, so I added this to be sure. I can see how it might be a bit redundant. I’ll remove it for now.

docs/user_guide/mitigation.rst
```diff
@@ -81,6 +94,707 @@ Postprocessing
...
+
+.. figure:: ../auto_examples/images/user_guide_reject_option_clf_binary_output.png
+    :align: center
+
+Changing the selection label
+*********************************
+By default, the selection label is assumed to be 1. It can happen of course, that the situation is
+such that being selected corresponds to receiving the label 0. If this is the case, setting the
+parameter :code:`selection_label` to 0 is the only thing that needs to be changed in the
+configuration of :class:`RejectOptionClassifier`. The critical region itself remains the same,
```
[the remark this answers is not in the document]
bramreinders97: Indeed all instances with prob <0.4 receive 0 and all instances with prob >0.6 receive 1. The only thing that changes is that `group_to_upselect` now gets 0 inside the critical region (and `group_to_downselect` gets 1). I added a sentence that explains what happens outside.

fairlearn/postprocessing/_plotting.py
```diff
@@ -104,3 +107,286 @@ def plot_threshold_optimizer(threshold_optimizer, ax=None, show_plot=True):
...
+    all_sf : list
+        List with all sensitive feature (combinations) observed in sensitive_features
+
+    samples_each_sf : pandas.Series
+        The number of samples observed of each sensitive feature (combination)
+
+    n_test_samples : int
+        The number of rows in the data
+    """
+    if len(sensitive_features.shape) > 1 and sensitive_features.shape[1] > 1:
```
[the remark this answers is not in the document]
bramreinders97: This is done to check whether there are multiple sensitive featues (so to check if there are columns to be merged), similar to [here](https://github.com/fairlearn/fairlearn/blob/ea3ec7e6a2a49dc4733901a8d22eb7854011eb14/fairlearn/utils/_input_validation.py#L97)

fairlearn/postprocessing/_plotting.py
```diff
@@ -104,3 +107,286 @@ def plot_threshold_optimizer(threshold_optimizer, ax=None, show_plot=True):
...
+
+    Parameters
+    ----------
+    sensitive_features : pandas.Series or pandas.DataFrame
+        Sensitive features to identify groups by
+
+    Y_pred_proba : numpy.ndarray
+        Probabilities returned by a classifier
+
+    specified_sf_to_plot : list or None, default=None
```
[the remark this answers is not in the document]
bramreinders97: The items in this list represent the values the sf can take, not the name of the sf itself. So the type of list will always be dependent on the dtype of the sensitive feature.

fairlearn/postprocessing/_plotting.py
```diff
@@ -104,3 +107,286 @@ def plot_threshold_optimizer(threshold_optimizer, ax=None, show_plot=True):
...
+        bottom += nmbr_probas_percentages
+
+    plt.ylabel('distribution of predict_proba')
+    plt.xlabel('Sensitive feature')
+    plt.xticks(rotation=45, ha='right')
+    plt.legend(labels=labels, handles=handles, bbox_to_anchor=(1, .9))
+    plt.show()
+
+
+def plot_positive_predictions(sensitive_features, Y_pred, specified_sf_to_plot=None):
```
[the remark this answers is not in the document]
bramreinders97: For `plot_positive_predictions`, I think it is possible to create a plot using `by_group.plot()` that gives the same information, but I think it will be hard to have it include the percentages in the bars without writing code for that (which would end up being more or less the same as in the function written now). So I guess the question is: do we want it to have the percentages included or not. Personally, I feel like the plots produced by the functions in _plotting.py are more intuitive than using `by_group.plot()`, but then again I’m obviously rather biased as I wrote the functions ;). Wdyt @romanlutz @hildeweerts

bramreinders97:
> Small comment: is "examples/filename.png" supposed to be part of this PR?

No

bramreinders97:
I have tried to fix and resolve all comments. Some are still left open so I can reply directly to  explain why I did something and/or to keep the discussion for that particular comment open. There were also some comments aimed at the maintainers, so I did not do anything about those. 

I would really love for `Thresholder` and `RejectOptionClassifier` to be merged before the next milestone. The algorithms themselves did not get any feedback that’s unresolved at this point, so from that perspective it should be possible to get them merged. 

However, the new changes made to the user guide are still to be checked, so the user guide is probably not ready for production? I’m not sure if the algorithms themselves can be merged and the user guide in a later stage?  @adrinjalali

Then, there is the matter of the plotting functions. Having the plots in a separate PR will mean that `Thresholder` and `RejectOptionClassifier` are not blocked. This would mean however that the user guide has to be rewritten, as the plots will not be there for now, only to be changed back to it’s current state later (when we finish the plotting PR). 

I think that if this is the only way to get `Thresholder` and `RejectOptionClassifier` included in the milestone, we should definitely do this. However, if the plotting is the only thing blocking this PR from getting merged, and we can get that done before the milestone deadline, I’d say we try to finish it in this PR? 

Wdyt is the wisest thing to do here @hildeweerts @romanlutz @adrinjalali ?

docs/user_guide/mitigation.rst
```diff
@@ -81,6 +94,707 @@ Postprocessing
...
+    >>> print(np.array_equal(Y_pred_no_prefit,Y_pred_prefit))    
+    True
+
+Specify thresholds for a regressor
+***********************************
+It is also possible to threshold the predicted output of a regressor.
+This can be useful if it is desired to transform the continuous output
+of a regressor into a binary prediction. To illustrate how this could work,
+consider a simple LinearRegression example on the
+:ref:`boston housing dataset<boston_dataset>`, where the sensitive feature
```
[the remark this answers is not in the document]
romanlutz: Same! I have a PR out (postprocessing docs) that that would have already been merged except that it's using the adult/census dataset and we held off on merging for that reason until I move over to something else.

fairlearn/postprocessing/_plotting.py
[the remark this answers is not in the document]
romanlutz: Can you attach a picture? It's a bit hard to imagine 🙂

docs/user_guide/datasets/boston_housing_data.rst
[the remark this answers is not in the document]
romanlutz: That seems reasonable. Seems related to #1086, or is that yet another dataset?

docs/user_guide/installation_and_version_guide/v0.7.1.rst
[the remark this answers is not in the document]
romanlutz: This file has been renamed in `main`: https://github.com/fairlearn/fairlearn/blob/main/docs/user_guide/installation_and_version_guide/v0.8.0.rst

docs/user_guide/mitigation.rst
```diff
@@ -194,6 +207,684 @@ Postprocessing
...
+the second element of each prediction obtained from :code:`predict_proba()` of a classifier, or on
+:code:`decision_function()` of, for example, :class:`sklearn.svm.SVC` or
+:class:`sklearn.linear_model.LogisticRegression`. Examples of these different applications
+are presented below.
+
+Why use it
+^^^^^^^^^^^^
+:class:`Thresholder` is an easy to interpret tool, that can help explore
+trade-offs between fairness metrics and predictive performance. Where an algorithm like
+:class:`ThresholdOptimizer` “blindly” satisfies
```
romanlutz, line 242: Instead of "blindly" I would suggest using "simply" or "merely."
From a cursory search I'd say the internet isn't quite in unanimous agreement about this yet, but there are people who have called the use of "blindly" as an adverb ableist. That was my hunch, too. I learned quite a bit about this topic recently and was shocked how many expressions that plenty of people (including myself) use quite frequently are ableist. So much to learn, so little time 🙁 

This may be a short starter on the topic: https://languageplease.org/ableism-ableist-language/

docs/user_guide/mitigation.rst
```diff
@@ -194,6 +207,684 @@ Postprocessing
...
+a specific constraint, :class:`Thresholder` aids the user in manually exploring
+different trade-offs, which can help gain more insight into the problem. Also,
+since :class:`Thresholder` is a post-processing technique, it is not necessary to alter
+an already trained estimator, which might be the only feasible unfairness
+mitigation approach when developers cannot influence training, due to
+practical reasons or security or privacy.
```
romanlutz, line 248:
```suggestion
practical reasons, security or privacy.
```

docs/user_guide/mitigation.rst
```diff
@@ -194,6 +207,684 @@ Postprocessing
...
+.. figure:: ../auto_examples/images/user_guide_thresholder_proba_distr.png
+    :align: center
+
+.. _pos_thresholds:
+
+Two ways of specifying positive thresholds
+*******************************************
+Say, you want to change the threshold of a specific group to 0.4 (all 
+probabilities greater than 0.4 will be predicted as 1). By default,
+:class:`Thresholder` assumes a provided threshold to be positive, which
```
romanlutz, line 387: What is a positive threshold?

docs/user_guide/mitigation.rst
```diff
@@ -194,6 +207,684 @@ Postprocessing
...
+means that there are two ways to specify this threshold in the
+:code:`threshold_dict`: 
+
+    - 0.4
+    - ('>', 0.4)
+
+Changing this threshold for groups 'Asian' and 'Other' shows the impact a
+change in threshold of just 0.1 can have on the false negative rates and the
+amount of positive predictions. Note that the purpose of this user guide is
```
romanlutz, line 396: If I read this as a user I'd be super curious to know what the debate is all about 😆 Why would I not want to do this? I'm not sure if it's easy to capture that in a somewhat concise manner, though...

docs/user_guide/mitigation.rst
```diff
@@ -194,6 +207,684 @@ Postprocessing
...
+    Other            Female    0.451613
+                     Male      0.303030
+    Name: false_negative_rate, dtype: float64
+
+    >>> plot_positive_predictions(A_multiple_test,Y_pred_multiple_sf_thresholded)
+
+.. figure:: ../auto_examples/images/user_guide_thresholder_multiple_sf_with_threshold.png
+    :align: center
+
+RejectOptionClassifier
```
romanlutz, line 741:
```suggestion
Reject Option Classifier
```

docs/user_guide/mitigation.rst
```diff
@@ -194,6 +207,684 @@ Postprocessing
...
+    ...                                     prefit=True)\
+    ...                                     .fit(X_train,Y_train,sensitive_features=sf_train_reject_clf)
+
+    >>> Y_pred_reject_clf = reject_clf.predict(X_test,sensitive_features=sf_test_reject_clf)
+
+    >>> MetricFrame(metrics=false_negative_rate, 
+    ...             y_true=Y_test,
+    ...             y_pred=Y_pred_reject_clf, 
+    ...             sensitive_features=sf_test_reject_clf).by_group
+    race
```
romanlutz, line 834: If you're using a contrived example then it's probably best not to call it "race"

fairlearn/postprocessing/_plotting.py
```diff
@@ -130,3 +133,257 @@ def plot_threshold_optimizer(threshold_optimizer, ax=None, show_plot=True):
 
     if show_plot:
         plt.show()
+
+
+"""Utilities to plot information about the output of an estimator"""
+
+
+def _check_sensitive_features(sensitive_features):
```
romanlutz, line 141: Should this live in the utilities file where merge_columns lives?

fairlearn/postprocessing/_plotting.py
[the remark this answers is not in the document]
romanlutz: I meant here 🙂 Thanks!

docs/user_guide/datasets/boston_housing_data.rst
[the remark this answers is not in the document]
rensoostenbach: That is the same dataset, so I would encourage you not to write it again :).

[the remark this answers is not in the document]
romanlutz: Oh great! I'm glad we figured that part out. Thanks @rensoostenbach for confirming!

docs/user_guide/mitigation.rst
```diff
@@ -194,6 +207,684 @@ Postprocessing
...
+    >>> Y = data.target
+
+    >>> df = df[df['gender'] != 'Unknown/Invalid']
+    >>> df = df[df['race'] != 'Unknown']
+
+    >>> Y = Y[df.index]
+
+    >>> sensitive_features =  df.loc[:, 'race']
+
+    >>> #Create dummy's, drop sf from X, drop columns that correlate with Y
```
romanlutz, line 305:
```suggestion
    >>> # Create dummy's, drop sf from X, drop columns that correlate with Y 
```

docs/user_guide/mitigation.rst
```diff
@@ -194,6 +207,684 @@ Postprocessing
...
+    ...
+    ...   return X_train, Y_train, sf_train
+
+    >>> X_train, Y_train, sf_train = resample_dataset(X_train, Y_train, sf_train)
+   
+
+.. doctest:: mitigation
+    :options:  +NORMALIZE_WHITESPACE
+
+    >>> #Train classifier, make and inspect default predictions
```
romanlutz, line 341:
```suggestion
    >>> # Train classifier, create and inspect default predictions
```

romanlutz (commented):
Made another pass over Thresholder. I felt like I was just nitpicking on tiny things, so I feel like we're close to merging that part. I don't know if the same applies to Reject Option Classifier since I didn't quite get that far.

docs/user_guide/mitigation.rst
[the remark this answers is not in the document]
adrinjalali: Some examples:

https://www.openml.org/search?type=data&sort=runs&id=1100&status=active
https://www.openml.org/search?type=data&sort=runs&id=203&status=active
https://www.openml.org/search?type=data&sort=runs&id=40536&status=active
https://www.openml.org/search?type=data&sort=runs&id=41084&status=active
https://www.openml.org/search?type=data&sort=runs&id=509&status=active

docs/user_guide/mitigation.rst
```diff
@@ -194,6 +207,684 @@ Postprocessing
 
 .. currentmodule:: fairlearn.postprocessing
 
+What is a decision threshold?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Many postprocessing algorithms make decisions based on thresholds. But what
+is a decision threshold?
+
+In binary classification, the decision threshold is a number which represents the border between
+positive and negative predictions. In many implementations of classification algorithms, the
```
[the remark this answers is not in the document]
hildeweerts: I get where you're coming from, but I think this terminology is extremely common so I don't think we need to change it to 'predictions with label 0' and 'predictions with label 1' (which can become quite tedious to read in a text).

hildeweerts (commented):
I just noticed I still had a bunch of comments I forgot to post...

bramreinders97:
> You can pass multiple metrics, so it doesn't take up more space.

I did it like this in order to show how to use the plot_positive_predictions function. With that in mind, is it okay like this?

bramreinders97:
> FWIW I'm fine with merging this as is and creating an issue as a follow-up (that does not necessarily need to be done by you, of course) since this is a bit of a rabbit hole in and of itself.

Awesome, is there an action I should take for this to happen?

docs/user_guide/mitigation.rst
```diff
@@ -194,6 +207,684 @@ Postprocessing
...
+
+    >>> plot_positive_predictions(sf_test,Y_pred_thresholded)
+
+.. figure:: ../auto_examples/images/user_guide_thresholder_decision_func_positive_pred_yes_threshold.png
+    :align: center    
+
+
+
+
+Specify threshold for multiple sensitive features
```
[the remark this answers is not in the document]
bramreinders97: I’d say we keep it in until we have the functionality, and we create a separate PR for the functionality. Should I create that PR, and if so what exactly should I specify in the PR description?

fairlearn/postprocessing/_reject_option_classifier.py
```diff
@@ -0,0 +1,256 @@
...
+        error. In that case, please use ``prefit=False``.
+
+    Notes
+    -----
+    The procedure is based on the algorithm of
+    `Kamiran et al. (2012) <https://ieeexplore.ieee.org/document/6413831>`_ [1]_.
+
+    References
+    ----------
+    .. [1] F. Kamiran, A. Karim and X. Zhang,
```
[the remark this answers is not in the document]
bramreinders97: If I understand correctly, If I add the bibtext using `:footcite:kamiran2012rejectoptionclassifier`, I should also include `:footbibliography:` at the end of the page. As this shows up in the API docs page of postprocessing, It doesn’t seem right to include this in the separate code files that end up o that page. I’m not sure if I should include `:footbibliography:` somewhere, and if so where exactly?

fairlearn/postprocessing/_reject_option_classifier.py
```diff
@@ -0,0 +1,256 @@
...
+    --------
+    >>> from fairlearn.postprocessing import RejectOptionClassifier
+    >>> import pandas as pd
+    >>> from sklearn.ensemble import RandomForestClassifier
+    >>> X_train = pd.DataFrame([[0, 4], [6, 2], [1, 3], [10, 5], [1, 7], [-2, 1]])
+    >>> y_train = pd.Series([1, 0, 1, 0, 1, 0])
+    >>> sensitive_features_train = pd.DataFrame(
+    ...                 [['A'], ['A'], ['A'], ['B'], ['B'], ['B']], columns=['SF1'])
+    >>> estimator = RandomForestClassifier(random_state=1)
+    >>> estimator.fit(X_train, y_train)  # doctest: +SKIP
```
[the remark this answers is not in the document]
bramreinders97: I’m not sure what exactly the error was, but something went wrong if the doctest called .fit. The results from the subsequent lines are correct though, so the code works as intended.

fairlearn/postprocessing/_reject_option_classifier.py
```diff
@@ -0,0 +1,256 @@
...
+            raise TypeError("critical_width should be of type 'float', but is of "
+                            "type '{}'".format(type(self.critical_width).__name__))
+
+        # check if between 0 and 1
+        if not 0 <= self.critical_width <= 1:
+            raise ValueError(
+                "critical_width should be between 0 and 1, but is {}".format(self.critical_width))
+
+    def _validate_selection_label(self):
+        """Check if selection_label is 0 or 1, raise error if not."""
```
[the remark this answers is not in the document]
bramreinders97: That might be something for the ‘follow-up’ on this PR after it gets merged as it is now?

fairlearn/postprocessing/_reject_option_classifier.py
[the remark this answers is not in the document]
romanlutz: Correct.
```
References
----------

.. footbibliography::
```

You can use https://raw.githubusercontent.com/fairlearn/fairlearn/main/docs/user_guide/fairness_in_machine_learning.rst as an example of how to do it.

docs/user_guide/mitigation.rst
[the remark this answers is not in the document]
romanlutz: Rereading the section after a long time, I'm mostly confused by the word "tuple." 
AFAIK the line
```
A_multiple = df.loc[:, ['race','gender']]
```
just creates a DataFrame with two columns, rather than 1 column with a tuple as the value, right? I'm guessing that's what you meant and I'm just being pedantic about wording.

If that's true, then I think we already have the best solution and you can ignore what I wrote above as it doesn't really apply.

TamaraAtanasoska:
Hi @bramreinders97, it has been a while 👋  Would you maybe still be interested in finishing this PR?

Issue #984, opened 2021-10-19
ENH: Incorporating folktables datasets from Ding et al.


millerjohnp:
#### Is your feature request related to a problem? Please describe.

Fairlearn currently exposes and makes use of the UCI Adult dataset in a few places, e.g. [in this example](https://github.com/fairlearn/fairlearn/blob/1247b081b5cbdf9373595b7e4627c142ae30ae44/examples/plot_grid_search_census.py). (Some recent work)[https://arxiv.org/abs/2108.04884] I've been apart of has highlighted a few limitations with UCI Adult and introduced a [few new datasets](https://github.com/zykls/folktables) for fair ML intended to serve as a replacement.

I discussed this briefly with @romanlutz, and I'd love to find a way to incorporate these datasets into fairlearn, perhaps as a dedicated example or as a replacement to UCI Adult in the current example. I'm happy to do most or all of the implementation work needed to make something like that happen, but I first wanted to discuss with the maintainers or others involved with the project what the best way of incorporating these datasets would be. 

Looking forward to discussing.

adrinjalali:
I just came to say what @hildeweerts said.

romanlutz:
Thanks for submitting this issue @millerjohnp! 

I'm very much in favor of this project since we're not doing a good job at providing the information on why this dataset may not be ideal to use as a benchmark. Of course, your paper (and repo) add several alternative datasets which could be added to `fairlearn.datasets`. More on that below.

I'd also encourage breaking it down into as many pieces as possible. In my own experience (see many PRs of mine) it's a lot easier to add a small change/addition rather than a big chunk since all reviewers need to be satisfied and the more content there is the more there is to disagree on.

I see several pieces here:
- documenting the existing Adult dataset - all we have right now is https://fairlearn.org/main/api_reference/fairlearn.datasets.html#fairlearn.datasets.fetch_adult . A more elaborate exploration would probably go into the user guide. 
- Uploading the new datasets to OpenML
- Adding functions to the `datasets` module to import the OpenML datasets with a basic docstring
- Adding context on why these new datasets are in some ways preferable to the existing adult dataset either in the user guide or through an example.

Note: #961 is currently adding a datasets section in the user guide to which some of this could be added.

I would be a bit reluctant to replace the adult dataset in the "get started" guide because we were planning to use a synthetic dataset there, based on #907. The rationale in short is that when you don't have the time to explain the context, don't use the data. In user guide or examples we do have the space to elaborate, so it's different there.

[...]

PR #1005, opened 2022-01-10
DOC: ACSIncome Documentation

kspieks:
This PR introduces documentation describing the ACSIncome dataset, which is 1 of 5 datasets created and published by Ding et al. (2021) [link](https://arxiv.org/pdf/2108.04884.pdf). The documentation summarizes the features and target variable of the dataset. It also reorganizes the documentation files in preparation for adding the other datasets from Ding et al. However, in the spirit of [Lutz' Law](https://twitter.com/romanlutz13/status/1451289545627086849), the other datasets will be left for future PRs.

Looking forward to your comments. Thanks in advance for reviewing!

### DIFF
--- a/docs/user_guide/datasets/acs_income.rst
+++ b/docs/user_guide/datasets/acs_income.rst
@@ -1,16 +1,23 @@
 .. _acsincome-data:
-Retiring Adult: ACSIncome
+ACSIncome
 -------------------------
 
 Introduction
 ^^^^^^^^^^^^^^^^^
 
 The ACSIncome dataset is one of five datasets created by Ding et al. [1]_ 
-as an improved alternative to the popular UCI Adult dataset.
-The authors compiled data from the American Community Survey (ACS) Public Use Microdata Sample (PUMS). 
+as an improved alternative to the popular UCI Adult dataset [2]_.
+Briefly, the UCI Adult dataset is commonly used as a benchmark dataset when comparing
+different algorithmic fairness interventions. ACSIncome offers a few improvements,
+such as providing more datapoints (1,664,500 vs. 48,842) and more recent data (2018 vs. 1994).
+Further, the binary labels in the UCI Adult dataset indicate whether an indivdual earned 
+more than $50k US dollars in that year. Ding et al. show that the choice of threshold impacts
+the amount of fairness violation, so they allow users to define any threshold.
+
+Ding et al. compiled data from the American Community Survey (ACS) Public Use Microdata Sample (PUMS). 
 Note that this is a different source than the Annual Social and Economic Supplement (ASEC) 
 of the Current Population Survey (CPS) used to construct the original UCI Adult dataset.
-Ding et al. [1]_ filtered the data such that ACSIncome only includes individuals above 16 years old 
+Ding et al. filtered the data such that ACSIncome only includes individuals above 16 years old 
 who worked at least 1 hour per week in the past year and had an income of at least $100.
 
 
@@ -18,7 +25,7 @@ who worked at least 1 hour per week in the past year and had an income of at lea
 
 Dataset Description
 ^^^^^^^^^^^^^^^^^^^
-The authors provide data from 2014-2018 for all 50 states and Puerto Rico.
+Ding et al. provide data from 2014-2018 for all 50 states and Puerto Rico.
 We uploaded the 2018 data to OpenML, which fairlearn can access.
 The dataset contains 1,664,500 rows. Each row describes a person and contains 10 features, which we describe below:
 
@@ -104,8 +111,8 @@ The dataset contains 1,664,500 rows. Each row describes a person and contains 10
          13. Unmarried partner
          14. Foster child
          15. Other nonrelative
-         16. Institutionalized group quarters population. Includes correctional facilities, nursing homes, and mental hopsitals. [2]_
-         17. Noninstitutionalized group quarters population. Includes college dormitories, military barracks, group homes, missions, and shelters. [2]_
+         16. Institutionalized group quarters population. Includes correctional facilities, nursing homes, and mental hopsitals. [3]_
+         17. Noninstitutionalized group quarters population. Includes college dormitories, military barracks, group homes, missions, and shelters. [3]_
 
    *  - WKHP
       - Usual hours worked per week in the past 12 months. Values are an integer from 1 to 99. Any hours above 99 are rounded down to 99
@@ -148,6 +155,8 @@ A threshold can be applied to PINCP to frame this as a binary classification tas
   .. [1] Frances Ding, Moritz Hardt, John Miller, Ludwig Schmidt `"Retiring Adult: New Datasets for Fair Machine Learning" <https://arxiv.org/pdf/2108.04884.pdf>`_,
       Advances in Neural Information Processing Systems 34, 2021.
 
-  .. [2] `"Group Quarters and Residence Rules for Poverty", <https://www.census.gov/topics/income-poverty/poverty/guidance/group-quarters.html>`_,
+  .. [2] R. Kohavi and B. Becker. "UCI Adult Data Set." UCI Meachine Learning Repository, 5, 1996.
+
+  .. [3] `"Group Quarters and Residence Rules for Poverty", <https://www.census.gov/topics/income-poverty/poverty/guidance/group-quarters.html>`_,
       United States Census Bureau.
 
--- a/docs/user_guide/datasets/acs_income.rst
+++ b/docs/user_guide/datasets/acs_income.rst
@@ -14,8 +14,8 @@ a few improvements, such as providing more datapoints (1,664,500 vs. 48,842)
 and more recent data (2018 vs. 1994). Further, the binary labels in the UCI 
 Adult dataset indicate whether an individual earned more than $50k US dollars 
 in that year. Ding et al. show that the choice of threshold impacts the 
-amount of disparity, so they allow users to define any threshold 
-rather than fixing it at $50k.
+amount of disparity in proportion of positives, so they allow users to 
+define any threshold rather than fixing it at $50k.
 
 Ding et al. compiled data from the American Community Survey (ACS) Public 
 Use Microdata Sample (PUMS). Note that this is a different source than the 
@@ -32,7 +32,7 @@ Dataset Description
 ^^^^^^^^^^^^^^^^^^^
 Ding et al. provide data from 2014-2018 for all 50 states and Puerto Rico.
 Note that Puerto Rico is the only US territory included in this dataset.
-We uploaded the 2018 data to OpenML, which anyone can access.
+We uploaded the 2018 data to `OpenML <https://www.openml.org/d/43141>`_.
 The dataset contains 1,664,500 rows. Each row describes a person and contains 
 10 features, which we describe below:
 
@@ -101,7 +101,7 @@ The dataset contains 1,664,500 rows. Each row describes a person and contains
       - Place of birth. There are over 200 categories, including the 50 US states and several countries. Please see the data dictionary at `ACS PUMS documentation <https://www.census.gov/programs-surveys/acs/microdata/documentation.2018.html>`_ for the full list.
 
    *  - RELP
-      - Relationship:
+      - Relationship to householder:
          0. Reference person
          1. Husband or wife
          2. Biological son or daughter
--- a/docs/user_guide/datasets/acs_income.rst
+++ b/docs/user_guide/datasets/acs_income.rst
@@ -5,23 +5,22 @@ Retiring Adult: ACSIncome
 Introduction
 ^^^^^^^^^^^^^^^^^
 
-The ACSIncome dataset is one of five datasets created by Ding et al. [#0]_ 
+The ACSIncome dataset is one of five datasets created by Ding et al. [1]_ 
 as an improved alternative to the popular UCI Adult dataset.
 The authors compiled data from the American Community Survey (ACS) Public Use Microdata Sample (PUMS). 
 Note that this is a different source than the Annual Social and Economic Supplement (ASEC) 
 of the Current Population Survey (CPS) used to construct the original UCI Adult dataset.
-Ding et al. [#0]_ filtered the data such that ACSIncome only includes individuals above 16 years old 
+Ding et al. [1]_ filtered the data such that ACSIncome only includes individuals above 16 years old 
 who worked at least 1 hour per week in the past year and had an income of at least $100.
-
-
 
 
 .. _acsincome_dataset_description:
 
 Dataset Description
 ^^^^^^^^^^^^^^^^^^^
-Data is provided from 2018 for all 50 states and Puerto Rico.
-There are 1,664,500 rows and 10 features, which are described below:
+The authors provide data from 2014-2018 for all 50 states and Puerto Rico.
+We uploaded the 2018 data to OpenML, which fairlearn can access.
+The dataset contains 1,664,500 rows. Each row describes a person and contains 10 features, which we describe below:
 
 .. list-table::
    :header-rows: 1
@@ -105,8 +104,8 @@ There are 1,664,500 rows and 10 features, which are described below:
          13. Unmarried partner
          14. Foster child
          15. Other nonrelative
-         16. Institutionalized group quarters population
-         17. Noninstitutionalized group quarters population
+         16. Institutionalized group quarters population. Includes correctional facilities, nursing homes, and mental hopsitals. [2]_
+         17. Noninstitutionalized group quarters population. Includes college dormitories, military barracks, group homes, missions, and shelters. [2]_
 
    *  - WKHP
       - Usual hours worked per week in the past 12 months. Values are an integer from 1 to 99. Any hours above 99 are rounded down to 99
@@ -141,25 +140,14 @@ A threshold can be applied to PINCP to frame this as a binary classification tas
       - Description
 
    *  - PINCP
-      - Total income, denoted as an integer ranging from 104 to 1,423,000.
-
-
-An additonal column for the state code is also provided so users to easily filter the data to focus on subsets by state.
-
-.. list-table::
-   :header-rows: 1
-   :widths: 7 30
-   :stub-columns: 1
-
-   *  - Column name
-      - Description
-
-   *  - ST
-      - State Code based on 2010 Census definitions. Please see the data dictionary at `ACS PUMS documentation <https://www.census.gov/programs-surveys/acs/microdata/documentation.2018.html>`_ for the full list.
+      - Total income per person, denoted as an integer ranging from 104 to 1,423,000.
 
 
 .. topic:: References:
 
-  .. [#0] Frances Ding, Moritz Hardt, John Miller, Ludwig Schmidt `"Retiring Adult: New Datasets for Fair Machine Learning" <https://arxiv.org/pdf/2108.04884.pdf>`_,
+  .. [1] Frances Ding, Moritz Hardt, John Miller, Ludwig Schmidt `"Retiring Adult: New Datasets for Fair Machine Learning" <https://arxiv.org/pdf/2108.04884.pdf>`_,
       Advances in Neural Information Processing Systems 34, 2021.
+
+  .. [2] `"Group Quarters and Residence Rules for Poverty", <https://www.census.gov/topics/income-poverty/poverty/guidance/group-quarters.html>`_,
+      United States Census Bureau.
 
--- a/docs/user_guide/datasets/acs_income.rst
+++ b/docs/user_guide/datasets/acs_income.rst
@@ -1,4 +1,4 @@
-.. _acsincome-data:
+.. _acsincome_data:
 ACSIncome
 ---------
 
@@ -26,11 +26,12 @@ years old who worked at least 1 hour per week in the past year and had an
 income of at least $100.
 
 
-.. _acsincome-dataset-description:
+.. _acsincome_dataset_description:
 
 Dataset Description
 ^^^^^^^^^^^^^^^^^^^
 Ding et al. provide data from 2014-2018 for all 50 states and Puerto Rico.
+Note that Puerto Rico is the only US territory included in this dataset.
 We uploaded the 2018 data to OpenML, which anyone can access.
 The dataset contains 1,664,500 rows. Each row describes a person and contains 
 10 features, which we describe below:
--- a/docs/user_guide/datasets/acs_income.rst
+++ b/docs/user_guide/datasets/acs_income.rst
@@ -140,7 +140,7 @@ A threshold can be applied to PINCP to frame this as a binary classification tas
       - Description
 
    *  - PINCP
-      - Total income per person, denoted as an integer ranging from 104 to 1,423,000.
+      - Total annual income per person, denoted as an integer ranging from 104 to 1,423,000.
 
 
 .. topic:: References:
--- a/docs/user_guide/datasets/acs_income.rst
+++ b/docs/user_guide/datasets/acs_income.rst
@@ -1,0 +1,165 @@
+.. _acsincome-data:
+Retiring Adult: ACSIncome
+-------------------------
+
+Introduction
+^^^^^^^^^^^^^^^^^
+
+The ACSIncome dataset is one of five datasets created by Ding et al. [#1]_ 
+as an improved alternative to the popular UCI Adult dataset.
+The authors compiled data from the American Community Survey (ACS) Public Use Microdata Sample (PUMS). 
+Note that this is a different source than the Annual Social and Economic Supplement (ASEC) 
+of the Current Population Survey (CPS) used to construct the original UCI Adult dataset.
+Ding et al. [#1]_ filtered the data such that ACSIncome only includes individuals above 16 years old 
+who worked at least 1 hour per week in the past year and had an income of at least $100.
+
+
+
+
+.. _acsincome_dataset_description:
+
+Dataset Description
+^^^^^^^^^^^^^^^^^^^
+Data is provided from 2018 for all 50 states and Puerto Rico.
+There are 1,664,500 rows and 10 features, which are described below:
+
+.. list-table::
+   :header-rows: 1
+   :widths: 7 30
+   :stub-columns: 1
+
+   *  - Column name
+      - Description
+
+   *  - AGEP
+      - Age as an integer from 0 to 99
+
+   *  - COW
+      - Class of worker:
+         1. Employee of a private for-profit company or business, or of an individual, for wages, salary, or commissions 
+         2. Employee of a private not-for-profit, tax-exempt, or charitable organization 
+         3. Local government employee (city, county, etc.) 
+         4. State government employee 
+         5. Federal government employee 
+         6. Self-employed in own not incorporated business, professional practice, or farm 
+         7. Self-employed in own incorporated business, professional practice or farm 
+         8. Working without pay in family business or farm 
+         9. Unemployed and last worked 5 years ago or earlier or never worked
+
+   *  - SCHL
+      - Educational attainment:
+         1. No schooling completed
+         2. Nursery school or preschool
+         3. Kindergarten
+         4. Grade 1
+         5. Grade 2
+         6. Grade 3
+         7. Grade 4
+         8. Grade 5
+         9. Grade 6
+         10. Grade 7
+         11. Grade 8
+         12. Grade 9
+         13. Grade 10
+         14. Grade 11
+         15. Grade 12 (no diploma)
+         16. Regular high school diploma
+         17. GED or alternative credential
+         18. Some college but less than 1 year
+         19. 1 or more years of college credit but no degree
+         20. Associate's degree
+         21. Bachelor's degree
+         22. Master's degree
+         23. Professional degree beyond a bachelor's degree
+         24. Doctorate degree
+
+   *  - MAR
+      - Marital status:
+         1. Married
+         2. Widowed
+         3. Divorced
+         4. Separated
+         5. Never married or under 15 years old
+
+   *  - OCCP
+      - Occupation. There are over 500 categories. Please see data dictionary at `ACS PUMS documentation <https://www.census.gov/programs-surveys/acs/microdata/documentation.2018.html>`_ for the full list of occupation codes.
+
+   *  - POBP
+      - Place of birth. There are over 200 categories, including the 50 US states and several countries. Please see the data dictionary at `ACS PUMS documentation <https://www.census.gov/programs-surveys/acs/microdata/documentation.2018.html>`_ for the full list.
+
+   *  - RELP
+      - Relationship:
+         0. Reference person
+         1. Husband or wife
+         2. Biological son or daughter
+         3. Adopted son or daughter
+         4. Stepson or stepdaughter
+         5. Brother or sister
+         6. Father or mother
+         7. Grandchild
+         8. Parent-in-law
+         9. Son-in-law or daughter-in-law
+         10. Other relative
+         11. Roomer or boarder
+         12. Housemate or roommate
+         13. Unmarried partner
+         14. Foster child
+         15. Other nonrelative
+         16. Institutionalized group quarters population
+         17. Noninstitutionalized group quarters population
+
+   *  - WKHP
+      - Usual hours worked per week in the past 12 months. Values are an integer from 1 to 99. Any hours above 99 are rounded down to 99
+
+   *  - SEX
+      - Sex code:
+         1. Male
+         2. Female
+
+   *  - RAC1P
+      - Race code
+         1. White alone
+         2. Black or African American alone
+         3. American Indian alone
+         4. Alaska Native alone
+         5. American Indian and Alaska native tribes specified; or American Indian or Alaska Native, not specified and no other races
+         6. Asian alone
+         7. Native Hawaiian and Other Pacific Islander alone
+         8. Some Other Race alone
+         9. Two or More races
+
+
+The target label is given by PINCP. For generalizability, the integer value is provided.
+A threshold can be applied to PINCP to frame this as a binary classification task.
+
+.. list-table::
+   :header-rows: 1
+   :widths: 7 30
+   :stub-columns: 1
+
+   *  - Column name
+      - Description
+
+   *  - PINCP
+      - Total income, denoted as an integer ranging from 104 to 1,423,000.
+
+
+An additonal column for the state code is also provided so users to easily filter the data to focus on subsets by state.
+
+.. list-table::
+   :header-rows: 1
+   :widths: 7 30
+   :stub-columns: 1
+
+   *  - Column name
+      - Description
+
+   *  - ST
+      - State Code based on 2010 Census definitions. Please see the data dictionary at `ACS PUMS documentation <https://www.census.gov/programs-surveys/acs/microdata/documentation.2018.html>`_ for the full list.
+
+
+.. topic:: References:
+
+  .. [#1] Frances Ding, Moritz Hardt, John Miller, Ludwig Schmidt `"Retiring Adult: New Datasets for Fair Machine Learning" <https://arxiv.org/pdf/2108.04884.pdf>`_,
+      Advances in Neural Information Processing Systems 34, 2021.
+
--- a/docs/user_guide/datasets/acs_income.rst
+++ b/docs/user_guide/datasets/acs_income.rst
@@ -14,7 +14,7 @@ a few improvements, such as providing more datapoints (1,664,500 vs. 48,842)
 and more recent data (2018 vs. 1994). Further, the binary labels in the UCI 
 Adult dataset indicate whether an individual earned more than $50k US dollars 
 in that year. Ding et al. show that the choice of threshold impacts the 
-amount of fairness violation, so they allow users to define any threshold 
+amount of disparity, so they allow users to define any threshold 
 rather than fixing it at $50k.
 
 Ding et al. compiled data from the American Community Survey (ACS) Public 
--- a/docs/user_guide/datasets/acs_income.rst
+++ b/docs/user_guide/datasets/acs_income.rst
@@ -1,33 +1,39 @@
 .. _acsincome-data:
 ACSIncome
--------------------------
+---------
+
 
 Introduction
-^^^^^^^^^^^^^^^^^
+^^^^^^^^^^^^
 
 The ACSIncome dataset is one of five datasets created by Ding et al. [1]_ 
 as an improved alternative to the popular UCI Adult dataset [2]_.
-Briefly, the UCI Adult dataset is commonly used as a benchmark dataset when comparing
-different algorithmic fairness interventions. ACSIncome offers a few improvements,
-such as providing more datapoints (1,664,500 vs. 48,842) and more recent data (2018 vs. 1994).
-Further, the binary labels in the UCI Adult dataset indicate whether an indivdual earned 
-more than $50k US dollars in that year. Ding et al. show that the choice of threshold impacts
-the amount of fairness violation, so they allow users to define any threshold.
+Briefly, the UCI Adult dataset is commonly used as a benchmark dataset 
+when comparing different algorithmic fairness interventions. ACSIncome offers 
+a few improvements, such as providing more datapoints (1,664,500 vs. 48,842) 
+and more recent data (2018 vs. 1994). Further, the binary labels in the UCI 
+Adult dataset indicate whether an individual earned more than $50k US dollars 
+in that year. Ding et al. show that the choice of threshold impacts the 
+amount of fairness violation, so they allow users to define any threshold 
+rather than fixing it at $50k.
 
-Ding et al. compiled data from the American Community Survey (ACS) Public Use Microdata Sample (PUMS). 
-Note that this is a different source than the Annual Social and Economic Supplement (ASEC) 
-of the Current Population Survey (CPS) used to construct the original UCI Adult dataset.
-Ding et al. filtered the data such that ACSIncome only includes individuals above 16 years old 
-who worked at least 1 hour per week in the past year and had an income of at least $100.
+Ding et al. compiled data from the American Community Survey (ACS) Public 
+Use Microdata Sample (PUMS). Note that this is a different source than the 
+Annual Social and Economic Supplement (ASEC) of the Current Population 
+Survey (CPS) used to construct the original UCI Adult dataset. Ding et al. 
+filtered the data such that ACSIncome only includes individuals above 16 
+years old who worked at least 1 hour per week in the past year and had an 
+income of at least $100.
 
 
-.. _acsincome_dataset_description:
+.. _acsincome-dataset-description:
 
 Dataset Description
 ^^^^^^^^^^^^^^^^^^^
 Ding et al. provide data from 2014-2018 for all 50 states and Puerto Rico.
-We uploaded the 2018 data to OpenML, which fairlearn can access.
-The dataset contains 1,664,500 rows. Each row describes a person and contains 10 features, which we describe below:
+We uploaded the 2018 data to OpenML, which anyone can access.
+The dataset contains 1,664,500 rows. Each row describes a person and contains 
+10 features, which we describe below:
 
 .. list-table::
    :header-rows: 1
@@ -111,7 +117,7 @@ The dataset contains 1,664,500 rows. Each row describes a person and contains 10
          13. Unmarried partner
          14. Foster child
          15. Other nonrelative
-         16. Institutionalized group quarters population. Includes correctional facilities, nursing homes, and mental hopsitals. [3]_
+         16. Institutionalized group quarters population. Includes correctional facilities, nursing homes, and mental hospitals. [3]_
          17. Noninstitutionalized group quarters population. Includes college dormitories, military barracks, group homes, missions, and shelters. [3]_
 
    *  - WKHP
@@ -135,8 +141,9 @@ The dataset contains 1,664,500 rows. Each row describes a person and contains 10
          9. Two or More races
 
 
-The target label is given by PINCP. For generalizability, the integer value is provided.
-A threshold can be applied to PINCP to frame this as a binary classification task.
+The target label is given by PINCP. For generalizability, the integer value 
+is provided. A threshold can be applied to PINCP to frame this as a binary 
+classification task.
 
 .. list-table::
    :header-rows: 1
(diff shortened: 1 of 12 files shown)


docs/user_guide/datasets/acs_income.rst
```diff
@@ -0,0 +1,165 @@
+.. _acsincome-data:
+Retiring Adult: ACSIncome
+-------------------------
+
+Introduction
+^^^^^^^^^^^^^^^^^
+
+The ACSIncome dataset is one of five datasets created by Ding et al. [#1]_
```
imatiach-msft, line 8: I see two [2] in the preview:

image

imatiach-msft (approved):

kspieks:
This PR addresses some of the tasks in Issue #984.  

Tagging @millerjohnp. Please feel free to leave any comments as this PR pertains to the dataset that you and others created

docs/user_guide/datasets/acs_income.rst
kspieks: Thanks for pointing this out. I pushed a fix. Once the PR is passing tests and ready to merge, I can squash the temporary fixup commits

romanlutz: Don't worry about commits. We're squashing anyway.

fairlearn/datasets/_fetch_acsincome.py
```diff
@@ -0,0 +1,90 @@
...
+    ==============   ==============
+    Samples total           1664500
+    Dimensionality               12
+    Features                   real
+    Target                     real
+    ==============   ==============
+
+    Source: Paper: Frances Ding (2021) [1]_
+
+    .. versionadded:: 0.8.0
```
imatiach-msft, line 25: maybe set to 0.7.1 for now since we don't know what the next release is

docs/user_guide/datasets/index.rst
```diff
@@ -0,0 +1,13 @@
+.. _datasets:
+
+Datasets
+==========
+
+This section contains pages describing datasets that have been
+incorporated into the Fairlearn package.
```
romanlutz, line 7:
```suggestion
In this section we dive deeper into various datasets that have fairness-related concerns.
```

fairlearn/datasets/_fetch_acsincome.py
```diff
@@ -0,0 +1,90 @@
...
+    Download it if necessary.
+
+    ==============   ==============
+    Samples total           1664500
+    Dimensionality               12
+    Features                   real
+    Target                     real
+    ==============   ==============
+
+    Source: Paper: Frances Ding (2021) [1]_
```
romanlutz, line 23: Perhaps it's best to point to the Folktables repository? In any case, let's write "Ding et al." to indicate that there are multiple authors.

docs/user_guide/datasets/index.rst
```diff
@@ -0,0 +1,13 @@
+.. _datasets:
```
romanlutz, line 1: If you create a PR with just this new file and not even the adult page it'll probably unblock @britneyting since she can base her changes on it. Definitely in the spirit of #LutzsLaw 😉

fairlearn/datasets/_fetch_acsincome.py
```diff
@@ -0,0 +1,90 @@
...
+        object.
+
+    Returns
+    -------
+    dataset : :obj:`~sklearn.utils.Bunch`
+        Dictionary-like object, with the following attributes.
+
+        data : ndarray, shape (1664500, 12)
+            Each row corresponding to the 10 feature values in order as well as
+            the state code (ST) and the target varaible (PINCP).
```
romanlutz, line 56:
```suggestion
            the state code (ST) and the target variable (PINCP).
```

fairlearn/datasets/_fetch_acsincome.py
```diff
@@ -0,0 +1,90 @@
...
+# Licensed under the MIT License.
+
+import pathlib
+
+from sklearn.datasets import fetch_openml
+from ._constants import _DOWNLOAD_DIRECTORY_NAME
+
+
+def fetch_acsincome(*, cache=True, data_home=None,
+                    as_frame=False, return_X_y=False):
```
romanlutz, line 11: Do you want to allow people to filter by state throw an argument?
Or should they do that themselves? The code is a number, not text abbreviation (such as MA), right? Perhaps it's worth copying the mapping over from folktables and map it to the text abbrevation, because nobody is going to understand this.

fairlearn/datasets/_fetch_acsincome.py
```diff
@@ -0,0 +1,90 @@
...
+    .. versionadded:: 0.8.0
+
+    Parameters
+    ----------
+    cache : bool, default=True
+        Whether to cache downloaded datasets using joblib.
+
+    data_home : str, default=None
+        Specify another download and cache folder for the datasets.
+        By default, all scikit-learn data is stored in '~/.fairlearn-data'
```
romanlutz, line 34:
```suggestion
        By default, all fairlearn data is stored in '~/.fairlearn-data'
```
This looks like a copy-paste error from when we first created one of these and that hasn't been detected. Perhaps we should fix this in all the existing _fetch_*.py files. That can totally be a separate PR if you feel like it. Otherwise let's open an issue for it.

docs/user_guide/datasets/acs_income.rst
```diff
@@ -0,0 +1,165 @@
+.. _acsincome-data:
+Retiring Adult: ACSIncome
+-------------------------
+
+Introduction
+^^^^^^^^^^^^^^^^^
+
+The ACSIncome dataset is one of five datasets created by Ding et al. [#0]_ 
+as an improved alternative to the popular UCI Adult dataset.
```
romanlutz, line 9: Are you planning on adding a description for UCI Adult as well? Just wondering because this leaves it unclear what the improvements actually are, and what the problems with UCI Adult were in the first place. It could totally be a separate page, of course. You can absolutely borrow heavily from the paper (with citation). No need to reinvent the wheel 😉

docs/user_guide/datasets/acs_income.rst
```diff
@@ -0,0 +1,165 @@
...
+         8. Parent-in-law
+         9. Son-in-law or daughter-in-law
+         10. Other relative
+         11. Roomer or boarder
+         12. Housemate or roommate
+         13. Unmarried partner
+         14. Foster child
+         15. Other nonrelative
+         16. Institutionalized group quarters population
+         17. Noninstitutionalized group quarters population
```
romanlutz, line 109: What do these actually mean? Any idea? I'm not familiar with these terms

docs/user_guide/datasets/acs_income.rst
```diff
@@ -0,0 +1,165 @@
...
+
+Dataset Description
+^^^^^^^^^^^^^^^^^^^
+Data is provided from 2018 for all 50 states and Puerto Rico.
+There are 1,664,500 rows and 10 features, which are described below:
+
+.. list-table::
+   :header-rows: 1
+   :widths: 7 30
+   :stub-columns: 1
```
romanlutz, line 29: It's a bit hard to imagine what this looks like when built. Can you run the doc build command and paste what this looks like locally for you?

docs/user_guide/datasets/acs_income.rst
britneyting: According to [disabilitystatistics](https://www.disabilitystatistics.org/glossary.cfm?g_id=266&view=true#nextTerm263), "institutionalized population includes people under formally authorized supervised care," and those people are generally referred to as "inmates" or "patients" and have restricted access to the surrounding community, whereas "the noninstitutionalized population includes all people who live in group quarters other than institutions, such as dormitories, halfway houses, and communes."

docs/user_guide/datasets/acs_income.rst
kspieks: Is the plan to squash everything into one commit or do we squash into the original 4 commits?

docs/user_guide/datasets/index.rst
kspieks: Fair point. I think this PR is almost ready to merge though? But I can do that if needed.

fairlearn/datasets/_fetch_acsincome.py
kspieks: That's correct. Currently, the state code is a number. I agree that's not as convenient though because the numbers are not just 1-51 for the 50 states + Puerto Rico. Should I replace it with the abbreviation and upload a new dataset to openml?

My initial thought was that since this function has to download all of the data anyways (because we're not uploading 51 separate datasets to openml), the function may as well return everything to the user. But let me know if you think we should add a keyword argument for filtering by state(s)

fairlearn/datasets/_fetch_acsincome.py
kspieks: Thanks for noticing this! For now, I just made this its own commit on this PR

docs/user_guide/datasets/acs_income.rst
kspieks: I wasn't planning to, at least not in this PR. To be honest, I was a bit confused what the improvements were. It's definitely more data (1.6M rows vs. ~40k), and it's more recent data (2018 vs. 1994). The paper questions the choice of $50k income cutoff, which is why we're providing the raw incomes here. I'd need to do more analysis to understand how the composition of these datasets differ. Let's discuss more as a team about what we think a good dataset for fairness benchmarking means

docs/user_guide/datasets/acs_income.rst
kspieks: Admittedly, I just copied the table settings from the boston housing file. This is how it looks on my end. Let me know if I should make changes
Screen Shot 2022-01-11 at 4 05 33 PM

kspieks:
@romanlutz Thanks for the review and thoughtful comments! I incorporated and pushed changes for most and replied to the remaining. Looking forward to your replies :)

docs/user_guide/datasets/acs_income.rst
alliesaizan: I believe the Census uses these terms to distinguish group quarters like university dorms from group quarters like prison facilities. Though I agree that a footnote explaining the distinction would be helpful! Here's more background: https://www.census.gov/topics/income-poverty/poverty/guidance/group-quarters.html

docs/user_guide/datasets/acs_income.rst
```diff
@@ -0,0 +1,165 @@
...
+
+
+.. _acsincome_dataset_description:
+
+Dataset Description
+^^^^^^^^^^^^^^^^^^^
+Data is provided from 2018 for all 50 states and Puerto Rico.
+There are 1,664,500 rows and 10 features, which are described below:
+
+.. list-table::
```
alliesaizan, line 26: This table looks pretty long. What do you think about just including the field names + descriptions in the table, and listing the unique values within each variable in an appendix?

docs/user_guide/datasets/acs_income.rst
alliesaizan: If we do end up adding more information to this page about issues with the UCI Adult dataset (and I agree that even a few paragraphs would provide useful context), it might be good to move the information about the unique values in each variable to an appendix/the end of the page and just include the field names + description in this table; otherwise, the length of the table might be unwieldy.

docs/user_guide/datasets/acs_income.rst
```diff
@@ -0,0 +1,165 @@
...
+who worked at least 1 hour per week in the past year and had an income of at least $100.
+
+
+
+
+.. _acsincome_dataset_description:
+
+Dataset Description
+^^^^^^^^^^^^^^^^^^^
+Data is provided from 2018 for all 50 states and Puerto Rico.
```
alliesaizan, line 23:
```suggestion
The authors provide 2018 data for all 50 states and Puerto Rico.
```

docs/user_guide/datasets/acs_income.rst
```diff
@@ -0,0 +1,165 @@
...
+There are 1,664,500 rows and 10 features, which are described below:
```
alliesaizan, line 24:
```suggestion
The dataset contains 1,664,500 rows and 10 features, which are described below:
```

docs/user_guide/datasets/acs_income.rst
alliesaizan: Removing passive voice :)

docs/user_guide/datasets/acs_income.rst
```diff
@@ -0,0 +1,165 @@
...
+   :stub-columns: 1
+
+   *  - Column name
+      - Description
+
+   *  - PINCP
+      - Total income, denoted as an integer ranging from 104 to 1,423,000.
+
+
+An additonal column for the state code is also provided so users to easily filter the data to focus on subsets by state.
```
alliesaizan, line 147:
```suggestion
An additional state code column allows users to easily filter the data to focus on subsets by state.
```

docs/user_guide/datasets/acs_income.rst
romanlutz: Well spotted! If you want to go a step further you can also replace the "are described" with "we describe" or something like that.

docs/user_guide/datasets/acs_income.rst
romanlutz: Looks fine! Thank you

docs/user_guide/datasets/acs_income.rst
romanlutz: Couldn't agree more with both of you here! I find this context very useful to understand as well.

docs/user_guide/datasets/acs_income.rst
romanlutz: One of the reasons I'm asking: UCI Adult documentation might be the foundation for documentation of these ACS datasets, and perhaps on a separate page.

Re: limitations: I definitely agree on dataset size and recency as well as the choice of cutoff. Others from the paper:
> The choice of the 50,000 dollar threshold is idiosyncratic and potentially limits the external validity of UCI Adult as a benchmark for algorithmic fairness. In 1994, the median USincome was 26,000 dollars, and 50,000 dollars corresponds to the 76th quantile of the income distribution, and the 88th and 89th quantiles of the income distribution for the Black and female populations, respectively. Consequently, almost all of the Black and female instances in the dataset fall below the threshold and models trained on UCI adult tend to have substantially higher accuracies on these subpopulations. For instance, a standard logistic regression model trained on UCI Adult dataset achieves 85% accuracy overall, 91.4% accuracy on the Black instances, and 92.7% on Female instances. This is a rather untypical situation since often machine learning models perform more poorly on historically disadvantaged groups.
The following chart also tells you how much influence this choice of threshold has on the fairness objectives that we're trying to optimize for: 
image

More: 
> A notable if obvious limitation of our work is that it is entirely US-centric. A richer dataset ecosystem covering international contexts within the algorithmic fairness community is still lacking. Although empirical work in the Global South is central in other disciplines, there continues to be much need for the North American fairness community to engage with it more strongly

Another limitation IMO is the task definition that doesn't strike me as realistic. After all, why would you want to predict people's income? There's probably a different motivation there. That's probably a larger point of discussion, though. The main ones I wanted to include are the ones mentioned in the paper. I agree that this may be better off in a separate PR, though.

@alliesaizan what do you mean by "unique values in each variable"? Did you mean the feature descriptions? I personally don't feel very strongly about including in the text or at the end, but it's certainly something we can consider as more content is added.

fairlearn/datasets/_fetch_acsincome.py
romanlutz: I think it's fine on OpenML. I just think it's best to copy the mapping from number to state abbreviation over from folktables and put it in the code. Then add an optional arg where people can pass a list of strings, and by default they get the whole dataset, otherwise just the states they selected (which you can filter down to using the mapping). In any case, I would always drop the state index column before returning to the user, since this is not at all a user-friendly or interpretable piece of information.

docs/user_guide/datasets/acs_income.rst
riedgar-ms: +1 on having more active voice.

docs/user_guide/datasets/acs_income.rst
```diff
@@ -0,0 +1,165 @@
...
+.. list-table::
+   :header-rows: 1
+   :widths: 7 30
+   :stub-columns: 1
+
+   *  - Column name
+      - Description
+
+   *  - PINCP
+      - Total income, denoted as an integer ranging from 104 to 1,423,000.
```
riedgar-ms, line 144: Income per ...... ?

docs/user_guide/datasets/acs_income.rst
riedgar-ms: Also 'easily' is splitting an infinitive (not the worst of sins, but I don't think it adds anything here), and is a bit redundant.

docs/user_guide/datasets/acs_income.rst
alliesaizan: @romanlutz my apologies if I was not clear! I mean putting the part of the feature description that reviews what values the variable can take at the end, but agree that we can reevaluate this suggestion when more content has been added.

docs/user_guide/datasets/acs_income.rst
kspieks: Based on @romanlutz comment above, I'm changing the function so that the state code is no longer returned. So I'll remove this part from the documentation

docs/user_guide/datasets/acs_income.rst
romanlutz: We do that upon merging, no need to worry about it on your end.

docs/user_guide/datasets/acs_income.rst
riedgar-ms: I was actually thinking weekly vs monthly vs annual, but also good to clarify personal vs household.

kspieks: Ah thanks for that clarification. I'll add the word annual now!

fairlearn/datasets/_fetch_acsincome.py
kspieks: Added functionality to filter by state. Let me know what you think!

docs/user_guide/datasets/acs_income.rst
```diff
@@ -0,0 +1,153 @@
...
+Ding et al. [1]_ filtered the data such that ACSIncome only includes individuals above 16 years old 
+who worked at least 1 hour per week in the past year and had an income of at least $100.
+
+
+.. _acsincome_dataset_description:
+
+Dataset Description
+^^^^^^^^^^^^^^^^^^^
+The authors provide data from 2014-2018 for all 50 states and Puerto Rico.
+We uploaded the 2018 data to OpenML, which fairlearn can access.
```
LeJit, line 22: Do we want to eventually support the ACS datasets from the other years (2014-2017) as well? One of the key benefits of these datasets, as listed in the paper by the authors, was to perform temporal evaluation of a fairness criterion and intervention.

fairlearn/datasets/_fetch_acsincome.py
```diff
@@ -0,0 +1,140 @@
...
+        return_X_y=False,
+    )
+
+    # filter by state
+    df_all = data_dict['data'].copy(deep=True)
+    df_all['PINCP'] = data_dict['target']
+    cols = df_all.columns
+    df = pd.DataFrame(np.zeros((0, len(cols))), columns=cols)
+    for state in states:
+        dfs = [df, df_all.query(f"ST == {int(_STATE_CODES[state])}")]
```
LeJit, line 121: `_STATE_CODES[state]` will throw an error if `state` is not a valid state abbreviation. We should probably include a validation check on the `states` input.

fairlearn/datasets/_fetch_acsincome.py
```diff
@@ -0,0 +1,143 @@
...
+    return_X_y : bool, default=False
+        If True, returns ``(data.data, data.target)`` instead of a Bunch
+        object.
+
+    states: list, default=None
+        List containing two letter (capitalized) state abbreviations.
+        If None, data from all 50 US states, including Puerto Rico, will be returned.
+        The state abbreviations and codes can be found on page 1 of the data dictionary at ACS PUMS
+        https://www2.census.gov/programs-surveys/acs/tech_docs/pums/data_dict/PUMS_Data_Dictionary_2018.pdf
+        and are reproduced in the `_STATE_CODES` dictionary below.
```
imatiach-msft, line 59: "and are reproduced in the `_STATE_CODES` dictionary below."
users reading this on the docs page won't know what "dictionary below" means.  Perhaps you can just remove that part, or say "defined in the dictionary used in the source code of this method".

docs/user_guide/datasets/acs_income.rst
mmadaio: Agreed with @romanlutz that even a sentence or two of context for what the UCI Adult dataset is and why an improved alternative might be needed will be helpful for readers who may not have that context!

docs/user_guide/datasets/acs_income.rst
```diff
@@ -0,0 +1,153 @@
...
+Introduction
+^^^^^^^^^^^^^^^^^
+
+The ACSIncome dataset is one of five datasets created by Ding et al. [1]_ 
+as an improved alternative to the popular UCI Adult dataset.
+The authors compiled data from the American Community Survey (ACS) Public Use Microdata Sample (PUMS). 
+Note that this is a different source than the Annual Social and Economic Supplement (ASEC) 
+of the Current Population Survey (CPS) used to construct the original UCI Adult dataset.
+Ding et al. [1]_ filtered the data such that ACSIncome only includes individuals above 16 years old 
+who worked at least 1 hour per week in the past year and had an income of at least $100.
```
mmadaio, line 14: A few words about why they filtered it that way could be helpful

docs/user_guide/datasets/acs_income.rst
```diff
@@ -0,0 +1,153 @@
+.. _acsincome-data:
+Retiring Adult: ACSIncome
```
mmadaio, line 2: I believe their paper title was meant to be a pun about "retiring" the UCI adult dataset, so I'm not sure that it makes sense to include that pun as the title here, since it may make readers think it's a dataset about retiring adults..

mmadaio (commented):
Exciting to see this being added! Great work.

fairlearn/datasets/_fetch_acsincome.py
```diff
@@ -0,0 +1,143 @@
...
+    cols = df_all.columns
+    df = pd.DataFrame(np.zeros((0, len(cols))), columns=cols)
+    for state in states:
+        dfs = [df, df_all.query(f"ST == {int(_STATE_CODES[state])}")]
+        df = pd.concat(dfs)
+    # drop the state column since it is not a feature in the published ACSIncome dataset
+    df.drop('ST', axis=1, inplace=True)
+
+    if as_frame:
+        data_dict['data'] = df.iloc[:, :10]
```
britneyting, line 130: What does the 10 represent? 

Relatedly, what is the team's policy on using magic numbers? @imatiach-msft

docs/user_guide/datasets/acs_income.rst
kspieks: Fair point. I'll just say ACSIncome here to improve clarity. Thanks!

fairlearn/datasets/_fetch_acsincome.py
kspieks: I'll just remove that phrase since the linked pdf already provides the codes and abbreviations.

fairlearn/datasets/_fetch_acsincome.py
kspieks: 10 refers to the number of features. It is hardcoded since the number of features is constant for this dataset. To improve readability, I'll update this by defining `_NUM_FEATS = 10` inside the function

docs/user_guide/datasets/acs_income.rst
kspieks: I'm actually unsure why these filters were chosen. For example, page 7 of the paper simply states

```
ACSIncome: predict whether an individual’s income is above $50,000, after filtering the ACS PUMS data sample to only include individuals above the age of 16, who reported usual working hours of at least 1 hour per week in the past year, and an income of at least $100.
```


Page 21 in the SI restates the filters, but again does not offer an explanation. In contrast, the ACSPublicCoverage is filtered to individuals under the age of 65, and the authors explicitly state that the purpose was to focus on people who are not eligible for Medicare.

fairlearn/datasets/_fetch_acsincome.py
kspieks: Thanks for pointing this out. I added a try except that prints a helpful error message

docs/user_guide/datasets/acs_income.rst
kspieks: Great question. Yes I think we want to support these in the future. My opinion was that the scope of this PR should only add 2018 data for one of the datasets (ACSIncome) as a starting point. Ideas for future PRs include 
1) Adding other datasets from the paper, such as ACSPublicCoverage which @britneyting is working on
2) Adding analysis of how the composition of the new ACSIncome dataset compares to UCI Adult. Some interesting questions could include: Are certain groups better represented in the updated dataset? Are any discrepancies larger or smaller? 
3) Adding data from other years

docs/user_guide/datasets/acs_income.rst
kspieks: I added a brief summary of UCI Adult and the improvements offered by ACSIncome based on the points made in this thread :)

fairlearn/datasets/_fetch_acsincome.py
```diff
@@ -0,0 +1,143 @@
...
+            Integer denoting each person's annual income.
+            A threshold can be applied as a postprocessing step to frame this
+            as a binary classification problem.
+            If ``as_frame`` is True, ``target`` is a pandas object.
+        feature_names : list of length 10
+            Array of ordered feature names used in the dataset.
+        DESCR : string
+            Description of the ACSIncome dataset.
+
+    (data, target) : tuple of (numpy.ndarray, numpy.ndarray) or (pandas.DataFrame, pandas.Series)
```
britneyting, line 79: This might be a copy-paste error, I think this should only return a tuple of (numpy.ndarray, numpy.ndarray) since `as_frame` is False?

kspieks: Good catch. I'll fix this in the other files as well

kspieks:
Thanks so much for all of the comments! I pushed more fixup commits to incorporate everyone's suggestions. Really value your feedback for improving this PR!

Also, please unresolve any comments if you have more to discuss. I resolve them to help me keep track of what I've pushed changes for but I apologize if I preemptively resolved any conversations.

romanlutz:
Not sure why I'm not able to respond to thread above, but this is the one I'm referring to:
@LeJit wrote:
> Do we want to eventually support the ACS datasets from the other years (2014-2017) as well? One of the key benefits of these datasets, as listed in the paper by the authors, was to perform temporal evaluation of a fairness criterion and intervention.

@kspieks responded:
> Great question. Yes I think we want to support these in the future. My opinion was that the scope of this PR should only add 2018 data for one of the datasets (ACSIncome) as a starting point. Ideas for future PRs include
> Adding other datasets from the paper, such as ACSPublicCoverage which @britneyting is working on
Adding analysis of how the composition of the new ACSIncome dataset compares to UCI Adult. Some interesting questions could include: Are certain groups better represented in the updated dataset? Are any discrepancies larger or smaller?
Adding data from other years

I think it's fair to restrict yourself to one of the years, but let's make sure that restriction is captured in the corresponding issue #984 so that someone else can revisit it in the future.

fairlearn/datasets/_fetch_acsincome.py
```diff
@@ -0,0 +1,158 @@
...
+            try:
+                _STATE_CODES[state]
+            except KeyError:
+                raise KeyError(f'Error with state code: {state}\n'
+                               f'State code must be a two letter abbreviation'
+                               f'from the list {list(_STATE_CODES.keys())}\n'
+                               f'Note that PR is the abbreviation for Puerto Rico.'
+                               )
+    else:
+        states = _STATE_CODES.keys()
```
britneyting, line 119: Not sure how much we care about performance, but your code will go through the whole dataset and re-add all the rows in this case, which will result in superfluous work given that `fetch_openml` retrieves the whole dataset

kspieks: Correct, it's technically inefficient if users will frequently request data for all states. However, in the other extreme, it should be fast if users frequently request data for just 1 or 2 states. I believe the later will be more common for a couple of reasons:

1) Smaller datasets lead to faster training times, which is convenient when one desires to focus on evaluating fairness rather than dealing with slow training times caused by ~1.7M datapoints. For example, this basically eliminates the possibility of ever training and evaluating an SVC due to terrible scaling with dataset size. Using a linear kernel helps, but practically it's still orders of magnitude slower than decision trees or logistic regression, which would only be compounded if using Exponentiated Gradient to train multiple models.

2) More data isn't necessary to properly evaluate fairness. As the authors conclude: `Increased dataset size does not necessarily help in reducing observed disparities`, so I think requesting data for just 1 or 2 states will more common.

Let me know your thoughts :)

docs/user_guide/datasets/acs_income.rst
```diff
@@ -0,0 +1,162 @@
+.. _acsincome-data:
+ACSIncome
+-------------------------
+
+Introduction
+^^^^^^^^^^^^^^^^^
+
+The ACSIncome dataset is one of five datasets created by Ding et al. [1]_ 
+as an improved alternative to the popular UCI Adult dataset [2]_.
+Briefly, the UCI Adult dataset is commonly used as a benchmark dataset when comparing
```
romanlutz, line 10: I'm not sure what "Briefly" is meant to indicate here tbh. Perhaps it's because English is not my native language (?)

docs/user_guide/datasets/acs_income.rst
```diff
@@ -0,0 +1,162 @@
...
+different algorithmic fairness interventions. ACSIncome offers a few improvements,
+such as providing more datapoints (1,664,500 vs. 48,842) and more recent data (2018 vs. 1994).
+Further, the binary labels in the UCI Adult dataset indicate whether an indivdual earned
```
romanlutz, line 13:
```suggestion
Further, the binary labels in the UCI Adult dataset indicate whether an individual earned 
```

docs/user_guide/datasets/acs_income.rst
```diff
@@ -0,0 +1,162 @@
...
+^^^^^^^^^^^^^^^^^
+
+The ACSIncome dataset is one of five datasets created by Ding et al. [1]_ 
+as an improved alternative to the popular UCI Adult dataset [2]_.
+Briefly, the UCI Adult dataset is commonly used as a benchmark dataset when comparing
+different algorithmic fairness interventions. ACSIncome offers a few improvements,
+such as providing more datapoints (1,664,500 vs. 48,842) and more recent data (2018 vs. 1994).
+Further, the binary labels in the UCI Adult dataset indicate whether an indivdual earned 
+more than $50k US dollars in that year. Ding et al. show that the choice of threshold impacts
+the amount of fairness violation, so they allow users to define any threshold.
```
romanlutz, line 15: I'd suggest adding at the end: "rather than fixing it at $50k." to be super explicit.

docs/user_guide/datasets/acs_income.rst
```diff
@@ -0,0 +1,162 @@
...
+of the Current Population Survey (CPS) used to construct the original UCI Adult dataset.
+Ding et al. filtered the data such that ACSIncome only includes individuals above 16 years old 
+who worked at least 1 hour per week in the past year and had an income of at least $100.
+
+
+.. _acsincome_dataset_description:
+
+Dataset Description
+^^^^^^^^^^^^^^^^^^^
+Ding et al. provide data from 2014-2018 for all 50 states and Puerto Rico.
```
romanlutz, line 28: I'm undecided on whether it makes sense to explicitly mention that other territories are excluded. What do you think?

docs/user_guide/datasets/acs_income.rst
```diff
@@ -0,0 +1,162 @@
...
+We uploaded the 2018 data to OpenML, which fairlearn can access.
```
romanlutz, line 29: Anyone can access it 🙂 Also, Fairlearn is capitalized unless we're in code `fairlearn`

docs/user_guide/datasets/acs_income.rst
```diff
@@ -0,0 +1,162 @@
...
+   *  - MAR
+      - Marital status:
+         1. Married
+         2. Widowed
+         3. Divorced
+         4. Separated
+         5. Never married or under 15 years old
+
+   *  - OCCP
+      - Occupation. There are over 500 categories. Please see data dictionary at `ACS PUMS documentation <https://www.census.gov/programs-surveys/acs/microdata/documentation.2018.html>`_ for the full list of occupation codes.
```
romanlutz, line 91: Can you try to keep the line lengths to 79 chars max? There are cases where a link is longer and links don't split well in ReST, so it's best to keep them on a single line, but for everything else it should be possible.

docs/user_guide/datasets/adult_data.rst
```diff
@@ -0,0 +1,18 @@
+.. _retiring-adult-data:
+Retiring Adult
+-------------------------
+
+This section contains datasets created by Ding et al. [1]_, which are described
+in their published paper "Retiring Adult: New Datasets for Fair Machine Learning" (2021).
```
romanlutz, line 6: Shall we make this `_adult_data`, header `Adult census dataset`, and then mention that the adult dataset is very popular benchmark for algorithmic interventions w.r.t. fairness metrics (like you did in the other file)?
After that as a brief intro, the current lines 5&6 are great since they give us an idea of what to expect in the other files.

romanlutz: I guess a citation for the adult dataset wouldn't hurt in that case either. You can find it in the API ref of fetch_adult

docs/user_guide/datasets/boston_housing_data.rst
```diff
@@ -1,6 +1,4 @@
-Datasets
-==========
-
+.. _boston-housing-data:
```
romanlutz, line 1: I don't know if dashes work in these tags, but let's stick to underscores.
```suggestion
.. _boston_housing_data:
```

fairlearn/datasets/_fetch_acsincome.py
romanlutz: If it's a huge issue we can always look into uploading the smaller datasets separately at a later point. I wouldn't worry too much about it as long as it doesn't take too long for you to run locally. Any rough estimate?

romanlutz (commented):
Just a few minor things. I think this is a fantastic start and gives us room to build on in the future as people might want to explore particular aspects of these datasets. I'm happy to "approve" once these minor points are resolved.

fairlearn/datasets/_fetch_acsincome.py
```diff
@@ -0,0 +1,140 @@
...
+        data. If ``return_X_y`` is True, then ``(data, target)`` will be pandas
+        DataFrames or Series as describe above.
+
+    return_X_y : bool, default=False
+        If True, returns ``(data.data, data.target)`` instead of a Bunch
+        object.
+
+    states: list, default=None
+        List containing state abbreviations. If None, data from all 50 US
+        states, including Puerto Rico, will be returned.
```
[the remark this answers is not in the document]
kspieks: No worries at all. It's an easy change to make 🙂

docs/user_guide/datasets/acs_income.rst
kspieks: Thanks for catching another spelling mistake! Odd that my editor didn't alert me. I'll check my settings (I actually just caught another)

docs/user_guide/datasets/acs_income.rst
kspieks: Good addition!

docs/user_guide/datasets/acs_income.rst
kspieks: In my opinion, the documentation should just describe what is included and not necessarily enumerate everything that is excluded, but I'm open

docs/user_guide/datasets/acs_income.rst
kspieks: Updated this description

docs/user_guide/datasets/acs_income.rst
kspieks: I edited the paragraphs to be < 79 characters. However, this section is a table so adding newlines was creating errors when compiling the documentation

docs/user_guide/datasets/adult_data.rst
kspieks: Updated the cross-reference tag, header, and added the citation for UCI Adult

docs/user_guide/datasets/boston_housing_data.rst
kspieks: My understanding was that tags for cross-referencing should use dashes, or at least that's what the documentation uses as an example. Please help me learn if I'm incorrect https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#cross-referencing-arbitrary-locations

fairlearn/datasets/_fetch_acsincome.py
kspieks: Yea the rate limiting step is downloading the data from OpenML, which takes about 16 seconds for me. The remaining processing takes well under a second.

britneyting: One suggestion is to use [Pandas's `query` function](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.query.html) when `states is not None`, although I imagine the bigger benefit here is code readability rather than performance, since it's not taking too long for you to run anyway.

docs/user_guide/datasets/boston_housing_data.rst
romanlutz: Interesting. Not sure why I thought this to be true then....

Since we're doing it this way everywhere else I'd still suggest sticking with underscore, though. If we ever decide to go to dashes we should do it everywhere.

docs/user_guide/datasets/acs_income.rst
romanlutz: I'm thinking from my perspective as someone who wasn't born in the US. I don't think I knew about anything beyond 50 states, let alone territories like Guam, until I moved to the US. From that mistaken viewpoint it's easy to assume you're querying all of the US and describe it as such. In other words, this nuance is potentially lost on people. Part of why we're documenting this is to avoid missing nuances like that. This is the application context that's so crucial. If we train a model on data from 50 states + PR we can't assume it works elsewhere (not even the territories). [In fact, we shouldn't necessarily assume it "works" anywhere unless we validate it in its real world context.]

So perhaps this makes sense as elaborating a little bit on why it's a slippery slope, and the more information the better.

romanlutz (approved):

docs/user_guide/datasets/boston_housing_data.rst
kspieks: Makes sense. I just updated everything to use underscores

docs/user_guide/datasets/acs_income.rst
```diff
@@ -0,0 +1,170 @@
...
+
+The ACSIncome dataset is one of five datasets created by Ding et al. [1]_ 
+as an improved alternative to the popular UCI Adult dataset [2]_.
+Briefly, the UCI Adult dataset is commonly used as a benchmark dataset 
+when comparing different algorithmic fairness interventions. ACSIncome offers 
+a few improvements, such as providing more datapoints (1,664,500 vs. 48,842) 
+and more recent data (2018 vs. 1994). Further, the binary labels in the UCI 
+Adult dataset indicate whether an individual earned more than $50k US dollars 
+in that year. Ding et al. show that the choice of threshold impacts the 
+amount of fairness violation, so they allow users to define any threshold
```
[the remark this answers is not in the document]
kspieks: Thanks for the suggestion. I updated this wording to use "amount of disparity"

docs/user_guide/datasets/acs_income.rst
kspieks: Thanks for clarifying. I added a sentence of "Note that Puerto Rico is the only US territory included in this dataset." Let me know if we should provide more detail like explicitly naming the territories that are excluded

docs/user_guide/datasets/acs_income.rst
```diff
@@ -0,0 +1,171 @@
...
+
+The ACSIncome dataset is one of five datasets created by Ding et al. [1]_ 
+as an improved alternative to the popular UCI Adult dataset [2]_.
+Briefly, the UCI Adult dataset is commonly used as a benchmark dataset 
+when comparing different algorithmic fairness interventions. ACSIncome offers 
+a few improvements, such as providing more datapoints (1,664,500 vs. 48,842) 
+and more recent data (2018 vs. 1994). Further, the binary labels in the UCI 
+Adult dataset indicate whether an individual earned more than $50k US dollars 
+in that year. Ding et al. show that the choice of threshold impacts the 
+amount of disparity, so they allow users to define any threshold
```
hildeweerts, line 17: Maybe mention what type of disparity you're referring to here, e.g., something like "disparity in proportion of positives".

docs/user_guide/datasets/acs_income.rst
```diff
@@ -0,0 +1,171 @@
...
+income of at least $100.
+
+
+.. _acsincome_dataset_description:
+
+Dataset Description
+^^^^^^^^^^^^^^^^^^^
+Ding et al. provide data from 2014-2018 for all 50 states and Puerto Rico.
+Note that Puerto Rico is the only US territory included in this dataset.
+We uploaded the 2018 data to OpenML, which anyone can access.
```
hildeweerts, line 35: Nitpicks:

Maybe add a link to openml:

```suggestion
We uploaded the 2018 data to [OpenML](https://www.openml.org), which anyone can access.
```

I'm also not super sure about the phrasing of this sentence, but that's more of a style thing :)

docs/user_guide/datasets/acs_income.rst
```diff
@@ -0,0 +1,171 @@
+.. _acsincome_data:
+ACSIncome
```
hildeweerts, line 2: [I haven't gone through the other PR yet so this may be irrelevant]

It seems to me that a user guide item should also provide guidance on when one should (not) use this data set (or at least refer to other resources that cover that).

docs/user_guide/datasets/acs_income.rst
```diff
@@ -0,0 +1,171 @@
...
+         5. Never married or under 15 years old
+
+   *  - OCCP
+      - Occupation. There are over 500 categories. Please see data dictionary at `ACS PUMS documentation <https://www.census.gov/programs-surveys/acs/microdata/documentation.2018.html>`_ for the full list of occupation codes.
+
+   *  - POBP
+      - Place of birth. There are over 200 categories, including the 50 US states and several countries. Please see the data dictionary at `ACS PUMS documentation <https://www.census.gov/programs-surveys/acs/microdata/documentation.2018.html>`_ for the full list.
+
+   *  - RELP
+      - Relationship:
```
hildeweerts, line 104: I think this feature could use a bit more explanation. Relationship to whom? The reference person?

fairlearn/datasets/_fetch_acs_income.py
```diff
@@ -0,0 +1,160 @@
...
+import pandas as pd
+from sklearn.datasets import fetch_openml
+from ._constants import _DOWNLOAD_DIRECTORY_NAME
+
+
+def fetch_acs_income(*, cache=True, data_home=None,
+                     as_frame=False, return_X_y=False,
+                     states=None,
+                     ):
+    """Load the ACS Income dataset.
```
hildeweerts, line 16: We should really refer to the relevant user guide item here.

fairlearn/datasets/_fetch_acs_income.py
```diff
@@ -0,0 +1,160 @@
...
+        If True, returns ``(data.data, data.target)`` instead of a Bunch
+        object.
+
+    states: list, default=None
+        List containing two letter (capitalized) state abbreviations.
+        If None, data from all 50 US states and Puerto Rico will be returned.
+        Note that Puerto Rico is the only US territory included in this dataset.
+        The state abbreviations and codes can be found on page 1 of the data
+        dictionary at ACS PUMS:
+        https://www2.census.gov/programs-surveys/acs/tech_docs/pums/data_dict/PUMS_Data_Dictionary_2018.pdf
```
hildeweerts, line 60: I think it might be better to put this link in the references.

hildeweerts (commented):
Thank you for getting this started! I just have a few small comments for this PR :)

I didn't build the docs locally so I have only reviewed the content.

fairlearn/datasets/_fetch_acs_income.py
kspieks: Added a reference!

docs/user_guide/datasets/acs_income.rst
kspieks: Thanks for this question. Page 41 of 2018 US Census Data Dictionary ([link](https://www2.census.gov/programs-surveys/acs/tech_docs/pums/data_dict/PUMS_Data_Dictionary_2018.pdf)) and page 20 of the paper's SI ([link](https://arxiv.org/pdf/2108.04884.pdf)) only say relationship with no other information provided. However, I found this separate appendix from the US Census ([link](https://www.census.gov/content/dam/Census/library/publications/2020/acs/acs_pums_handbook_2020_appendix_a.pdf)) that says this is "relationship to householder". I'll update the description.

docs/user_guide/datasets/acs_income.rst
kspieks: We agree. However, we were thinking that the scope of this PR is just to report the facts of what this dataset is. Any recommendations/interpretations would be a good candidate for a future PR since it's more of an open discussion that definitely requires further analysis to justify any recommendations, especially since UCI Adult (and this updated version) are so central to the fairness literature; UCI Adult appeared in 300+ papers related to fairness according to Ding et al. It's something we're still discussing as a team.

docs/user_guide/datasets/acs_income.rst
kspieks: Thanks for the suggestion! I added the url for this dataset

docs/user_guide/datasets/acs_income.rst
kspieks: Thanks! Updated the wording

docs/user_guide/datasets/acs_income.rst
hildeweerts: That sounds very reasonable. Did you happen to already open an issue for this? That might help to get the discussion started and gather some views from the community.

kspieks: I can open one this afternoon :)

romanlutz:
@hildeweerts is there anything missing from this PR that you want to include before merging?

hildeweerts (approved):

Issue #911, opened 2021-07-20
Enable Reweighing to handle categorical output instead of only 0/1 


adrinjalali:
From: https://github.com/fairlearn/fairlearn/pull/906#discussion_r673118210

Some of our algorithms require the output to be 0/1. However, they only need the output to be categorical, and in case of binary classification they need to know which one is the positive label, i.e. `pos_lable`. This issue is to fix those instances.

romanlutz:
We may want to separate this into several issues (one per algorithm) if it turns out to be a lot of work. I do expect there to be a lot of common foundational work that can be shared by all of them since they already share the input validation code.

adrinjalali:
This issue certainly doesn't have to be fixed in a single PR, we could update the description and list the algorithms which are relevant and go through them one by one, or we could have a separate issue for each, I'm easy either way.

SeanMcCarren:
Hi @adrinjalali @romanlutz, I think scikit-learn's `LabelEncoder` can be used? I could give it a try if this issue is still relevant :)

romanlutz:
@SeanMcCarren it's definitely still relevant! I would suggest starting with one algorithm only and trying it out with LabelEncoder. Thanks for volunteering!

SeanMcCarren:
I've delved a bit deeper into the issue now. Starting with a list of algorithms to update (as @adrinjalali mentioned):
- [ ] threshold optimizer
- [ ] exponentiated gradient
- [ ] grid search

Then, my thoughts so far... My initial thought was that we could just translate strings to numbers in the shared input validation and be done with it. But I now think that this does not work, because we need to persist this encoding throughout calls to fit & predict. 
I assume we should follow [this sklearn guideline](https://scikit-learn.org/stable/developers/develop.html#specific-models) which tells us that we should store a list of classes in `self.classes_` on every estimator. Seems clear!

Side notes: 
1) preferrably, I'd avoid writing the same piece of code to do this in every algorithm, but then I see that sklearn actually does repeat the initialization of `self.classes_` at times, instead of capturing this in some mixin or base class. I'd rather follow sklearn here then, but I wonder why sklearn made this decision (@adrinjalali maybe you know?)
2)  In `interpolated_thresholder.py` line 141 there is `enforce_binary_labels = False` while in `threshold_optimizer.py` line 288 there is `enforce_binary_labels = True`. I can not infer whether one of the two is an error or whether I can assume that threshold optimizer only works on binary data.

[...]

PR #1006, opened 2022-01-12
ENH WIP Arbitrary target labels instead of 0/1

SeanMcCarren:
WIP on issue #911 (don't need to call the police)

This PR involves:
- [x] threshold optimizer
- [ ] exponentiated gradient. (somewhat finished)
- [ ] grid search

Issues:
- If all supplied labels are one of the two possible labels, then we can't infer self.classes_. So, now, we should raise an error. But this breaks the testcases `test_threshold_optimization_degenerate_labels` and `test_single_y_class` (the latter is of gridsearch).
- In Exponentiated gradient, how can we know if y is binary or continuous? If it is continuous, we should not apply the encoding. Can we do the same as in gridsearch: `isinstance(self.constraints, ClassificationMoment)` ?
- Exponentiated gradient testcase is dependent on 0/1 predictions in Moments, I don't know how to fix this. For now, I'll leave the prediction as 0/1s.


romanlutz:
> Exponentiated gradient testcase is dependent on 0/1 predictions in Moments, I don't know how to fix this. For now, I'll leave the prediction as 0/1s.

You can adjust those test cases to run multiple times, of course. Just parametrize it to have a variety of labels, e.g.,
- 0 and 1
- -1 and 1
- 0 and 2
- (Are text labels allowed? I'm not sure, feel free to try, but it's not mandatory to have that IMO)

SeanMcCarren:
> You can adjust those test cases to run multiple times, of course. Just parametrize it to have a variety of labels, e.g.,

Thanks for the quick response! Yes, I did parameterize, but the problem here is that I added preprocessing in Exponentiated Gradient, and not in all the individual load_data functions. In `test_argument_types_difference_bound` there are some operations on `y` that assume it is 0/1 and don't work for other values/strings, but this should not be the case I think. Additionally, there are some operations that assume that the prediction outputs are 0/1, which also shouldn't necessarily be true. I guess I could change the testcases, but I'm wondering if instead I should more fundamentally change `load_data` (but this seems wrong)

SeanMcCarren:
@adrinjalali @romanlutz Is this important to be done quick? If so, you should know I am a bit stuck, see the issues in the top post.

adrinjalali:
> - If all supplied labels are one of the two possible labels, then we can't infer self.classes_. So, now, we should raise an error. But this breaks the testcases `test_threshold_optimization_degenerate_labels` and `test_single_y_class` (the latter is of gridsearch).

I'd say that's the expected behavior. We should raise if there is only one class.

> - In Exponentiated gradient, how can we know if y is binary or continuous? If it is continuous, we should not apply the encoding. Can we do the same as in gridsearch: `isinstance(self.constraints, ClassificationMoment)` ?

Generally, `is_classifier` should work, if it doesn't we should change the implementation for it to work. Meaning you may end up having two different classes for a regressor and a classifier.


Also, this is not in the milestone, so we're not in too much of a rush for this PR.

TamaraAtanasoska:
since a lot has changed since this PR was open, it might make more sense to approach this issue if necessary from scratch. I will close it and we can reopen it if there is interest.

PR #1007, opened 2022-01-13
Adding ACSPublicCoverage to Fairlearn

britneyting:
This PR adds documentation for the ACSPublicCoverage dataset, which is 1 out of 5 datasets created by [Ding et al](https://arxiv.org/pdf/2108.04884.pdf) to provide an improved alternative to the UCI Adult dataset. In particular, ACSPublicCoverage was created for the task of predicting whether an individual has public health coverage or not.

Note: This pull request depends on @kspieks changes on ACSIncome and thus can't be merged until ACSIncome is incorporated. (Although his commits are included in this pull request because I pulled from his branch into mine.) I created this pull request so my code can be reviewed. Thanks @kspieks for setting the foundation to include all five datasets in the future, and thanks to everyone who reviewed his PR; the comments were very helpful for this PR too! Looking forward to your reviews!

Related to #984


docs/user_guide/datasets/acs_public_coverage.rst
```diff
@@ -0,0 +1 @@
...

   *  - PUBCOV
      - Public health coverage, with PUBCOV == 1 if the individual has public health coverage, else 0

.. topic:: References:

  .. [1] Frances Ding, Moritz Hardt, John Miller, Ludwig Schmidt `"Retiring Adult: New Datasets for Fair Machine Learning" <https://arxiv.org/pdf/2108.04884.pdf>`_,
      Advances in Neural Information Processing Systems 34, 2021.

  .. [2] R. Kohavi and B. Becker. "UCI Adult Data Set." UCI Meachine Learning Repository, 5, 1996.
```
MiroDudik, line 1: There are some issues with end-of-line encoding.

fairlearn/datasets/_fetch_acspubliccoverage.py
```diff
@@ -0,0 +1 @@
...

    if not as_frame:
        data.data = data.data.to_numpy()
        data.target = data.target.to_numpy()
        data.frame = None

    if return_X_y:
        return (data.data, data.target)

    return data
```
MiroDudik, line 1: End-of-line issues as well.

docs/user_guide/datasets/acs_public_coverage.rst
imatiach-msft, line 1: the raw formatting of this file looks really strange, not sure how to fix that

fairlearn/datasets/_fetch_acspubliccoverage.py
imatiach-msft, line 1: same here, the formatting looks very strange

docs/user_guide/datasets/acs_public_coverage.rst
britneyting: Thank you for spotting it; it's been fixed!

docs/user_guide/datasets/acs_public_coverage.rst
```diff
@@ -0,0 +1,197 @@
...
+as an improved alternative to the popular UCI Adult dataset [2]_.
+Briefly, the UCI Adult dataset is commonly used as a benchmark dataset when comparing
+different algorithmic fairness interventions. While UCI Adult is used to predict 
+whether an individual's annual income is above $50,000, the authors created 
+ACSPublicCoverage to predict whether an individual is covered by public health insurance.
+The authors compiled data from the American Community Survey (ACS) Public Use Microdata Sample (PUMS). 
+Note that this is a different source than the Annual Social and Economic Supplement (ASEC) 
+of the Current Population Survey (CPS) used to construct the original UCI Adult dataset.
+Ding et al. [#0]_ filtered the data such that ACSPublicCoverage only includes individuals under 65 years old 
+had an income of less than $30,000.
```
romanlutz, line 18:
```suggestion
Ding et al. [#0]_ filtered the data such that ACSPublicCoverage only includes individuals under 65 years old 
that had an income of less than $30,000.
```
Also, any idea why $30k? Is that relevant to who can get public coverage?

docs/user_guide/datasets/acs_public_coverage.rst
```diff
@@ -0,0 +1,197 @@
...
+
+
+.. _acspubliccoverage_dataset_description:
```
romanlutz, line 21:
```suggestion
.. _acs_public_coverage_dataset_description:
```
just for consistency with the first one

docs/user_guide/datasets/acs_public_coverage.rst
```diff
@@ -0,0 +1,197 @@
...
+
+   *  - PUBCOV
+      - Public health coverage, with PUBCOV == 1 if the individual has public health coverage, else 0
+
+.. topic:: References:
+
+  .. [1] Frances Ding, Moritz Hardt, John Miller, Ludwig Schmidt `"Retiring Adult: New Datasets for Fair Machine Learning" <https://arxiv.org/pdf/2108.04884.pdf>`_,
+      Advances in Neural Information Processing Systems 34, 2021.
+
+  .. [2] R. Kohavi and B. Becker. "UCI Adult Data Set." UCI Meachine Learning Repository, 5, 1996.
```
romanlutz, line 196: Did you copy this from somewhere? If so, it must have that typo there, too.
```suggestion
  .. [2] R. Kohavi and B. Becker. "UCI Adult Data Set." UCI Machine Learning Repository, 5, 1996.
```

fairlearn/datasets/__init__.py
```diff
@@ -5,11 +5,15 @@
 """This module contains datasets that can be used for benchmarking and education."""
 
 
+from ._fetch_acsincome import fetch_acsincome
+from ._fetch_acspubliccoverage import fetch_acspubliccoverage
```
romanlutz, line 9: I had missed that in #1005 @kspieks @britneyting perhaps we should add underscores between these? 
`fetch_acs_income` and `fetch_acs_public_coverage`
The only other dataset so far with multiple words is `fetch_bank_marketing`. Definitely more readable IMO.

docs/user_guide/datasets/acs_public_coverage.rst
```diff
@@ -0,0 +1,197 @@
...
+
+Introduction
+^^^^^^^^^^^^^^^^^
+
+The ACSPublicCoverage dataset is one of five datasets created by Ding et al. [1]_ 
+as an improved alternative to the popular UCI Adult dataset [2]_.
+Briefly, the UCI Adult dataset is commonly used as a benchmark dataset when comparing
+different algorithmic fairness interventions. While UCI Adult is used to predict 
+whether an individual's annual income is above $50,000, the authors created 
+ACSPublicCoverage to predict whether an individual is covered by public health insurance.
```
romanlutz, line 13: Can you elaborate on what public health insurance means? Is that Medicare, Medicaid, or something else? If you find a good resource you can also link to that or cite it.

docs/user_guide/datasets/acs_public_coverage.rst
```diff
@@ -0,0 +1,197 @@
+.. _acs-public-coverage:
```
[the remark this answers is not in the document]
kspieks: My understanding was that tags for cross-referencing should use dashes, or at least that's what the documentation uses as an example. Please help me learn if I'm incorrect https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#cross-referencing-arbitrary-locations

docs/user_guide/datasets/acs_public_coverage.rst
britneyting: Yes, public health insurance in the U.S. means either Medicare or Medicaid. People under age 65 may qualify for Medicare (which is normally reserved for people over age 65) if they have certain disabilities. I will elaborate on this in the documentation

docs/user_guide/datasets/acs_public_coverage.rst
britneyting: Ah, thanks for catching that! @kspieks Think the typo must also be in ACSIncome too then

kspieks: This typo was in my PR at some point because my text editor wasn't checking spelling and I was typing quickly 😬 It's been fixed in my PR now

fairlearn/datasets/__init__.py
kspieks: @romanlutz This change was in my PR. You didn't mess anything :) It's just that @britneyting's PR is based on mine so until mine is merged in, everything related to acsincome is present in her PR

fairlearn/datasets/_fetch_acspubliccoverage.py
```diff
@@ -0,0 +1,146 @@
...
+
+    References
+    ----------
+    .. [1] Ding, F., Hardt, M., Miller, J., & Schmidt, L. (2021).
+       "Retiring Adult: New Datasets for Fair Machine Learning."
+       Advances in Neural Information Processing Systems, 34.
+
+    """
+
+    if not data_home:
```
imatiach-msft, line 116: There is some code to validate abbreviations in fetch above for ACSIncome.  Perhaps that can be extracted into a common utility and re-used for both datasets, once the initial PR is merged, as that validation logic is needed in this new fetch method as well.

docs/user_guide/datasets/acs_public_coverage.rst
britneyting: Upon further research, I actually am not sure. It seems like the income requirements differ by state, so the $30,000 cutoff seems rather arbitrary, but I could have missed something.

kspieks: Oh that's interesting. Page 25 of their [paper](https://arxiv.org/pdf/2108.04884.pdf) just says that the filter was just that PINCP > $30,000.

docs/user_guide/datasets/acs_public_coverage.rst
```diff
@@ -0,0 +1,262 @@
+.. _acs_public_coverage:
+ACSPublicCoverage
+-------------------------
+
+Introduction
+^^^^^^^^^^^^
+
+The ACSPublicCoverage dataset is one of five datasets created by Ding et al. [1]_ 
+as an improved alternative to the popular UCI Adult dataset [2]_.
+Briefly, the UCI Adult dataset is commonly used as a benchmark dataset when comparing
```
kspieks, line 10: Maybe I'm misunderstanding but I thought only ACSIncome was presented as a replacement for UCI Adult while the other datasets were just new datasets to evaluate algorithmic fairness (not direct improvements on UCI Adult?). The paper was titled Retiring Adult and thus only analyzed models trained on ACSIncome since the others have different prediction tasks and thus aren't really comparable? My initial thought is that describing UCI Adult isn't really relevant here, but I welcome other opinions :)

kspieks:
@britneyting don't forget to add a short summary of this PR to `docs/user_guide/installation_and_version_guide/v0.7.1.rst` :)

fairlearn/datasets/_fetch_acs_public_coverage.py
```diff
@@ -0,0 +1,169 @@
...
+                'PR': '72'}
+
+
+class InvalidStateException(Exception):
+    """Exception thrown for invalid state codes."""
+
+    pass
+
+
+def check_states_valid(states):
```
imatiach-msft, line 29: maybe move this to a separate utility file? then both this and the other function can re-use it

docs/user_guide/datasets/acs_public_coverage.rst
britneyting: That's a good point! I have clarified it as an improved fairness benchmark rather than a direct alternative. Thanks for pointing that out!

fairlearn/datasets/_utils.py
```diff
@@ -0,0 +1,43 @@
+# Copyright (c) Microsoft Corporation and Fairlearn contributors.
+# Licensed under the MIT License.
+
+# State Code based on 2010 Census definitions
```
romanlutz, line 4: "... for ACS datasets"
Perhaps also add a link to the folktables repo.

britneyting: Clarified the description!

imatiach-msft (approved):

fairlearn/datasets/_utils.py
```diff
@@ -0,0 +1,46 @@
...
+                'MA': '25', 'MI': '26', 'MN': '27', 'MS': '28', 'MO': '29',
+                'MT': '30', 'NE': '31', 'NV': '32', 'NH': '33', 'NJ': '34',
+                'NM': '35', 'NY': '36', 'NC': '37', 'ND': '38', 'OH': '39',
+                'OK': '40', 'OR': '41', 'PA': '42', 'RI': '44', 'SC': '45',
+                'SD': '46', 'TN': '47', 'TX': '48', 'UT': '49', 'VT': '50',
+                'VA': '51', 'WA': '53', 'WV': '54', 'WI': '55', 'WY': '56',
+                'PR': '72'}
+
+
+class InvalidStateException(Exception):
```
adrinjalali, line 21: not sure if a custom exception just for this is needed. Probably a `ValueError` with a descriptive text would be enough?

adrinjalali (commented):
Please keep the lines in the documentation at max 79 chars long.

You can also apply `black` to the new code you're adding.

Otherwise implementation-wise it looks quite good to me.

adrinjalali (commented):
DUPLICATE

imatiach-msft:
@adrinjalali the author @britneyting finished her internship several weeks ago - would it be possible to merge this PR as is and then apply the 79 max line change and ```black``` afterward?

adrinjalali:
I've applied the changes I requested. @mmadaio would you be able to leave a review?

docs/user_guide/datasets/acs_public_coverage.rst
```diff
@@ -0,0 +1,281 @@
...
+      - Public health coverage, with PUBCOV == 1 if the individual has public
+        health coverage, else 0
+
+
+.. _acs_public_coverage_discussion:
+
+Discussion
+^^^^^^^^^^
+
+As Ding et al. [1]_ notes, algorithmic (un)fairness must be evaluated in
```
hildeweerts, line 213: This paragraph is a bit vague. If with "intervention" we refer to an "algorithmic fairness intervention", it is unclear what is meant with the "context of the intervention". How about something like:

```suggestion
A meaningful evaluation of algorithmic (un)fairness considers the context in which the machine learning model will be used. In this way, we can avoid falling into abstraction traps (see :ref:`abstraction_traps`). For example, the relationship between demographics and healthcare coverage may vary widely by state - a model that works well in one state may not be suitable to use in a different state. 

To investigate how viable ACSPublicCoverage is as a fairness benchmark dataset, we must first ask what real-world tasks might require a prediction of whether an individual is covered by public health insurance.
```

docs/user_guide/datasets/acs_public_coverage.rst
```diff
@@ -0,0 +1,281 @@
...
+context of the intervention, and the effects of different interventions vary
+widely by state. Similarly, datasets should be carefully designed for the
+specific prediction task in mind.
+
+Therefore, to investigate how viable ACSPublicCoverage is as a fairness
+benchmark dataset, we must first ask what real-world tasks might require a
+prediction of whether an individual is covered by public health insurance.
+Usually, the entity that needs this information (e.g. the government checking
```
hildeweerts, line 221:
```suggestion
Usually, the entity that needs this information (e.g., the government checking
```

docs/user_guide/datasets/acs_public_coverage.rst
```diff
@@ -0,0 +1,281 @@
...
+for eligibilty in a social program) will simply ask the individual directly.
+Ding et al. do not provide any examples, so it is difficult to evaluate what
+"fair" even means without context. For now, the rest of this discussion focuses
```
hildeweerts, line 224:
```suggestion
notion of fairness is applicable without further context. 

The remainder of this discussion focuses
```

docs/user_guide/datasets/acs_public_coverage.rst
```diff
@@ -0,0 +1,281 @@
...
+problems that users should be aware of beforehand.
+
+1. A uniform $30,000 threshold ignores differences between states.
+
+    Although the purpose of ACSPublicCoverage is to "predict whether a
+    low-income individual, not eligible for Medicare, has coverage from public
+    health insurance," this dataset should not be blindly used for all
+    prediction tasks that fall in that category. Ding et al. filters ACS PUMS
+    for all individuals whose income is below $30,000, but the 'low-income'
+    threshold differs for every state, so applying one threshold to all states
```
hildeweerts, line 235:
```suggestion
    threshold differs for every state. Consequently, applying one threshold to all states
```

docs/user_guide/datasets/acs_public_coverage.rst
```diff
@@ -0,0 +1,281 @@
...
+"fair" even means without context. For now, the rest of this discussion focuses
+on preliminary data analysis of ACSPublicCoverage and identifies potential
+problems that users should be aware of beforehand.
+
+1. A uniform $30,000 threshold ignores differences between states.
+
+    Although the purpose of ACSPublicCoverage is to "predict whether a
+    low-income individual, not eligible for Medicare, has coverage from public
+    health insurance," this dataset should not be blindly used for all
+    prediction tasks that fall in that category. Ding et al. filters ACS PUMS
```
hildeweerts, line 233: what category?

docs/user_guide/datasets/acs_public_coverage.rst
```diff
@@ -0,0 +1,281 @@
...
+    for all individuals whose income is below $30,000, but the 'low-income'
```
hildeweerts, line 234: What is meant with 'low-income' threshold here? Is this a specific threshold that is set by the government or are we referring more generally to the fact that what is "low" depends on the distribution of income, which can vary across states?

docs/user_guide/datasets/acs_public_coverage.rst
```diff
@@ -0,0 +1,281 @@
...
+    threshold differs for every state, so applying one threshold to all states
+    can lead to flawed results. The paper does not justify the choice of
```
hildeweerts, line 236: Can we be more specific what the consequences would be?

docs/user_guide/datasets/acs_public_coverage.rst
```diff
@@ -0,0 +1,281 @@
...
+    $30,000, but the authors answer in `this Github issue
+    <https://github.com/zykls/folktables/issues/14>`_ that it was calculated
+    approximately in accordance with Affordable Care Act (ACA) eligibilty
+    guidelines qualifying adults with income up to 133% of the 2021 Federal
+    Poverty Line for Medicaid. Note that we examine the 2018 dataset in this
+    article, and 2018 had a different poverty line. Furthermore, not all states
+    expanded Medicaid under the ACA, so some states have a much lower poverty
```
hildeweerts, line 243: Is this poverty line what is meant with the "low-income threshold" mentioned earlier?

docs/user_guide/datasets/acs_public_coverage.rst
```diff
@@ -0,0 +1,281 @@
...
+As Ding et al. [1]_ notes, algorithmic (un)fairness must be evaluated in
+context of the intervention, and the effects of different interventions vary
+widely by state. Similarly, datasets should be carefully designed for the
+specific prediction task in mind.
+
+Therefore, to investigate how viable ACSPublicCoverage is as a fairness
+benchmark dataset, we must first ask what real-world tasks might require a
+prediction of whether an individual is covered by public health insurance.
+Usually, the entity that needs this information (e.g. the government checking
+for eligibilty in a social program) will simply ask the individual directly.
```
hildeweerts, line 222: I would expect that eligibility is based on a predetermined policy, does the government need to ask individuals directly or will they just determine eligibility based on provided information?

fairlearn/datasets/_utils.py
```diff
@@ -0,0 +1,90 @@
...
+}
+
+
+class InvalidStateException(Exception):
+    """Exception thrown for invalid state codes."""
+
+    pass
+
+
+def check_states_valid(states):
```
hildeweerts, line 69: I am not sure why this should be included in Fairlearn - wouldn't it be easier to upload a version of the data set to OpenML that is already in the right format?

fairlearn/datasets/_fetch_acs_public_coverage.py
```diff
@@ -0,0 +1,133 @@
...
+        a pandas DataFrame or Series depending on the number of target_columns.
+        The Bunch will contain a ``frame`` attribute with the target and the
+        data. If ``return_X_y`` is True, then ``(data, target)`` will be pandas
+        DataFrames or Series as describe above.
+
+    return_X_y : bool, default=False
+        If True, returns ``(data.data, data.target)`` instead of a Bunch
+        object.
+
+    states : list, default=None
```
hildeweerts, line 62: I am not sure whether we need to provide this functionality in Fairlearn. It's quite trivial to filter the data set after it is fetched.

docs/user_guide/datasets/acs_public_coverage.rst
```diff
@@ -0,0 +1,281 @@
...
+    for Medicare, but this is not strictly true. While 65 traditionally is the
+    eligibilty age for Medicare, and the data filters for people under 65 years
+    old, younger people with disabilities may still qualify [3]_.
+
+3. Missing features that contribute to low-income status.
+
+    The dataset focuses on individuals, but low-income status also depends on
+    household size, which is not one of the 19 features in ACSPublicCoverage.
+    This could affect prediction tasks.
+
```
hildeweerts, line 265: Although all these points (1,2,3) are valid critiques of the dataset, the current phrasing feels a bit like we are nitpicking the authors, rather than providing valuable insights for people who are learning about fairness or consider using the data set. 

I think point (1) is very important, as this questions the validity of the task as a whole: is a single threshold of 30.000 a valid proxy for low income? We can make a connection to some of our text on construct validity here, I think. 

The other two points might be better off as some final small remarks.

hildeweerts (changes requested):
I have added a few textual suggestions!

I am not sure whether we should add the functionality of automatically filter by state in the `fetch` function - it seems to me that that is quite a trivial thing to do manually, but perhaps I am missing something here (@romanlutz ?).

docs/user_guide/datasets/acs_public_coverage.rst
```diff
@@ -0,0 +1,281 @@
+.. _acs_public_coverage:
+ACSPublicCoverage
+-------------------------
+
+Introduction
+^^^^^^^^^^^^
+
+The ACSPublicCoverage dataset is one of five datasets created by Ding et al.
+[1]_ as an improved fairness benchmark dataset to the popular UCI Adult dataset
```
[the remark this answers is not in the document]
adrinjalali: done, hope they're okay.

docs/user_guide/datasets/acs_public_coverage.rst
adrinjalali: not sure about the reference, please check

docs/user_guide/datasets/acs_public_coverage.rst
adrinjalali: do we have to clarify?

docs/user_guide/datasets/acs_public_coverage.rst
adrinjalali: how would I know :D

docs/user_guide/datasets/acs_public_coverage.rst
adrinjalali: cc @MiroDudik

fairlearn/datasets/_fetch_acs_public_coverage.py
adrinjalali: if it was a single liner I'd agree with you, but I'm happy to keep this since it's a bit more than that.

fairlearn/datasets/_utils.py
adrinjalali: Agreed. Does that mean I need to do it?

adrinjalali (commented):
I don't have answers to most of your questions @hildeweerts , I'm afraid you may be the one who can answer them? OR maybe @MiroDudik?

docs/user_guide/datasets/acs_public_coverage.rst
```diff
@@ -0,0 +1,274 @@
...
+To investigate how viable ACSPublicCoverage is as a fairness benchmark dataset,
+we must first ask what real-world tasks might require a prediction of whether
+an individual is covered by public health insurance. Usually, the entity that
+needs this information (e.g., the government checking for eligibility in a
+social program) will simply ask the individual directly. Ding et al. do not
+provide any examples, so it is difficult to evaluate what notion of fairness is
+applicable without further context. 
+
+The remainder of this discussion focuses on preliminary data analysis of
+ACSPublicCoverage and identifies potential problems that users should be aware
```
mmadaio, line 229: Who are the "users" here, and what are they using it for? People (e.g., data scientists?) who might use this dataset for fairness benchmarks? That may be worth calling out explicitly.

docs/user_guide/datasets/acs_public_coverage.rst
```diff
@@ -0,0 +1,274 @@
...
+example, the relationship between demographics and healthcare coverage may vary
+widely by state - a model that works well in one state may not be suitable to
+use in a different state.
+
+To investigate how viable ACSPublicCoverage is as a fairness benchmark dataset,
+we must first ask what real-world tasks might require a prediction of whether
+an individual is covered by public health insurance. Usually, the entity that
+needs this information (e.g., the government checking for eligibility in a
+social program) will simply ask the individual directly. Ding et al. do not
+provide any examples, so it is difficult to evaluate what notion of fairness is
```
mmadaio, line 225: Definitely agreed that this is difficult for the author of this documentation to do in the abstract, but is the implication also that this would be difficult for users of this dataset?

docs/user_guide/datasets/acs_public_coverage.rst
mmadaio: So, the US Census Bureau does have a definition of a [poverty line](https://www.census.gov/topics/income-poverty/poverty/guidance/poverty-measures.html), but it's calculated at the household level, not individually.

docs/user_guide/datasets/acs_public_coverage.rst
mmadaio: Hmm, my understanding (from [here](https://www.census.gov/topics/income-poverty/poverty/about.html)) is that there _isn't_ a state-level poverty line, only a federal one, and it's updated annually to keep track with inflation via the CPI... but this may not be what the authors are referring to.

docs/user_guide/datasets/acs_public_coverage.rst
```diff
@@ -0,0 +1,274 @@
...
+    Poverty Line for Medicaid. Note that we examine the 2018 dataset in this
+    article, and 2018 had a different poverty line. Furthermore, not all states
+    expanded Medicaid under the ACA, so some states have a much lower poverty
+    line.
+
+    The authors acknowledge that the calculation was a very rough
+    approximation, and ACSPublicCoverage is not designed to accommodate all the
+    subtleties that contribute to the low-income designation and subsequent
+    Medicaid eligibility. As a result, users of ACSPublicCoverage should be
+    aware of the dataset's limitations and possibly modify the data for their
```
mmadaio, line 254: By modify the data, do you mean the low-income threshold, specifically? Or other features too

docs/user_guide/datasets/acs_public_coverage.rst
```diff
@@ -0,0 +1,274 @@
...
+    task.
+
+2. Slight misstatement of data contents.
+
+    The paper states that ACSPublicCoverage excludes individuals not eligible
+    for Medicare, but this is not strictly true. While 65 traditionally is the
+    eligibilty age for Medicare, and the data filters for people under 65 years
+    old, younger people with disabilities may still qualify
```
mmadaio, line 262: This may be a helpful nuance to communicate earlier - lines 14-15 are a bit confusing on this point.

docs/user_guide/datasets/acs_public_coverage.rst
mmadaio: Agreed with @hildeweerts about the framing, but I also think point 3 belongs with point 1, in a larger point about the construct validity of "low-income," given the state-level differences, and that it's calculated at a household level, not individually.

mmadaio (commented):
Thanks so much for your work on this @britneyting! I left a few questions and clarifying comments. 

So sorry for the delay on this review!

romanlutz:
Just to make sure: @britneyting if you're still planning to continue to wrap this up please respond here. I totally understand if you don't have time anymore after your internship. In that case we'll just see if anyone else can continue. No worries at all! The limiting factor was definitely the timely feedback.

In case there's no response I'll just say that we'll wait for at least two weeks (i.e., until 4/20). Looking forward to hearing from you!

adrinjalali:
@romanlutz there hasn't been an activity by the author here for a while, also: https://github.com/fairlearn/fairlearn/pull/1007#issuecomment-1049897451, that's why I've taken over to apply suggestions. But reviewers need to answer the questions themselves here rather than asking for me to apply them.

docs/user_guide/datasets/acs_public_coverage.rst
hildeweerts: To be honest I am still very confused about the prediction task and I feel like we could maybe expand a bit on why it is important that the task makes sense before we can even talk about fairness.

It reminds me a bit about the "studying up" paper; why are we focussing on low-income individuals. How would something like this be used to help them? Or could this prediction task also be misused by, I don't know, predatory health insurance companies or "fraud" detection by the government (i.e., the government uses the model to check whether people are "honest" about whether they are covered by public health insurance)? I have no idea whether these examples make any sense, because I don't know much about the US context, but reading this paragraph did not clear anything up for me.

@mmadaio @MiroDudik @romanlutz being more familiar with the US context, do you have thoughts?

docs/user_guide/datasets/acs_public_coverage.rst
hildeweerts: I think the main audience would be either researchers who want to benchmark some fair-ml approach, or people who want to use the dataset as an educational resource for educators/blog post writers who are looking for a dataset to use in examples/tutorials/etc. 

I have attempted to rewrite the piece a bit, but I am not sure whether everything I have written is correct. It would be great if some of our US citizens could check whether my interpretation is correct.

```
Another potential issue that researchers and educators should be aware 
of before using this dataset as a benchmark or educational resource, is the
construct validity of the measurements (see also :ref:`construct_validity`).
The purpose of ACSPublicCoverage is to "predict whether a low-income 
individual, not eligible for Medicare, has coverage from public health 
insurance".
Here, Ding et al. define 'low-income' as individuals whose income is 
below $30.000.
This income threshold `was calculated
<https://github.com/zykls/folktables/issues/14>`_ 
approximately in accordance with Affordable Care Act (ACA) eligibility 
guidelines, qualifying adults with income up to 133% of the 2021 Federal 
Poverty Line for Medicaid.
An important question is to what extent the $30.000 threshold is a valid
measurement of low income, in relation to public health insurance 
eligibility.
Firstly, the dataset contains data from 2018, in which the federal poverty 
line was slightly different compared to 2021.
More importantly, however, the federal poverty line is the same across
all states, whereas the low-income threshold used to determine 
eligibility for Medicaid can differ per state.
In particular, not all states expanded Medicaid under the ACA. 
Consequently, some states have a much lower low-income threshold 
compared to the federal poverty line.
Additionally, the dataset concerns individuals, but low-income status also
depends on household size, which is not one of the 19 features in 
ACSPublicCoverage.
We observe a similar issue with the eligibility for Medicare, which 
is implemented through the exclusion of individuals of 65 years and older.
While 65 traditionally is the eligibilty age for Medicare, younger 
people with disabilities may still qualify :footcite:`usgov2014who`.
Depending on how the predictions are used, the mismeasurement of 
low income and Medicare eligibility may result in allocation harms 
based on state or disability status (see also :ref:`types_of_harms`).
```

From this page, it looks like Medicaid eligibility may differ per state: https://www.healthcare.gov/lower-costs/

fairlearn/datasets/_utils.py
hildeweerts: I think I read this incorrectly. It seems to check whether the states supplied by the user (for filtering) are valid, not the dataset itself, which makes sense if we keep the `states` argument.

hildeweerts (commented):
A few more comments and suggestions.

docs/user_guide/datasets/acs_public_coverage.rst
romanlutz: @britneyting @kspieks @imatiach-msft and I were discussing this back in January as well. I share your concerns. In fact, I feel the same about the income prediction and all the other tasks mentioned in the paper. The context in which this task is useful is not obvious to me. That said, I do believe that this is something we can discuss in the user guide. It leads to the bigger question of "Should benchmarks always be realistic?" I have thoughts on that... Can we perhaps push people to more realistic datasets by outlining shortcomings?

Anyway, curious to hear what you and others think, @hildeweerts .

adrinjalali: we can also open an issue and discuss it further and not block this PR for this.

adrinjalali (commented):
I'm not really sure how to move forward here.

docs/user_guide/datasets/acs_public_coverage.rst
hildeweerts: @romanlutz given that you have thought about this extensively, do you have a suggestion to rephrase this paragraph to make it clearer what our concerns are?

I like @adrinjalali's suggestion to open a new issue to discuss the more general problem more thoroughly.

fairlearn/datasets/_fetch_acs_public_coverage.py
hildeweerts: Fair enough. In that case let's leave it in :)

fairlearn/datasets/_utils.py
hildeweerts: Agreed!

docs/user_guide/datasets/acs_public_coverage.rst
romanlutz: I'll have to think about this a bit more to phrase it right. It may take me a while, though. 

I'm not sure whether we should proceed with this anyway (as we already have Adult and the other variation) and open an issue for describing this sort of problem, or block this PR instead. I see the concern (as mentioned earlier), but given that we have others with the same issue it feels like an odd point to block. On the other hand, that opens the door for more and more contributions that don't interrogate this problem (at least until we capture this in some way).

hildeweerts: I am also not sure... if we block the PR now I don't think it will get merged in the foreseeable future (if ever), but if we do not touch upon this issue at all we also create more "technical" debt.

I think it should be possible to rephrase the current text a bit to make our concerns clearer, without interrogating the problem in detail. E.g., I think some examples can go a long way, e.g., who is using it (e.g., a governmental organization versus industry) and for what purpose (e.g., a model used to distribute benefits versus a model that is used for fraud detection) already makes a huge difference in how we think about fairness.

romanlutz: I agree 100% in the general case. 

In this particular one I'm still wondering what that might be. Any government org that needs to know your insurance status can just ask. Same for companies. And if they don't feel like they can ask they probably aren't meant to know and it's a nefarious or at least borderline/grey area activity. Besides, I just can't wrap my head around why one would have ALL THE OTHER features but just not this one? That seems like a significant amount of detail we know about these individuals, how come we know all that and still need to guess the insurance status? Again, if we lawfully knew this much for a good reason they'd probably tell us that status, too. It's entirely possible my reasoning is flawed somewhere, of course.

hildeweerts: You raise very good points - especially because it is based on such an apparently well-defined concept, the need for a predictive model seems... odd. I'm inclined to agree with your former comment and say we should just not proceed with this at all.

It's a pity of all the work that went into this, but perhaps sunk cost is not the best argument for including the dataset.

adrinjalali:
Based on https://github.com/fairlearn/fairlearn/pull/1007#discussion_r847279809, I guess we're not going ahead with this PR then. Closing, happy to reopen if it resurfaces.

PR #1011, opened 2022-01-24
Fixed quickstart page

britneyting:
Before:
- Fairlearn quickstart page showed both recall and true positive rate, which are the same metrics, on its plots.

Now:
- Replaced true positive rate with true negative rate on the plots.

### DIFF
--- a/examples/plot_quickstart.py
+++ b/examples/plot_quickstart.py
@@ -12,6 +12,7 @@ from fairlearn.metrics import (
     MetricFrame,
     false_positive_rate,
     true_positive_rate,
+    true_negative_rate,
     selection_rate,
     count,
 )
@@ -35,7 +36,7 @@ metrics = {
     "precision": precision_score,
     "recall": recall_score,
     "false positive rate": false_positive_rate,
-    "true positive rate": true_positive_rate,
+    "true negative rate": true_negative_rate,
     "selection rate": selection_rate,
     "count": count,
 }
--- a/examples/plot_quickstart.py
+++ b/examples/plot_quickstart.py
@@ -11,14 +11,13 @@ MetricFrame visualizations
 from fairlearn.metrics import (
     MetricFrame,
     false_positive_rate,
-    true_positive_rate,
-    true_negative_rate,
+    false_negative_rate,
     selection_rate,
     count,
 )
 import pandas as pd
 from sklearn.datasets import fetch_openml
-from sklearn.metrics import accuracy_score, precision_score, recall_score
+from sklearn.metrics import accuracy_score, precision_score
 from sklearn.tree import DecisionTreeClassifier
 
 data = fetch_openml(data_id=1590, as_frame=True)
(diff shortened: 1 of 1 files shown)


docs/user_guide/datasets/acs_income.rst
```diff
@@ -0,0 +1,162 @@
+.. _acsincome-data:
```
romanlutz, line 1: Thanks for your PR! However, you need to remove all these other files from the other PR. You can do that manually or by running `git rebase -i HEAD~10` (or replace 10 with whatever number of commits you want to drop) and then write "drop" for the corresponding commits. Whatever suits you better.

romanlutz:
Thanks @britneyting for the update to remove the redundancy!

Perhaps it's just me, but I find FPR, FNR, TPR, TNR way more intuitive than recall and precision. So if you ask me I'd get rid of those two and keep just FPR, FNR (since they contain the information on TPR, TNR). One can perhaps make the argument to keep precision since it's different.

@fairlearn/fairlearn-maintainers any thoughts?

hildeweerts:
> Perhaps it's just me, but I find FPR, FNR, TPR, TNR way more intuitive than recall and precision. So if you ask me I'd get rid of those two and keep just FPR, FNR (since they contain the information on TPR, TNR). One can perhaps make the argument to keep precision since it's different.

Do you mean intuitive in general or for this case? If you mean the latter I agree, precision and recall don't make a lot of sense unless we frame the Adult dataset as some sort of recommendation/information retrieval task. In the spirit of ROC curve analysis you could argue for keeping tpr/fpr, but comparing fpr/fnr in a single figure seems more sensible to me as well.

adrinjalali:
I would agree with Roman in general :D

britneyting:
> Thanks @britneyting for the update to remove the redundancy!
> 
> Perhaps it's just me, but I find FPR, FNR, TPR, TNR way more intuitive than recall and precision. So if you ask me I'd get rid of those two and keep just FPR, FNR (since they contain the information on TPR, TNR). One can perhaps make the argument to keep precision since it's different.
> 
> @fairlearn/fairlearn-maintainers any thoughts?

Thanks for bringing this up, that makes sense! I've changed the metrics to include fpr/fnr and removed recall but kept precision. Is there a reason to do fpr/fnr as opposed to tpr/tnr?

romanlutz (approved):
Thanks for adjusting this @britneyting !

I personally prefer the false positive/negative rates, but it's the same as using true positive/negative rates since they're complements (TPR = 1-FNR and TNR = 1-FPR). I just like imagining what a certain type of error means in our application scenario. For example, if I misclassify the patient as healthy even though they have a disease then that can cause the patient harm (because the disease can potentially progress). That would be a false negative. Or if we misclassify a healthy patient as having a disease then we're wasting medical resources / staff time on someone who doesn't need them (and perhaps don't give them to others that do), which would be a false positive. In many cases one might also want to weigh them differently, e.g., a false negative seems worse to me here, but if the hospital is already at capacity it might be worse to waste the scarce resources (false positive).

romanlutz:
I think the PR gate failure is unrelated to your PR. I'll try to look into that asap

hildeweerts (approved):

PR #1014, opened 2022-01-28
Fix other ML packages tests

riedgar-ms:
Fix a couple of issues with our tests with other ML packages:

- The latest version of `pandas` objects to how `xgboost` was using it (specifically `FutureWarning: pandas.Int64Index is deprecated and will be removed from pandas in a future version. Use pandas.Index with the appropriate dtype instead.`). Temporarily exclude this version of `pandas` with an upper bound
- It appears that `conda-forge` had the latest `scikeras` but not the latest `tensorflow` to match it. Switch to installing both from `pip`

Have filed #1015 as a reminder to watch `xgboost` for an update to be compatible with the latest `pandas`.

### DIFF
--- a/test_othermlpackages/conda-tensorflow.yaml
+++ b/test_othermlpackages/conda-tensorflow.yaml
@@ -7,4 +7,5 @@ dependencies:
 - pip>=19.1.1
 - pytest
 - pip:
-   - scikeras[tensorflow] # Latest TF not on conda-forge
+   - scikeras
+   - tensorflow # Latest TF not on conda-forge
--- a/test_othermlpackages/conda-tensorflow.yaml
+++ b/test_othermlpackages/conda-tensorflow.yaml
@@ -6,6 +6,5 @@ dependencies:
 - python==3.8
 - pip>=19.1.1
 - pytest
-- tensorflow>=2.4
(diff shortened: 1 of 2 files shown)


riedgar-ms:
Build: https://dev.azure.com/fairlearn/Fairlearn/_build/results?buildId=1336&view=results

romanlutz (approved):

riedgar-ms:
Since this is just a build fix, I'll merge on Monday unless @fairlearn/fairlearn-maintainers object.

test_othermlpackages/conda-tensorflow.yaml
```diff
@@ -6,6 +6,6 @@ dependencies:
 - python==3.8
 - pip>=19.1.1
 - pytest
-- tensorflow>=2.4
 - pip:
    - scikeras
+   - tensorflow # Latest TF not on conda-forge
```
adrinjalali, line 11: Or alternatively just `scikeras[tensorflow]`

riedgar-ms: Doesn't work for some reason :-(

adrinjalali (approved):

PR #1032, opened 2022-02-22
Add sphinx bibtex to provide bib file based citations

### LINKED ISSUE #936: DOC Use BibTeX for citations

When we wish to cite papers in our documentation, we are currently writing out the citations manually. Not only is this tedious, but it is also highly error prone (since the links in the text are numbers along the lines of `[4]_`). This is particularly acute if merge conflicts occur in the 'References' section, since then the whole page must be cross checked, to ensure that all the links to the references point at the right paper.

Somewhat over 30 years ago, BibTex was invented to help with precisely this issue. Furthermore, online paper databases often provide BibTeX snippets for inclusion in `.bib` files. There is also a [BibTeX package for Sphinx](https://sphinxcontrib-bibtex.readthedocs.io/en/latest/quickstart.html) which looks to be under active development. A possible solution suggests itself.

iofall:
Fixes #936 

This PR adds the `sphinxcontrib-bibtex` extension which allows us to use citations from a `.bib` file.

**Usage**
1. Check if your required reference already exists in the `docs/refs.bib` file. If not add your bibtex snippet to the `refs.bib` file. To keep the `.bib` file organized, please ensure that you do not introduce duplicate entries.
2. The convention for citation keys should be `<FirstAuthorSurname><Year><PaperKeyword>`.
3. After you have added to the `refs.bib` file, you can reference your citation using the specified citation key using the syntax:
```
:footcite:`citation-key` 
```
The above citation will be rendered as `[Number]`.
4. To add the bibliography, you need to use `:footbibliography:` directive at the bottom of your file. This will list all the citations for the current document automatically!

**Assumptions**
- I used `:footcite:` instead of the normal `:cite:` because of the way references are resolved.
- I have currently used the `<FirstAuthorSurname><Year><PaperKeyword>` convention for citation keys.

**Caveats**
- If you see multiple URLs, change your bibtex to only show one URL. In case of `arXiv` references, the library somehow creates URLs for the `arXiv` paper and shows both the generated URL as well as the URL under the URL parameter in `.bib` file. To avoid this, either only specify a `URL` parameter or `eprint` along with `eprinttype`.

**To Do**
- Change references throughout the documentation to the new format. This could be a good sprint issue especially for beginners.
- Change the version check function. See the comments below.

**Look out for**
- While testing using `sphinx multiversion build`, remember to whitelist the PR branch.
- For now only `mitigations.rst` has the new citations.

**Example Output**

- **Before**

- **After**


docs/conf.py
```diff
@@ -245,3 +245,22 @@ def linkcode_resolve(domain, info):
...
+# by comparing version numbers.
+def check_if_v071dev0():
+    """Check to see if current version being built is v0.7.1.dev0"""
+    result = False
+
+    if fairlearn.__version__ == "0.7.1.dev0":
+        print("Detected 0.7.1.dev0 in fairlearn.__version__")
+        result = True
+
+    return result
```
iofall, line 259: Before the PR gets merged, we will have to change this version check. I was thinking of using a `greater than comparison` here. I did look into this and realized we might have to use the `packaging` library as suggested in [Stackoverflow](https://stackoverflow.com/questions/11887762/how-do-i-compare-version-numbers-in-python). However I am not sure if we want to introduce another dependency?

riedgar-ms:
Thank you for starting to look at this! The versioning of the docs is something we've talked on and off about for a while; this might help animate those discussions.

docs/conf.py
adrinjalali: since this is a doc build dependency, I don't mind adding it.

docs/refs.bib
```diff
@@ -0,0 +1,51 @@
...
+  title     = {Equality of Opportunity in Supervised Learning},
+  journal   = {NeurIPS},
+  volume    = {abs/1610.02413},
+  year      = {2016},
+  url       = {https://papers.nips.cc/paper/6374-equality-of-opportunity-in-supervised-learning.pdf},
+  eprinttype = {arXiv},
+  timestamp = {Mon, 13 Aug 2018 16:47:05 +0200},
+  biburl    = {https://dblp.org/rec/journals/corr/HardtPS16.bib},
+  bibsource = {dblp computer science bibliography, https://dblp.org}
+}
```
adrinjalali, line 51: new line at the end of the file missing

adrinjalali (commented):
I'm happy to use packaging for the version check, and it LGTM otherwise.

docs/refs.bib
iofall: Thanks, I have added this.

docs/conf.py
iofall: Yes, I later realized that packaging is already used by `setuptools` so most people should already have it installed. But to be explicit, I have added it to the `requirements-dev.txt`.

docs/conf.py
```diff
@@ -245,3 +247,22 @@ def linkcode_resolve(domain, info):
         }
     }
 }
+
+
+def check_if_v071dev0():
+    """Check to see if current version being built is >= v0.7.1.dev0."""
+    result = False
+
+    if parse(fairlearn.__version__) >= parse("0.7.1.dev0"):
```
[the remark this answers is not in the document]
iofall: Yes that will work, thank you. I will make these changes and we should be good.

adrinjalali (approved):

docs/refs.bib
```diff
@@ -0,0 +1,51 @@
+Mitigations.rst
```
romanlutz, line 1: I'm in favor of this, but the comments in this file don't add anything from my point of view. In fact, these papers are cited all over the place rather than just in one file. Keeping such a mapping in sync sounds tricky and I don't really see the value it would add since you can just search for the identifier to find where it's cited. Again, just the comments (including paper comments below), everything else is great! I assume you just copied from Google Scholar or somewhere similar?

docs/user_guide/mitigation.rst
```diff
@@ -649,17 +649,9 @@ Exponentiated Gradient
 Grid Search
 ~~~~~~~~~~~
 
-.. topic:: References:
+.. _references:
```
romanlutz, line 652: Is this smart enough to include only the cited ones on this page? It will be important moving forward when there are perhaps dozens of citations across the website, but only say, a handful, on this particular page.

docs/conf.py
```diff
@@ -245,3 +247,22 @@ def linkcode_resolve(domain, info):
...
+    if parse(fairlearn.__version__) > parse("0.7"):
+        print("Detected version > 0.7 in fairlearn.__version__")
+        result = True
+
+    return result
+
+
+# Setup for sphinx-bibtex
+
+# Only use sphinx-bibtex if version is above 0.7
```
romanlutz, line 265: It's not used in other versions, so it should just be fine(?) or am I missing something?

Oh or is this about the refs file not existing?

Just trying to make sure I get it 🙂

romanlutz (commented):
Thanks for getting the bibtex conversion started! This is great!!!

Make sure to add a line to the changelog (part of docs/user_guide, there's a file somewhere for each version, so just add to the latest one).

docs/refs.bib
MiroDudik: Agreed that we probably don't need comments like this. The entries look like a merge of DBLP versions, which I think is sensible. We shouldn't use Googe Scholar as a source for citations. However, the main decision here is perhaps whether we are okay with the citation ids of the form
```
<last-name-of-first-author><year><first-important-title-word>
```
I'm fine with that.

docs/refs.bib
```diff
@@ -0,0 +1,51 @@
...
+
+Agarwal, Beygelzimer, Dudik, Langford, Wallach “A Reductions Approach to Fair Classification”, ICML, 2018.
+@article{agarwal2018reductions,
```
MiroDudik, line 4: I think we should be using something more similar to this: https://dblp.uni-trier.de/rec/conf/icml/AgarwalBD0W18.html?view=bibtex&param=0
(so `@inproceedings` rather than `@article`)

docs/refs.bib
```diff
@@ -0,0 +1,51 @@
...
+  year      = {2018},
+  eprinttype = {arXiv},
+  eprint    = {1803.02453},
+  timestamp = {Mon, 13 Aug 2018 16:47:08 +0200},
+  biburl    = {https://dblp.org/rec/journals/corr/abs-1803-02453.bib},
+  bibsource = {dblp computer science bibliography, https://dblp.org}
+}
+
+Agarwal, Dudik, Wu “Fair Regression: Quantitative Definitions and Reduction-based Algorithms”, ICML, 2019.
+@article{agarwal2019fair,
```
MiroDudik, line 22: see: https://dblp.uni-trier.de/rec/conf/icml/AgarwalDW19.html?view=bibtex&param=0

docs/refs.bib
```diff
@@ -0,0 +1,51 @@
...
+  year      = {2019},
+  eprinttype = {arXiv},
+  eprint    = {1905.12843},
+  timestamp = {Mon, 03 Jun 2019 13:42:33 +0200},
+  biburl    = {https://dblp.org/rec/journals/corr/abs-1905-12843.bib},
+  bibsource = {dblp computer science bibliography, https://dblp.org}
+}
+
+Hardt, Price, Srebro “Equality of Opportunity in Supervised Learning”, NeurIPS, 2016.
+@article{hardt2016equality,
```
MiroDudik, line 38: see: https://dblp.uni-trier.de/rec/conf/nips/HardtPNS16.html?view=bibtex&param=0
[with NIPS -> NeurIPS]

MiroDudik (changes requested):
I just checked the specific bib entries that are being used.

docs/refs.bib
iofall: > Keeping such a mapping in sync sounds tricky and I don't really see the value it would add since you can just search for the identifier to find where it's cited. 

Ok that makes sense, I think the step 1 should cover the duplicates, no need for comments. If someone is adding the a new entry, they should check if an entry already exists. I'll modify the PR and the description.

> 1. Check if your required reference already exists in the `docs/refs.bib` file.

iofall: > I assume you just copied from Google Scholar or somewhere similar?

I used the bibtex entries available through arXiv and modified some of the entries to showcase how different formats look.

> The bibtex entries below look like a merge of DBLP versions, which I think is sensible. We [shouldn't use Googe Scholar](https://scholar.hasfailed.us/) as a source for citations. The main decision is perhaps whether we are okay with the citation ids of the form

Thank you for the reference, I did not know that about Google Scholar. I used the ones available through arXiv, they link to the DBLP hosted records. For example - [arXiv Page Fairlearn](https://arxiv.org/abs/1803.02453) and its [DBLP Bibtex](https://dblp.uni-trier.de/rec/journals/corr/abs-1803-02453.html?view=bibtex). However, they always link back to the arXiv page for the paper.

**Takeaway**: It might be helpful to establish where we source our bibtex snippets from.

docs/user_guide/mitigation.rst
iofall: I mirrored what we already had in the documentation. Right now the structure is - Each page like `mitigation.rst` will only list citations that are used on this page.

docs/conf.py
iofall: Yes, you are right, it will be fine. However, sphinx-multiversion uses the latest `conf.py` and for other versions we don't have a `refs.bib` file which generates a warning. Although it will not stop the documentation build, this can worry some new users, so I made it explicit to use `sphinx-bibtex` only when the required version is available.

image
It is really minor but it looks something like this.

docs/refs.bib
iofall: Got it I will change to `@inproceedings`, do you mean to use the `condensed` format? It does not have URLs so the references will not link back to the papers.

iofall: For reference, the citations will look like this:

**Standard**


**Condensed**

docs/user_guide/mitigation.rst
MiroDudik: That sounds great. I'm pretty sure that's what @romanlutz had in mind too.

docs/refs.bib
MiroDudik: I think that having a URL is definitely nice to have, but the full standard citation seems very verbose. I would go for something in between, maybe (also fixing my name, which is getting a bit mangled):
```
@inproceedings{agarwal2018reductions,
  author    = {Alekh Agarwal and
               Alina Beygelzimer and
               Miroslav Dudík and
               John Langford and
               Hanna M. Wallach},
  title     = {A Reductions Approach to Fair Classification},
  booktitle = {Proceedings of the 35th International Conference on Machine Learning,
               {ICML} 2018},
  pages     = {60--69},
  year      = {2018},
  url       = {http://proceedings.mlr.press/v80/agarwal18a.html},
  biburl    = {https://dblp.org/rec/conf/icml/AgarwalBD0W18.bib},
  bibsource = {dblp computer science bibliography, https://dblp.org}
}
```

romanlutz: FWIW I'd rather add more info than less. It's at the bottom of the page anyway...

docs/refs.bib
romanlutz: I believe 2016 preceeds the renaming of the conference (?). Is it okay to call it by the new name regardless? I'm obviously very much in favor, just trying to understand the conventions.

docs/refs.bib
iofall: > (also fixing my name, which is getting a bit mangled

Sorry about that, I'll change it.

I think let's go with the condensed version, we will add in the `url` field to it. The verbosity mainly comes from the `booktitle`.

It does not look too verbose, nor too less -
image

adrinjalali:
So it seems there's quite a bit of discussion on how to reference papers. What I've always done is to cope the bibtex entry from google scholar and let it be formatted by whatever library's rendering them.

But it seems @MiroDudik doesn't like that. I understand in a paper that may be important, but I'm not sure if this much discussion and work is worth it everytime we'd be adding new citations. This library is not an academic article, so I'm not sure why the sensitivity on the details of the citations are raised here. For me, as long as people can copy/paste our reference and find it wherever they find their articles, it's good enough.

MiroDudik (approved):
Looks good to me.

docs/refs.bib
MiroDudik: In this case, retroactive renaming is okay... we're providing a link, so it shouldn't matter.

Issue #999, opened 2021-11-30
Error Bars/Uncertainty Quantification for MetricFrame


Harsha-Nori:
## Is your feature request related to a problem? Please describe.

There is currently no built-in way of assessing the volatility or uncertainty associated with metrics generated via `MetricFrame`. This has important implications for the interpretation of any observed values or differences between groups. For example, if a metric changes significantly under small changes (e.g. through resampling), the difference might be explained fully by something like [sampling error](https://en.wikipedia.org/wiki/Sampling_error). 

For example, here is a plot from the FairLearn documentation: 

[<img src="https://fairlearn.org/v0.7.0/_images/sphx_glr_plot_quickstart_002.png" width="600" />](https://fairlearn.org/v0.7.0/_images/sphx_glr_plot_quickstart_002.png)

The differences between the two groups in `accuracy` and `selection_rate` appear to be larger than the differences in `precision` and `recall`. However, it could just be that metrics like `selection_rate` are inherently higher variance metrics than `precision`, which might contextualize why the magnitude of the differences is larger.

## Describe the solution you'd like

I'd like to propose a [bootstrap](https://en.wikipedia.org/wiki/Bootstrapping_(statistics)) based solution built directly into `MetricFrame`. Bootstrap is a conceptually simple procedure that involves building an empirical distribution of a given metric/dataset by repeatedly:

1. Sampling up to the full size of the dataset _with replacement_
2. Re-calculating the metric on this new sample

When repeated hundreds or thousands of times, we can study the empirical distribution and produce useful summaries like [confidence intervals](https://en.wikipedia.org/wiki/Bootstrapping_(statistics)#Deriving_confidence_intervals_from_the_bootstrap_distribution) for the metric.  

There are several reasons why bootstrap might be an appropriate solution here:

- It is widely adopted and studied, making it a statistically defensible choice
- Bootstrap is agnostic to characteristics of the dataset and metric, making it largely assumption-free and widely applicable
- It is straightforward to explain and implement as a higher level layer on top of the core MetricFrame subroutines

However, it does come with a few downsides. Primarily, bootstrap is very computationally intensive. By definition, it is 100x - 1000x+ as intensive as running the core MetricFrame routine, which makes performance a key concern. This might be mitigated by:

- Having the bootstrap procedure be opt-in, and defaulted to "off". This has the added benefit of being completely backwards compatible.
[...]

PR #1038, opened 2022-03-03
[WIP] Uncertainty Intervals for MetricFrame via Bootstrap

Harsha-Nori:
This PR is an implementation of the discussion around issue #999. Currently it supports the following operations:

-Use of [bootstrap](https://en.wikipedia.org/wiki/Bootstrapping_(statistics)) to estimate the empirical distribution of any scalar metric passed to MetricFrame. Bootstrap settings are controlled by two new init parameters: `n_boot`, which determines how many replications of the dataset to generate with replacement, and `n_jobs`, which controls how many parallel processes to use in the bootstrap operation.

-Confidence interval estimation as a summary of the empirical bootstrap distribution. This is controlled by the `ci` and `ci_method` parameters, which offer options of which quantiles to sample and whether or not estimates should be bias-corrected. Many users can use simple parameter settings like `ci=0.95` for a 95% confidence interval on each MetricFrame statistic, while advanced users can query specific quantiles (e.g. `ci=[0.25, 0.5, 0.75]`) for further study. These results are calculated during initialization and are stored in new attributes on MetricFrame suffixed in `.*_ci`. 

-----------

Some significant changes have been made to MetricFrame internals in order to support this set of operations. Firstly, because we build the bootstrap distribution during initialization and chose not to pin it on the object, some operations like `.difference` are calculated during init. Secondly, because bootstrapping involves calculating the same operation on different (resampled) inputs, I've changed some internal functions like `def __group()` to work on any set of inputs, and not be hard-coded to use pinned attributes like `self.overall` and `self.by_group`. I've also refactored some general code into new private functions to make repeated calls easier. These should have no breaking changes to an end user. 

The new `.*_ci` attributes are all lists of tuples, with each tuple representing the quantile of the distribution in the first element, and the corresponding values for each metric in the second. For example, if a user requests a 90% confidence interval with `ci=0.9`, the output of `.overall_ci` will look like:

```python
.overall_ci = [
   (0.05, pd.DataFrame({...})),
   (0.95, pd.DataFrame({...}))
]
```

where each of the inner frames exactly parallel the structure of `.overall`. This API design hopefully makes it easy for users to adopt confidence intervals into their workflow -- any code that runs on the "main" MetricFrame objects will also work on the .ci versions.

-----------

This PR is still a work in progress -- there's undoubtedly more testing, documentation, and potential refactors to do. But I wanted to put it up to start soliciting some feedback from the maintainers and community. 

There's a few concrete questions I have:

1. How do people feel about the user facing API portions (init parameters, exposed attributes, datastructures of output objects, etc.)?
2. There is some gnarly code in the bias corrected bootstrap quantile calculation (in ._bootstrap.py) that I can't seem to vectorize. It doesn't seem to be a performance bottleneck on datasets I've tested with, but I'd appreciate an extra set of eyes on potentially making it better.
3. Currently the code automatically calculates and detects metrics with high (statistical) bias as part of the bootstrap procedure. I currently throw a generic warning right now that these exist, but some guidance on how to better format these warnings to users would be appreciated!
4. What type of documentation should accompany this PR, and where? 

And of course, any general thoughts would be appreciated as well 🙂!


fairlearn/metrics/_metric_frame.py
```diff
@@ -307,46 +331,41 @@ class MetricFrame:
...
+        all_data = pd.DataFrame(dtype='object')
+
         y_t = _convert_to_ndarray_and_squeeze(y_true)
         y_p = _convert_to_ndarray_and_squeeze(y_pred)
 
-        all_data = pd.DataFrame.from_dict(
-            {"y_true": list(y_t), "y_pred": list(y_p)}
-        )
+        all_data['y_true'] = list(y_t)
+        all_data['y_pred'] = list(y_p)
```
riedgar-ms, line 355: Curious.... why this change?

fairlearn/metrics/_metric_frame.py
```diff
@@ -307,46 +331,41 @@ class MetricFrame:
...
-        annotated_funcs = self._process_functions(
-            metrics, sample_params, all_data
-        )
+        annotated_funcs = self._process_functions(metrics, sample_params, all_data)
 
         # Now, prepare the sensitive features
-        sf_list = self._process_features(
-            "sensitive_feature_", sensitive_features, y_t
-        )
+        sf_list = self._process_features("sensitive_feature_", sensitive_features, y_t)
```
riedgar-ms, line 360: I think some of these changes may be `black` and `autopep8`/`flake8` arguing

test/unit/metrics/test_metricframe_comprehensive.py
```diff
@@ -5,6 +5,7 @@
 import pandas as pd
 import pytest
 import sklearn.metrics as skm
+import time
```
riedgar-ms, line 8: I think you can get rid of the things in here. Those look like leftovers from when I was benchmarking the performance improvements.

fairlearn/metrics/_bootstrap.py
```diff
@@ -0,0 +1,110 @@
...
+    bootstrap_std = np.sqrt(np.var(bootstrap_runs, axis=0, dtype='float64'))
+    bias = bootstrap_mean - sample_estimate
+    bias_adjusted_mean_estimate = 2*sample_estimate - bootstrap_mean # Might be useful to publish?
+
+    # Efron and Tibshirani (1993) guidance on identifying "large" bias via bootstrap
+    high_bias_regions = np.abs(bias) >= (0.25 * bootstrap_std)
+    if np.any(high_bias_regions):
+        #TODO: Add bias indication regions into warning. Simply printing all "high_bias_regions" is overwhelming.
+        warn(
+            f"""Some statistics calculated by MetricFrame are detected to have high bias
```
riedgar-ms, line 66: Please be more definite in the error message - you know which metrics came back with large errors, so tell the user which ones.... is that what you're getting at in the TODO?

riedgar-ms (commented):
Thanks.... only a few comments so far, but this is not something with which I'm familiar.

fairlearn/metrics/_bootstrap.py
Harsha-Nori: > is that what you're getting at in the TODO?

Yes exactly -- I was struggling a bit with how to format the warning properly without creating a (potentially) overwhelmingly large message. We're computing so many different disaggregated metrics -- we calculate bias in each cell of .overall, .by_group, difference to overall/between groups, and ratio to overall/between groups -- and any one of the cells in any of the frames might have high statistical bias. Actionable information for a user would include: 

- Name of the frame (.overall vs. difference_between_groups, etc.)
- Name of the metric
- index values (control features and sensitive attributes that make up the calculation)

and there might be potentially dozens of these. Wasn't sure if I should try to format all this info line by line, which could lead to a large warning message, or throw a generic warning and capture all of the details in a property that a user could inspect later

riedgar-ms: Understood - I hadn't realised just how large the list of potential 'hits' could be. I can see that extra discussion is warranted around that. Could you format them into a list, attach it as an extra field to the object, and tell users to inspect it?

riedgar-ms:
Circling back on this.... I would like to get this in (although I don't think we should hold v0.8.0 for it). I have already taken a little bit of the refactoring in #1110 )

@MiroDudik , I assume that the API was developed in collaboration with yourself, and so you like it?

@adrinjalali what do you think of the API?

For my part.... I'm not overjoyed at making the constructor even more expensive. However, within the `MetricFrame` paradigm, I'm short on alternative suggestions.

adrinjalali (commented):
should we introduce an `metricframe.evaluate()` method to remove computations in the constructor then?

riedgar-ms:
It's a possibility @adrinjalali , but has its own downsides. Mainly (as I recall from the discussions), since we generally expect a new `MetricFrame` to be created each time a user wants metrics, having an `evaluate()` method just adds an extra step. Even if users did want to keep the same object, then the only real 'state' for the constructor is the collection of metric functions. More likely, users are going to construct the frame, look at some of the metrics (perhaps via `difference()` and `ratio()`) and then move on to the next thing. That's not like estimators, where it's reasonable to expect `predict()` to be called multiple times.

The other major alternative is to have a `create_metric_frame()` factory function which effectively replaces the constructor. Then `MetricFrame` is only the return type of that function (its value is in the `difference()` and `ratio()` methods). However, that's a bunch of complication just to avoid a 'heavy' constructor.

If you feel differently (or I've missed an alternative option), please say, but.... at the end of the day, a 'heavy' constructor is only a code smell, and not something which is actually *wrong*. My aesthetic sensibilities aren't worth breaking changes - especially given the *other* issues the alternatives would introduce.

riedgar-ms:
I'll work on the current API (probably in a fresh PR, though). I think I'm the only one with more-than-minor concerns, and even then, I can't concoct anything better than what is proposed here.

riedgar-ms:
Superseded by #1242 Although we might want to pull the edge correction code out in future. @MiroDudik  has also been doing some related work which is applicable for small _n_ cases.

PR #1039, opened 2022-03-04
Wrap roc_auc_score in _create_group_metric_set()

riedgar-ms:
Add a wrapper to the call to `roc_auc_score()` in `_create_group_metric_set()`. This deals with the case where only one class is present in the data (especially for subgroups). If this is encountered, then a warning is issued and a value of zero substituted. This is similar to how `recall_score()` operates (and the test picks up on the warning issued by `recall_score()` as well).

### DIFF
--- a/fairlearn/metrics/_group_metric_set.py
+++ b/fairlearn/metrics/_group_metric_set.py
@@ -1,7 +1,7 @@
 # Copyright (c) Microsoft Corporation and Fairlearn contributors.
 # Licensed under the MIT License.
 
-import logging
+import warnings
 
 import sklearn.metrics as skm
 from sklearn import preprocessing
@@ -18,8 +18,6 @@ from ._extra_metrics import (_balanced_root_mean_squared_error,
                              count)
 from ._metric_frame import MetricFrame
 from ._input_manipulations import _convert_to_ndarray_and_squeeze
-
-logger = logging.getLogger(__name__)
 
 _Y_TRUE = 'trueY'
 _Y_PRED = 'predictedY'
@@ -79,7 +77,7 @@ class func_wrapper:
             result = self.metric_function(y_true, y_pred)
         except ValueError:
             msg = "Evaluation of {0} failed. Substituting 0"
-            logger.warning(msg.format(self.metric_function.__name__))
+            warnings.warn(msg.format(self.metric_function.__name__))
         return result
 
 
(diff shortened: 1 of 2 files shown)


MiroDudik (approved):
@riedgar-ms : Any plans to move this module out of Fairlearn?

fairlearn/metrics/_group_metric_set.py
```diff
@@ -64,6 +70,27 @@
 SPECIFICITY_SCORE = "specificity_score"
 ZERO_ONE_LOSS = "zero_one_loss"
 
+
+class _func_wrapper:
+    """
+    Metric function wrapper, so they return 0 rather than raising ValueError.
```
romanlutz, line 76: Does 0 make sense? I'm thinking this could be misleading... if my dataset doesn't have 0s perhaps it's just not defined (NaN? "Undefined"?) rather than 0. Wdyt?

riedgar-ms: Can the dashboard cope with `np.NaN`? If so, I'd be happy with that.

riedgar-ms:
To @MiroDudik 's query, yes, I ought to move this to one of our Microsoft repos which contain the dashboard these days. However, for now, this fix needs to be here as well.

fairlearn/metrics/_group_metric_set.py
riedgar-ms: I've just checked @romanlutz , and if I substitute `np.nan` then the upload goes OK, but the dashboard itself doesn't load.

romanlutz: That sounds like a shortcoming of the dashboard that needs fixing. 0 still doesn't make sense, does it?

riedgar-ms: I agree that it's not great, but zero does have the benefit of being a valid `roc_auc_score()` and a _bad_ `roc_auc_score()`. It is going to give a 'balanced accuracy score difference' of zero, which is unfortunate; I'm not sure what will happen with the ratio when it hits the TS code and a divide-by-zero is thrown.

riedgar-ms: This is also copying the default behaviour of `recall_score()` which can suffer a similar problem. Unlike `roc_auc_score()` though, `recall_score()` does provide a `zero_division` argument to allow users to control what happens.

riedgar-ms: Ping.... @romanlutz ? I think that '0' is the best default for now - it's what `recall_score()` will also be returning in this sort of situation (and _has_ been returning to date).

romanlutz (approved):

PR #1044, opened 2022-03-16
Convert 'Fairness in ML' to BibTeX

riedgar-ms:
Bring civilised citations to the 'Fairness in Machine Learning' page in the documentation. Also do some reformatting of the ReST file, to reduce the length of a selection of lines.

Related to #936


riedgar-ms:
This should probably go in before #1042  (because when I add citations to that, I _really_ don't want to do so by hand)

docs/refs.bib
```diff
@@ -38,3 +38,53 @@ @inproceedings{hardt2016equality
   year      = {2016},
   url       = {https://proceedings.neurips.cc/paper/2016/hash/9d2682367c3935defcb1f9e247a97c0d-Abstract.html}
 }
+
+@inproceedings{10.1145/3287560.3287598,
+author = {Selbst, Andrew D. and Boyd, Danah and Friedler, Sorelle A. and Venkatasubramanian, Suresh and Vertesi, Janet},
```
romanlutz, line 43: danah boyd prefers to be spelled in lowercase. That doesn't come through in the paper as the authors are shown in a font that has only caps.

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -75,29 +81,32 @@ Key Term Examples  - Unobservable theoretical constructs and Measurement models
...
 
 What is construct validity? 
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-Though the Measurement and Fairness [#4]_ paper explores both construct reliability and construct validity, we focus our 
-exploration below on construct Validity. We note that both play an important role in understanding fairness in sociotechnical 
-contexts. With that said, Measurement and Fairness [#4]_ offers a fairness-oriented conceptualization of construct validity, that 
-is helpful in thinking about fairness in sociotechnical contexts. We capture the idea in seven key parts that when combined 
-can serve as a framework for analyzing an AI task and attempting to establish construct validity:
+Though :footcite:cts:`10.1145/3442188.3445901` explore both construct reliability and construct validity, we focus our
```
romanlutz, line 96: Do we want the citation to use this complex ID rather than the simpler ines we've been using so far? I am more familiar with the simple ones like agarwal2016reductions.

docs/refs.bib
riedgar-ms: I'll fix that.

docs/user_guide/fairness_in_machine_learning.rst
hildeweerts: I prefer the format @romanlutz suggested! The format is easy to reproduce for other journals/conferences/ArXiv. I also expect that this format makes it less likely that people add redundant copies to the .bib.

riedgar-ms: Can you clarify what you mean by 'format is easy to reproduce' @hildeweerts ?

romanlutz: I prefer it for the simple reason that I can read the text without checking the literature section and I have a hunch which paper it is 🙂 Agarwal 2016 reductions tells me all I need to know. That number thingy... not so much.

I agree with "easy to reproduce" in terms of just taking first author's last name + year + first (real) word of title. I believe that's what @hildeweerts meant (?)

docs/refs.bib
```diff
@@ -38,3 +38,53 @@ @inproceedings{hardt2016equality
...
...arning based systems that, once introduced into a social context, can achieve social and legal outcomes such as fairness, justice, and due process. Bedrock concepts in computer science---such as abstraction and modular design---are used to define notions of fairness and discrimination, to produce fairness-aware learning algorithms, and to intervene at different stages of a decision-making pipeline to produce "fair" outcomes. In this paper, however, we contend that these concepts render technical interventions ineffective, inaccurate, and sometimes dangerously misguided when they enter the societal context that surrounds decision-making systems. We outline this mismatch with five "traps" that fair-ML work can fall into even as it attempts to be more context-aware in comparison to traditional data science. We draw on studies of sociotechnical systems in Science and Technology Studies to explain why such traps occur and how to avoid them. Finally, we suggest ways in which technical designers can mitigate the traps through a refocusing of design in terms of process rather than solutions, and by drawing abstraction boundaries to include social actors rather than purely technical ones.},
```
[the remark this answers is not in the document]
romanlutz: But... shouldn't we? Do we really want to keep abstracts of papers here? It's not like we're going to render those...

docs/user_guide/fairness_in_machine_learning.rst
adrinjalali: I really think we should just copy/paste from our source and be done with it. I don't think citations warrant this much discussion 😁 I'm happy if we come up with a script that replaces all citations we have.

hildeweerts: > I prefer it for the simple reason that I can read the text without checking the literature section and I have a hunch which paper it is 🙂 Agarwal 2016 reductions tells me all I need to know. That number thingy... not so much.

Same!

> I agree with "easy to reproduce" in terms of just taking first author's last name + year + first (real) word of title. I believe that's what @hildeweerts meant (?)

Yeah, exactly :)

docs/refs.bib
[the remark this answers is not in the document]
riedgar-ms: We could trim it; I was just doing the simplest thing based on the citation source.

docs/user_guide/fairness_in_machine_learning.rst
riedgar-ms: I would agree that ACM is using a particularly obscure key (the ADS format at least gives you a chance of figuring out the paper). But as @adrinjalali says, in the long run, it will be easier just to copy/paste the citations from the source. And to catch duplicates, since BibTeX should whine if the same key appears twice (we might concoct our own 'approved' format, but if creation is manual then there's room for error).

hildeweerts: I don't think this will resolve duplicates, because some people may refer to the ArXiV preprint instead of a published paper and others may copy the citation from Google Scholar or something like doi2bib.org instead of the original publication venue.

[PS. I agree with @adrinjalali that citation probably do not warrant this much discussion but don't think it's that much effort to try to keep things consistent :D]

docs/refs.bib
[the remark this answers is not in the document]
romanlutz: If we're not rendering it let's just cut it(?)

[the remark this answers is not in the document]
adrinjalali: I'm happy to have as few modifications as possible on the copy pasted chunk. Also, this doesn't take that much space, why bother.

docs/refs.bib
```diff
@@ -39,8 +39,8 @@ @inproceedings{hardt2016equality
   url       = {https://proceedings.neurips.cc/paper/2016/hash/9d2682367c3935defcb1f9e247a97c0d-Abstract.html}
 }
 
-@inproceedings{10.1145/3287560.3287598,
-author = {Selbst, Andrew D. and boyd, danah and Friedler, Sorelle A. and Venkatasubramanian, Suresh and Vertesi, Janet},
+@inproceedings{selbst2019fairnessabstraction,
```
MiroDudik, line 42:
```suggestion
@inproceedings{selbst2019fairness,
```
for consistency

docs/refs.bib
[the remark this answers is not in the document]
MiroDudik: Since we are providing URL, I would personally lean towards skipping most of the extra meta-information here like isbn, abstract, numpages, keywords. But I don't care too much about extra meta-information that doesn't get displayed.

Two things that I think we should care about are:
* consistency in the naming of the bibtex ids (format <author><year><keyword>, where keyword is most of the time the first word in the paper title)
* consistency in the appearance of the references at the end of files (this is something that is actually not too hard to fix later, because we only need to edit .bib file...)
* linking to peer-reviewed versions of papers rather than arXiv whenever peer-reviewed versions are available

MiroDudik (commented):
I cannot check how these render... can you take a snippet and post in a comment or one of the replies?
* https://5032-133444044-gh.circle-artifacts.com/0/docs/_build/html/main/user_guide/fairness_in_machine_learning.html#id16

But, based on looking at bibtex entries below, I'd be in favor of something more condensed (e.g., that refers to the venue as FAT* and the publisher as ACM). The easiest way to construct that is to basically take the condensed DBLP entry:
* https://dblp.uni-trier.de/rec/conf/fat/SelbstBFVV19.html?view=bibtex&param=0

But take the URL from the standard DBLP entry:
* https://dblp.uni-trier.de/rec/conf/fat/SelbstBFVV19.html?view=bibtex&param=1

Sorry for being a pest, but I think it would be good to refer to academic papers consistently, and this one is not too hard.

riedgar-ms:
The current rendering on my machine:
image

Is ACM not considered a canoncial source of bibliographic information?

docs/refs.bib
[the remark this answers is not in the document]
riedgar-ms: Skipping metainformation is the business of the bibstyle file. Isn't that pretty much the whole point of BibTeX?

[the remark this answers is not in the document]
MiroDudik: The only issue that I see re. consistency is that these bibtex entries are generating multiple URLs when rendered. I'm in favor of dropping "eprint" and "doi" and retaining URL. (I think that's simpler that trying to tweak the stylefile?)

MiroDudik (approved):
This looks good... (modulo the merge). But if you end up ensuring that there's only one URL associated with each reference (by removing doi and eprint), I definitely won't mind :-)... but it's a low order concern for me.

romanlutz (approved):

adrinjalali (approved):

hildeweerts:
@MiroDudik feel free to make any small adjustments in a new PR!

PR #1046, opened 2022-03-16
Change CircleCI images

riedgar-ms:
Following up on notification from CircleCI:
https://discuss.circleci.com/t/legacy-convenience-image-deprecation/41034

Signed-off-by: Richard Edgar <riedgar@microsoft.com>

PR #1052, opened 2022-03-31
Doc build bug fix

riedgar-ms:
Fixing the publish stage of the doc build, which has been broken since #1046 .

The new image lacked `npm`, so install it via `apt-get`. When `npm` itself is run, it also needs to be superuser.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,15 +30,13 @@ jobs:
 
       - run:
           name: 'BUG REPRO'
-          command: |
-            sudo apt-get install npm
+          command: sudo apt-get install npm
 
       - run:
           name: 'BUG REPRO'
-          command: |
-            sudo npm install -g --silent gh-pages@2.0.1
+          command: sudo npm install -g --silent gh-pages@2.0.1
 
-       - run:
+      - run:
           name: 'BUG REPRO'
           command: |
             git config user.email "ci-build@fairlearn.org"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,7 +31,7 @@ jobs:
       - run:
           name: 'BUG REPRO'
           command: |
-            sudo apt-get install npm
+            # sudo apt-get install npm
 
       - run:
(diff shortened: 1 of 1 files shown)


adrinjalali:
shouldn't we be using `setup-node` instead? (https://www.voorhoede.nl/en/blog/super-fast-npm-install-on-github-actions/)

I'm not an expert here, but I worry about running npm in sudo and I've heard npm shouldn't be installed globally usually.

hildeweerts:
After we merge this, are there still any actions we should take to ensure all the merged PRs that did not go through are still published (in addition to re-running the checks)?

riedgar-ms:
@adrinjalali I'm not sure either (perhaps @romanlutz , who as been journeying to the even-darker-side recently can comment). This is just me prodding at things until they worked. In terms of speed, I'm not that concerned. The `apt-get` command runs in 11 seconds, and the `npm install` runs in three. For something which only runs after merging a PR (which we're not doing dozens of times a day), that doesn't seem onerous.

romanlutz:
> Ok, I'm happy to merge if @romanlutz doesn't have a much better solution at hand.

I don't to be honest. I'm sure there's a better way, but given that nobody runs this locally and it takes only a few seconds it doesn't seem worth the time to research it further IMO.

@hildeweerts there are no actions needed other than checking that the website deployment step runs correctly. We'd want to make sure the website does indeed update. It always builds the entire thing from scratch, so there's no risk of leaving an individual commit out.

romanlutz:
If you feel like following along live: https://app.circleci.com/pipelines/github/fairlearn/fairlearn/2684/workflows/185bd83b-55cc-40f7-9e6c-4f77f5e8e7de

riedgar-ms:
I am....

riedgar-ms:
And there's going to be another bugfix coming

romanlutz:
Oh no :-( failed

riedgar-ms:
I expect that #1054 will deal with it

PR #1060, opened 2022-04-04
Update known_hosts for doc build

riedgar-ms:
## Description

Need to put the GitHub SSH keys into the right `known_hosts` file before we can push our docs.

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,10 +61,6 @@ jobs:
     steps:
       # check out PR branch
       - checkout
-      
-      - run:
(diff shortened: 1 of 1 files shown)


romanlutz (approved):
I don't have a good intuition on whether this will work, but seeing as it can't break things that are already broken I'd say go ahead.

adrinjalali (approved):
BTW, you can ssh to your circleCI instance and test these.

PR #1071, opened 2022-04-26
[WIP] CI fix doc deploy

adrinjalali:
Write our own doc push like scikit-learn and numpy do.


adrinjalali:
Will send the PR from a branch from the main repo to test if the key is available then. It doesn't seem to be available here, and might be because it's a from from a forked repo (which makes sense from a security perspective).

PR #1072, opened 2022-04-26
MNT speed up plot_grid_search_census.py

adrinjalali:
The example is really slow, this PR speeds up the example.

The non-formating changes are: 0.4 test set size instead of 0.2, and 30 grid size instead of 71.


romanlutz (approved):
No objections, of course.

Did you check that the text where the resulting plot is shown doesn't reference any specific values? Because those may change with the new train-test-split...

examples/plot_grid_search_census.py
```diff
@@ -143,9 +151,11 @@
 # We are using this metric for the sake of simplicity; in general, the appropriate fairness
 # metric will not be obvious.
 
-sweep = GridSearch(LogisticRegression(solver='liblinear', fit_intercept=True),
-                   constraints=DemographicParity(),
-                   grid_size=71)
+sweep = GridSearch(
+    LogisticRegression(solver="liblinear", fit_intercept=True),
+    constraints=DemographicParity(),
+    grid_size=30,
```
riedgar-ms, line 157: Small point, but if this is an odd number, then the grid includes an unmitigated model.

riedgar-ms:
No objections from me; the comment about grid size is minor.

riedgar-ms (approved):
Comment about grid size is minor. If none of the text mentions the unmitigated model, it's fine to skip.

adrinjalali:
Cool, then I'll merge when CI goes green with two approvals.

PR #1080, opened 2022-04-29
CI move build/test to GH actions

### LINKED ISSUE #847: Add code coverage test to the CI
We can add https://about.codecov.io/ to the CI, to make sure the PRs are all tested.

adrinjalali:
Enabling GH actions for build and tests

Also fixes #847


adrinjalali:
@riedgar-ms WDYT of this one? My idea is to start simple, merge, migrate the next bits in future PRs and then remove the azure pipelines.

.github/workflows/build-test.yml
```diff
@@ -0,0 +1,62 @@
...
+        python: ["3.8", "3.9", "3.10"]
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: actions/setup-python@v3
+        with:
+          python-version: ${{ matrix.python }}
+      - run: pip install .
+      - run: pip install -r requirements-dev.txt
+      - run: pytest --ignore=test/install --cov=fairlearn --cov-report=xml -m "not notebooks" test/
```
riedgar-ms, line 26: Quite a few of these steps appear to be shared.... can that be abstracted?

romanlutz (approved):

.github/workflows/build-test.yml
adrinjalali: Except the codecov upload, none of them are exactly the same, and having them in the same file would mean accepting input args and conditioning on those and at the end it'd be more code than what it is at this stage. I don't want to overengineer things at this stage, happy to abstract things away once it starts to actually save lines of code.

romanlutz: Having worked with the previous system a bit I will say I'd take a bit of duplication if it gives us readability (as in, being able to read the entire workflow end-to-end in one file).

riedgar-ms (approved):

PR #1083, opened 2022-05-02
MNT Lint the codebase

adrinjalali:
This PR adds linter pre-commit hooks and runs them on the codebase.

It also changes the line length to 88 which is the default for `black`, since it seems most libraries are adopting this.

We can then add the hash of this commit to the ignored ones so that `git blame` doesn't complain.

This is to move our codebase to a more consistent state, and so that we can add a linter to the CI and avoid running CI if the liter fails.

cc @fairlearn/fairlearn-maintainers


.pre-commit-config.yaml
```diff
@@ -0,0 +1,23 @@
+repos:
```
riedgar-ms, line 1: I'm guessing that this is the critical file, with the rest being `black` doing its worst?

pyproject.toml
```diff
@@ -1,5 +1,10 @@
 [tool.black]
-line-length = 79
+line-length = 88
```
riedgar-ms, line 2: Ahh, this too?

riedgar-ms:
I've no objections in principle to this, but please can you highlight which files you've changed, since most of the changes are simply `black` being let off its leash.

adrinjalali:
Not sure why we get the circular import errors here, haven't changed any imports, I'll try to find it.

adrinjalali:
@riedgar-ms CI is green :)

riedgar-ms (approved):
I'm happy with this; I doubt there will ever be a "good" time to do this, so we may as well go with "now."

MiroDudik (approved):

PR #1092, opened 2022-05-19
Notes on the Four-Fifths Rule

riedgar-ms:
## Description

This adds a short discussion of the "Four-Fifths Rule" to the "Fairness in Machine Learning" section of the user guide. It is an explanation of why users probably don't want to use the rule as their fairness criterion.

This replaces #1042 , which was getting old and had a bunch of merge conflicts.

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

### DIFF
--- a/docs/user_guide/fairness_in_machine_learning.rst
+++ b/docs/user_guide/fairness_in_machine_learning.rst
@@ -480,7 +480,11 @@ not be appropriate in most cases.
 Even in cases where it is appropriate, it does not automatically
 avoid legal jeopardy, much less ensure that results are fair.
 
-
+For a much deeper discussion of the issues involved, we suggest
+:footcite:ct:`watkins2022fourfifths`.
+A higher level look at how legal concepts of fairness can collide
+with mathematical measures of disparity, see
+:footcite:ct:`Xiang2019legalcompatibility`.
 
 References
 ----------
--- a/docs/user_guide/fairness_in_machine_learning.rst
+++ b/docs/user_guide/fairness_in_machine_learning.rst
@@ -435,6 +435,56 @@ incorporate the processes associated with crime reporting, the offense-trial
 pipeline, and an awareness of how the relationship between various social actors and
 the algorithm may impact the intended outcomes of a given model.
 
+
+The 4/5ths Rule: A Common Trap
+------------------------------
+
+In the literature around fairness in machine learning, one will often find
+the so-called "4/5ths rule" or "80% rule" used to assess whether a model
+(or mitigation technique) has produced a 'fair' result.
+Typically, the rule is implemented by using the demographic parity ratio introduced
+in the :ref:`disparity_metrics` section, with a result considered 'fair' if the ratio
+exceeds 80% for all identified subgroups.
+*Unthinking application of this threshold is wrong in many scenarios.*
+
+There are three particular things which should be considered before applying
+the "4/5ths rule" to a particular scenario:
+
+1. What does the rule mean, legally?
+2. Where is the rule valid?
+3. Why 80%?
+
+**What exactly does the rule mean?**
+Legally, the 80% threshold (relevant to a chosen, relevant control population)
+is a limit for
+`prima facie evidence <https://en.wikipedia.org/wiki/Prima_facie>`_
+that illegal discrimination has occurred relative to a chosen,
+relevant control population.
+A causal link between the statistic and the alleged discrimination must also
+be demonstrated.
+However, the 'Four-Fifths Rule' does not prove or disprove any case.
+It is possible to argue successfully that a demographic parity
+ratio of 50% is still fair.
+It is equally possible to argue that even at a 95% ratio, illegal
+discrimination has occurred.
+Conflating the *disparate impact metric* with the *legal finding of disparate impact*
+is a form of epistemic trespassing.
+
+**Where is the 4/5ths rule valid?**
+The answer is that it is only valid
+within a specific area of US federal employment law.
+Outside this context, the rule has no particular claim to validity.
+
+**Why is the ratio set at 80%?**
+The reasoning for selecting 80% is unclear.
+
+Taken together, we see that applying the 4/5ths rule will
+not be appropriate in most cases.
+Even in cases where it is appropriate, it does not automatically
+avoid legal jeopardy, much less ensure that results are fair.
+
+
+
 References
 ----------
 
--- a/docs/user_guide/fairness_in_machine_learning.rst
+++ b/docs/user_guide/fairness_in_machine_learning.rst
@@ -475,9 +475,6 @@ The answer is that it is only valid
 within a specific area of US federal employment law.
 Outside this context, the rule has no particular claim to validity.
 
-**Why is the ratio set at 80%?**
-The reasoning for selecting 80% is unclear.
-
 Taken together, we see that applying the 4/5ths rule will
 not be appropriate in most cases.
 Even in cases where it is appropriate, it does not automatically
--- a/docs/user_guide/fairness_in_machine_learning.rst
+++ b/docs/user_guide/fairness_in_machine_learning.rst
@@ -447,16 +447,8 @@ in the :ref:`disparity_metrics` section, with a result considered 'fair' if the 
 exceeds 80% for all identified subgroups.
 *Unthinking application of this threshold is wrong in many scenarios.*
 
-There are three particular things which should be considered before applying
-the "4/5ths rule" to a particular scenario:
-
-1. What does the rule mean, legally?
-2. Where is the rule valid?
-3. Why 80%?
-
 **What exactly does the rule mean?**
-Legally, the 80% threshold (relevant to a chosen, relevant control population)
-is a limit for
+Legally, the 80% threshold is a limit for
 `prima facie evidence <https://en.wikipedia.org/wiki/Prima_facie>`_
 that illegal discrimination has occurred relative to a chosen,
 relevant control population.
@@ -471,8 +463,8 @@ Conflating the *disparate impact metric* with the *legal finding of disparate im
 is a form of epistemic trespassing.
 
 **Where is the 4/5ths rule valid?**
-The answer is that it is only valid
-within a specific area of US federal employment law.
+The answer is that it is only valid within a specific area of US
+federal employment law.
 Outside this context, the rule has no particular claim to validity.
 
 Taken together, we see that applying the 4/5ths rule will
--- a/docs/refs.bib
+++ b/docs/refs.bib
@@ -100,3 +100,26 @@ doi = {https://doi.org/10.2307/1229039},
 URL = {https://www.jstor.org/stable/1229039},
 eprint = {https://www.jstor.org/stable/1229039}
 }
+
+@misc{Xiang2019legalcompatibility,
+  doi = {10.48550/ARXIV.1912.00761},
+  url = {https://arxiv.org/abs/1912.00761},
+  author = {Xiang, Alice and Raji, Inioluwa Deborah},
+  keywords = {Computers and Society (cs.CY), Artificial Intelligence (cs.AI), Machine Learning (cs.LG), Machine Learning (stat.ML), FOS: Computer and information sciences, FOS: Computer and information sciences},
+  title = {On the Legal Compatibility of Fairness Definitions},
+  publisher = {arXiv},
+  year = {2019},
+  copyright = {Creative Commons Attribution 4.0 International}
+}
+
+@misc{watkins2022fourfifths,
+  doi = {10.48550/ARXIV.2202.09519},
+  url = {https://arxiv.org/abs/2202.09519},
+  author = {Watkins, Elizabeth Anne and McKenna, Michael and Chen, Jiahao},
+  keywords = {Computers and Society (cs.CY), Artificial Intelligence (cs.AI), Machine Learning (cs.LG), Logic in Computer Science (cs.LO), FOS: Computer and information sciences, FOS: Computer and information sciences, K.4; K.5; F.4; I.2, 68T27, 03B70},
+  title = {The four-fifths rule is not disparate impact: a woeful tale of epistemic trespassing in algorithmic fairness},
+  publisher = {arXiv},
+  year = {2022},
+  copyright = {arXiv.org perpetual, non-exclusive license}
+}
+


riedgar-ms:
This doesn't seem to be appearing on the docs? And building locally is note working but not giving me any sort of error :-(

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -435,6 +435,49 @@ incorporate the processes associated with crime reporting, the offense-trial
...
+The 4/5ths Rule: A Common Trap
+------------------------------
+
+In the literature around fairness in machine learning, one will often find
+the so-called "4/5ths rule" or "80% rule" used to assess whether a model
+(or mitigation technique) has produced a 'fair' result.
+Typically, the rule is implemented by using the demographic parity ratio introduced
+in the :ref:`disparity_metrics` section, with a result considered 'fair' if the ratio
+exceeds 80% for all identified subgroups.
+*Unthinking application of this threshold is wrong in many scenarios.*
```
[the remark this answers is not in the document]
riedgar-ms: Fair enough. Have adjusted.

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -435,6 +435,49 @@ incorporate the processes associated with crime reporting, the offense-trial
...
+*Application of this threshold is wrong in many scenarios.*
+
+**What exactly does the rule mean?**
+Legally, the 80% threshold is a limit for
+`prima facie evidence <https://en.wikipedia.org/wiki/Prima_facie>`_
+that illegal discrimination has occurred relative to a chosen,
+relevant control population.
+A causal link between the statistic and the alleged discrimination must also
+be demonstrated.
+However, the 'Four-Fifths Rule' does not prove or disprove any case.
```
LeJit, line 457: Does 'Four-Fifths Rule' need to be capitalized?

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -435,6 +435,49 @@ incorporate the processes associated with crime reporting, the offense-trial
...
+It is possible to argue successfully that a demographic parity
+ratio of 50% is still fair.
```
LeJit, line 459: This may be better explained with an example (e.g. the fire department exam case).

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -435,6 +435,49 @@ incorporate the processes associated with crime reporting, the offense-trial
 pipeline, and an awareness of how the relationship between various social actors and
 the algorithm may impact the intended outcomes of a given model.
 
+
+The 4/5ths Rule: A Common Trap
```
hildeweerts, line 439: As we really want people to read this, perhaps we should come up with a slightly more click-baity title, like "Why you should (almost) never apply the 4/5ths rule". 

My personal preference would go to something even more general, e.g., something like: "How 'fair' should my model be?". This is a common question data scientists have, and with this phrasing they might be more likely to read the text even if they've never heard of the 4/5ths rule. (This would also provide a nice space to add more guidelines on how to translate harms into fairness metrics, in accordance with [learning objective 1.6](https://github.com/orgs/fairlearn/projects/3/views/1))

Additionally, instead of introducing this as a separate trap, an alternative position of this paragraph would be right after the explanation of the fairness constraints. In a way, blindly applying the 4/5ths rule could be seen as an example of the formalism trap and/or a portability trap. Instead of introducing it as a separate trap, we could refer to it as an example of (one of) those traps.

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -435,6 +435,49 @@ incorporate the processes associated with crime reporting, the offense-trial
...
+------------------------------
+
+In the literature around fairness in machine learning, one will often find
+the so-called "4/5ths rule" or "80% rule" used to assess whether a model
+(or mitigation technique) has produced a 'fair' result.
```
hildeweerts, line 444: I would add a line here that says that this 4/5th rule originates from US labor law.

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -435,6 +435,49 @@ incorporate the processes associated with crime reporting, the offense-trial
...
+Typically, the rule is implemented by using the demographic parity ratio introduced
```
hildeweerts, line 445: I would emphasize that this is the interpretation/translation computer scientists/fairness researchers have given to the rule.

This might be a good place to emphasize the difference between fairness and illegal discrimination.

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -435,6 +435,49 @@ incorporate the processes associated with crime reporting, the offense-trial
...
+in the :ref:`disparity_metrics` section, with a result considered 'fair' if the ratio
+exceeds 80% for all identified subgroups.
+*Application of this threshold is wrong in many scenarios.*
+
+**What exactly does the rule mean?**
+Legally, the 80% threshold is a limit for
```
hildeweerts, line 451: I would prefer something like "In US labor law" to emphasize that this is specific to the US.

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -435,6 +435,49 @@ incorporate the processes associated with crime reporting, the offense-trial
...
+`prima facie evidence <https://en.wikipedia.org/wiki/Prima_facie>`_
+that illegal discrimination has occurred relative to a chosen,
+relevant control population.
+A causal link between the statistic and the alleged discrimination must also
```
hildeweerts, line 455: From this sentence it is not really clear to me what a "A causal link between the statistic and the alleged discrimination" would look like - perhaps an example would help.

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -435,6 +435,49 @@ incorporate the processes associated with crime reporting, the offense-trial
...
+be demonstrated.
+However, the 'Four-Fifths Rule' does not prove or disprove any case.
+It is possible to argue successfully that a demographic parity
+ratio of 50% is still fair.
+It is equally possible to argue that even at a 95% ratio, illegal
+discrimination has occurred.
+Conflating the *disparate impact metric* with the *legal finding of disparate impact*
+is a form of epistemic trespassing.
```
hildeweerts, line 463: In my opinion, "epistemic trespassing", sounds too formal for a user guide, especially without further explanation on what we mean by it (case in point: if you Google the term you mostly find academic papers).

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -435,6 +435,49 @@ incorporate the processes associated with crime reporting, the offense-trial
...
+
+**Where is the 4/5ths rule valid?**
```
hildeweerts, line 465: Nitpick: something like "In which case" might be a bit more encompassing than "where" (which mostly implies location to me)

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -435,6 +435,49 @@ incorporate the processes associated with crime reporting, the offense-trial
...
+The answer is that it is only valid within a specific area of US
+federal employment law.
+Outside this context, the rule has no particular claim to validity.
```
hildeweerts, line 468: Nitpick: "having no particular claim to validity" sounds very formal

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -435,6 +435,49 @@ incorporate the processes associated with crime reporting, the offense-trial
 pipeline, and an awareness of how the relationship between various social actors and
 the algorithm may impact the intended outcomes of a given model.
 
+
+The 4/5ths Rule: A Common Trap
+------------------------------
+
```
hildeweerts, line 441: Maybe add some introductory sentence(s) here that says that a common question data scientists have is what value their fairness metrics should be for the model to be fair.

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -435,6 +435,49 @@ incorporate the processes associated with crime reporting, the offense-trial
...
+the so-called "4/5ths rule" or "80% rule" used to assess whether a model
+(or mitigation technique) has produced a 'fair' result.
+Typically, the rule is implemented by using the demographic parity ratio introduced
+in the :ref:`disparity_metrics` section, with a result considered 'fair' if the ratio
+exceeds 80% for all identified subgroups.
+*Application of this threshold is wrong in many scenarios.*
+
+**What exactly does the rule mean?**
+Legally, the 80% threshold is a limit for
+`prima facie evidence <https://en.wikipedia.org/wiki/Prima_facie>`_
```
hildeweerts, line 452: Maybe add a small sentence to explain what prima facie evidence is.

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -435,6 +435,49 @@ incorporate the processes associated with crime reporting, the offense-trial
...
+that illegal discrimination has occurred relative to a chosen,
```
hildeweerts, line 453: Is the word "chosen" particularly relevant here?

docs/user_guide/fairness_in_machine_learning.rst
hildeweerts: +1 to adding more examples (same for the 95%)

hildeweerts, line 459: We need to make sure that we clearly distinguish between "fairness" and "illegal discrimination" in this paragraph.

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -435,6 +435,49 @@ incorporate the processes associated with crime reporting, the offense-trial
...
+relevant control population.
+A causal link between the statistic and the alleged discrimination must also
+be demonstrated.
+However, the 'Four-Fifths Rule' does not prove or disprove any case.
+It is possible to argue successfully that a demographic parity
+ratio of 50% is still fair.
+It is equally possible to argue that even at a 95% ratio, illegal
+discrimination has occurred.
+Conflating the *disparate impact metric* with the *legal finding of disparate impact*
```
hildeweerts, line 462: As we typically do not say "disparate impact metric", perhaps we should say something like:
"Conflating the *demographic parity ratio*, sometimes referred to as *disparate impact*, with the *legal finding of disparate impact*". Or already mention this earlier in the text, e.g., when we first refer to the demographic parity ratio.

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -435,6 +435,49 @@ incorporate the processes associated with crime reporting, the offense-trial
...
+is a form of epistemic trespassing.
+
+**Where is the 4/5ths rule valid?**
+The answer is that it is only valid within a specific area of US
+federal employment law.
+Outside this context, the rule has no particular claim to validity.
+
+Taken together, we see that applying the 4/5ths rule will
+not be appropriate in most cases.
+Even in cases where it is appropriate, it does not automatically
```
hildeweerts, line 472: "Appropriate" sounds a bit odd in this sentence (because we actually say that it it isn't in the next sentence). Maybe something like:
```suggestion
Even in cases where the rule is applicable, it does not automatically
```

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -435,6 +435,49 @@ incorporate the processes associated with crime reporting, the offense-trial
...
+avoid legal jeopardy, much less ensure that results are fair.
```
hildeweerts, line 473: I like this sentence a lot! I would be great if we could emphasize the difference between fairness and illegal discrimination more explicitly earlier on in the text.

docs/user_guide/fairness_in_machine_learning.rst
hildeweerts, line 457: I assume "prove"/"disprove" in the legal sense of the word?

docs/user_guide/fairness_in_machine_learning.rst
```diff
@@ -435,6 +435,49 @@ incorporate the processes associated with crime reporting, the offense-trial
...
+
```
hildeweerts, line 474: I think we can add to this paragraph that this type of situation is why we try to avoid legal terminology as much as possible in Fairlearn, which explains why we do not call it "disparate impact" but "demographic parity".

hildeweerts (commented):
I love that we are adding this!!!

docs/user_guide/fairness_in_machine_learning.rst
riedgar-ms: Can you remind me of that, please? Do you mean:
https://en.wikipedia.org/wiki/Ricci_v._DeStefano

docs/user_guide/fairness_in_machine_learning.rst
riedgar-ms: Clickbaity would be something like "You'll never guess what 90% of fairness experts say about the 4/5ths rule!!!"

I've moved it up in the doc, but am still mulling a title.

docs/user_guide/fairness_in_machine_learning.rst
riedgar-ms: Mentioned (slightly further down).... of course, calling it 'labor' implicitly means it's 'US' ;-)

docs/user_guide/fairness_in_machine_learning.rst
riedgar-ms: That's what the link is for.

hildeweerts: In my opinion we should avoid jargon as much as possible, and if we use at least explain it in the text if we do. I'm probably spending too much time on such a detail, but we've seen in other texts (e.g., on construct validity) that using a lot of (new) terminology can make texts less accessible for our primary audience (data scientists). I would definitely count "prima facie" as jargon, but that's perhaps because I'm a non-native English speaker. But it's not a hill I'm willing to die on so if others disagree we can keep it in the current form.

docs/user_guide/fairness_in_machine_learning.rst
hildeweerts: The intended the emphasis was on *here*! Mostly related to my next comment re. differentiation between legal definitions of discrimination and "fairness".

hildeweerts: PS. In an unexpected turn of events there are legal systems outside of the Anglosphere (what?!), but I suppose "arbeidsrecht" won't be very understandable for most of our readers ;)

docs/user_guide/fairness_in_machine_learning.rst
riedgar-ms: Have removed this subhead

docs/user_guide/fairness_in_machine_learning.rst
riedgar-ms: Simplified

docs/user_guide/fairness_in_machine_learning.rst
riedgar-ms: Removed

docs/user_guide/fairness_in_machine_learning.rst
riedgar-ms: Done.

riedgar-ms:
@hildeweerts  and @LeJit I've pushed another draft, which takes on some of your comments, but also changes enough that addresses your comments but not quite directly (since I've done some restructuring of the text). As such, I've left several of your comments open so that we have them for reference.

Also, can you confirm that I had the right case for the firefighters, please?

docs/user_guide/fairness_in_machine_learning.rst
LeJit: Yes. This is the correct legal case. Although now that I re-read the case, I think it would be a better example of the opposite idea: How reducing the disparate impact required an illegal intervention.

docs/user_guide/fairness_in_machine_learning.rst
riedgar-ms: Fixed.

romanlutz (approved):

Issue #1089, opened 2022-05-16
Attributes not rendering properly in documentation


rensoostenbach:
While working on #1043 and implementing the docstrings, I noticed that attributes are not properly rendered in the documentation. Perhaps related, but I also noticed that no other class has attributes in the API docs.

As can be seen in the below screenshot, there is no separate header for the attributes, while I have included it in the [code](https://github.com/rensoostenbach/fairlearn/blob/90a11a868c96658821b35e1fec6551bb6c13573e/fairlearn/linear_model/_constrained_logistic.py#L502).
![attributes missing](https://user-images.githubusercontent.com/33757199/168590266-80a53d35-9f1d-46e3-b9b1-ab1f8f44d292.jpg)

Is it a conscious decision that the attributes aren't handled properly, or is it a bug?

Also, this probably explains the RST306 errors regarding unknown target names that occur in the [PR gate](https://dev.azure.com/fairlearn/Fairlearn/_build/results?buildId=2080&view=logs&j=011e1ec8-6569-5e69-4f06-baf193d1351e&t=e5bbc628-0816-5281-89e5-f3e62109ab4a&l=22) for #1043. Future contributors might run into similar issues, which is why I think it is good to have a look at this.

romanlutz:
My initial reaction was that attributes are internal and shouldn't really matter to the caller. Of course, that's only half true... sometimes we want to take attributes that were set during `fit` and use them for something further.

I'm guessing it's something like this: https://stackoverflow.com/questions/72220924/sphinx-how-to-show-attributes-as-in-scipy
It would require some more investigation to be sure, though.

hildeweerts:
Good catch @rensoostenbach! In my opinion attributes should absolutely be rendered in the docs. For many machine learning models attributes can give insights into what was learned, think e.g., `coef_` in `sklearn.linear_model.LogisticRegression` (note that scikit-learn API docs also renders these attributes). In Fairlearn an example would be `ThresholdOptimizer`: it would make a lot of sense if the API docs indicate under what name and in what format the learned thresholds are stored. In my opinion, *not* showing these things does not really go with our vision of a well thought out ML development process.

adrinjalali:
I agree that they should be rendered. Only the public ones are rendered and private ones are not @romanlutz . But you're right in that there should be a distinction between public and private attributes.

PR #1093, opened 2022-05-20
CI add numpydoc to doc build

adrinjalali:
I suspect this should fix https://github.com/fairlearn/fairlearn/issues/1089

Let's see.

Closes https://github.com/fairlearn/fairlearn/issues/1089

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -79,6 +79,7 @@ extensions = [
     "sphinx_gallery.gen_gallery",
     "sphinx_multiversion",
     "sphinx_autodoc_typehints",  # needs to be AFTER napoleon
+    "numpydoc",
 ]
 
 source_suffix = [".rst"]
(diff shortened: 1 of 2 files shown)


riedgar-ms (approved):

rensoostenbach:
After merging this PR to my branch in #1043, the attributes are still being rendered as in the screenshot in #1089 unfortunately.

rensoostenbach:
@adrinjalali Is there any chance you could look into why attributes are still not rendering (as per my above comment)? Something else I noticed, is that the `total_samples` attribute is being rendered [here](https://fairlearn.org/main/api_reference/fairlearn.reductions.html) (and also on previous versions of the docs), but I am not sure how that is done since I don't see anything related to the attributes [in the code](https://github.com/fairlearn/fairlearn/blob/a481cff8313b7c02a837d544adba121dd5de37e8/fairlearn/reductions/_moments/bounded_group_loss.py).

PR #1096, opened 2022-06-02
CI checkout the right branch in circleci

### LINKED ISSUE #819: CircleCI doc build does not use PR branch but main branch
We set up the CircleCI doc build to build the documentation based on the branch of a Pull Request. However, in the last few weeks it's often (note: not always!) produced the documentation based on the `main` branch rather than the PR branch. This means we can't validate that the new docs will look right. Instead, we have to check out the branch locally and build it which wastes a ton of time.

Ideally we should figure out what's wrong with CircleCI and fix it. Perhaps they've changed something about this behavior recently, perhaps the code checkout logic needs to be adjusted or something like that. This issue requires a bit of investigation from someone who's ideally already somewhat familiar with CircleCI and sphinx doc builds.

Alternatively, we could also switch to a different provider, but the advantage of CircleCI so far was that the artifacts render right in the browser without the need to download anything. GitHub Actions and Azure DevOps don't seem to provide that as of now.

adrinjalali:
By default CircleCI checks out the default branch, this PR adds a script to checkout the right branch.

Fixes #819


adrinjalali:
So it seems the issue wasn't checking out the wrong branch, rather it was that the multiversion was ignoring the PR branch. This PR makes it so that on the PRs we only build a bare minimum of the current branch w/o multiversion and no landing page (since the landing page has links to the wrong folders w/o multiversion)

cc @riedgar-ms @romanlutz @hildeweerts

adrinjalali:
On the plus side, the doc build on PRs now takes just about 3 minutes

adrinjalali:
Should be ready for a review/merge.

riedgar-ms:
I'm really wondering if `sphinx-multiversion` is worth the trouble :-/

riedgar-ms (approved):

adrinjalali:
> I'm really wondering if `sphinx-multiversion` is worth the trouble :-/

It isn't, I'm working on removing it.

.circleci/config.yml
```diff
@@ -35,20 +35,25 @@ jobs:
...
           command: python scripts/install_requirements.py --pinned False
 
       - run:
           name: "Build Documentation"
-          command: python scripts/build_documentation.py --documentation-path=docs --output-path=docs/_build/html
+          command: |
+            if [[ "${CIRCLE_BRANCH}" =~ ^main$|^[0-9]+\.[0-9]$|^[0-9]+\.[0-9]+\.[0-9]+$ ]]; then
+              bash python scripts/build_documentation.py --documentation-path=docs --output-path=docs/_build/html
+            else
+              make doc
```
romanlutz, line 49: What I understand here is: if the branch is main or a release version then build the versioned docs, otherwise just the single-version-webpage. But then how would we know if we break the versioned docs on a PR?

romanlutz: That's fair. In combination with the other PR that completely addresses my concern.

PR #1097, opened 2022-06-03
MNT update pydata-sphinx-theme and on remove multiversion

adrinjalali:
This solves the issue of not being able to build the docs on python>3.8


adrinjalali:
This also fixes https://github.com/fairlearn/fairlearn/pull/1096.

cc @riedgar-ms @romanlutz

adrinjalali:
The idea is to build only one version, and push to the right branch on the website's repo. We would need a few future PRs to fix the minor issues though after/while the release happens.

romanlutz:
So that means that the older versions would all be gone? And there would no longer be any links between versioned websites since the older versions are gone?

adrinjalali:
Those ones would be a header we inject in all files and sphinx can do that. It's the same way as we do in sklearn. There's a js file you can ask sphinx to include in every single page. If this PR makes those vanish, we can fix that quite easily.

romanlutz (approved):
Thanks a lot for this!!!

This also just reminded me that we won't need things like bokeh anymore! https://github.com/fairlearn/fairlearn/blob/main/requirements-dev.txt

adrinjalali:
@riedgar-ms WDYT?

Makefile
```diff
@@ -22,8 +22,5 @@ test-unit:
 test-deprecation:
 	$(PYTEST) -Werror::DeprecationWarning -Werror::FutureWarning ./test/unit
 
-doc-multiversion:
-	python scripts/build_documentation.py --documentation-path=docs --output-path=docs/_build/html
```
riedgar-ms, line 26: Can we also delete `build_documentation.py` ?

adrinjalali: done

adrinjalali:
@riedgar-ms we can build the entire doc repo, but either manually or to write a script to do so. Since we don't really touch previous versions, I don't think it's something we'd be doing anyway.

adrinjalali:
Merging this with 2 approvals. I'll have an eye on the deployment to see if anything breaks to fix it today.

PR #1103, opened 2022-06-09
MNT add theme's version switcher

adrinjalali:
This updates the theme's version, which natively now supports switching between versions.

If this works, we can also update the older versions and push a build with a new theme version.

Documentation for this feature: https://pydata-sphinx-theme.readthedocs.io/en/latest/user_guide/configuring.html#add-a-dropdown-to-switch-between-docs-versions

cc @romanlutz @hildeweerts @riedgar-ms


adrinjalali:
The good thing is now that we don't have multiversion build/deploy, we can experiment more on the `main` branch and it only affects that branch of the documentation.

adrinjalali:
As long as people stay on the old docs, they don't see any change as a result of this PR. For the old ones to change we'd need to do a separate commit to those branches to update them if we want them to look like this one (but I don't think that has a high priority, they're old docs)

adrinjalali:
The new theme even respects the "dark mode" setting of the browser 🤯

adrinjalali:
To really test this one, we'd need to merge, testing it before merge is kinda tricky (but since it only affects the `main` version of the docs, risk is low)

hildeweerts:
Looking at the CircleCI artefacts (🤯) it does look like this update adjusts the color schemes quite a bit, so we might want to readjust some of those if possible (and use a different logo in dark mode).

I like the switcher a lot though! It's a lot less intrusive to the left-hand side navigation compared to our previous solution.

adrinjalali:
The changes to the light mode are not that invasive I find, and I quite like them the way they are now.

For the dark mode, I'm happy for them to be improved, but I don't think it's in the scope of this PR, at least I'm not the best person to do them.

riedgar-ms (approved):
OK with me.

hildeweerts:
Fair! I'll just open another issue then to make sure we don't forget to update the logo and possibly the colors as well :)

PR #1106, opened 2022-06-09
FIX fix multiversion links

adrinjalali:
It seems the documentation [here](https://pydata-sphinx-theme.readthedocs.io/en/latest/user_guide/configuring.html#add-a-dropdown-to-switch-between-docs-versions) might be misleading and the links shouldn't have the `index.html` at the end.


adrinjalali:
Merging, since this one I also can't test w/o merging :/

Issue #1061, opened 2022-04-04
Add examples to API documentation


alliesaizan:
Related to issue #730, this issue covers adding links to relevant example notebooks to the API documentation. The fix would be for someone to go through the [notebooks](https://fairlearn.org/v0.7.0/auto_examples/index.html) and add links in the API documentation for the classes/functions used in the example. Tagging @hildeweerts and @romanlutz in case they have additional ideas!

yanhong-zhao-ef:
would love to have a go at this if it is not covered already

hildeweerts:
For whoever will work on this, note @adrinjalali's suggestion in [this comment](https://github.com/fairlearn/fairlearn/pull/1109#pullrequestreview-1011501704):

> One way to fix the original issue is to automatically have links to all examples from the API guides the same way scikit-learn does (example: https://scikit-learn.org/stable/modules/generated/sklearn.compose.ColumnTransformer.html#examples-using-sklearn-compose-columntransformer)

romanlutz:
The latest task to do here is to find out how sklearn is getting the "gallery" of examples into the API reference. That would automatically tackle this problem. I believe this configuration is responsible: https://sphinx-gallery.github.io/dev/configuration.html#references-to-examples

So someone would need to try this (using the website build) and can check with how sklearn is doing it as a comparison, and validate that it works.

miacodehub:
hi! would love to work on this if it is still available

romanlutz:
Absolutely! Do you know what to do?

miacodehub:
I believe I understand. Correct me if I'm wrong:
Currently there's one example notebook for each of the APIs documented. In the API reference pages, at the bottom of the page there should be a section with title "examples that use API #1" followed by the "gallery" representation to that example. Since there's currently five APIs with five example notebooks, each API reference page will have its resepective example notebook.
I'll follow the suggestion you mentioned w.r.t the config file modifications. Do let me know if I missed anything.

alliesaizan:
> I believe I understand. Correct me if I'm wrong: Currently there's one example notebook for each of the APIs documented. In the API reference pages, at the bottom of the page there should be a section with title "examples that use API #1" followed by the "gallery" representation to that example. Since there's currently five APIs with five example notebooks, each API reference page will have its resepective example notebook. I'll follow the suggestion you mentioned w.r.t the config file modifications. Do let me know if I missed anything.

[...]

PR #1099, opened 2022-06-08
MNT remove Mitigating Disparities in Ranking from Binary Data.ipynb

adrinjalali:
I think we have covered the contents of the file in other user guides and the dataset used here is no longer available from that website, therefore it's making our CI fail.

cc @fairlearn/fairlearn-maintainers

PR #1109, opened 2022-06-16
Add hyperlink from example notebooks (sphinx-gallery) to API doc

yanhong-zhao-ef:
towards #1061, a simple update to sphinx-gallery seems to do the trip. Keen to see if more needs to be done @romanlutz @hildeweerts

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -166,6 +166,7 @@ pygments_style = "sphinx"
 # included in the gallery, but its plot is available for
 # the quickstart
 sphinx_gallery_conf = {
+    "reference_url": {"fairlearn": None},
     "examples_dirs": "../examples",
     "gallery_dirs": "auto_examples",
     # pypandoc enables rst to md conversion in downloadable notebooks


romanlutz:
Perhaps I'm misreading the issue #1061 but I thought it's the reverse. Let me elaborate:
If you find something (a class or function, for example) in the API reference and want to see the examples for it, there's no way to find them other than checking all examples. I believe that to be the problem statement of #1061. @alliesaizan can confirm 🙂 
What your PR does (I think) is adding links into examples that can be used to find the API reference for classes or functions used in the examples. So that's what I mean by reverse, one goes from API ref --> examples, the other backwards. 

I believe both are very useful! So I'd definitely be in favor of merging this, but I don't think it addresses the issue quite yet.

yanhong-zhao-ef:
> Perhaps I'm misreading the issue #1061 but I thought it's the reverse. Let me elaborate: If you find something (a class or function, for example) in the API reference and want to see the examples for it, there's no way to find them other than checking all examples. I believe that to be the problem statement of #1061. @alliesaizan can confirm 🙂 What your PR does (I think) is adding links into examples that can be used to find the API reference for classes or functions used in the examples. So that's what I mean by reverse, one goes from API ref --> examples, the other backwards.
> 
> I believe both are very useful! So I'd definitely be in favor of merging this, but I don't think it addresses the issue quite yet.

Thanks for taking a look at this @romanlutz and I believed @alliesaizan already added the user guide section to the API documentation. Are you suggesting I can add the relevant notebook to relevant API? Wouldn't it be better to reference the notebooks from the user guide? Or maybe both?

romanlutz:
> > Perhaps I'm misreading the issue #1061 but I thought it's the reverse. Let me elaborate: If you find something (a class or function, for example) in the API reference and want to see the examples for it, there's no way to find them other than checking all examples. I believe that to be the problem statement of #1061. @alliesaizan can confirm 🙂 What your PR does (I think) is adding links into examples that can be used to find the API reference for classes or functions used in the examples. So that's what I mean by reverse, one goes from API ref --> examples, the other backwards.
> > I believe both are very useful! So I'd definitely be in favor of merging this, but I don't think it addresses the issue quite yet.
> 
> Thanks for taking a look at this @romanlutz and I believed @alliesaizan already added the user guide section to the API documentation. Are you suggesting I can add the relevant notebook to relevant API? Wouldn't it be better to reference the notebooks from the user guide? Or maybe both?

Here's what I *think* @alliesaizan meant: If you go to, say, [this API reference](https://fairlearn.org/main/api_reference/fairlearn.reductions.html#fairlearn.reductions.ExponentiatedGradient) you should see all the examples in which it is used. Either we don't have that yet (otherwise it would show up) OR sphinx is somehow misconfigured (somewhat unlikely). I believe this task to be somewhat of a manual effort of going through all the examples and cross-linking them, but maybe I'm wrong.

Now what you bring up in your latest comment is linking user guide section to the API reference. True, that has been done. There's three things here:
- user guide
- API ref
- examples
and we can cross-link them all one way or the other 🙂 I think all the cross-links are useful in their own way. After all, why search manually when you can have a link presented to you? What you're adding here (and I checked: it works!) is one of the cross-links: the one from examples to API reference. So I'm absolutely in favor of merging your change. I just don't know if it's 100% what @alliesaizan asked for, so perhaps we want to leave the issue open. I'll leave it to @alliesaizan to decide.

Any other member of @fairlearn/fairlearn-maintainers can merge this PR assuming gates pass (which they will after #1099 is merged).

alliesaizan:
> Perhaps I'm misreading the issue #1061 but I thought it's the reverse. Let me elaborate: If you find something (a class or function, for example) in the API reference and want to see the examples for it, there's no way to find them other than checking all examples. I believe that to be the problem statement of #1061. @alliesaizan can confirm 🙂 What your PR does (I think) is adding links into examples that can be used to find the API reference for classes or functions used in the examples. So that's what I mean by reverse, one goes from API ref --> examples, the other backwards.
> 
> I believe both are very useful! So I'd definitely be in favor of merging this, but I don't think it addresses the issue quite yet.

Yes I am sorry, this issue is worded vaguely and arose after a discussion in early April. Roman is correct, and I recall that the intent of the issue is to make it easier for a user to see examples of how to implement the methods in the API documentation.  However, I believe the reverse (linking references to the API in the example notebooks) is also useful, though that is not what's asked for here.

adrinjalali (commented):
Thanks for the PR @yanhong-zhao-ef 

As others pointed out, this is a nice addition, but doesn't fix the original issue.

One way to fix the original issue is to automatically have links to all examples from the API guides the same way scikit-learn does (example: https://scikit-learn.org/stable/modules/generated/sklearn.compose.ColumnTransformer.html#examples-using-sklearn-compose-columntransformer)

But I think for us, since we don't have many examples, the idea was to manually add links to examples from the user guide and API docs whenever appropriate, if I'm not mistaken.

romanlutz:
> Thanks for the PR @yanhong-zhao-ef
> 
> As others pointed out, this is a nice addition, but doesn't fix the original issue.
> 
> One way to fix the original issue is to automatically have links to all examples from the API guides the same way scikit-learn does (example: https://scikit-learn.org/stable/modules/generated/sklearn.compose.ColumnTransformer.html#examples-using-sklearn-compose-columntransformer)
> 
> But I think for us, since we don't have many examples, the idea was to manually add links to examples from the user guide and API docs whenever appropriate, if I'm not mistaken.

I think we just didn't know there's an automatic way to do this @adrinjalali 😆 That's even better! How do we turn that on?

PR #1113, opened 2022-07-18
[NOCHECKIN] Confidence Intervals using bootstrap

riedgar-ms:
## Description

This is an update of #1038 to sort out some git confusion. It will address #999 . The following description is taken from the original PR:

- Use of [bootstrap](https://en.wikipedia.org/wiki/Bootstrapping_(statistics)) to estimate the empirical distribution of any scalar metric passed to MetricFrame. Bootstrap settings are controlled by two new init parameters: `n_boot`, which determines how many replications of the dataset to generate with replacement, and `n_jobs`, which controls how many parallel processes to use in the bootstrap operation.

- Confidence interval estimation as a summary of the empirical bootstrap distribution. This is controlled by the `ci` and `ci_method` parameters, which offer options of which quantiles to sample and whether or not estimates should be bias-corrected. Many users can use simple parameter settings like `ci=0.95` for a 95% confidence interval on each MetricFrame statistic, while advanced users can query specific quantiles (e.g. `ci=[0.25, 0.5, 0.75]`) for further study. These results are calculated during initialization and are stored in new attributes on MetricFrame suffixed in `.*_ci`. 

Some significant changes have been made to MetricFrame internals in order to support this set of operations. Firstly, because we build the bootstrap distribution during initialization and chose not to pin it on the object, some operations like `.difference` are calculated during init. Secondly, because bootstrapping involves calculating the same operation on different (resampled) inputs, I've changed some internal functions like `def __group()` to work on any set of inputs, and not be hard-coded to use pinned attributes like `self.overall` and `self.by_group`. I've also refactored some general code into new private functions to make repeated calls easier. These should have no breaking changes to an end user. 

The new `.*_ci` attributes are all lists of tuples, with each tuple representing the quantile of the distribution in the first element, and the corresponding values for each metric in the second. For example, if a user requests a 90% confidence interval with `ci=0.9`, the output of `.overall_ci` will look like:

```python
.overall_ci = [
   (0.05, pd.DataFrame({...})),
   (0.95, pd.DataFrame({...}))
]
```

where each of the inner frames exactly parallel the structure of `.overall`. This API design hopefully makes it easy for users to adopt confidence intervals into their workflow -- any code that runs on the "main" MetricFrame objects will also work on the .ci versions.

## Tests

- [ ] no new tests required
- [x] new tests added
- [ ] existing tests adjusted

## Documentation

### DIFF
--- a/docs/user_guide/assessment.rst
+++ b/docs/user_guide/assessment.rst
@@ -621,6 +621,10 @@ Bootstrapping works as follows:
 3. Fit a distribution to this set of metric values
 4. Obtain the desired confidence intervals from that distribution
 
+Since bootstrapping involves random resampling, results will not
+be precisely reproducible between runs.
+However, as :code:`n_boot` is increased, the values found should
+converge.
 Let us look at an example:
 
 .. doctest:: assessment_metrics
--- a/docs/user_guide/assessment.rst
+++ b/docs/user_guide/assessment.rst
@@ -664,17 +664,39 @@ intervals:
     C    0.625000
     Name: accuracy_score, dtype: float64
     >>> metric_ci.by_group_ci
-    [(0.025, SF
+    [(0.025,
+    SF
     A    0.000000
-    B    0.090909
-    C    0.285714
-    Name: accuracy_score, dtype: float64), (0.975, SF
-    A    0.600
-    B    0.625
-    C    1.000
+    B    0.105278
+    C    0.333333
+    Name: accuracy_score, dtype: float64),
+    (0.975,
+    SF
+    A    0.621429
+    B    0.608077
+    C    0.952500
     Name: accuracy_score, dtype: float64)]
 
+So here, we see that group 'B' has a nominal value of 0.333, with its 95% confidence
+interval running from 0.105 to 0.608.
+Finally, confidence intervals are also provided for the :method:`MetricFrame.difference`
+and :method:`MetricFrame.ratio` methods.
+For example:
 
+
+.. doctest:: assessment_metrics
+    :options:  +NORMALIZE_WHITESPACE
+
+    >>> metric_ci.difference(method='between_groups')
+    0.325
+    >>> metric_ci.difference_group_ci
+    [(0.025, 0.10214826839826842), (0.975, 0.7585227272727271)]
+
+What we see here is that the largest difference between any two of the groups
+marked by the sensitive feature is nominally 0.325, but the 95% confidence interval
+ranges from 0.102 to 0.759. This is quite a range, but not unexpected when
+we have such a small number of samples in our original dataset and we are taking
+a difference between two similar values.
 
 
 .. _plot:
--- a/docs/user_guide/assessment.rst
+++ b/docs/user_guide/assessment.rst
@@ -589,6 +589,57 @@ For more examples, please
 see the :ref:`sphx_glr_auto_examples_plot_new_metrics.py` notebook in the
 :ref:`examples`.
 
+.. _bootstrap:
+
+Confidence Intervals with Bootstrapping
+---------------------------------------
+
+It is a truth universally acknowledged, that a person in possession of
+a statistic, must be in want of an error bar.
+The only question is whether or not the person is aware of their want.
+Two aspects of fairness metrics make this particularly acute:
+
+1. Some of the subgroups defined by the sensitive features may be much
+   smaller than the overall sample size, leading to high levels of
+   random noise
+2. We often then wish to measure the disparity between two such subgroups.
+   While this means taking a difference or a ratio of the metrics,
+   the errors will always add (in quadrature)
+
+To address the need for error estimates, :class:`MetricFrame` provides
+a means of estimating confidence intervals for scalar metrics by using
+a process known as bootstrapping.
+This avoids the need for an analytic form for the error to be provided
+for each metric computed.
+
+Bootstrapping works as follows:
+
+1. Generate a large number of synthetic datasets by sampling the
+   target dataset *with replacement* (i.e. a given sample from the target
+   dataset may appear more than once in a synthetic dataset)
+2. Compute the metric value for each of these synthetic datasets
+3. Fit a distribution to this set of metric values
+4. Obtain the desired confidence intervals from that distribution
+
+Let us look at an example:
+
+.. doctest:: assessment_metrics
+    :options:  +NORMALIZE_WHITESPACE
+
+    >>> # Enable bootstrapping with the n_boot parameter
+    >>> metric_ci = MetricFrame(metrics=skm.accuracy_score,
+    ...                         y_true=decision,
+    ...                         y_pred=prediction,
+    ...                         sensitive_features={'SF' : sensitive_feature},
+    ...                         n_boot=100)
+    >>> metric_ci.overall
+    0.1
+    >>> metric_ci.overall_ci
+    0.1
+
+
+
+
 .. _plot:
 
 Plotting
--- a/docs/user_guide/assessment.rst
+++ b/docs/user_guide/assessment.rst
@@ -633,9 +633,46 @@ Let us look at an example:
     ...                         sensitive_features={'SF' : sensitive_feature},
     ...                         n_boot=100)
     >>> metric_ci.overall
-    0.1
+    0.4
     >>> metric_ci.overall_ci
-    0.1
+    [(0.025, 0.23333333333333334), (0.975, 0.6)]
+
+We have turned on bootstrapping by setting the :code:`n_boot` parameter;
+in this case, we have requested that 100 synthetic datasets be generated
+to create the statistics.
+We see that the :attr:`MetricFrame.overall` property is unchanged.
+However, we can also access the :attr:`MetricFrame.overall_ci` property
+which contains estimates of the confidence interval.
+By default, the bootstrapping algorithm computes the 95% confidence
+interval, which ranges from the 25th millile to the 975th millile.
+The :attr:`MetricFrame.overall_ci` has returned a list of binary tuples.
+The first entry in each tuple indicates the position within the distribution
+(which are 0.025 and 0.975 for the default confidence interval), while
+the second entry in each tuple is the metric value.
+For this case, we see that we can be 95% certain that the overall
+accuracy lies between 0.23 and 0.6 (and has a nominal value of 0.4).
+We can also examine the statistics for each group, and their confidence
+intervals:
+
+.. doctest:: assessment_metrics
+    :options:  +NORMALIZE_WHITESPACE
+
+    >>> metric_ci.by_group
+    SF
+    A    0.300000
+    B    0.333333
+    C    0.625000
+    Name: accuracy_score, dtype: float64
+    >>> metric_ci.by_group_ci
+    [(0.025, SF
+    A    0.000000
+    B    0.090909
+    C    0.285714
+    Name: accuracy_score, dtype: float64), (0.975, SF
+    A    0.600
+    B    0.625
+    C    1.000
+    Name: accuracy_score, dtype: float64)]
 
 
 
--- a/fairlearn/metrics/_bootstrap.py
+++ b/fairlearn/metrics/_bootstrap.py
@@ -9,102 +9,136 @@ import numpy as np
 from scipy.stats import norm
 import numbers
 
+
 def process_ci_bounds(ci, ci_method, precision=6):
-    '''Parses user inputs to construct confidence intervals from bootstrap.
-       This function interprets integer and float inputs as the desired
-       width of a two-sided confidence interval, and lists and tuples as
-       exact quantiles to retrieve.
+    """Parses user inputs to construct confidence intervals from bootstrap.
+    This function interprets integer and float inputs as the desired
+    width of a two-sided confidence interval, and lists and tuples as
+    exact quantiles to retrieve.
 
-       Parameters
-        ----------
-        ci : float, list, tuple
-        Determines the quantiles reported for confidence intervals. All entries must
-        be in the range [0, 1]. A single float is interpreted as the width of the 
-        desired confidence interval (e.g. 0.95 is a 95% interval), while a List or Tuple
-        reports exact quantiles from the empirical bootstrap distribution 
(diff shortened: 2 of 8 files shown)


fairlearn/metrics/_metric_frame.py
```diff
@@ -364,6 +392,138 @@ def __init__(
...
+                    self.by_group, group_min_output
+                )
+                difference_overall_output = self.__difference(
+                    self.by_group, self.overall
+                )
+                ratio_group_output = group_min_output / group_max_output
+                ratio_overall_output = self._calc_ratio_overall(
+                    self.by_group, self.overall
+                )
+            except:
```
riedgar-ms, line 457: Having been bitten by this when bring the PR over, I think it would be best not to completely throw out the actual exception. Given that the raised `ValueError` will likely be the most common for users, I can see the sense in keeping it, but I think we should nest the actual exception inside it.

test/unit/metrics/test_metricframe_bootstrap.py
```diff
@@ -0,0 +1,261 @@
...
+        mf.ratio(method="to_overall").shape == ci_output.shape
+        for _, ci_output in mf.ratio_overall_ci
+    )
+    assert all(
+        mf.ratio(method="between_groups").shape == ci_output.shape
+        for _, ci_output in mf.ratio_group_ci
+    )
+
+
+def check_all_bootstrap_directions(mf):
```
riedgar-ms, line 58: This should probably check that the values reported by `overall`, `difference()` etc. lie within the confidence intervals

fairlearn/metrics/_metric_frame.py
```diff
@@ -651,19 +866,39 @@ def difference(
...
+                result = tmp.transform(lambda x: min(x, 1 / x)).min()
         else:
-            result = (mf - subtrahend).abs().groupby(level=self.control_levels).max()
+            if self.control_levels:
+                # It's easiest to give in to the DataFrame columns preference
+                ratios = self.by_group.unstack(
+                    level=self.control_levels
+                ) / self.overall.unstack(level=self.control_levels)
+            else:
+                ratios = self.by_group / self.overall
```
riedgar-ms, line 889: Similarly here - and `overall_frame` too @harsha-nori ?

riedgar-ms:
Also should add some API docs, and possibly an example notebook.

fairlearn/metrics/_metric_frame.py
riedgar-ms: Fixed this mone.

test/unit/metrics/test_metricframe_aggregates.py
```diff
@@ -125,6 +125,17 @@ def test_ratio_to_overall(self, metric_fn):
             expected_ratio_overall, rel=1e-10, abs=1e-16
         )
 
+    def test_ratio_zero_divide(self):
```
riedgar-ms, line 128: This is for a lurking bug... I can put into a separate PR if there are strong feelings, but it would be simpler to keep here.

fairlearn/metrics/_metric_frame.py
```diff
@@ -651,19 +866,39 @@ def difference(
...
-        # Can assume errors='coerce', else error would already have been raised in .group_min
-        # Fill all non-scalar values with NaN
-        if isinstance(mf, pd.Series):
-            mf = mf.map(lambda x: x if np.isscalar(x) else np.nan)
-        else:
-            mf = mf.applymap(lambda x: x if np.isscalar(x) else np.nan)
+        return self.__difference(self.by_group, subtrahend)
+
+    def _calc_ratio_overall(self, by_group_frame, overall_frame):
+        def ratio_sub_one(x):
```
riedgar-ms, line 872: This is part of the bugfix mentioned for the aggregates. Basically, this makes everything use `ratio_sub_one()` rather than using `min(x, 1/x)` in some places.

docs/user_guide/assessment.rst
```diff
@@ -589,6 +589,94 @@ For more examples, please
...
+    [(0.025, 0.23333333333333334), (0.975, 0.6)]
+
+We have turned on bootstrapping by setting the :code:`n_boot` parameter;
+in this case, we have requested that 100 synthetic datasets be generated
+to create the statistics.
+We see that the :attr:`MetricFrame.overall` property is unchanged.
+However, we can also access the :attr:`MetricFrame.overall_ci` property
+which contains estimates of the confidence interval.
+By default, the bootstrapping algorithm computes the 95% confidence
+interval, which ranges from the 25th millile to the 975th millile.
```
riedgar-ms, line 647: Feedback on whether 'millile' is the right word invited.

riedgar-ms: Or should I just go with 'quantile' and give decimals?

test/unit/metrics/test_metricframe_bootstrap_values.py
```diff
@@ -0,0 +1,180 @@
...
+import math
+
+import numpy as np
+import pytest
+
+from sklearn.metrics import accuracy_score
+
+from fairlearn.metrics import MetricFrame
+
+# The setup is slightly unorthodx, to allow for splitting of
```
riedgar-ms, line 19: Comments welcome on this pattern, since I've not used it before. I'm trying to avoid having a huge test case with lots of assertions in it, but at the same time, having a bunch of stuff critical to the test cases defined at the top of the file like this isn't ideal either.

test/unit/metrics/test_metricframe_bootstrap_values.py
```diff
@@ -0,0 +1,180 @@
...
+        for i in range(len(sigma_vals)):
+            v_A = mean_A + (sigma_vals[i] * sigma_A)
+            vf_A = v_A / n_A
+
+            v_B = mean_B + (sigma_vals[i] * sigma_B)
+            vf_B = v_B / n_B
+
+            # Slightly wider tolerance here.... not sure if this should be required
+            assert mf.by_group_ci[i][1]["A"] == pytest.approx(
+                vf_A, rel=2 / math.sqrt(n_A)
```
riedgar-ms, line 144: @MiroDudik and @Harsha-Nori, your feedback here would be much appreciated. This is a wider tolerance than I was expecting to need, so I'm worried that I'm doing something wrong in the calculation.

test/unit/metrics/test_metricframe_bootstrap_values.py
```diff
@@ -0,0 +1,180 @@
...
+        # Do the error calculation
+        nominal_acc_AB = abs(p_A - p_B)
+
+        sigma_A = math.sqrt(n_A * p_A * (1 - p_A))
+        sigma_B = math.sqrt(n_B * p_B * (1 - p_B))
+        acc_sigma_A = sigma_A / n_A
+        acc_sigma_B = sigma_B / n_B
+
+        # The actual quadrature
+        acc_sigma_AB = math.sqrt(acc_sigma_A**2 + acc_sigma_B**2)
```
riedgar-ms, line 169: Again @MiroDudik  and @harsha-nori, I'd appreciate it if you double check my work here. It's been a while since I did error calculations.

riedgar-ms:
Trying to add some checks on the actual values produced by the bootstrapping. I'd be very grateful if my approach could be double checked (before I go and fill out ratios etc.). @MiroDudik  and @harsha-nori especially.

riedgar-ms:
Ping @MiroDudik  and @harsha-nori

riedgar-ms:
Ping again.... @MiroDudik @harsha-nori

riedgar-ms:
Ping @harsha-nori and @MiroDudik . I'd really appreciate some feedback on this before I expand the testing

fairlearn/metrics/_metric_frame.py
```diff
@@ -309,6 +333,10 @@ def __init__(
         sample_params: Optional[
             Union[Dict[str, Any], Dict[str, Dict[str, Any]]]
         ] = None,
+        n_boot: Optional[int] = None,
+        ci: Union[float, List[float], Tuple[float]] = 0.95,
```
MiroDudik, line 337: I think that the default should be [0.025, 0.975]

riedgar-ms:
Spoke briefly with @MiroDudik about this yesterday. He suggests that not returning tuples, but instead:

- Having a `MetricFrame.quantiles_` property which stores the list of quantiles computed
- Convert the existing lists of Tuples into simple lists. These then are 'indexed' by `quantiles_`

I quite like this idea (stop returning lots of redundant data), but it will be a big change. So before I do so, I'm soliciting thoughts @Harsha-Nori, @adrinjalali @hildeweerts ?

docs/user_guide/assessment.rst
riedgar-ms: Will do. Although this particular part of the PR is not going to survive the other changes I'm making to the MetricFrame documentation.

fairlearn/metrics/_bootstrap.py
```diff
@@ -0,0 +1,144 @@
...
+"""Utility functions for bootstrap calculations (used for confidence intervals)."""
+
+from warnings import warn
+import pandas as pd
+import numpy as np
+from scipy.stats import norm
+import numbers
+
+
+def process_ci_bounds(ci, ci_method, precision=6):
```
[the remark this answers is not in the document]
riedgar-ms: @harsha-nori can you comment? This bit was copied straight from your PR.

fairlearn/metrics/_metric_frame.py
```diff
@@ -364,6 +392,138 @@ def __init__(
             all_data, annotated_funcs, grouping_features
         )
 
+        # Calculate uncertainty intervals using bootstrapped versions of data
+        def _bootstrap_metrics_calc(data):
```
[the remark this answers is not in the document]
riedgar-ms: Let me think about some possibilities. This was just a direct copy of @harsha-nori 's PR, to get it moving again. I'm certainly not a huge fan of how things are calculated right now.

fairlearn/metrics/_metric_frame.py
```diff
@@ -405,6 +565,29 @@ def overall(self) -> Union[Any, pd.Series, pd.DataFrame]:
         else:
             return self._overall
 
+    @property
+    def overall_ci(
```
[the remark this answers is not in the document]
riedgar-ms: Have been applying this change today.

test/unit/metrics/test_metricframe_bootstrap_values.py
riedgar-ms: Hmm.... this has been lost from the 'Files' tab unfortunately :-/


I think I get what you're saying here @MiroDudik , but I'd like to be certain.... Basically, what I was aiming to do here was verify that sensible values were being produced, because nothing in the existing tests did that. What you're saying @MiroDudik  is that the way I've gone about this is also testing whether or not bootstrapping works statistically (because I'm relying on certain properties of binomial and normal distributions, and then hammering down uncertainties with a large n_boot)? So a better approach would be to create a 'known' set of bootstrap samples, and then perform the quantile calculations on those?

test/unit/metrics/test_metricframe_bootstrap_values.py
riedgar-ms: It sounds like you're suggesting testing the guts of `create_ci_output()` in `_bootstrap.py` - the stuff around `np.quantile()` specifically. I would actually be happy to do that; this was an option because I understood what was supposed to be going on at this high level. The expected output of `create_ci_output()` was not so easy to discern...

fairlearn/metrics/_metric_frame.py
```diff
@@ -223,6 +226,27 @@ class MetricFrame:
...
+        not supported on complex metrics that yield multiple outputs.
+    ci : float, list, tuple
+        Determines the quantiles reported for confidence intervals. All entries must
+        be in the range [0, 1]. A single float is interpreted as the width of the
+        desired confidence interval (e.g. 0.95 is a 95% interval), while a List or Tuple
+        reports exact quantiles from the empirical bootstrap distribution
+        (e.g. (0, 0.33, 0.5, 0.66, 1) will calculate the min, 33rd, 50th, 66th, and max quantiles).
+    ci_method: str
+        One of {'percentile', 'bias-corrected'}. This method determines the algorithm used to
+        produce valid confidence intervals from the bootstrap distribution. 'percentine' returns
```
hildeweerts, line 243:
```suggestion
        produce valid confidence intervals from the bootstrap distribution. 'percentile' returns
```

fairlearn/metrics/_metric_frame.py
```diff
@@ -405,6 +565,29 @@ def overall(self) -> Union[Any, pd.Series, pd.DataFrame]:
...
+                    ]
+                else:
+                    return [
+                        (quantile, output.iloc[0])
+                        for quantile, output in self._overall_ci
+                    ]
+            else:
+                return self._overall_ci
+
+        return None
```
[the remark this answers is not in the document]
hildeweerts: To me `None` makes sense, because that it throws an error if you try to iterate over `_overall_ci` which might be useful for debugging purposes. But I don't know what the best practices are :)

docs/user_guide/assessment.rst
```diff
@@ -815,6 +815,120 @@ For more examples, please
...
+
+.. doctest:: assessment_metrics
+    :options:  +NORMALIZE_WHITESPACE
+
+    >>> metric_ci.difference(method='between_groups')
+    0.325
+    >>> metric_ci.difference_group_ci
+    [(0.025, 0.10214826839826842), (0.975, 0.7585227272727271)]
+
+What we see here is that the largest difference between any two of the groups
```
hildeweerts, line 925: Should we also mention somewhere that fairness metrics such as max-min difference are (upwards) statistically biased estimates (see Section 4 / Figure 2 of  [Lum et al. 2022](https://arxiv.org/pdf/2205.05770.pdf)), or would that make things too complicated for our user guide?

hildeweerts (commented):
Just a few small comments :)

riedgar-ms:
Will be closing this in favour of #1125 . Keeping until I've retrieved the docs

riedgar-ms:
Superseded

PR #1120, opened 2022-08-24
Drafting a workflow for releasing to PyPI

riedgar-ms:
## Description

This starts the process of creating a GitHub workflow to release Fairlearn to PyPI. The actual release will be part of a later update; this just gets things started.

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation


.github/workflows/release-wheel.yml
```diff
@@ -0,0 +1,63 @@
...
+
+env:
+  VERSION_FILE_NAME: version.txt
+  WHEEL_ARTIFACT: fairlearn-wheels
+
+jobs:
+  build-wheel:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
```
adrinjalali, line 20: should we be checking out a specific branch?

.github/workflows/release-wheel.yml
```diff
@@ -0,0 +1,63 @@
...
+    - name: Remove Fairlearn source directory
+      uses: JesseTG/rm@v1.0.3
+      with:
+        path: fairlearn
+    - name: Download wheel artifact
+      uses: actions/download-artifact@v3
+      with:
+        name: env.WHEEL_ARTIFACT
+    - name: Install Fairlearn from wheel
+      run: python scripts/install_fairlearn_from_wheel.py --wheel-dir=.
```
adrinjalali, line 59: can this file not be removed and have here just  simple command?

.github/workflows/release-wheel.yml
```diff
@@ -0,0 +1,63 @@
...
+    - name: Install testing requirements
+      run: pip install pytest==7.1.1 pytest-mock pytest-mpl>=0.11 lightgbm
```
adrinjalali, line 61: do we need to pin pytest version?

adrinjalali (commented):
we can also add a workflow which needs to run manually, like the one here: 

https://github.com/skops-dev/skops/blob/main/.github/workflows/publish-pypi.yml

.github/workflows/release-wheel.yml
riedgar-ms: I spent some time trying to do that. However, to build the name of the `.whl` I'd have to figure out the version we're about to release.

adrinjalali:
Can we make things a bit more standard so that `python -m build` would work, and then test in the same step as build so that we don't have to deal with artifacts then?

riedgar-ms:
> Can we make things a bit more standard so that `python -m build` would work, and then test in the same step as build so that we don't have to deal with artifacts then?

I'd be happy to make our setup closer to the standard (I'd have to look up `python -m build` since I've not seen that before). However, I don't think that would remove the need for the Artifact. Firstly, we need o test the wheel we've built on multiple platforms and Python versions. Secondly, having the artifact will give us a record of what we think we uploaded to PyPI, should we every have to track down issues.

adrinjalali:
I'm assuming you're not waiting for my review here. Please let me know if that's not the case.

riedgar-ms:
> I'm assuming you're not waiting for my review here. Please let me know if that's not the case.

I was actually..... @adrinjalali . I do need to write up the release proposal we discussed too, since this workflow would be part of it.

adrinjalali (commented):
I'd be happy to merge this, but rather not introduce a new python file to install the package, w/o that I'm happy to have this merged and then have next steps to make sure things work as they should.

scripts/install_fairlearn_from_wheel.py
```diff
@@ -0,0 +1,48 @@
...
+    with _LogWrapper("Finding wheel file"):
+        target_dir = pathlib.Path(args.wheel_dir)
+        wheel_list = list(target_dir.glob("fairlearn*.whl"))
+        assert len(wheel_list) == 1, f"Bad wheel_list: {wheel_list}"
+        wheel_path = wheel_list[0].resolve()
+        msg = f"Path to wheel: {wheel_path}"
+        _logger.info(msg)
+
+    with _LogWrapper("Installing wheel"):
+        fairlearn_spec = f"fairlearn[customplots] @ {wheel_path.as_uri()}"
```
riedgar-ms, line 43: This script mainly exists for this line @adrinjalali

adrinjalali (approved):
I really dislike a python file doing the install, but don't want to block this. Happy for it to be merged and then you can move forward. I personally wouldn't have done it this way. But I'm not the one doing it, so ...

riedgar-ms:
@adrinjalali I've been trying to install the wheel without using a script, but I'm not having much luck.
```bash
pip install --no-index --find-links <dir> fairlearn[customplots]
```
Doesn't work because there aren't any dependencies available. It appears that there's no easy way to extract the dependencies from the wheel file and then install them with a regular `pip` call.

The key problem that the script is solving is installing from the wheel when we don't know the full name of the file in advance (since it includes the version) and we also want to include 'extras' for the package.

riedgar-ms:
Thanks for the help @adrinjalali  - extra script has been removed.

.github/workflows/release-wheel.yml
```diff
@@ -0,0 +1,94 @@
...
+    - name: Download wheel artifact
+      uses: actions/download-artifact@v3
+      with:
+        name: env.WHEEL_ARTIFACT
+    - name: Install Fairlearn from wheel
+      # Do this in three steps to make sure that fairlearn wasn't pulled from PyPI
+      run: |
+        pip install --pre --find-links . "fairlearn[customplots]"
+        pip uninstall fairlearn --yes
+        pip install --no-index --find-links . "fairlearn[customplots]"
```
adrinjalali, line 69: 🤣 one way to do it!

PR #1123, opened 2022-09-08
[Docs] Split Assessment section

riedgar-ms:
## Description

Split the 'Assessment' section of the User Guide into multiple files, to keep each to a more manageable size.

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation


riedgar-ms:
@fairlearn/fairlearn-maintainers this is a first pass, strictly just splitting up the doc into multiple ReST files. I'm expecting some broken builds, but would like feedback on the split.

Tagging @alliesaizan @hildeweerts @mmadaio in particular.

adrinjalali (approved):
Nice!

riedgar-ms:
Ping @mmadaio @hildeweerts @alliesaizan

hildeweerts:
Nice! We might need to fine-tune the details (e.g., we start "More complex metrics" with "So far", when not everyone may have read all that came before) but generally these splits make sense to me.

I think most of the headers are quite self-contained, but 
* Intersecting groups
* Extra arguments to metric functions
* More complex metrics
might be harder to grasp for (new) users. Would it make sense to group these together in a page called "Advanced usage of MetricFrame"?

riedgar-ms:
Thanks @hildeweerts ! Redrafting will definitely be needed; I was just trying to make minimal changes. I can certainly look at how the sections start before merge.

I had been wondering about those last three sections. I see your point; at the same time, I also think that they should probably be expanded (especially Intersecting Metrics), and would then be worth having their own page.

Any thoughts @alliesaizan ? @mmadaio ?

alliesaizan:
> Thanks @hildeweerts ! Redrafting will definitely be needed; I was just trying to make minimal changes. I can certainly look at how the sections start before merge.
> 
> I had been wondering about those last three sections. I see your point; at the same time, I also think that they should probably be expanded (especially Intersecting Metrics), and would then be worth having their own page.
> 
> Any thoughts @alliesaizan ? @mmadaio ?

I think overall the new layout makes sense, and agree with Hilde's feedback.  Maybe we could combine _extra arguments to metric functions_ and _more complex metrics_ and add a little more context to the _intersecting groups_ section? We introduce the topic of intersecting groups under "Conditional Group Fairness" in the _common fairness_ metrics page; maybe moving this section to the intersecting groups page would provide useful background for readers?

adrinjalali:
I like the suggestions here, but I think we shouldn't halt this PR for content which we don't have here. 

For things like "we talked above" where "above" doesn't make sense, the requested changes are fine, but for things like "we should cover this thing as well", I think would be outside the scope of this PR.

riedgar-ms:
OK @fairlearn/fairlearn-maintainers  and @alliesaizan , I've merged the two 'advanced' sections as agreed, and fixed up `doctest`. Unless you spot issues, I think this is ready to merge.

alliesaizan:
> OK @fairlearn/fairlearn-maintainers and @alliesaizan , I've merged the two 'advanced' sections as agreed, and fixed up `doctest`. Unless you spot issues, I think this is ready to merge.

I agree that the merged advanced section looks good layout-wise!

hildeweerts (approved):
Awesome! Please feel free to merge.

PR #1125, opened 2022-09-16
[NOCHECKIN] Bootstrapped Confidence Intervals for MetricFrame

riedgar-ms:
## Description

Introduce bootstrapping for `MetricFrame`, based on #1038 by @Harsha-Nori. This is currently under development.

## Tests

- [ ] no new tests required
- [x] new tests added
- [ ] existing tests adjusted

## Documentation


riedgar-ms:
I will probably skip the 'edge correction' and parallelisation from the original PR, at least initially; For right now, I want to make sure that things work.

fairlearn/metrics/_metric_frame.py
```diff
@@ -462,7 +504,9 @@ def __group(
         else:
             return result
 
-    def group_max(self, errors: str = "raise") -> Union[Any, pd.Series, pd.DataFrame]:
+    def group_max(
+        self, errors: str = "raise", quantiles: Optional[List[float]] = None
```
riedgar-ms, line 508: This is my main thought for the new API @MiroDudik  - have an optional `quantiles=` argument, which will trigger the usage of the bootstrap samples computed in the constructor

riedgar-ms (commented):
@fairlearn/fairlearn-maintainers comments welcome. So far, this sketches out the approach and shows a couple of changes in `MetricFrame` itself

fairlearn/metrics/_metric_frame.py
```diff
@@ -367,6 +382,20 @@ def overall(self) -> Union[Any, pd.Series, pd.DataFrame]:
         else:
             return self._result.overall
 
+    def overall_quantiles(self, quantiles: List[float]):
```
riedgar-ms, line 385: In a way, I'm regretting that `overall` and `by_group` are properties now.

docs/user_guide/assessment/error_estimation.rst
```diff
@@ -0,0 +1,358 @@
...
+    ...         'bootstrap_sum' : bootstrap_sum,
+    ...         'count' : count
+    ... }
+    >>> mf = MetricFrame(
+    ...         metrics=metrics,
+    ...         y_true=y_true,
+    ...         y_pred=y_true,
+    ...         sensitive_features={ 'SF0' : s_f },
+    ...         n_bootstrap_samples=100,
+    ...         bootstrap_random_state=13489623,
```
adrinjalali, line 224: interesting random state?

docs/user_guide/assessment/error_estimation.rst
```diff
@@ -0,0 +1,358 @@
...
+    >>> bootstrap_sum(y_true, y_true)
+    500.0
+
+Activating Bootstrapping in :code:`MetricFrame`
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+With our data in place, we can create our :class:`MetricFrame`
+object.
+In order to use bootstrapping, we specify the
+:code:`n_bootstrap_samples` and :code:`bootstrap_random_state`
```
adrinjalali, line 205: why does it need to be specified why the random_state is passed? On the API, we can simply have a `random_state` and use it wherever we need randomness.

adrinjalali (commented):
Looking at the user API, it's pretty nice. But the code is really sparse on docs.

docs/user_guide/assessment/error_estimation.rst
```diff
@@ -0,0 +1,358 @@
+.. _error_estimation:
```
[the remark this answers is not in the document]
riedgar-ms: I've been considering that option too. This is very preliminary.

docs/user_guide/assessment/error_estimation.rst
riedgar-ms: There is an alternative to banging my head on the keyboard:
https://xkcd.com/221/

:-)

docs/user_guide/assessment/error_estimation.rst
riedgar-ms: It was concern about the user passing an instance of the old or new numpy random API - and that we can't be sure that the user has a version of numpy installed which has the new API (which makes testing for types tricky. There's probably a way around this.

This comes into play when setting up the PRNGs, where each bootstrap iteration expects to get a fresh RNG, so if a single item (and not a list) is passed, I instead use the provided RNG to seed a new RNG for each iteration.

This can probably all be resolved, but for simplicity I imposed this restriction at first.

adrinjalali: As for which RNG, we certainly should simply support both, and also support a simple integer (as a seed).

But I was referring more to the name of the argument: `random_state` instead of `bootstrap_random_state`.

riedgar-ms: I misunderstood your question :-) I'm pretty agnostic on the name, but I'm sure @MiroDudik will have thoughts :-)

riedgar-ms:
Ping..... @MiroDudik trying to get this back to the top of your queue

riedgar-ms:
Ping @MiroDudik again.... it would be *really* nice to have some feedback on this, so that I can finish it up and get it merged.

docs/user_guide/assessment/error_estimation.rst
```diff
@@ -0,0 +1,360 @@
...
+the relative effects of random noise.
+We are then generally interested in the difference or ratio
+of function evaluations on these groups, and errors always
+accumulate, even when the target values (the difference or ratio
+in this case) are getting smaller.
+:ref:`Intersecting groups <intersecting_groups>` make the
+problem worse, since some intersections can have very low
+sample counts, or even be empty.
+
+In Fairlearn, we offer bootstrapping as a means of estimating
```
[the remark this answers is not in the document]
hildeweerts: I also think we need to mention here how bootstrap estimates are supposed to be used and that bootstraps can't be used directly to make claims about statistical significance, particularly when you consider issues with multiple hypothesis testing (as discussed in #999)

docs/user_guide/assessment/error_estimation.rst
```diff
@@ -0,0 +1,360 @@
...
+bootstrapping capabilities.
+For pedagogical purposes, we are going to work on a rigged
+data set, where we know the 'true' answers in advance.
+Specifically, we are going to generate data according to a binomial
+distribution with :math:`p=q=0.5` and use the well known
+gaussian approximations :math:`\mu = n p` and :math:`\sigma^2 = n p q`.
+Our data set will have 1000 samples.
+We will have two subgroups within the data set, 'A' and 'B'.
+Subgroup 'A' will comprise 60% of the data, and 80% of its
+values will be correct (so :math:`p_A = 0.8`).
```
hildeweerts, line 94: "correct" or "positive"?

docs/user_guide/assessment/error_estimation.rst
hildeweerts: I agree with @adrinjalali we can just call it `random_state` (and mention in the API docs that it's only used for bootstrapping). This is also in line with the API / docs in scikit learn, e.g., [KFold](https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.KFold.html#sklearn.model_selection.KFold).

docs/user_guide/assessment/error_estimation.rst
```diff
@@ -0,0 +1,360 @@
...
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Now on to the bootstrap results themselves.
+We access the bootstrapped values for the overall values via
+:meth:`MetricFrame.overall_quantiles`.
+This method takes a single argument, :code:`quantiles`,
+which is a list of the quantiles we want extracted, each
+in the range :math:`[0, 1]`.
+We are going to look at the bootstrapped mean and standard
+deviation, so the quantiles we want are 0.159, 0.5 and 0.841
```
hildeweerts, line 250: I'm probably missing something obvious, but what's so special about 0.159?

docs/user_guide/assessment/error_estimation.rst
```diff
@@ -0,0 +1,360 @@
...
+    B             20.0  400.0
+    >>> mf.by_group_quantiles(quantiles=my_quantiles)
+                 bootstrap_sum                    count
+    SF0
+    A    [465.0, 480.0, 495.0]  [582.741, 600.5, 615.0]
+    B       [16.0, 19.5, 23.0]  [385.0, 399.5, 417.259]
+
+We can see that the nominal values from :attr:`MetricFrame.by_group`
+are as expected, given how we constructed our dataset.
+The :meth:`MetricFrame.by_group_quantiles` method has
```
hildeweerts, line 302: I was wondering whether it would make sense to use hierarchical indexing (rather than a dataframe with lists), but I suppose MultiIndex is asking for headaches.

hildeweerts (commented):
The API in init makes a lot of sense (with `random_state` and `n_bootstrap_samples`) and being able to input the quantiles in a list in later stages also makes sense to me.

It does feel a bit odd that we need to use a separate method for `by_group_quantiles()` and `overall_quantiles`, but not for `group_max()`. I understand why (turning `by_group`/`overall` into a method would not be backwards compatible), but it would be more consistent (even though without setting `quantiles` nothing is really computed).

`by_group_quantiles()` returning a dataframe of lists also feels a bit suboptimal, because it makes it a lot more difficult to do indexing / plotting / etc. Using hierarchical indexing is an option, though I admit that can result in some headaches as well.

docs/user_guide/assessment/error_estimation.rst
riedgar-ms: For a gaussian, the 1-\sigma quantile (similarly for 0.841)

riedgar-ms:
Thanks @LeJit  and @hildeweerts !

I very much get the comment about the asymmetric API. Do you think it would be better to introduce `group_max_quantiles()` etc. to match `by_group_quantiles()` etc.?

As for the return value, if I changed the 'DataFrame element is a list' behaviour to inserting a `MultiIndex` (well, adding an extra layer to the index, to be precise), does that make plotting into a one-liner? I'm open to suggestions, but I'd like to understand the advantages and disadvantages of the options a little better.

hildeweerts:
> I very much get the comment about the asymmetric API. Do you think it would be better to introduce `group_max_quantiles()` etc. to match `by_group_quantiles()` etc.?

Good question. I think that might make more sense indeed, though I'm curious to hear what others think @LeJit @fairlearn/fairlearn-maintainers 

> As for the return value, if I changed the 'DataFrame element is a list' behaviour to inserting a `MultiIndex` (well, adding an extra layer to the index, to be precise), does that make plotting into a one-liner? I'm open to suggestions, but I'd like to understand the advantages and disadvantages of the options a little better.

I don't think it would be a one liner, but it is generally not advised to put lists in a `DataFrame`. To retrieve the data you need to either iterate over all the rows, or use some variation of`apply()` to access the data (e.g., `df.metric.apply(lambda x: x[0]) )` to access the first item in the list), which is generally very slow - though that is probably a minor concern as `MetricFrame.by_group` will be rather small in most use case.  I'll try to put together some code examples (likely next week) to see which format would be easier to turn into a plot with error bars.

riedgar-ms:
@adrinjalali @LeJit did you have opinions on the handling of `group_max()` etc.? I can see arguments both ways. Since `.overall` and `by_group` are properties, I have to introduce `overall_quantiles()` and `by_group_quantiles()`. However, `group_max()` etc. are already methods, and it's easy to add a extra `None`-defaulting parameter. Then again, for consistency, `group_max_quantiles()` might be preferred, even if both it and `group_max()` just immediately passed the call on to a universal internal method.

And @hildeweerts 's argument in favour of an extra level of index rather than embedding a list for the quantiles also makes sense, unless anyone has contrary experience.

riedgar-ms:
BTW, I will address the documentation comments after the API is sorted.

hildeweerts:
> I don't think it would be a one liner, but it is generally not advised to put lists in a `DataFrame`. To retrieve the data you need to either iterate over all the rows, or use some variation of`apply()` to access the data (e.g., `df.metric.apply(lambda x: x[0]) )` to access the first item in the list), which is generally very slow - though that is probably a minor concern as `MetricFrame.by_group` will be rather small in most use case. I'll try to put together some code examples (likely next week) to see which format would be easier to turn into a plot with error bars.

So I haven't been able to come up with a straightforward code example quickly for either option, mostly because `pandas.plot.bar()` wants asymmetrical errors to be in an annoying format. _(If you want to make a plot where bars are grouped per metric, the means (or more accurately in case of the 0.5 quantile, medians) will be in MxN matrix, where M is metrics and N is groups. Then, `pandas.plot.bar()` wants `yerr` to be formatted as an Mx2xN array with the errors **relative to the median** rather than the absolute values. That's not really doable in a one or two liner in either of the formats...)_

One advantage of hierarchical indexing, though, is that (1) it is somewhat easier to interpret, (2) you can very easily get _some_ kind of plot without further processing. E.g., if you have the following dataframes:

``` python
import pandas as pd 

# create df with lists
df_l = pd.DataFrame({'fpr' : [[0.1, 0.22, 0.3], [0.05, 0.1, 0.17]],
              'tpr' : [[0.8, 0.9, 0.95], [0.85,0.92,0.97]]}, index=['group A', 'group B'])

# df with multi index, metrics first
cols = pd.MultiIndex.from_tuples([("fpr", "q1"), 
                                  ("fpr", "q2"),
                                  ("fpr", "q3"),
                                  ("tpr", "q1"),
                                  ("tpr", "q2"),
                                  ("tpr", "q3"),
                                 ])
data=[[0.1, 0.22, 0.3,  0.8, 0.9, 0.95], 
      [0.05, 0.1, 0.17, 0.85,0.92,0.97]]
df_mm = pd.DataFrame(data, columns=cols,index=['group A', 'group B'])
```

image

```python
df_mm.plot.bar()
```

image

riedgar-ms:
Hmmm.... are you suggesting that the quantile index should go on the columns, not the rows? I can see some sense in that, although it will mean repeating the quantile values each time.

hildeweerts:
> Hmmm.... are you suggesting that the quantile index should go on the columns, not the rows? I can see some sense in that, although it will mean repeating the quantile values each time.

I'm not sure what you mean - I tried to keep the dataframe as consistent with `MetricFrame` and your example. Putting the quantile index in the rows seems a bit odd when you consider multiple sensitive features or control features. We could also swap the metrics and quantiles, in which case you'd repeat the metrics. In any case, I don't think there's really a way around repeating some variable if we are going for a hierarchical index, right?

riedgar-ms:
*thinks* Yes of course (unless we can persuade Pandas to add `DataCube` as a type.....).

docs/user_guide/assessment/error_estimation.rst
```diff
@@ -0,0 +1,360 @@
...
+We are then generally interested in the difference or ratio
+of function evaluations on these groups, and errors always
+accumulate, even when the target values (the difference or ratio
+in this case) are getting smaller.
+:ref:`Intersecting groups <intersecting_groups>` make the
+problem worse, since some intersections can have very low
+sample counts, or even be empty.
+
+In Fairlearn, we offer bootstrapping as a means of estimating
+errors.
```
MiroDudik, line 25: The language needs to be more precise here. Many of the metrics reported in metric frame are estimates of errors (like FPR, FNR, classification error, mean square error, etc.)--the bootstrap is used to estimate confidence intervals of these metrics.

docs/user_guide/assessment/error_estimation.rst
MiroDudik: I'd be in favor of `n_boot` (following seaborn) and `random_state` (following sklearn and pandas conventions)

fairlearn/metrics/_metric_frame.py
MiroDudik: Let's discuss on #999 (or suggest an alternative)--otherwise we will have an exploding number of comments.

MiroDudik:
Let's move the discussion of the API for `overall`, `by_group`, `group_max()` (etc.) to #999.

riedgar-ms:
Pulling the discussion from #999  back here. My understanding from the meeting is that we agreed:

- To avoid mention of `quantiles` in the API, but instead have the API refer to confidence intervals. This is to avoid confusion as to which distribution (since the bootstrapped distribution is *not* the true distribution) we perform calculations
- To have the results be list-of-DataFrames (*mutatis mutandis*) rather than DataFrame-of-lists or DataFrame-with-extra-level-of-index
- To calculate everything in the constructor, rather than generate quantiles on demand. This side-steps the property-vs-method conundrum

Does that match other people's recollections.... @hildeweerts ?

MiroDudik:
@riedgar-ms -- this sounds good to me, and it doesn't preclude more generality in future.

riedgar-ms:
Since I have a bit of time to write, some more API thoughts....

@adrinjalali may not be overjoyed, but I think we need a new type to specify how bootstrapping is going to work. The following is an abstract base class:

```python
class BootstrapStatisticifier
  @property
  def num_bootstrap_samples(self) -> int:
    """Exactly what it says."""
    pass

  @property
  def bootstrap_labels(self) -> List[str]:
    """
    The labels to be applied to the return lists of values.
    Things like ' Upper 95% confidence' and 'Lower 95% confidence' or
    whatever is appropriate.
    """
    pass

  def bootstrap_statisticify(self, samples: List[Any]) -> List[Any]:
    """
    This is the method which takes a set of bootstrap values
    of length `num_bootstrap_samples` (most likely floats)
    and does the resampling/quantiling/whatever
    required to produce the final results.
    The returned list should be the same length as `bootstrap_labels`
    """
    pass
```
The `MetricFrame` API then becomes:
```python
class MetricFrame
   def __init__(
    self,
    # etc.
    bootstrap: Optional[BootstrapStatisticifier] = None ):
     # etc.

  @property
  def bootstrap_labels -> List[str]:
     # Access the labels from the BootstrapStatisticifier
      pass
```
If `None` is passed, then `MetricFrame` acts normally. Otherwise, it goes into bootstrap mode. The user would have to construct a concrete `BootstrapStatisticifier` object beforehand, and I would provide a simple one, to provide the median and 95% confidence intervals. An alternative one could do the quantile tweaking which @harsha-nori had in his original PR.

Technically, `num_bootstrap_samples` could be a separate argument, but I think it's cleaner to put everything bootstrap related into a single object. I think it's better to have the object, rather than start adding endless arguments to `MetricFrame` as we concoct alternative bootstrapping samplings.

Thoughts @hildeweerts (or @MiroDudik )?

adrinjalali:
@riedgar-ms we might be going around circles, but if that's what you think we need, WDYT of having a separate class alltogether for `BootstrappedMetricFrame`? Seems like it would solve our API issues.

riedgar-ms:
I wouldn't object to a `BootstrappedMetricFrame` on the grounds that otherwise we're adding yet another return type to a class which has enough already. However, that's not the issue I was getting at with the `BoostrapStatisticifier` idea. The really key part os the `bootstrap_statisticify()` method which converts a collection of bootstrap samples into the specified confidence intervals. That's the part I'm concerned about acquiring an unknown collection of possible configuration arguuments.

adrinjalali:
One of the arguments you have is that by adding `BootstrapStatisticifier`, we avoid adding too many constructor args to `MetricFrame`, and I agree with that. But if we don't do the bootstrap in `MetricFrame`, then we don't necessary have the issue, as all those args go to the new `BootstrappedMetricFrame` class.

I'm thinking from the user's perspective, creating two objects to create a metricframe, seems more cumbersome than creating a single object with all required args.

riedgar-ms:
I think we may be talking at cross purposes - or at least, I'm not quite sure what you're proposing. Let me try to explain my concern again.

There's more than one way of converting a collection of bootstrap samples into some form of error estimate, and each of these is going to need its own set of configuration parameters. Even in the most basic case of assuming that the bootstrap distribution matches the 'true' distribution and calling `np.quantiles()` directly, you need to specify _which_ quantiles. So the `BootstrapStatisticifier` class serves as the common interface for all of these, and each implementation can define whatever parameters it needs.

Even if we have a separate `BootstrappedMetricFrame` class, it faces the same problem: there are many possible ways one can turn the bootstrap samples into the error estimates (although having a `BootstrappedMetricFrame` class would at least restore some order to the proliferation of return types). It either ends up needing to accept `**kwargs` or use a `BootstrapStatisticifier`, and have those `**kwargs` passed in as (properly named) constructor arguments.

I suppose one alternative would be to say that, rather than a class, users have to supply a `bootstrap_statisticify()` function.... and if they need extra arguments, then `functools.partial()` exists. However, I'm not convinced that that's any easier than giving them a base class to implement (plus a few starter implementations, of course).

Am I making sense?

adrinjalali:
@riedgar-ms I think we can start with a default single way of bootstraping, and then either add alternate ways via a string argument, or an argument which can accept a callable, but for the initial version I think we can skip that, don't you think?

riedgar-ms:
We could certainly defer this decision for now. We'd have a single `bootstrap_sample_count` argument to the `MetricFrame` constructor, and if present, we'd hand back the 95% confidence intervals (or perhaps median, 67% and 95% confidence intervals - we discuss that separately).

For the future, are you envisioning that Callable taking a list of bootstrap samples, and returning _both_ the designated confidence intervals _and_ the appropriate labels for those intervals in a tuple?

Of course, I may be overestimating the number of different ways bootstrap statistics can be collated; I know it as a technique, but it's not one I've used myself. If you (or @hildeweerts , @MiroDudik  or anyone else) have more experience I'm very happy to be guided.

MiroDudik:
I would be fine to start with one method (percentile bootstrap). The API could be something like this:

* `MetricFrame(..., ci_quantiles=None, n_boot=None)`
  * user can specify something like `ci_quantiles=[0.03, 0.97]` (or some other list of size two) and then `n_boot` is set to a default value (e.g., 1000)
  * user can specify something like `n_boot=100` and then `ci_quantiles` is set to a default value (e.g., [0.025, 0.975])
  * or user can specify both `ci_quantiles` and `n_boot`

Instead of `ci_quantiles`,  we could have something similar to [seaborn's convention](http://seaborn.pydata.org/tutorial/error_bars.html) and just use symmetric confidence intervals specified via `ci`, for example `ci=95` would mean quantiles [0.025, 0.975].

Also, we could have an additional argument, something like `save_bootstrap_replicates` to save bootstrap replicates internally. Then the user could access and process those replicates however they see fit...

MiroDudik:
> I thought we were going to try to avoid 'quantiles' in the API, since it got a bit ambiguous as to which distribution the quantiles applied?
>
I didn't think `ci_quantiles` was ambiguous--is it? I'd be also okay using seaborn's convention, e.g., 'ci=95' for 95% confidence interval. It only yields symmetric confidence intervals, but that's the main use case anyway.

> And I don't see how that avoids my main concern of variations in the quantile extraction method winding up spinning off endless extra arguments. That was my original reason for proposing a `BootstrapStatisticifier` concept.
>
My proposal has two extra arguments in the constructor. I personally don't find that too excessive (but it's definitely matter of taste). From my side, I don't mind having a lot of optional arguments--that's how matplotlib and seaborn work; not ideal, but something that data scientists are already used to. My guess is that a typical user will copy-paste an example and then adjust to taste.

MiroDudik:
I'm not suggesting to use `**kwargs`. The current constructor API:
```python
    def __init__(
        self,
        *,
        metrics,
        y_true,
        y_pred,
        sensitive_features,
        control_features=None,
        sample_params=None
    )
```
My proposed API is this:
```python
    def __init__(
        self,
        *,
        metrics,
        y_true,
        y_pred,
        sensitive_features,
        control_features=None,
        sample_params=None,
        n_boot=None,
        ci_quantiles=None  # or ci=None
    )
```
Does this seem confusing? Standard python APIs for plotting / data analysis tend to follow this pattern, e.g.:
https://seaborn.pydata.org/generated/seaborn.pointplot.html

riedgar-ms:
I can live with that API, although it is deferring the question of handling different 'quantile extraction' methods to later (as I said, I can also live with that). In general, I think we should avoid endlessly growing argument lists, even if we avoid `**kwargs`.  They make documentation pages grow incomprehensibly long, and are also a nightmare to validate.

First stage is going to be changing `MetricFrame` to calculate _everything_ in the constructor. I'd like to do that separately (once I have time again), since it's a big change which will also affect the non-bootstrapped version.

adrinjalali:
It seemed that @riedgar-ms 's suggestion was going to be an easier decision and a smoother path down the line though. @MiroDudik 's suggestion brings up quite a few hard decisions and I think it would make it harder to get this on.

hildeweerts:
To stir things up a bit, I actually quite like @MiroDudik's suggestion. It is a simple solution that doesn't require the user to construct another object first, which often leads to confusion (at least with my students, lol). Adding 2 additional arguments to be able to get uncertainty estimates does not seem excessive to me. 

I do not think many people would expect more complex types of bootstrapping to be included in Fairlearn. I suspect that even if we get this in, it would take quite some time before we would expand it further to other types of bootstrapping - especially because writing a decent user guide on when to use which type of bootstrapping approach seems like a non-trivial task to me without turning into an applied statistics manual.

riedgar-ms:
Superseded

PR #1126, opened 2022-09-22
[Doc] Reorganisation of Common Fairness Metrics

riedgar-ms:
## Description

Reorganise the Common Fairness Metrics section, and move the discussion of the Four-Fifths rule here.

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

### DIFF
--- a/docs/user_guide/assessment/common_fairness_metrics.rst
+++ b/docs/user_guide/assessment/common_fairness_metrics.rst
@@ -9,7 +9,7 @@ In the sections below, we review the most common fairness metrics, as well
 as their underlying assumptions and suggestions for use. Each metric requires
 that some aspects of the predictor behavior be comparable across groups.
 Note that *common* usage does not imply *correct* usage; we discuss
-one very common misue in the
+one very common misuse in the
 :ref:`section on the Four-Fifths Rule <assessment_four_fifths>`.
 In code examples, we will use the following input arrays:
 
--- a/docs/user_guide/assessment/common_fairness_metrics.rst
+++ b/docs/user_guide/assessment/common_fairness_metrics.rst
@@ -109,7 +109,11 @@ phenomena are valid. For demographic parity, one relevant question would be
 whether demographic parity meets the criteria for establishing "fairness", 
 itself an unobservable theoretical construct. Further, it's important 
 to ask whether satisfying demographic parity actually brings us closer 
-to the world we'd like to see. 
+to the world we'd like to see.
+
+Finally, note that demographic parity underlies the 'Four-Fifths' rule.
+This is a common metric used to assess datasets but
+:ref:`as we discuss below <assessment_four_fifths>`, it is often misapplied.
 
 
 .. _conditional_group_fairness:
@@ -299,7 +303,79 @@ capture the costs of missclassification disparities.
 
 
 
+.. _assessment_four_fifths:
+
+The Four Fifths Rule: Often Misapplied
+--------------------------------------
+
+In the literature around fairness in machine learning, one will often find
+the so-called "four fifths rule" or "80% rule" used to assess whether a model
+(or mitigation technique) has produced a 'fair' result.
+Typically, the rule is implemented by using the demographic parity ratio introduced
+in the :ref:`assessment_demographic_parity` section above, with a result considered
+'fair' if the ratio exceeds 80% for all identified subgroups.
+*Application of this threshold is wrong in many scenarios.*
+
+As we note in many other places in the Fairlearn documentation, 'fairness'
+must be assessed by examining the entire sociotechnical context of a machine
+learning system.
+In particular, it is important to start from the harms which can occur to real
+people, and work inwards towards the model.
+The demographic parity ratio is simply a metric by which a particular model
+may be measured (on a particular dataset).
+Given the origin of the 'four-fifths rule' (which we will discuss next), its
+application may also give an unjustified feeling of legal invulnerability by
+conflating fairness with legality.
+In reality, 'fairness' is not always identical to 'legally allowable,' and
+the former may not even be a strict subset of the latter. [#f1]_
+
+The 'four fifths rule' has its origins in a specific area of US
+federal employment law.
+It is a limit for
+`prima facie evidence <https://en.wikipedia.org/wiki/Prima_facie>`_
+that illegal discrimination has occurred relative to a 
+relevant control population.
+A violation of the rule is still not sufficient to demonstrate that
+illegal discrimination has occurred - a causal link between the
+statistic and alleged discrimination must still be shown, and the
+court would examine the particulars of each case.
+For an example of the subtleties involved, see
+`Ricci v. Stefano <https://en.wikipedia.org/wiki/Ricci_v._DeStefano>`_
+which resulted from an attempt to 'correct' for disparate impact.
+*Outside* its particular context in US federal employment law,
+the 'four fifths rule' has no validity.
+
+Taken together, we see that applying the 'four fifths rule' will
+not be appropriate in most cases.
+Even in cases where it is applicable, the rule does not automatically
+avoid legal jeopardy, much less ensure that results are fair.
+The use of the 'four fifths rule' in this manner is an indefensible
+example of epistemic trespassing. [#f2]_
+It is for this reason that we try to avoid the use of legal
+terminology in our documentation.
+
+For a much deeper discussion of the issues involved, we suggest
+:footcite:ct:`watkins2022fourfifths`.
+A higher level look at how legal concepts of fairness can collide
+with mathematical measures of disparity, see
+:footcite:ct:`Xiang2019legalcompatibility`.
+
+
+
+
 References
 ----------
 
 .. footbibliography::
+
+
+.. rubric:: Footnotes
+
+.. [#f1] For a related example, see the discussion on 'law' and 'justice' in
+         *The Caves of Steel* (Asimov, 1953)
+
+.. [#f2] Epistemic trespassing is the process of taking expertise in one field and
+         applying it to another. This is not an intrinsically bad thing - one could
+         label all interdisciplinary research a form of epistemic trespassing.
+         However, doing so successfully requires a willingness to learn the subtleties
+         of the new field.
--- a/docs/user_guide/assessment/common_fairness_metrics.rst
+++ b/docs/user_guide/assessment/common_fairness_metrics.rst
@@ -8,11 +8,13 @@ Common fairness metrics
 In the sections below, we review the most common fairness metrics, as well
 as their underlying assumptions and suggestions for use. Each metric requires
 that some aspects of the predictor behavior be comparable across groups.
-Note that *common* usage does not imply *correct* usage; we discuss
-one very common misuse in the
-:ref:`section on the Four-Fifths Rule <assessment_four_fifths>`.
-In code examples, we will use the following input arrays:
 
+.. note::
+  Note that *common* usage does not imply *correct* usage; we discuss
+  one very common misuse in the
+  :ref:`section on the Four-Fifths Rule <assessment_four_fifths>`
+
+In the code examples presented below, we will use the following input arrays:
 
 .. doctest:: common_fairness_metrics_code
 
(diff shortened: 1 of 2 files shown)


docs/user_guide/assessment/common_fairness_metrics.rst
```diff
@@ -171,6 +173,8 @@ lowest selection rates :math:`a \in A` so a result of 0 indicates
 that demographic parity has been achieved.
 The second reports the ratio of the lowest and highest selection rates,
 so a result of 1 means there is demographic parity.
+This can potentially be used to implement the 'Four-Fifths' Rule,
+but :ref:`read our discussion below <assessment_four_fifths>`.
```
hildeweerts, line 177:
```suggestion
but :ref:`read our discussion below <assessment_four_fifths>` to understand whether this is an appropriate metric for your use case.
```

docs/user_guide/assessment/common_fairness_metrics.rst
```diff
@@ -171,6 +173,8 @@ lowest selection rates :math:`a \in A` so a result of 0 indicates
 that demographic parity has been achieved.
 The second reports the ratio of the lowest and highest selection rates,
 so a result of 1 means there is demographic parity.
+This can potentially be used to implement the 'Four-Fifths' Rule,
```
hildeweerts, line 176:
```suggestion
This metric can potentially be used to implement the 'Four-Fifths' Rule,
```

docs/user_guide/assessment/common_fairness_metrics.rst
```diff
@@ -8,6 +8,9 @@ Common fairness metrics
 In the sections below, we review the most common fairness metrics, as well
 as their underlying assumptions and suggestions for use. Each metric requires
 that some aspects of the predictor behavior be comparable across groups.
+Note that *common* usage does not imply *correct* usage; we discuss
```
hildeweerts, line 11: Important remark! Perhaps it should be even in bold or in a separate `note` environment thingie.

docs/user_guide/assessment/common_fairness_metrics.rst
```diff
@@ -299,7 +303,122 @@ capture the costs of missclassification disparities.
 
 
 
+.. _assessment_four_fifths:
+
+The Four Fifths Rule: Often Misapplied
```
hildeweerts, line 308: Following the discussion of the paper from the folks at Spotify, I can imagine that this would at some point become a subsection of a new section "Determine thresholds"/"How fair is fair enough"/whatever we want to call it under "Performing a fairness assessment". But for now I'm also happy to have this here.

docs/user_guide/assessment/common_fairness_metrics.rst
```diff
@@ -299,7 +303,122 @@ capture the costs of missclassification disparities.
...
+
+We have introduced three commonly used fairness metrics in this section,
+which can be summed up as follows:
+
+* Demographic Parity
+
+    * *What it compares:* Predictions between different groups
+      (true values are ignored)
+
+    * *Reason to use:* If the input data are known to contain historical
```
hildeweerts, line 375:
```suggestion
    * *Reason to use:* If the input data are known to contain
```

docs/user_guide/assessment/common_fairness_metrics.rst
```diff
@@ -299,7 +303,122 @@ capture the costs of missclassification disparities.
...
+      biases, demographic parity may be appropriate
```
hildeweerts, line 376:
```suggestion
      biases, demographic parity may be appropriate to measure fairness
```

docs/user_guide/assessment/common_fairness_metrics.rst
```diff
@@ -299,7 +303,122 @@ capture the costs of missclassification disparities.
...
+
+    * *Caveats:* By only using the predicted values, information is thrown
```
hildeweerts, line 378: Another important caveat is that selection rate is a very course way to quantify the distribution of predictions within a group, which makes it tricky to use as a constraint in an optimization task.

docs/user_guide/assessment/common_fairness_metrics.rst
```diff
@@ -299,7 +303,122 @@ capture the costs of missclassification disparities.
...
+      away
+
+* Equalized Odds
+
+    * *What it compares:* True and False Positive rates between different groups
+
+    * *Reason to use:* If historical data are useful, and true and false positives
```
hildeweerts, line 385:
```suggestion
    * *Reason to use:* If historical data does not contain measurement bias or historical bias we want to account for, and true and false positives
```

docs/user_guide/assessment/common_fairness_metrics.rst
```diff
@@ -299,7 +303,122 @@ capture the costs of missclassification disparities.
...
+      have roughly equal weight, equalized odds may be useful
+
+    * *Caveats:* If there are historical biases in the data, then the original labels
+      may hold little value. Also, large label imbalances (e.g. 95% of the data have 
+      a positive label) may make equalized odds an inappropriate measure
```
hildeweerts, line 390: The point about label imbalance is not really (which is usually referred to as class imbalance, I think) is not mentioned in the main text. In what sense would it be inappropriate? If false / true positives are of equal importance, it should not really matter what balance we're talking about. If the 5% of the data is too small to make accurate estimates, that's definitely an issue, but mostly because of the small sample size rather than the imbalance per se.

docs/user_guide/assessment/common_fairness_metrics.rst
```diff
@@ -299,7 +303,122 @@ capture the costs of missclassification disparities.
...
+
+    * *Caveats:* By only using the predicted values, information is thrown
+      away
+
+* Equalized Odds
+
+    * *What it compares:* True and False Positive rates between different groups
+
+    * *Reason to use:* If historical data are useful, and true and false positives
+      have roughly equal weight, equalized odds may be useful
```
hildeweerts, line 386:
```suggestion
      are considered to be (roughly) of the same importance, equalized odds may be useful
```

A nitpick just to highlight that this is not an innate characteristic of the problem, but something we need to decide based on the use case

docs/user_guide/assessment/common_fairness_metrics.rst
```diff
@@ -299,7 +303,122 @@ capture the costs of missclassification disparities.
...
+
+    * *Caveats:* If there are historical biases in the data, then the original labels
+      may hold little value. Also, large label imbalances (e.g. 95% of the data have 
+      a positive label) may make equalized odds an inappropriate measure
+
+* Equal opportunity
+
+    * *What it compares:* True Positive rates between different groups
+
+    * *Reason to use:* If historical data are useful, and false positive rates
```
hildeweerts, line 396: Nitpick: I think you'd want to use it not because FPR is *less* important but because true positives (and therefore TPR) correspond to something that is important for decision subjects, e.g., because it quantifies whether somebody received a benefit they are deserving of and we want to avoid false negatives (i.e., people "missing out" on a positive.

hildeweerts (commented):
I think you'll be happy to see that the number of comments is rather limited! :)

docs/user_guide/assessment/common_fairness_metrics.rst
riedgar-ms: Rephrased. LMK what you think.

docs/user_guide/assessment/common_fairness_metrics.rst
riedgar-ms: Adjusted to talk about harms

riedgar-ms (commented):
Thanks @hildeweerts . Taken a number of your suggestions, but also have some follow up questions.

docs/user_guide/assessment/common_fairness_metrics.rst
```diff
@@ -299,7 +305,128 @@ capture the costs of missclassification disparities.
...
+.. rubric:: Footnotes
+
+.. [#f1] For a related example, see the discussion on 'law' and 'justice' in
+         *The Caves of Steel* (Asimov, 1953)
+
+.. [#f2] Epistemic trespassing is the process of taking expertise in one field and
+         applying it to another. This is not an intrinsically bad thing - one could
+         label all interdisciplinary research a form of epistemic trespassing.
+         However, doing so successfully requires a willingness to learn the subtleties
+         of the new field.
```
alliesaizan, line 432: It's my understanding that epistemic trespassing involves some aspect of applying one's expertise in their own field to another *in which they do not have equivalent (or any) competence*, and that therefore the expertise or judgement may be incorrectly applied.  I just suggest making it a little clearer why this might be a problematic thing in the context of fairness assessment.

docs/user_guide/assessment/common_fairness_metrics.rst
```diff
@@ -299,7 +305,128 @@ capture the costs of missclassification disparities.
...
+
+The Four Fifths Rule: Often Misapplied
+--------------------------------------
+
+In the literature around fairness in machine learning, one will often find
+the so-called "four fifths rule" or "80% rule" used to assess whether a model
+(or mitigation technique) has produced a 'fair' result.
+Typically, the rule is implemented by using the demographic parity ratio introduced
+in the :ref:`assessment_demographic_parity` section above, with a result considered
+'fair' if the ratio exceeds 80% for all identified subgroups.
```
alliesaizan, line 318: Would it be helpful to include a brief example of calculating the four fifths rule here? Understand if we just briefly want to touch on this topic but don't want to go into too much detail!

docs/user_guide/assessment/common_fairness_metrics.rst
```diff
@@ -299,7 +305,128 @@ capture the costs of missclassification disparities.
...
+conflating fairness with legality.
+In reality, 'fairness' is not always identical to 'legally allowable,' and
+the former may not even be a strict subset of the latter. [#f1]_
+
+The 'four fifths rule' has its origins in a specific area of US
+federal employment law.
+It is a limit for
+`prima facie evidence <https://en.wikipedia.org/wiki/Prima_facie>`_
+that illegal discrimination has occurred relative to a 
+relevant control population.
```
alliesaizan, line 339:
```suggestion
relevant control population. The four-fifths rule is one of many test statistics that can be used to establish a prima facie case, but it is generally only used within the context of employment regulation. 
```

My intent with this suggestion is to further narrow the scope of when the four fifths rule is generally used/reasonable to use. We could also cite the [Uniform Guidelines on Employee Selection Procedures (1978)](https://www.ecfr.gov/current/title-29/subtitle-B/chapter-XIV/part-1607) so readers can get more info.

docs/user_guide/assessment/common_fairness_metrics.rst
```diff
@@ -299,7 +305,128 @@ capture the costs of missclassification disparities.
...
+which resulted from an attempt to 'correct' for disparate impact.
+*Outside* its particular context in US federal employment law,
+the 'four fifths rule' has no validity.
+
+Taken together, we see that applying the 'four fifths rule' will
+not be appropriate in most cases.
+Even in cases where it is applicable, the rule does not automatically
+avoid legal jeopardy, much less ensure that results are fair.
+The use of the 'four fifths rule' in this manner is an indefensible
+example of epistemic trespassing. [#f2]_
```
alliesaizan, line 355: Would it also make sense to reference the section on construct validity here?

docs/user_guide/assessment/common_fairness_metrics.rst
```diff
@@ -299,7 +305,128 @@ capture the costs of missclassification disparities.
...
+relevant control population.
+A violation of the rule is still not sufficient to demonstrate that
+illegal discrimination has occurred - a causal link between the
+statistic and alleged discrimination must still be shown, and the
+(US) court would examine the particulars of each case.
+For an example of the subtleties involved, see
+`Ricci v. Stefano <https://en.wikipedia.org/wiki/Ricci_v._DeStefano>`_
+which resulted from an attempt to 'correct' for disparate impact.
+*Outside* its particular context in US federal employment law,
+the 'four fifths rule' has no validity.
```
alliesaizan, line 348: Would it also make sense to reference the section on the portability trap in Fairness and ML here, in the sense that though the four fifths rule is valid in one context, it's invalid when used in this context?

docs/user_guide/assessment/common_fairness_metrics.rst
hildeweerts: Yeah, with the current state of the documentation I think the place in your PR makes sense.

hildeweerts (commented):
Just a response to your question. I agree with all comments @alliesaizan posted as well!

docs/user_guide/assessment/common_fairness_metrics.rst
riedgar-ms: Agreed.

docs/user_guide/assessment/common_fairness_metrics.rst
riedgar-ms: Accepted, with the addition of 'US Federal'

docs/user_guide/assessment/common_fairness_metrics.rst
riedgar-ms: I've provided a link to the function people could use, but I think it would be odd to give a code example of 'how to do this' in a section which is saying "don't do this."

docs/user_guide/assessment/common_fairness_metrics.rst
riedgar-ms: Tweaked.

riedgar-ms:
Thank you @alliesaizan  and @hildeweerts . I've updated the PR.

hildeweerts (approved):

PR #1130, opened 2022-09-29
[BUILD] Add code coverage to OtherML build

riedgar-ms:
## Description

Some new PRs have code which targets packages such as PyTorch. To make our code coverage useful, we therefore need to have the `otherml_packages` build upload its coverage as well.

## Tests

- [ ] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted
- [x] Existing test run tweaked

## Documentation

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -3,4 +3,5 @@ codecov:
   branch: main
   require_ci_to_pass: true
   notify:
+    after_n_builds: 3
     wait_for_ci: true
--- a/codecov.yml
+++ b/codecov.yml
@@ -3,5 +3,5 @@ codecov:
   branch: main
   require_ci_to_pass: true
   notify:
-    after_n_builds: 3
+    after_n_builds: 12
     wait_for_ci: true
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,1 +1,6 @@
 comment: false
+codecov:
+  branch: main
+  require_ci_to_pass: true
+  notify:
+    wait_for_ci: true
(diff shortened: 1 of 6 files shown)


adrinjalali:
Should we then setup codecov to wait for the correct number of reports before reporting anything?

riedgar-ms:
> Should we then setup codecov to wait for the correct number of reports before reporting anything?

I know that codecov tends to change its status during builds, but I have no idea how to configure that.

adrinjalali:
Example:

https://github.com/skops-dev/skops/blob/be34ea2d2606e5883c2edcb112b5292c29dc6a13/.codecov.yml#L5-L7

codecov.yml
```diff
@@ -1 +1,6 @@
 comment: false
+codecov:
+  branch: main
+  require_ci_to_pass: true
+  notify:
+    wait_for_ci: true
```
adrinjalali, line 6: we need to also tell it how many reports it should wait for.

riedgar-ms: Added.... although I'm not sure there's a good way of setting that number

adrinjalali: this should be the number of builds we send to codecov, and I think we send a lot more than 3. Don't we send on each node of our CI matrix? Isn't this more like 12 or so?

riedgar-ms: I've bumped to 12 (I'm assuming that each individual matrix entry counts as a build to CodeCov?) I'm a bit reluctant to try calculating the limit exactly, since this strikes me as a good way of getting weird behaviour in six months when we've all forgotten about this setting and we decide to drop a Python version.

adrinjalali (approved):
Ok, let's see how this works, I usually set the number as the number of uploads, but we can also test this as you got here.

PR #1135, opened 2022-10-17
[BUILD] Incorporate otherml workflow into PR-Gate

riedgar-ms:
## Description

Having the otherml workflow trigger on PRs in addition to the PR-Gate may be causing some confusing behaviour. Try folding it into the same workflow.

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation


.github/workflows/build-test.yml
```diff
@@ -1,6 +1,10 @@
 name: build-test
 
-on: [push, pull_request]
+on:
+  pull_request:
+  push:
+    branches:
+      - 'main'
```
adrinjalali, line 7: you don't want the tests to run on feature branches?

adrinjalali (approved):
otherwise LGTM

.github/workflows/build-test.yml
riedgar-ms: On a pull request, I was seeing things double up - once for the pull_request trigger and once for the push trigger. Or did you mean something else?

adrinjalali: But we do that everywhere, as in, once we run things on the PR, and then once the PR is merged.

adrinjalali: That's because you shouldn't be creating branches on the main repo 😁 if we put it here as is, then the CI wouldn't run on PRs merged into release branches (after merge)

.github/workflows/build-test.yml
```diff
@@ -1,6 +1,8 @@
 name: build-test
 
-on: [push, pull_request]
+on:
+  pull_request:
+  push:
```
adrinjalali, line 5:
```suggestion
on: [push, pull_request]
```

adrinjalali (approved):
Otherwise LGTM!

riedgar-ms:
I had been thinking of adding a schedule too, but for now.....

PR #1138, opened 2022-10-28
[Release] Update main for v0.8.0

riedgar-ms:
## Description

Updating bits & pieces to reflect that `v0.8.0` is (about to be) released, and `v0.9.0` is the new development version.

There will be a follow up PR to get the docs straightened out (which will be cherry picked to the `release/v0.8.X` branch)

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

### DIFF
--- a/.github/workflows/release-wheel.yml
+++ b/.github/workflows/release-wheel.yml
@@ -1,13 +1,9 @@
 name: Release Wheel
 
 # Following will be changed in due course
-on: # [push, pull_request]
+on:
   workflow_dispatch:
-    inputs:
-      use_test_pypi:
-        description: 'Release to Test-PyPI'
-        required: true
-        default: true
+
 
 concurrency: 
   # this ensures after each commit the old jobs are cancelled and the new ones
@@ -23,9 +19,9 @@ jobs:
   build-wheel:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - name: Set up Python
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v4
       with:
         python-version: "3.10"
     - name: Update pip and setuptools
@@ -48,9 +44,9 @@ jobs:
         os: [ubuntu-latest, windows-latest, macos-latest]
         python-version: ['3.8', '3.9', '3.10']
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     - name: Set up Python ${{ matrix.python-version }} on ${{ matrix.os }}
-      uses: actions/setup-python@v2
+      uses: actions/setup-python@v4
       with:
         python-version: ${{ matrix.python-version }}
     - name: Remove Fairlearn source directory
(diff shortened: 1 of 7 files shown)


riedgar-ms:
This is likely to have a few more bits & pieces added as I debug the release process. If I've missed anything which needs changing, please let me know too.

adrinjalali:
I think after the release you'd need to manually change the stable branch on the website repo, since we only push for one branch from CI now.

adrinjalali (approved):
LGTM.

riedgar-ms:
@adrinjalali thanks - but what do I need to do to create the v0.8.0 documentation in the documentation repo?

adrinjalali:
I see, there are a few places we need to fix:

1. In the circle config in these two places we have a regex checking for the branch name:

https://github.com/fairlearn/fairlearn/blob/bc122366afecf3adacde3ebbf75815c92c2c4081/.circleci/config.yml#L115

https://github.com/fairlearn/fairlearn/blob/bc122366afecf3adacde3ebbf75815c92c2c4081/.circleci/config.yml#L133

But it seems we're doing our releases under branches with names such as `release/v0.8` and `release/v0.6.0`. So these two regex need to be fixed to match what we have.

2. To get the folder name under which the docs need to be pushed, we have this line in the script:

https://github.com/fairlearn/fairlearn/blob/bc122366afecf3adacde3ebbf75815c92c2c4081/devops/circle/push_doc.sh#L31

This needs to be fixed so that it converts `release/v0.6.0` to `v0.6` (we probably want one doc version per major release). Right now our folders on the website repo are:

image

We can leave them as they are, but maybe do `v0.8` for the new release.

2.1. The same line above which strips the last two chars, wouldn't work on `v0.8` since there's no minor version there. If we're going to have the minor bug fix releases under the same branch, we should name the branch something like `release/v0.8.X`.


I'm not sure why all of this was overlooked by me last time 😁

riedgar-ms:
@adrinjalali before I go running off and making more changes, I'd like to make sure I understand what you're proposing. I do like your suggestion of having patch releases on a single branch, which is why I followed it for this release. However, I should have asked about the docs first :-/

First off, it sounds like I should create a `release/0.8.X` branch based on my existing `release/0.8` branch.

Then I can modify the regexes you highlighted. When it comes to `push_doc.sh` no change will actually be required; indeed changing the branch naming convention will make it more robust, since if I'm understanding that line correctly, it would break on a `.10` patch.

Do I make these changes just on `main`, or do they need to go into the release branch as well, to make sure the automation is triggered?

adrinjalali:
> First off, it sounds like I should create a `release/0.8.X` branch based on my existing `release/0.8` branch.

Or rename the existing branch.

> Then I can modify the regexes you highlighted. When it comes to `push_doc.sh` no change will actually be required; indeed changing the branch naming convention will make it more robust, since if I'm understanding that line correctly, it would break on a `.10` patch.

You're correct about the `.10` issue, but I don't think we'll realistically ever have that issue ;)

Regarding changing the regex, in the push script, it'd be nice to explicitly handle the `release/` part of the branch name.

> Do I make these changes just on `main`, or do they need to go into the release branch as well, to make sure the automation is triggered?

We should do them first in main, then cherry pick them in the release branch. They need to be triggered on the release branch since the action on `main` won't be pushing anything to a release branch doc folder.

adrinjalali:
Sure, that sounds good to me.

PR #1147, opened 2022-11-09
CI Try to improve handling of JS directory

riedgar-ms:
## Description

The script was not pushing docs if there were no changes to the JS directory, since in those cases, the `git` commands on the directory were returning errors back to the shell.

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

### DIFF
--- a/.circleci/push_doc.sh
+++ b/.circleci/push_doc.sh
@@ -73,10 +73,15 @@ git config user.name $USERNAME
 git config push.default matching
 git add -f $dir/
 git commit -m "$MSG" $dir
-if [ "$CIRCLE_BRANCH" = "main" ]
+
+# See if the JS dir changed
+modified_js = `git status -s -uno js | wc -l`
+if [ "$CIRCLE_BRANCH" = "main"  && $modified_js -ne 0 ]
 then
     git add -f js/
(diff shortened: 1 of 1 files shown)


.circleci/push_doc.sh
```diff
@@ -73,10 +73,15 @@ git config user.name $USERNAME
...
 git commit -m "$MSG" $dir
-if [ "$CIRCLE_BRANCH" = "main" ]
+
+# See if the JS dir changed
+modified_js = `git status -s -uno js | wc -l`
+if [ "$CIRCLE_BRANCH" = "main"  && $modified_js -ne 0 ]
 then
     git add -f js/
-    git commit -m "$MSG" js/
+    git commit -m "JS dir update: $MSG" js/
```
adrinjalali, line 82: could also add --allow-empty here ;) but this works as well.

adrinjalali (approved):

.circleci/push_doc.sh
riedgar-ms: Turns out is doesn't.... not quite sure why `bash` is saying that `modified_js` is not found. Expect another PR shortly.

PR #1151, opened 2022-11-11
[DOC] fix bad source links

### LINKED ISSUE #1150: [DOC] Bad 'source' links v0.8.0

The "source" links in the API reference point to 404 because the `v` before the version is missing. E.g.,:

https://github.com/fairlearn/fairlearn/blob/0.8.0/fairlearn/datasets/_fetch_acs_income.py#L13-L211

instead of

https://github.com/fairlearn/fairlearn/blob/v0.8.0/fairlearn/datasets/_fetch_acs_income.py#L13-L211

I think this happens because we [used to get the `tag_or_branch` from sphinx multi version](https://github.com/fairlearn/fairlearn/blob/release/v0.7.0/docs/conf.py#L226) but in the current release `tag_or_branch ` is derived [from the fairlearn package version](https://github.com/fairlearn/fairlearn/blob/37e55bd9cdef858e4b62c379638c718b1e8c9b2b/docs/conf.py#L46).

It seems an easy fix would be to change [this line](https://github.com/fairlearn/fairlearn/blob/main/docs/conf.py#L231) in conf.py to include `v` before `{tag_or_branch}` in main and v0.8.0 - but I'm not sure if I understand our current sphinx setup well enough to understand if that would break other things in our doc build :D

hildeweerts:
## Description

this should resolve #1150 

@adrinjalali: I wasn't sure why 0.9.0 in `versions.json` was called `"v0.9.0"` instead of `"0.9.0"` like the others so I assumed this was a typo - let me know if I'm missing something here :)

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

### DIFF
--- a/docs/_static/versions.json
+++ b/docs/_static/versions.json
@@ -26,7 +26,7 @@
     },
     {
         "name": "v0.9.0 (dev)",
-        "version": "v0.9.0",
+        "version": "0.9.0",
         "url": "https://fairlearn.org/main/"
     }
 ]
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -228,7 +228,10 @@ def linkcode_resolve(domain, info):
     fn = os.path.relpath(fn, start=os.path.dirname(fairlearn.__file__)).replace(
         os.sep, "/"
     )
-    return f"http://github.com/fairlearn/fairlearn/blob/{tag_or_branch}/fairlearn/{fn}{linespec}"
+    if tag_or_branch == 'main':
+        return f"http://github.com/fairlearn/fairlearn/blob/{tag_or_branch}/fairlearn/{fn}{linespec}"
+    else:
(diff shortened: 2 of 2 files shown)


adrinjalali:
since this affects only the current (and maybe 0.8 release), it should work I think.

WDYT @riedgar-ms

riedgar-ms (approved):
This seems reasonable, although we don't find out until it's deployed - and probably cherry picked to the `v0.8.X` branch :-/

riedgar-ms:
I think that we may need to cherry pick this onto the release branch @hildeweerts @adrinjalali

hildeweerts:
@adrinjalali could you perhaps assist with cherry picking to the release branch? I am a bit worried I will ✨break✨things 😁

PR #1152, opened 2022-11-15
DOC Upgrade pydata_sphinx_theme to v0.12.0

### LINKED ISSUE #1112: DOC Fairlearn logo does not link to correct home page
With *main* selected, clicking on the Fairlearn logo on the top left corner of the website directs you to `https://fairlearn.org/main/https%3A//fairlearn.org.html` rather than `https://fairlearn.org`.

### LINKED ISSUE #1134: [Doc Bug] Bad link from Fairlearn logo
#### Describe the bug

Going to the 'main' branch of the documentation, the Fairlearn logo has a bad link

#### Steps/Code to Reproduce

image

This image has a link target of:

https://fairlearn.org/main/https%3A//fairlearn.org.html

iofall:
Resolves #1134, Resolves #1112 

## Description
- Upgrade pydata_sphinx_theme to v0.12.0
- Remove search field from sidebar. The field is now linked with the search icon in the `navbar` which can cause problems with mobile views. See [link](https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/search.html#configure-the-search-field-position).

## Tests

- [x] no new tests required

## Documentation


hildeweerts:
Thanks @iofall! 

A quick question re. the search button. The link you posted says this: 

>  It is strongly recommended that you use either search button and the hidden/overlaid field that comes with it, or use a persistent search field in a place that makes sense for your layout.

I believe we currently only have the persistent search field and not a button - so I'm not sure why this change is required? Not that it matters much, I like either option - just checking whether I'm missing something :)

hildeweerts:
I just noticed in the artifacts that the pages that don't have a section navigation (Get Started / About Us / FAQ) look a bit weird now that the persistent search bar is removed, especially when you navigate from a page that has one to another. It is not super disturbing but something to take into account when we decide between persistent search bar and a button.

image
image


It also seems the updated theme automatically collapses "FAQ" and "About Us" under "More", as the maximum number of header links is 5 by default. We should check whether increasing it to 7 looks okay.

```
html_theme_options = {
  "header_links_before_dropdown": 7
}
```

adrinjalali (approved):
Other than @hildeweerts 's comment, LGTM.

iofall:
> I just noticed in the artifacts that the pages that don't have a section navigation (Get Started / About Us / FAQ) look a bit weird now that the persistent search bar is removed, especially when you navigate from a page that has one to another. It is not super disturbing but something to take into account when we decide between persistent search bar and a button.

Yes, I agree. I think having the persistent search field, like we have now will be more consistent across the entire website. So, let's go with that.

iofall:
> I believe we currently only have the persistent search field and not a button - so I'm not sure why this change is required? Not that it matters much, I like either option - just checking whether I'm missing something :)

You are right, the problem was that v11.0 did not have any way of removing the `search-button` through the theme, so the only option was to remove the persistent search instead. However, it has now been fixed in [v12.0](https://github.com/pydata/pydata-sphinx-theme/releases/tag/v0.12.0), [PR](https://github.com/pydata/pydata-sphinx-theme/pull/1022). So now we can remove the `search-button` and have the persistent version of the `search-field` instead of the button.

docs/conf.py
```diff
@@ -141,6 +141,8 @@
         "version_match": tag_or_branch,
     },
     "navbar_start": ["navbar-logo", "version-switcher"],
+    "navbar_persistent": [],
```
iofall, line 144: This line helps with removing the persistent search button. It was not available in v11.0 so upgraded the theme to v12.0. Now it should be consistent with what we had on the website before.

docs/_static/css/custom.css
```diff
@@ -52,3 +52,8 @@ div.sponsor-div-box {
    font-size: 0.9rem;
    text-align: center;
  }
+
+ /* Fix version switcher button margin bottom */
+ button.version-switcher__button{
+   margin-bottom: 0em;
+ }
```
iofall, line 59: The version switcher is misaligned on mobile displays due to `margin-bottom` being set as `1em`. This CSS block fixes it.

**Before**
image


**After**
image

hildeweerts:
Thanks @iofall!

It seems there is now a horizontal scrollbar in the navigation on the left (I'm using Safari) :( 
It seems to be a side effect of the persistent search bar, because it is also visible in the theme docs: https://pydata-sphinx-theme.readthedocs.io/en/stable/examples/persistent-search-field.html

I generally prefer the persistent search bar but the horizontal scrollbar looks a bit unprofessional, so perhaps we should use the search button after all (or at least until the issue is fixed in the theme). What do you think would be best @iofall?

Two minor things:
* I think we don't need the "Show source" button on the righthandside but I'm not too bothered by it
* All of the "socials" buttons are small except for the "light / dark" mode button. IMO we don't have to stall this PR to fix that, I will just open a new issue if we don't tackle it here.

Old:
image

New:
image

iofall:
Thank you for the feedback @hildeweerts, the latest pull should address all the changes needed.

docs/_static/css/custom.css
```diff
@@ -52,3 +52,13 @@ div.sponsor-div-box {
...
+
+ /* Fix scrollbars on primary sidebar */
+ .bd-sidebar-primary .sidebar-end-items{
+   margin-top: unset;
+ }
```
iofall, line 64: This custom CSS block will remove the scrollbars. I was able to test on Chrome, Edge and Firefox. I think it will work on Safari as well, but please check and let me know @hildeweerts.

docs/_static/css/custom.css
```diff
@@ -52,3 +52,18 @@ div.sponsor-div-box {
...
+
+ /* Fix scrollbars on primary sidebar */
+ .bd-sidebar-primary .sidebar-end-items {
+   margin-top: unset;
+ }
+
+ /* Fix Sphinx-Glr page bottom border */
+ p.sphx-glr-signature a.reference.external {
+   padding-bottom: 0px;
+ }
```
iofall, line 69: Fixes the inconsistent bottom border on the Sphinx Gallery Examples page.

**Before**
image

**After**
image

docs/_static/css/custom.css
hildeweerts: Unfortunately the horizontal scrollbar is still there in Safari, though the vertical one is now only there when it's actually necessary, so that's nice!

image

hildeweerts:
@iofall Are the CircleCI artifacts slow for you as well or is this just my internet connection?

@riedgar-ms should codecov is complaining, but should it check things like conf.py and CSS files at all?

iofall:
> @iofall Are the CircleCI artifacts slow for you as well or is this just my internet connection?

No, it's working fine for me

docs/_static/css/custom.css
iofall: I did try to look into this using BrowserStack. It seems the issue is only on the latest Safari build for macOS Ventura; I am not sure which version you are using. 

Unfortunately, the access for the free account lasts for only a minute so I was not able to check much. If you can inspect the padding using the Inspect Tool, that might help us figure out where we need to make a change.

hildeweerts: Thanks for checking - I'm indeed using Safari on Ventura. It's not *super* disturbing and sounds more like a Safari + pydata theme issue that might be resolved over time.

hildeweerts:
There is still a tiny issue with the horizontal scrollbar in Safari, but IMO the failing home button is a lot more pressing, so I think we should just go ahead and merge @fairlearn/fairlearn-maintainers

hildeweerts (approved):

adrinjalali (approved):

PR #1157, opened 2022-11-24
CI set matplotlib backend to agg

### LINKED ISSUE #1156: [CI] TclError
In #1152, windows build 3.9 [failed](https://github.com/fairlearn/fairlearn/actions/runs/3525473951/jobs/5927445026) on a `TclError`  in `test_plot_equalized_odds_ex3` saying "This probably means that Tcl wasn't installed properly.". 

It doesn't look like it should be related to that PR so I did some digging. It seems to be a more common problem and [this issue](https://github.com/microsoft/azure-pipelines-tasks/issues/16426) provides some pointers to fix it, including "setting the `matplotlib` backend prior" and "selecting a non-interactive backend".

Tagging @adrinjalali @riedgar-ms

hildeweerts:
Resolves #1156 

## Description
Sets matplotlib backend to Agg so our windows builds (hopefully) don't fail anymore.

@riedgar-ms @adrinjalali 

## Tests
- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -16,6 +16,8 @@ jobs:
         os: [ubuntu-latest, windows-latest, macos-latest]
         python: ["3.8", "3.9", "3.10"]
     runs-on: ${{ matrix.os }}
+    env:
+      MPLBACKEND: Agg
     steps:
       - uses: actions/checkout@v3
       - uses: actions/setup-python@v3


adrinjalali:
(updated the branch to get the CI running)

hildeweerts:
Hmm, so that didn't work. 

Some variations I found would be to use either `variables` (instead of `env`) and `agg` (instead of `Agg`), but it is not like I understand our Azure workflows well enough to know which one should work in advance :D

hildeweerts:
Ah, the previous tests failed because codecov was complaining. After re-running all checks are green except for the linting issue (same as #1158)

.github/workflows/build-test.yml
```diff
@@ -16,6 +16,8 @@ jobs:
         os: [ubuntu-latest, windows-latest, macos-latest]
         python: ["3.8", "3.9", "3.10"]
     runs-on: ${{ matrix.os }}
+    env:
```
adrinjalali, line 19: you should be able to set it on the workflow level like here https://docs.github.com/en/actions/learn-github-actions/environment-variables

and I think it needs to be `agg`: https://matplotlib.org/stable/users/explain/backends.html

adrinjalali (commented):
I think we need to fix the linting issue though. Does it need an update in our requirements or something? @riedgar-ms

riedgar-ms:
@hildeweerts Linting now looks OK

.github/workflows/build-test.yml
hildeweerts: I put it in this job because it's the only one that uses matplotlib right now, but I agree setting it on the workflow level would be more future proof. 

On the page you sent me it says in a note: "The names of builtin backends case-insensitive; e.g., 'QtAgg' and 'qtagg' are equivalent." Happy to change it into `agg` though if you like that better.

adrinjalali (approved):
A green CI makes me happy :)

PR #1163, opened 2022-12-20
BUG fix bugs in tests that surfaced with sklearn 1.2 and autopep8 2.0.1

MiroDudik:

## Description

Three sklearn 1.2 related issues are being fixed:

* A typo in our test scripts: mismatched feature names between fit and predict. This was previously FutureWarning in sklearn, but with sklearn 1.2 it is now an error.
* Default behavior of log loss in sklearn 1.2 changes, affecting the evaluation of log loss in derived metrics. The default only kicks in when prediction=0 and truth=1 or vice versa, so the tests are adjusted to avoid this case (and give the log loss value equal to the previous default behavior).
* Warnings issued by fetch_openml when `parser` keyword argument is not present.

Additionally:

* autopep8 version is being kept lower to keep our linting scripts working

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

### DIFF
--- a/examples/plot_plotting_metrics_with_error.py
+++ b/examples/plot_plotting_metrics_with_error.py
@@ -26,7 +26,7 @@ from fairlearn.experimental.enable_metric_frame_plotting import plot_metric_fram
 from fairlearn.metrics import MetricFrame
 
 # %%
-# We download the data set using :func:`fetch_adult` function in :mod:`fairlearn.datasets`.
+# We download the data set using the function :func:`fetch_adult` in :mod:`fairlearn.datasets`.
 # The original Adult data set can be found at https://archive.ics.uci.edu/ml/datasets/Adult
 # There are some caveats to using this dataset, but we will use it solely as an example
 # to demonstrate the functionality of plotting metrics with error bars.
--- a/examples/plot_plotting_metrics_with_error.py
+++ b/examples/plot_plotting_metrics_with_error.py
@@ -9,7 +9,7 @@ Plotting Metrics with Errors
 import numpy as np
 from sklearn.compose import ColumnTransformer
 from sklearn.compose import make_column_selector as selector
-from sklearn.datasets import fetch_openml
+from fairlearn.datasets import fetch_adult
 from sklearn.impute import SimpleImputer
 from sklearn.metrics import accuracy_score, confusion_matrix, recall_score
 from sklearn.model_selection import train_test_split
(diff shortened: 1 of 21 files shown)


test/unit/metrics/test_generated_metrics.py
```diff
@@ -7,8 +7,12 @@
 
 # ======================================================
 
+apx0 = 1e-15
+apx1 = 1 - apx0
```
riedgar-ms, line 11: Should this be handled through `pytest.approx()` ?

MiroDudik: This is not used in assertions, this is used in the construction of an alternate y_pred. The goal is to avoid taking sklearn.log_loss([0], [1]) (which would mathematically result in +infinity, but sklearn has a default behavior that just changed between two versions).

riedgar-ms: Could you add a comment about that?

romanlutz (approved):

test/unit/metrics/test_generated_metrics.py
MiroDudik: done!

examples/plot_plotting_metrics_with_error.py
```diff
@@ -26,14 +26,14 @@
 from fairlearn.metrics import MetricFrame
 
 # %%
-# We download the data set using :meth:`fetch_openml` function in :class:`sklearn.datasets`.
+# We download the data set using :func:`fetch_adult` function in :mod:`fairlearn.datasets`.
```
romanlutz, line 29: using *the*

romanlutz (approved):

examples/plot_plotting_metrics_with_error.py
MiroDudik: done.

MiroDudik:
The current version of fairlearn.datasets actually requires sklearn >= 1.2. Are we comfortable with that requirement?

For example, errors in other-ml job are due to the fact that their conda environments use sklearn<1.2.

I'm happy to tweak fairlearn.datasets so they work with both sklearn>=1.2 as well as with the previous versions of sklearn. Do we want that?

fairlearn/datasets/_fetch_acs_income.py
```diff
@@ -181,6 +181,7 @@ def fetch_acs_income(
         cache=cache,
         as_frame=True,
         return_X_y=False,
+        parser="liac-arff",
```
adrinjalali, line 184: `liac-arff` is much slower than pandas though, do we really need it?

PR #1166, opened 2023-01-05
add sklearn-version-dependent parser kwarg

MiroDudik:

## Description

In sklearn 1.2, the function fetch_openml added a new keyword argument 'parser', and also changed the default behavior. To
maintain compatibility, we therefore set the 'parser' argument to 'liac-arff', which matches the behavior before version 1.2.

I haven't added any tests, but the compatibility behavior is partly tested by the 'other-ml' jobs which require older sklearn versions.

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

### DIFF
--- a/fairlearn/datasets/_constants.py
+++ b/fairlearn/datasets/_constants.py
@@ -2,4 +2,17 @@
 # Licensed under the MIT License.
 
 
+from sklearn.datasets import fetch_openml
+import inspect
+
 _DOWNLOAD_DIRECTORY_NAME = ".fairlearn-data"
+
+# In sklearn 1.2, the function fetch_openml added a new keyword
+# argument 'parser', and also changed the default behavior. To
+# maintain compatibility, we therefore set the 'parser' argument
+# to 'liac-arff', which matches the behavior before version 1.2.
+
+if 'parser' in inspect.getfullargspec(fetch_openml).kwonlyargs:
+    _PARSER_KWARG_COMPATIBILITY = {'parser': 'liac-arff'}
+else:
+    _PARSER_KWARG_COMPATIBILITY = {}
--- a/fairlearn/datasets/_fetch_acs_income.py
+++ b/fairlearn/datasets/_fetch_acs_income.py
@@ -7,7 +7,7 @@ import numpy as np
 import pandas as pd
 from sklearn.datasets import fetch_openml
 
-from ._constants import _DOWNLOAD_DIRECTORY_NAME
+from ._constants import _DOWNLOAD_DIRECTORY_NAME, _PARSER_KWARG_COMPATIBILITY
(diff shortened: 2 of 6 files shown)


adrinjalali:
Have you checked if things would work with pandas parser, which is much faster?

MiroDudik:
I did--that's the new default behavior, but it was failing our tests, because the pandas parser seems to deal differently with whitespaces in CSV files. This is what happens with adult:

```pycon
>>> import sklearn.datasets as skd
>>> d = skd.fetch_openml(data_id=1590, as_frame=True, parser='pandas')
>>> d.target
0         <=50K
1         <=50K
2          >50K
3          >50K
4         <=50K
          ...  
48837     <=50K
48838      >50K
48839     <=50K
48840     <=50K
48841      >50K
Name: class, Length: 48842, dtype: category
Categories (2, object): [' <=50K', ' >50K']
```
In the last line you see that there's an extra space in how the categorical variables are parsed. The old parser trims the leading space.

adrinjalali:
cc @glemaitre. Do you think this is a bug, or should we be dealing with the whitespace ourselves?

glemaitre:
You can check the original ARFF file [here](https://www.openml.org/search?type=data&status=active&id=1590&sort=runs). This is not a bug in the code but rather an inconsistency in the ARFF file where extra whitespace is added after each comma.

I would not add a new rule to the parser but rather correct the dataset because the issue is there.

romanlutz (approved):

MiroDudik:
If there are no plans to tweak the parser (or at least expose something like `skipinitialspace` in [pandas.read_csv](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html)?) then we can maybe merge the current version (with the slow parser) and create an issue to upload a new version of adult + change to the fast parser?

glemaitre:
That's a good point. I did not recall the ARFF spec. If this is actually different depending on the parser then we can consider that as a regression in scikit-learn. However, I don't know if we can easily correct it since pandas will probably not let us know if the space was within quotes or not. But this is worth to open an issue.

glemaitre:
I assume that we could accept a keyword argument to propagate to the `read_csv` to resolve those inconsistencies (hoping that there is no dataset that has a mixed of leading space with and without quotes).

glemaitre:
I will open an issue on scikit-learn.

glemaitre:
Looking more in-depth, the `skipinitialspace` does the job without the ambiguity of spaces being within quotes. So I assume that we just need to turn the option on.

glemaitre:
I open https://github.com/scikit-learn/scikit-learn/pull/25312

PR #1177, opened 2023-01-13
Update quickstart.rst

shimst3r:
## Description

In order to make the quickstart guide more accessible, I streamlined a few things:

- Add explicit installation of `matplotlib`.
- Rename `gm` to ~`as` for `accuracy_score` to mirror `sr` for `selection-rate`~ `mf` to be consistent with the remaining documentation.
- Remove redundant calls to `print()`.
- Order imports following `isort` standards.

NOTE: Because `auto_examples/plot_quickstart.py` is included instead of spelled out, the imports are implicit and missing from the final rendered page. Should this be challenged? :)

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation


docs/quickstart.rst
```diff
@@ -65,6 +65,15 @@ code snippet examples of how to use basic Fairlearn functionality for those
...
 is about binary classification, but we similarly support regression.
 
+Prerequisites
+^^^^^^^^^^^^^
+
+In order to follow the quickstart guide, you need to install additional dependencies:
+
+.. code-block:: bash
+
+    pip install matplotlib
```
romanlutz, line 75: My initial reaction was "But wait! You can just install fairlearn with `pip install fairlearn[customplots]`" but 

- other than me almost nobody would think of that (since I added that extra at some point)
- the name `customplots` isn't great and should probably change (just `plots`?) or we should remove that extra altogether
- Maybe we should also mention `pip install fairlearn`?

So I guess now I wonder if just the installation for fairlearn & matplotlib is best?

docs/quickstart.rst
```diff
@@ -107,15 +116,15 @@ we can evaluate metrics for subgroups within the data as below:
...
     >>> from sklearn.tree import DecisionTreeClassifier
-    >>> 
+    >>>
     >>> classifier = DecisionTreeClassifier(min_samples_leaf=10, max_depth=4)
     >>> classifier.fit(X, y_true)
     DecisionTreeClassifier(...)
     >>> y_pred = classifier.predict(X)
-    >>> gm = MetricFrame(metrics=accuracy_score, y_true=y_true, y_pred=y_pred, sensitive_features=sex)
-    >>> print(gm.overall)
+    >>> as = MetricFrame(metrics=accuracy_score, y_true=y_true, y_pred=y_pred, sensitive_features=sex)
```
romanlutz, line 124: I think `gm` is a remnant from what we used to call group metric(s) (?)

`as` is a word in itself... Maybe `mf` like elsewhere in the docs (?)

romanlutz (commented):
Thanks for taking the time to fix these! We just mentioned the quickstart the other day in the community call as it's still using the adult/census dataset which should change at some point. 🙂

docs/quickstart.rst
shimst3r: > other than me almost nobody would think of that

😄 

> Maybe we should also mention `pip install fairlearn`?

To avoid any confusion I would add it as well, yes. I didn't because it's already mentioned earlier on the same page, but not logically grouped within the guide itself. Will add it and rephrase the paragraph above. 👍

docs/quickstart.rst
shimst3r: > `as` is a word in itself.

I was pondering that problem too. 🤓 If the remainder of the docs use `mf`, I would suggest following suite in this case as well. `sr` as a special case for a second metric still makes sense that way.

shimst3r:
> [...] as it's still using the adult/census dataset which should change at some point. 🙂

Any ideas what to use instead?

docs/quickstart.rst
romanlutz: Amazing! I appreciate it, especially since a fresh set of eyes sometimes has a much clearer perspective on this 🙂

docs/quickstart.rst
romanlutz: Ah, there's another one... Well, I don't think I'm going to advocate for `asmf` and `srmf` 😆 I guess it's fine as is, it's really a nitpicky request anyway... let me resolve this.

romanlutz:
> > [...] as it's still using the adult/census dataset which should change at some point. 🙂
> 
> Any ideas what to use instead?

Yes! @MiroDudik had two ideas:
- the diabetes dataset from https://github.com/fairlearn/talks/tree/main/2021_scipy_tutorial (I think it's now part of fairlearn.datasets, but for the tutorial they added something synthetic) - it needs a lot of context to be explained which makes it less suitable for a quickstart where we can't spend much time (I think the page says 10 minutes), BUT we could just point to the tutorial for a holistic view.
- generate something entirely synthetic with a fairly obvious context. @MiroDudik's idea was this: rolling out a new version of a website happens slowly to avoid having everyone experience downtime if something goes wrong. So a model can select who sees the new version (1) or the old version (0) and we have some characteristics like perhaps location (country?) or browser or such and want to achieve equal risk distribution across groups (risk of having a bad QoS). Thinking about this a bit more, I'm wondering what features the original model without mitigation would use to make its prediction. Hmm, perhaps @MiroDudik can elaborate.

So if the latter is a bit unclear or requires a lot of work to set up first we can always go with the former.

shimst3r:
> rolling out a new version of a website happens slowly to avoid having everyone experience downtime if something goes wrong.

So more a canary rollout than an A/B testing experiment? While I agree that sounds exciting, it's also a bit out of the comfort zone for a lot of people. 😄 Did you create personas for the Quickstart tutorial by any chance and check whether this dataset would align with their interests?

romanlutz:
> > rolling out a new version of a website happens slowly to avoid having everyone experience downtime if something goes wrong.
> 
> 
> 
> So more a canary rollout than an A/B testing experiment? While I agree that sounds exciting, it's also a bit out of the comfort zone for a lot of people. 😄 Did you create personas for the Quickstart tutorial by any chance and check whether this ~dataset~ model would align with their interests?

I guess it could work for both. I'll let @MiroDudik respond since I don't know if he thought about this more. I certainly hadn't, just heard about it yesterday for the first time. 😄

shimst3r:
> I guess it could work for both. I'll let @MiroDudik respond since I don't know if he thought about this more.

Alright, let's wait for @MiroDudik. 👍

My gut feeling would be to use an existing dataset to highlight that `fairlearn` is shipping with usable datasets. I can also empathise with having a "more synthetic" dataset like canary rollouts though.

romanlutz (approved):
Regardless, I think the dataset change is a bit bigger than what this PR aimed at. Let's instead try to merge this and move the dataset discussion to an issue to get clarity before someone tackles it. I'll see if there is an existing issue and otherwise create a new one.

shimst3r:
@romanlutz I assume the broken CI pipeline has nothing to do with my changes? 😅

romanlutz:
> @romanlutz I assume the broken CI pipeline has nothing to do with my changes? 😅

Yes, that's correct. I requeued them. Looks like some connection issue with codecov.

MiroDudik (approved):
This looks fine to me. I think that we can hold off with replacing the dataset for another issue.

PR #1179, opened 2023-01-13
Add filename.png to gitignore

iofall:
## Description
`filename.png` gets generated every time we run `plot_quickstart.py`. On the last line we save a figure which leads to the file being saved in the `examples` directory.

## Tests

- [x] no new tests required

## Documentation

## Changes
- Added `filename.png` to gitignore.


examples/plot_quickstart.py
```diff
@@ -92,3 +92,8 @@
     title="Show count metric in pie chart",
 )
 fig[0][0].figure.savefig("filename.png")
+
+# sphinx_gallery_start_ignore
+import os
```
romanlutz, line 97: Instead of this code in the notebook, perhaps adding an exception for 'filename.png' in `.gitignore` is simpler?

iofall: Yep, I think that will be better.

.gitignore
```diff
@@ -118,3 +118,6 @@ docbuild/
 
 # OSX directory info files
 .DS_Store
+
+# plot_quickstart.py
+filename.png
```
romanlutz, line 123: nit: I believe it gets generated under `examples` so perhaps `**/filename.png` helps to account for such files.

romanlutz (approved):
LGTM! @fairlearn/fairlearn-maintainers please review and merge if you agree, or provide feedback 🙂 Thanks @iofall !

adrinjalali:
It would have been much nicer to fix the tests so that it doesn't produce the file :D

PR #1183, opened 2023-01-18
MNT Remove Azure DevOps pipeline definitions and add GitHub workflow for running tests with pinned requirements

### LINKED ISSUE #1180: CI update / delete nightly build pipelines
We currently don't have any nightly builds, do we want to reinstitute them?

Related items:
* It's probably safe to delete devops directory (which was previously used for nightly builds and Azure Pipelines more broadly)
* Azure Pipelines sticker in our readme should be deleted too

romanlutz:
## Description

closes #1180 

This PR adds a schedule in addition to the pull request and push to main triggers. This means we run all pipelines at least once a day which helps in figuring out when something breaks due to changing dependencies. Additionally, I've extended the matrix to run all tests once with default/latest dependencies and once with dependencies pinned to the oldest allowed version using @riedgar-ms 's `install_requirements.py` script. I've also removed the entire `devops` directory with Azure DevOps pipelines since they're of no use to us anymore, and removed the corresponding badge from the `README.md` since it's not useful.

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation


riedgar-ms:
OK in principle with me, but your workflow isn't running....

romanlutz:
> OK in principle with me, but your workflow isn't running....

Good point 🤣 probably the trigger yaml formatting...

riedgar-ms:
I'm actually a little worried that 'All checks passed' when a workflow was broken :-(

riedgar-ms:
That's the sort of situation which keeps me awake at night

.github/workflows/build-test.yml
```diff
@@ -1,6 +1,16 @@
 name: build-test
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - "main"
+  pull_request:
+    branches:
+      - "main"
```
adrinjalali, line 9: we also need to run them on release branches

.github/workflows/build-test.yml
```diff
@@ -1,6 +1,16 @@
...
+  schedule:
+    - cron: "0 8 * * *"
```
adrinjalali, line 11: do we want to run this job or do we want a separate job for nightly runs with different requirements? I think a separate job would make sense.

adrinjalali (commented):
Thanks for starting this @romanlutz

.github/workflows/build-test.yml
romanlutz: opened #1188 for that

adrinjalali (commented):
this is a rather large refactoring, but I don't mind it.

The only condition we need to have is for jobs to run only on this repo and not other repos, otherwise we can remove conditions for push and pull_request triggers.

.github/workflows/build-test.yml
romanlutz: Hmm, I was of the opinion that it doesn't run unless I do that. Let me try removing it.

adrinjalali (approved):
Ok then, let's see how this works. You can merge whenever you want, and see if it's working :)

Issue #1174, opened 2023-01-12
DOC Split out API reference docs


alliesaizan:
#### Describe the issue linked to the documentation

Currently, the API docs for each module include all the components of that module on one page, as shown below:

![image](https://user-images.githubusercontent.com/15962391/212165898-c8135afb-a7df-4c56-b0f5-754fb03363fa.png)

#### Suggest a potential alternative/fix

Suggest splitting the docs up so that each component in the module gets its own page, like in the image below:

![image](https://user-images.githubusercontent.com/15962391/212166132-3fd6f7f6-fbb1-4d8b-8ede-f023c43e7181.png)

This issue was brought up during the weekly community call on 1/12/23, and I found a package that can automatically split up Sphinx docs, [sphinx-audomodapi](https://sphinx-automodapi.readthedocs.io/en/latest/). 

Looping in @hildeweerts @MiroDudik @romanlutz to see if this package would work for this use case (also @hildeweerts, let me know if I misinterpreted the issue you mentioned!). I'd be happy to research this more and submit a PR if you think this would be a good solution!

romanlutz:
I check what scikit-learn does for this since they have exactly what we want in a working state. I think I got something promising with just "autosummary". I pushed it to a little branch. Feel free to build on that if you want! https://github.com/fairlearn/fairlearn/tree/romanlutz/separate_page_per_class_and_func
[Sorry for pushing the branch to the main repo, that was a mistake, but I guess it'll work....]

alliesaizan:
Oh np, autosummary was another way I saw to do this! If that's working then I will close this issue :)

romanlutz:
TBH I don't know about pros and cons of each option. `autosummary` just looks pretty straightforward and staying in somewhat close alignment with scikit-learn is typically a good idea as we can keep adopting their innovations 🙂 Let's leave the issue open until a PR is out and merged (?)
I'll leave it up to you whether you want to try this or not. If not, I might be able to try it over the next few weeks, but no guarantee.

hildeweerts:
This is exactly the issue I was referring to @alliesaizan! 

About a million years ago I opened an issue for this #720, I will close that one in favor of this new issue.

alliesaizan:
[...]

PR #1184, opened 2023-01-18
DOC Breaking out information in the API Reference into separate pages.

alliesaizan:
## Description
This PR splits out documentation on the functions and classes within fairlearn's modules into separate stub pages using the Sphinx [autosummary](https://www.sphinx-doc.org/en/master/usage/extensions/autosummary.html) extension, with the goal of improving the readability of the API reference.

Addresses issue #1174.

Tagging @hildeweerts, @romanlutz 

## Tests

- [X] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

- [X] API docs added or updated

## Screenshots
image


.gitignore
```diff
@@ -118,3 +118,5 @@ docbuild/
 
 # OSX directory info files
 .DS_Store
+docs/api_reference/index.rst
+examples/filename.png
```
romanlutz, line 122: #1179 added `filename.png` to `.gitignore`, so it's probably worth merging with main and removing the duplicate.

nit: These should probably in a different section since they're not related to OSX.

romanlutz (approved):
Amazing! Thanks @alliesaizan for taking this up.

I was initially thinking that everything should be on a single page from where you can navigate to the individual class/function pages, in other words a hierarchy with just 2 levels (overview and class/function pages).

This PR looks like we're getting a 3-level hierarchy. Since the navigation bar on the left has the full hierarchy and we can go to the module-level pages in the main part of the page I am not so sure anymore 😄 Perhaps this is actually better (?) I certainly don't feel very strongly about it either way. Maybe @hildeweerts does?

@fairlearn/fairlearn-maintainers FYI

adrinjalali (approved):
This looks pretty nice to me :)

romanlutz:
@hildeweerts @MiroDudik and I discussed this in the community call today. @alliesaizan it's awesome that you've taken this up! @hildeweerts has been asking for this for years and is delighted 🤣 

We talked about the hierarchy options as mentioned before and would prefer the "shallow" option, that is to have the `api_reference/index.rst` be the "overview" page that lists all modules as sections with all classes and functions therein. In the current version it's more hierarchical. A good example is scikit-learn: https://scikit-learn.org/stable/modules/classes.html
They list everything in one page.

The other thing that's notable about scikit-learn is that they have `classes.rst` under https://github.com/scikit-learn/scikit-learn/tree/main/doc/modules which generates the page linked above, but they don't have the `generated` individual files. Is that something we need to add to the repo or is it auto-generated?

@alliesaizan please let me know if this makes sense to you. If you prefer talking we can always hop on a call anytime. If you have other thoughts or disagree we're obviously interested in hearing about it. I'm just trying to summarize what happened in the call so that you have an idea and can respond. Again, thank you 🙂

docs/api_reference/fairlearn.metrics.rst
```diff
@@ -1,7 +1,31 @@
...
\ No newline at end of file
+.. currentmodule:: fairlearn.metrics
+    
+.. autosummary:: 
+   :nosignatures:
+   :toctree: generated/
+   :template: function.rst
+
+    count
+    demographic_parity_difference
```
romanlutz, line 12: were these the ones you meant @MiroDudik ?

alliesaizan:
> @hildeweerts @MiroDudik and I discussed this in the community call today. @alliesaizan it's awesome that you've taken this up! @hildeweerts has been asking for this for years and is delighted 🤣
> 
> We talked about the hierarchy options as mentioned before and would prefer the "shallow" option, that is to have the `api_reference/index.rst` be the "overview" page that lists all modules as sections with all classes and functions therein. In the current version it's more hierarchical. A good example is scikit-learn: https://scikit-learn.org/stable/modules/classes.html They list everything in one page.
> 
> The other thing that's notable about scikit-learn is that they have `classes.rst` under https://github.com/scikit-learn/scikit-learn/tree/main/doc/modules which generates the page linked above, but they don't have the `generated` individual files. Is that something we need to add to the repo or is it auto-generated?
> 
> @alliesaizan please let me know if this makes sense to you. If you prefer talking we can always hop on a call anytime. If you have other thoughts or disagree we're obviously interested in hearing about it. I'm just trying to summarize what happened in the call so that you have an idea and can respond. Again, thank you 🙂

Yes, I understand! I mocked that format up when I was testing locally, but I will commit the change to the index file asap!

romanlutz:
> Yes, I understand! I mocked that format up when I was testing locally, but I will commit the change to the index file asap! For point #2, the individual files are auto-generated under the folder you specify in the autosummary command. I might need to discuss this via a call if that isn't clear or if we need to change that setting.

Awesome! If they're auto-generated, then let's just add the directory they live in to the `.gitignore` since sphinx will regenerate them whenever needed. I'm assuming by "autosummary command" you mean the settings in `conf.py`. If you manually had to run something to generate them (other than the sphinx doc generation command to build the website) then I'm probably misinterpreting this.

docs/api_reference/index.rst
```diff
@@ -13,3 +13,108 @@ API Docs
    fairlearn.reductions
    fairlearn.adversarial
    fairlearn.experimental
```
romanlutz, line 15: This is exactly what I was thinking of, excellent! I am wondering if we need the module-specific pages, i.e., everything above this line (`fairlearn.*`) as separate pages at all. After all, we have all their contents below.

Wdyt?

alliesaizan:
> > Yes, I understand! I mocked that format up when I was testing locally, but I will commit the change to the index file asap! For point #2, the individual files are auto-generated under the folder you specify in the autosummary command. I might need to discuss this via a call if that isn't clear or if we need to change that setting.
> 
> Awesome! If they're auto-generated, then let's just add the directory they live in to the `.gitignore` since sphinx will regenerate them whenever needed. I'm assuming by "autosummary command" you mean the settings in `conf.py`. If you manually had to run something to generate them (other than the sphinx doc generation command to build the website) then I'm probably misinterpreting this.

Oh no, sorry, should have been more specific. In each module's rst file, you can use _toctree_ to specify a folder where the generated stub pages should be stored. I can add this folder to .gitignore.

docs/api_reference/index.rst
alliesaizan: I don't have a preference, it would be a pretty easy change - but I suppose we don't need them since readers can just use the side menu to navigate to a page for a specific module?

romanlutz: I guess I'm wondering if we should remove the module-specific pages altogether, so that they wouldn't even show up in the side menu. They aren't really useful unless they have additional content. AFAICT that's what scikit-learn does as well. There's only https://scikit-learn.org/stable/modules/classes.html#module-sklearn.covariance but no sub-pages per module. They do link relevant user guide sections in the section description for each module (if any).

hildeweerts: >I guess I'm wondering if we should remove the module-specific pages altogether, so that they wouldn't even show up in the side menu. 

I'm not sure if I fully understand your suggestion @romanlutz. Scikit-learn has module-level navigation on the side menu which directs to the specific location in the (single page) API reference overview. So while they don't have module specific pages, the side menu does have module level navigation. Are you suggesting no module-level navigation at all or just no module-level pages?

.gitignore
alliesaizan: Not sure why the index file also ended up in gitignore, but I have removed them both.

docs/api_reference/index.rst
romanlutz: Correct. I'm just saying let's not have module-specific pages unless we actually put useful content there (which we currently don't have). Otherwise it's just one additional click. So as long as the module-specific links just take you to the section on the same page that's fine, but I had understood the PR differently. I should probably run this locally and see what it looks like, but I don't have time before next week...

romanlutz:
Finally got down to trying it out locally!

image

First of all, this is great! Being able to go from API reference directly to the single-page for any class or function is exactly what we wanted all along and we got it.

Still, I don't understand why there's the list of modules at the top. We have all the modules in order right below. And it's not even like the ones on top link to the ones below. If you click on one it opens a module-specific page that isn't needed because it has no extra information compared to the full Fairlearn API reference overview page.

If instead we removed that part and all the module-specific pages (and added `automodule` similar to [sklearn](https://raw.githubusercontent.com/scikit-learn/scikit-learn/main/doc/modules/classes.rst) we'd get this:
image


Note that the left-side nav is for the individual pages because the module subpages are gone. We still have the right-side nav, though, which navigates within the page to all the modules.

If I'm really nitpicky I'd say perhaps we should order them in some other way, but this makes me quite happy. Wdyt @alliesaizan @hildeweerts  @MiroDudik @adrinjalali  @riedgar-ms ?

alliesaizan:
@romanlutz I agree that removing the module list improves the documentation! I tend to slightly favor the drop-down menus on the left-side over having the list of individual pages, but I think that because the modules are listed on the right-hand side, having the left-hand side as proposed (with the individual pages) is fine. Pushing directly to my branch is fine too!

romanlutz:
Sounds good. I've pushed those changes now and will wait for another maintainer to chime in and/or potentially merge (assuming it's passing the doc build).

adrinjalali (approved):
I don't love that everything is listed on the left side, but this is much better than what we have anyway. Thanks for the work :)

Issue #1185, opened 2023-01-19
BUG fetch_diabetes_hospital does not have as_frame argument


MiroDudik:
#### Describe the bug
The function [`fairlearn.datasets.fetch_diabetes_hospital`](https://fairlearn.org/main/api_reference/fairlearn.datasets.html#fairlearn.datasets.fetch_diabetes_hospital) does not have `as_frame` argument, unlike all other fetch functions in the datasets module.

#### Suggested fix
Add `as_frame` argument to `fetch_diabetes_hospital`. We could possibly also rethink the default behavior of the fetch functions. My sense is that defaulting to `as_frame=False` is actually the opposite of what I'd like whenever I use the fetch functions (which is basically just in educational resources).

shimst3r:
I'd like to tackle this so I can continue working on #1182 🤓

shimst3r:
@MiroDudik given that `as_frame=False` is the default for the other fetch functions, but `fetch_diabetes_hospital` has an implicit `as_frame=True` (because the current logic assumes the return value to be a data frame), I would propose to implement the solution for this issue using `as_frame=True` (to avoid breaking changes for people who haven't set the function argument explicitly).

Regarding the other fetch functions:
- In my opinion, **not** returning a data frame as the default is sensible. It's a more complex data structure than a `numpy.ndarray` and hence *feels* like setting an additional flag is okay.
- Would you say that change needs a dedicated issue with proper change management (potentially even a logger info with a grace period)?

MiroDudik:
I think that we should have a consistent default across all of `fetch_*` functions.

But you are definitely right that we shouldn't change the default behavior on `fetch_diabetes_hospital`, where it's effectively `True`, or on all other functions, where it is `False`, without some transition period. Once we decide what the new default should be, we should make sure that those functions where the default will change issue a warning when called without specifying `as_frame`. For example, the warning would be something like:

>You have not specified the value for `as_frame`, which currently defaults to `True`. Starting with version 0.10.0, the default is changing to `False`. Specify the value of `as_frame` to suppress the warning.

[...]

Issue #1165, opened 2023-01-03
BUG feature names missing in fetch_bank_marketing and fetch_credit_card


MiroDudik:
#### Describe the bug

The function `fairlearn.datasets.fetch_bank_marketing` loads data set with dummy feature names of the form 'V1',...,'V16' although the UCI dataset does list feature names: https://archive.ics.uci.edu/ml/datasets/Bank+Marketing

#### Steps/Code to Reproduce
```python
import fairlearn.datasets as fld
fld.fetch_bank_marketing(as_frame=True).feature_names
```

#### Expected Results
```
['job', 'marital', 'education', ...]
```

#### Actual Results
```python
['V1', 'V2', 'V3', 'V4', 'V5', 'V6', 'V7', 'V8', 'V9', 'V10', 'V11', 'V12', 'V13', 'V14', 'V15', 'V16']
```

MiroDudik:
A similar issue might arise with `fetch_credit_card` once #1008 is merged

romanlutz:
Thanks! I pointed this out in the credit card example PR, too. Is the fix to not pass the as_frame argument, extract the column names from the metadata, and then create the frame ourselves?

hildeweerts:
An easier fix is to just use (or upload) a version of the bank marketing dataset on OpenML where the feature names are filled out correctly, e.g., https://www.openml.org/search?type=data&status=active&id=44234

romanlutz:
Oh great, that is definitely preferable!

MiroDudik:
Heh... the linked version has feature names, but the description field is severely abridged and categorical features are incorrectly typed...

MiroDudik:
@hildeweerts , @romanlutz : Is this a good first issue?

romanlutz:
It's certainly one we want help on if someone is interested. I only vaguely remember uploading a dataset to OpenML once and it was exceedingly painful, but I can't remember if it was because of the dataset (format) I was uploading or because of OpenML's UI. They've done a lot of work to improve their UI in recent years, so perhaps it wouldn't be quite as difficult. I would hesitate to point someone new to this, though. Perhaps a good second or third issue? 😉

hildeweerts:
@romanlutz did you have any luck uploading the dataset to OpenML?

romanlutz:
[...]

PR #1194, opened 2023-01-21
BUG fetch diabetes hospital does not have as frame argument

shimst3r:
## Description

As proposed by @MiroDudik in #1185, I implement the `as_frame` argument for `datasets._fetch_hospital_diabetes()`. I opted for a default value of `True`, to reflect the current API behaviour. I decided against implementing a general change to how the fetch functions deal with `as_frame` to not bloat this PR.

## Tests

- [ ] no new tests required
- [ ] new tests added
- [x] existing tests adjusted

## Documentation


fairlearn/datasets/_fetch_diabetes_hospital.py
```diff
@@ -69,19 +73,22 @@ def fetch_diabetes_hospital(
     dataset : :obj:`~sklearn.utils.Bunch`
         Dictionary-like object, with the following attributes.
 
-        data : ndarray, shape (101766, 25)
-            Each row corresponding to the 25 feature values in order.
+        data : ndarray, shape (101766, 24)
```
romanlutz, line 76: I'm a bit surprised that his changed.

shimst3r: Me as well, I triple checked it. 😅 
```python
>>> data_dict['data'].values.shape
(101766, 24)
```

romanlutz: Ah I think the reason is that we count the target column. So either we need to change that everywhere or leave it as n+1

shimst3r: Well, if the options are "being right" and "being lazy", I will always opt for "being lazy". 😂

shimst3r: Is this something to put up for discussion and create a follow-up issue for it? 🤔

romanlutz: I agree with at least having a discussion. I have stumbled upon this (and over it) a couple of times, too. Let's create an issue so that people can comment.

MiroDudik: The new value (24) is correct. The old value (25) was incorrect. Consistently with all other datasets, we should be reporting the size of data.data, and not data.frame.

MiroDudik:
Before merging, we should settle the question of what default value of `as_frame` we would like to have across all `fetch_*` functions in future. I think it's okay to discuss on #1185? (Or feel free to create its own issue...)

fairlearn/datasets/_fetch_diabetes_hospital.py
shimst3r: I would suggest putting that change into a new PR, do you agree?

romanlutz: It's fine either way. Are you planning to make that change, too? If not, lmk and I'll create the issue.

shimst3r: I'll do it. 👍

shimst3r: @romanlutz @MiroDudik each dataset has a block description in the docstring, e.g.
```python
    ==============   ====================
    Samples total                   45211
    Dimensionality                     17
    Features         numeric, categorical
    Classes                             2
    ==============   ====================
```
for the `bank_marketing` dataset. Would you count the target against the `Dimensionality` listed here, i.e. keeping the `n+1` or list the dimensionality of the feature vector?

romanlutz: As @MiroDudik wrote earlier it's probably best to list the number of features, i.e., excluding the target column.

MiroDudik: @shimst3r -- if you are making the documentation change in this PR, can you update all three data sets that currently list n+1, so they list n? In all three data sets, the incorrect value n+1 appears in three places -- at the top of the doc, and also in the returned value section (where it appears twice).

shimst3r: It's 4 places, right? Because the `feature_names` is also listing n+1 if I'm not mistaken. :)

shimst3r: @MiroDudik anyway, changes are pushed. 😄

fairlearn/datasets/_fetch_diabetes_hospital.py
```diff
@@ -91,11 +98,21 @@ def fetch_diabetes_hospital(
     if not data_home:
         data_home = pathlib.Path().home() / _DOWNLOAD_DIRECTORY_NAME
 
-    return fetch_openml(
+    data_dict = fetch_openml(
         data_id=43874,
         data_home=data_home,
         cache=cache,
         as_frame=True,
```
MiroDudik, line 105: Can you use `as_frame=as_frame` to get the expected behavior without all of the ifs below?

MiroDudik (approved):
This looks good to me. Just double check whether `as_frame=as_frame` in the fetch_openml call could be substituted to simplify things.

fairlearn/datasets/_fetch_diabetes_hospital.py
MiroDudik: (What a can of worms... sorry about this... I'm just realizing through this issue as well as #1165 that we have a non-trivial dependence on OpenML.)

test/unit/datasets/test_datasets.py
```diff
@@ -52,8 +54,19 @@ def test_dataset_as_bunch(self, as_frame, fetch_function):
         ],
     )
     def test_dataset_as_X_y(self, as_frame, fetch_function):
+        if as_frame is False and fetch_function is fetch_diabetes_hospital:
```
romanlutz, line 57: nit: `pytest.skip` could be used here I think because `return` marks it as passed.

romanlutz (approved):

test/unit/datasets/test_datasets.py
shimst3r: Fixed, thanks for the nit 😌

MiroDudik (commented):
Looks good except for a small documentation tidbit (remove the old note, and appropriately format the new note).

fairlearn/datasets/_fetch_diabetes_hospital.py
```diff
@@ -52,6 +52,15 @@ def fetch_diabetes_hospital(
 
     Parameters
     ----------
+    as_frame : bool, default=True
+        If True, the data is a pandas DataFrame including columns with
+        appropriate dtypes (numeric, string or categorical).
+
+        NOTE: If set to False, this will raise an exception because of a
```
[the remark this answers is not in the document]
shimst3r: Oh yes, you're right. Thanks for being this thorough! I should take the time to actually render the docs locally every time I make an update. 🙈

PR #1202, opened 2023-02-05
[ENH] merge metric files into _base_metrics.py, rename _disparities.py to fairness_metrics.py, and move _input_manipulations.py to the utils module

### LINKED ISSUE #455: Improve organization of metrics module
#### Describe the issue
I think the organization of files in the fairlearn.metrics module could be improved, 

~~both in API reference and the repository itself. My main issue regarding the documentation is that the distinction between performance metrics (e.g. recall) and fairness metrics (e.g. demographic parity) is not clear in the [API reference](https://fairlearn.github.io/api_reference/fairlearn.metrics.html). Additionally,~~ 

navigating the repository was not very intuitive for me. For example, some of the performance metrics not directly available in scikit-learn have their own file (e.g. _balanced_root_mean_squared_error.py), whereas others are grouped in (e.g. _extra_metrics.py).

I think merging and/or renaming some of the seperate files could help to improve. E.g. something like this:
* _fairness.py
    * _disparities.py
*  _base_metrics.py ( ~~_performance.py (or perhaps _classification.py and _regression.py to mimic scikit-learn)~~
    * _extra_metrics.py
    * _balanced_root_mean_squared_error.py 
    * _mean_predictions.py
    * _selection_rate.py
* _input_manipulations.py might be better off in a utils folder in the main repository.
* I am not sure what the best place would be for _group_metric_set.py. 

~~PS. I suppose giving each function their own page as in the scikit-learn API reference is already implicit in the documentation proposal, right?~~

EDIT: documentation issues moved to issue #720

romanlutz:
## Description

closes #455 

This PR merges metric files into `_base_metrics.py`, renames `_disparities.py` to `_fairness.py`, and moves `_input_manipulations.py` to the `utils` module. None of the contents of the migrated functions was changed.

## Tests

- [x] no new tests required
- [ ] new tests added
- [x] existing tests adjusted

## Documentation


hildeweerts:
Other-ml tests seem to have failed due to a deprecation error in xgboost: "UserWarning: `use_label_encoder` is deprecated in 1.7.0." According to [the deprecation tracker](https://github.com/dmlc/xgboost/issues/3986) users need to do label encoding manually. We set it to `False` so we just need to remove `use_label_encoder=False` from our code.

hildeweerts (commented):
Thanks, @romanlutz! The renaming makes things so much clearer., IMO. There are a few codecov warnings we should look into, other than that this looks good to me!

riedgar-ms:
When it comes to naming, I thought we preferred `disparity` to `fairness` on the grounds that the former was definitely measurable, whereas the latter was not?

romanlutz:
> When it comes to naming, I thought we preferred `disparity` to `fairness` on the grounds that the former was definitely measurable, whereas the latter was not?

I'm following the instructions from the issue #455 🙂 I don't care either way tbh. It's just a file name 🙂 @hildeweerts  do you have thoughts?

hildeweerts:
Given that we refer to this type of metric as fairness metric a lot throughout the user guide I think using "fairness metric" is perfectly sensible :)

romanlutz:
> There are a few codecov warnings we should look into, other than that this looks good to me!

I didn't change any code, so codecov is probably angry that file coverage went down or something like that. 

Actually, looking at the specific files it points out... I didn't touch those or anything related at all. Must be a codecov bug?

hildeweerts:
> Actually, looking at the specific files it points out... I didn't touch those or anything related at all. Must be a codecov bug?

I think these files probably predate our installment of codecov, so maybe it's not a bug but an actual lack of coverage? In any case I don't think it is something that needs to be fixed in this PR.

hildeweerts (approved):

riedgar-ms:
I've no objection to the consolidation and reorganisation of files, but as I recall, calling them 'disparity metrics' was a pretty deliberate choice. That said, it's not user facing either.

fairlearn/metrics/_base_metrics.py
```diff
@@ -257,3 +260,122 @@ def count(y_true, y_pred) -> int:
...
+    """
+    y_p = _convert_to_ndarray_and_squeeze(y_pred)
+    s_w = np.ones(len(y_p))
+    if sample_weight is not None:
+        s_w = _convert_to_ndarray_and_squeeze(sample_weight)
+
+    return np.dot(y_p, s_w) / s_w.sum()
+
+
+def _mean_overprediction(y_true, y_pred, sample_weight=None) -> float:
```
MiroDudik, line 290: I had meant to delete `_mean_overprediction` and `_mean_underprediction` when I was dealing with dashboard ghosts. (We could do it in this PR or a separate PR.)

MiroDudik:
My sense is that we have since departed from calling them "disparity metrics". I'm more comfortable calling them "fairness metrics"... To be parallel with `_base_metrics.py`, maybe we should call the old `_disparities.py` something like `_fairness_metrics.py`?

fairlearn/metrics/_base_metrics.py
romanlutz: Done! Also found some test data related to group metric set that I deleted now.

riedgar-ms:
OK, I'm fine with the renaming then.

PR #1204, opened 2023-02-06
[MNT] Fix failing xgboost tests in the otherml test pipeline

romanlutz:
## Description

The xgboost tests are currently failing because `use_label_encoder` is deprecated. From v1.7.0 onwards it should be omitted.

While I check the requirements I also found that we had several packages with minimum versions that are apparently not supported by Python 3.8, so I bumped them.

## Tests

- [x] no new tests required
- [ ] new tests added
- [x] existing tests adjusted

## Documentation

### DIFF
--- a/test_othermlpackages/conda-xgboost.yaml
+++ b/test_othermlpackages/conda-xgboost.yaml
@@ -8,7 +8,7 @@ dependencies:
 - pytest
 - pytest-cov
 - xgboost>=1.7.0
-- pandas<1.4 # Capped for xgboost only
+- pandas
 
 # Note xgboost not available for Windows on conda-forge
 # See
--- a/test_othermlpackages/conda-xgboost.yaml
+++ b/test_othermlpackages/conda-xgboost.yaml
@@ -7,7 +7,7 @@ dependencies:
 - pip>=19.1.1
 - pytest
 - pytest-cov
-- xgboost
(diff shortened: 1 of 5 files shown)


test_othermlpackages/conda-xgboost.yaml
```diff
@@ -7,7 +7,7 @@ dependencies:
 - pip>=19.1.1
 - pytest
 - pytest-cov
-- xgboost
+- xgboost>=1.7.0
 - pandas<1.4 # Capped for xgboost only
```
hildeweerts, line 11: Do you know if the pandas cap is still required if we set xgboost >=1.7.0?

hildeweerts (approved):

test_othermlpackages/conda-xgboost.yaml
romanlutz: Let's remove it and see what happens...

romanlutz, line 11:
```suggestion
- pandas
```

romanlutz: Still works, keeping the change 🙂

hildeweerts: Awesome. Feel free to merge then if you agree with your own PR ;)

PR #1219, opened 2023-03-02
Caching for MetricFrame

riedgar-ms:
## Description

Change `MetricFrame` so that all results (including aggregations) are computed in the constructor and then cached. This is a preliminary for the introduction of bootstrapping, as discussed in a number of other venues.

This may well technically be a breaking change. Although I believe that it will only change the way in which the calling code may be broken. In a nutshell, if any of the aggregation functions raise an exception, this exception has to be caught, cached and rethrown if the user invokes the aggregation which causes it. This means that if a user metric function returns an exception (for some bizarre reason), then we would raise that exception, rather than the `ValueError` which users would have received before.

## Tests

- [ ] no new tests required
- [x] new tests added
- [ ] existing tests adjusted

## Documentation

### DIFF
--- a/fairlearn/metrics/_metric_frame.py
+++ b/fairlearn/metrics/_metric_frame.py
@@ -368,6 +368,7 @@ class MetricFrame:
                         raw_result, v, err_string
                     )
                 except Exception as e:  # noqa: B902
+                    # Store any exception for later
                     self._result_cache[k][err_string] = e
 
         # Differences
@@ -397,6 +398,7 @@ class MetricFrame:
                     else:
                         self._result_cache["difference"][c_m][err_string] = result
                 except Exception as e:  # noqa: B902
+                    # Store any exception for later
                     self._result_cache["difference"][c_m][err_string] = e
 
         # Ratios
@@ -426,6 +428,7 @@ class MetricFrame:
                     else:
                         self._result_cache["ratio"][c_m][err_string] = result
                 except Exception as e:  # noqa: B902
+                    # Store any exception for later
                     self._result_cache["ratio"][c_m][err_string] = e
 
     @property
--- a/fairlearn/metrics/_metric_frame.py
+++ b/fairlearn/metrics/_metric_frame.py
@@ -331,6 +331,22 @@ class MetricFrame:
             control_feature_names=self._cf_names,
         )
 
+        self._result_cache=dict()
+        self._populate_results(self._result)
+
+    def _populate_results(self, result: DisaggregatedResult):
+        # Start with overall
+        if self._user_supplied_callable:
+            if self.control_levels:
+                self._result_cache['overall'] = result.overall.iloc[:, 0]
+            else:
+                self._result_cache['overall'] = result.overall.iloc[0]
+        else:
+            self._result_cache['overall'] = result.overall
+
+        # Now
+
+
     @property
     def overall(self) -> Union[Any, pd.Series, pd.DataFrame]:
         """Return the underlying metrics evaluated on the whole dataset.
@@ -364,13 +380,7 @@ class MetricFrame:
             interface when calling programatically, while also reducing
             typing for those using Fairlearn interactively.
         """
-        if self._user_supplied_callable:
-            if self.control_levels:
-                return self._result.overall.iloc[:, 0]
-            else:
-                return self._result.overall.iloc[0]
-        else:
-            return self._result.overall
+        return self._result_cache['overall']
 
     @property
     def by_group(self) -> Union[pd.Series, pd.DataFrame]:
--- a/fairlearn/metrics/_metric_frame.py
+++ b/fairlearn/metrics/_metric_frame.py
@@ -342,6 +342,22 @@ class MetricFrame:
         self._result_cache = dict()
         self._populate_results(result)
 
+    def _extract_result(self, underlying_result, no_control_levels: bool):
+        """
+        Change result types for those who dislike consistency.
+
+        The `no_control_levels` parameter determines whether the presence
+        of control levels will affect the result. This is the case for
+        overall, but not the other cases.
+        """
+        if self._user_supplied_callable:
+            if self.control_levels or no_control_levels:
+                return underlying_result.iloc[:, 0]
+            else:
+                return underlying_result.iloc[0]
+        else:
+            return underlying_result
+
     def _populate_results(self, raw_result: DisaggregatedResult):
         """
         Populate the :code:`_result_cache`.
@@ -355,21 +371,15 @@ class MetricFrame:
         Note that if exceptions are thrown, we cache those, and they are thrown
         if the user calls the corresponding method (and arguments).
         """
-
         # Start with overall
-        if self._user_supplied_callable:
-            if self.control_levels:
-                self._result_cache["overall"] = raw_result.overall.iloc[:, 0]
-            else:
-                self._result_cache["overall"] = raw_result.overall.iloc[0]
-        else:
-            self._result_cache["overall"] = raw_result.overall
+        self._result_cache["overall"] = self._extract_result(
+            raw_result.overall, no_control_levels=False
+        )
 
         # Now do by_group
-        if self._user_supplied_callable:
-            self._result_cache["by_group"] = raw_result.by_group.iloc[:, 0]
-        else:
-            self._result_cache["by_group"] = raw_result.by_group
+        self._result_cache["by_group"] = self._extract_result(
+            raw_result.by_group, no_control_levels=True
+        )
 
         # Next up, group_min and group_max
         group_functions = {"group_min": "min", "group_max": "max"}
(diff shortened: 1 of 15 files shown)


adrinjalali:
A big caveat of this approach is that we do a lot more computation than necessary for most cases. Most users are not going to be calling all the methods we have, and moving the computation to the constructor, adds all of those computations which were not needed before. WDYT?

riedgar-ms:
I agree that it's doing extra computations - although not _that_ many, relatively speaking. The expensive ones are `.overall` and `.by_group`, and those are always done in the contructor.

However, this is the paradigm we need if we're not going to store each of the bootstrap `.overall` and `.by_group` results, but only the final confidence intervals. We could only do the pre-computation for the bootstrap case, but I'd rather have the same program flow for both regular and bootstrap.

adrinjalali:
Could we maybe have some benchmarks on real sized datasets for this? That would make it easier for us to make a decision.

riedgar-ms:
Did we ever define a benchmark set? What would you consider a realistic size - both for total number of samples, and the total number of subgroups (remember it's the latter which will affect the building of the cache, not the former).

adrinjalali:
I'd be happy if we have something for a dataset of 50-500k rows and a few different groups.

We should have at least an idea of what this change does.

riedgar-ms:
OK, using the following test code:
```python
import timeit

import numpy as np
from sklearn.metrics import accuracy_score

n_samples = 100_000
n_sensitive_classes = 4
n_iterations = 100
n_repeats = 5

rng = np.random.default_rng(abs(-440315))

y_true = rng.integers(low=0, high=2, size=n_samples)
y_pred = rng.integers(low=1, high=2, size=n_samples)

choices = [chr(ord("a") + i) for i in range(n_sensitive_classes)]
s_f = rng.choice(choices, size=n_samples)

result = timeit.repeat(
    stmt=(
        "MetricFrame(metrics=accuracy_score, y_true=y_true, y_pred=y_pred,"
        " sensitive_features=s_f)"
    ),
    number=n_iterations,
    repeat=n_repeats,
    globals=globals(),
)

print(f"Time taken: {result}")
```
For `main`, I get the following output:
```bash
Time taken: [108.91859960000001, 109.54478879999999, 113.20957380000002, 116.77203079999998, 113.9796627]
```
and for this branch, I get:
```bash
Time taken: [114.2391846, 112.0890944, 122.76237759999998, 110.1880256, 114.28243989999999]
```
so, a couple of percent slower. This isn't too surprising, given that the `by_group` computation is the expensive part. Once that exists, building the cache depends on the number of rows in that DataFrame (which is sized by the number of sensitive features) rather than the number of samples.

fairlearn/metrics/_metric_frame.py
```diff
@@ -323,14 +330,107 @@ def __init__(
...
             sensitive_feature_names=self._sf_names,
             control_feature_names=self._cf_names,
         )
 
+        # Build into cache
+        self._result_cache = dict()
+        self._populate_results(result)
+
+    def _populate_results(self, raw_result: DisaggregatedResult):
+        # Start with overall
```
adrinjalali, line 346: seems like each of these sections could be their own method/func

fairlearn/metrics/_metric_frame.py
```diff
@@ -323,14 +330,107 @@ def __init__(
...
             annotated_functions=annotated_funcs,
             sensitive_feature_names=self._sf_names,
             control_feature_names=self._cf_names,
         )
 
+        # Build into cache
+        self._result_cache = dict()
+        self._populate_results(result)
+
+    def _populate_results(self, raw_result: DisaggregatedResult):
```
adrinjalali, line 345: docstring would help a mortal like me understand the code better ;)

fairlearn/metrics/_metric_frame.py
```diff
@@ -438,7 +529,10 @@ def sensitive_levels(self) -> List[str]:
         return self._sf_names
 
     def __group(
```
adrinjalali, line 531: hadn't noticed we have something starting with `__`, instead of `_`, why is that?

fairlearn/metrics/_metric_frame.py
```diff
@@ -438,7 +529,10 @@ def sensitive_levels(self) -> List[str]:
...
-        self, grouping_function: str, errors: str = "raise"
+        self,
+        disagg_result: DisaggregatedResult,
+        grouping_function: str,
+        errors: str = "raise",
```
adrinjalali, line 535: docstring seems to be not up-to-date

adrinjalali (commented):
I don't completely understand this PR, but looks like with the refactoring the code looks nicer.

Question: would there be a case where we'd need to invalidate the cache but we miss it?

Otherwise looks quite nice, and I'm happy with the benchmark results we have.

fairlearn/metrics/_metric_frame.py
riedgar-ms: Done

fairlearn/metrics/_metric_frame.py
riedgar-ms: I don't remember. It might have been because this was considered an _extra_ private method. \Bbut that's not part of this PR :-)

fairlearn/metrics/_metric_frame.py
riedgar-ms: I'm not convinced of the value of that, since the method isn't that long and will only be called once. However, I have refactored out a little of the common functionality, to make this method a bit shorter (and to avoid some almost-duplicated code).

riedgar-ms:
I don't believe that the cache can ever be invalid - users can't add data after creating a `MetricFrame`.

fairlearn/metrics/_metric_frame.py
```diff
@@ -323,14 +330,105 @@ def __init__(
...
             annotated_functions=annotated_funcs,
             sensitive_feature_names=self._sf_names,
             control_feature_names=self._cf_names,
         )
 
+        # Build into cache
+        self._result_cache = dict()
+        self._populate_results(result)
+
+    def _extract_result(self, underlying_result, no_control_levels: bool):
```
adrinjalali, line 345: I suppose this is here also to keep backward compatibility?

fairlearn/metrics/_metric_frame.py
```diff
@@ -323,14 +330,105 @@ def __init__(
...
+            raw_result.by_group, no_control_levels=True
+        )
+
+        # Next up, group_min and group_max
+        group_functions = {"group_min": "min", "group_max": "max"}
+        for k, v in group_functions.items():
+            self._result_cache[k] = dict()
+            for err_string in _VALID_ERROR_STRING:
+                try:
+                    self._result_cache[k][err_string] = self.__group(
```
adrinjalali, line 400: why is it called `__group` and not `_group`?

fairlearn/metrics/_metric_frame.py
```diff
@@ -493,7 +580,14 @@ def group_max(self, errors: str = "raise") -> Union[Any, pd.Series, pd.DataFrame
             The maximum value over sensitive features. The exact type
             follows the table in :attr:`.MetricFrame.overall`.
         """
-        return self.__group("max", errors)
+        if errors not in _VALID_ERROR_STRING:
+            raise ValueError(_INVALID_ERRORS_VALUE_ERROR_MESSAGE)
+
+        value = self._result_cache["group_max"][errors]
+        if isinstance(value, Exception):
+            raise value
```
adrinjalali, line 588: what does this look like? It's odd to be raising something somewhere else without `from` maybe? And should we have like a `_raise_or_return(value)` method instead?

adrinjalali (commented):
I like how this is mostly simplifying the code actually.

fairlearn/metrics/_metric_frame.py
riedgar-ms: You could say that. Most of the work has been shuffled off into the `DisaggregatedResult`, and that tends towards my preference for type safety. The functionality here isn't new, but was previously distributed, to supply the more 'user friendly' return values of `MetricFrame`

fairlearn/metrics/_metric_frame.py
riedgar-ms: As before, I don't remember exactly why that was; I think that `__group()` was actually refactored out by someone other than me.

fairlearn/metrics/_metric_frame.py
adrinjalali: yeah I got the logic, what I was wondering was, imagine there's a non-fairlearn exception happening there, and we raise it here. How odes that look to the end user, and would they be able to figure out what to do, since the stack trace is now not very obvious to them? Could we make up an example where we could see how it looks like maybe?

fairlearn/metrics/_metric_frame.py
riedgar-ms: Changed (rather than putting in a separate PR).

PR #1220, opened 2023-03-09
MNT add codecov token

adrinjalali:
As suggested (https://github.com/codecov/codecov-action/issues/837#issuecomment-1453877750), this PR adds the codecov token to our config so that the codecov uploads become more reliable (they're very annoying).

Note that we can't put the token as a secret since secrets are not exposed to PRs from forks.

This does not add a security issue since people can upload coverage to the codecov anyway.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -5,3 +5,4 @@ codecov:
   notify:
     after_n_builds: 12
     wait_for_ci: true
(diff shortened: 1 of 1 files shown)


riedgar-ms (approved):
Not overjoyed, but I saw the linked comment saying that the token is siloed. Bad coverage reports aren't great but if that's the limit of the damage....

PR #1229, opened 2023-04-04
MNT Repository blackening and upgrade black version to 23.3.0

riedgar-ms:
## Description

Run `black` on the repository, and add the requiste GitHub Action to keep this enforced.

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

### DIFF
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -11,7 +11,7 @@ on:
         type: string
 
 jobs:
-  linting:
+  linting-flake8:
     runs-on: ${{ inputs.os }}
     steps:
       - uses: actions/checkout@v3
@@ -20,3 +20,8 @@ jobs:
           python-version: ${{ inputs.python }}
       - run: pip install -r requirements-dev.txt
       - run: flake8 .
+  linting-black:
+    runs-on: ${{ inputs.os }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: psf/black@stable
--- a/.github/workflows/linting.yml
+++ b/.github/workflows/linting.yml
@@ -21,7 +21,11 @@ jobs:
       - run: pip install -r requirements-dev.txt
       - run: flake8 .
   linting-black:
+    # Need to keep black version consistent in
+    # requirements-dev.txt
+    # linting.yml
+    # .pre-commit-config.yaml
     runs-on: ${{ inputs.os }}
     steps:
       - uses: actions/checkout@v3
-      - uses: psf/black@stable
+      - uses: psf/black@23.3.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,6 +9,10 @@ repos:
     -   id: check-case-conflict
     -   id: check-merge-conflict
 -   repo: https://github.com/psf/black
+# Need to keep black version consistent in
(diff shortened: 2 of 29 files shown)


adrinjalali:
why only one folder? we could do the whole code base in one PR/commit, and add them to precommit hooks (or did we do that already)?

riedgar-ms:
> why only one folder? we could do the whole code base in one PR/commit, and add them to precommit hooks (or did we do that already)?

You did it once already, and then told `git` to ignore that commit in diffs. However, things have drifted again, since we don't have the commit hook.

I have extended to cover all of `fairlearn` and `tests`, although I haven't accepted _all_ of the changes from `black`. There are a few nasty edge cases where `black` reformats something to keep a line length, but `flake8` wants to have a single line quoted string on a single like. Or the `black` reformatting means that a `flake8` suppression now doesn't apply properly.

adrinjalali:
You can also use a `# fmt: off / # fmt: on` tag to disable black formatting fora certain region.

I think we should fix those few issues you mention, and enable `black` in our CI / pre-commit hooks, and only use `flake8` for cases which black doesn't detect maybe.

riedgar-ms:
I think that might need wider discussion. I can make this work by rephrasing some of the one-line docstrings, to trim out characters. Or we could bump up the maximum line length slightly. Neither of those feel like good solutions.

Turning off the `flake8` rule which is giving grief here ("single line docstrings must fit on single line") may well be the better option, but I think that would merit wider discussion.

riedgar-ms:
Scratch that. I just needed to update the version of `black` I was running. Have added the matching GitHub action to the linting build.

.github/workflows/linting.yml
```diff
@@ -20,3 +20,8 @@ jobs:
           python-version: ${{ inputs.python }}
       - run: pip install -r requirements-dev.txt
       - run: flake8 .
+  linting-black:
+    runs-on: ${{ inputs.os }}
+    steps:
+      - uses: actions/checkout@v3
+      - uses: psf/black@stable
```
adrinjalali, line 27: do we want to use the latest always, or do we want to pin the version? 

This should be the same as what we have in `pre-commit` hooks, which right now is:

```
-   repo: https://github.com/psf/black
    rev: 22.3.0
    hooks:
    -   id: black
```

riedgar-ms: Good point; let's discuss in the community call, since I think that we should bump `black` up to v23, which appears to resolve some of the arguments between `black` and `flake8`.

adrinjalali: I might not make it to the call today, but I'm in favor of upgrading black here.

riedgar-ms: @fairlearn/fairlearn-maintainers , since we didn't really have a quorum at the community call, do you have any objections to bumping up the version of `black` to 23?

riedgar-ms: Have bumped.... will wait until after the community call, but at this point I'm leaning towards merging _unless_ there's an objection.

adrinjalali, line 27: we should still use the hardcoded version here.

adrinjalali (commented):
Otherwise LGTM

PR #1235, opened 2023-04-27
FEAT Create optimal transport preprocessing module by Berk et al.

mberk06:
## Description

This PR creates a data preprocessing method that transforms "disadvantaged" data to have a distribution that reflects that of a specified "advantaged" group.

## Tests

- [ ] no new tests required
- [x] new tests added
- [ ] existing tests adjusted

## Documentation


romanlutz:
Hi @mberk06 , can you share the paper where this was published?

mberk06:
> Hi @mberk06 , can you share the paper where this was published?

Yup. [Link](https://arxiv.org/abs/2111.09211) and it's also in the class docs.

mberk06:
@romanlutz just wondering if there are any updates. Can I help answer any questions?

hildeweerts:
Hi, @mberk06. Thank you for your PR and for considering integrating your research in Fairlearn! 

Unfortunately, we are a bit picky regarding the mitigation algorithms we include in our library as we have only limited capacity to work on the project. To avoid essentially replicating a peer-reviewing process, we decided a while ago to only include algorithms published in peer-reviewed papers that have been adopted in the research community (e.g., often used benchmarks or well-received) or in practice.

As a side note, you may have noticed that we (try to) emphasize a sociotechnical perspective on fairness in our documentation. We strive to include user guides that help practitioners understand the suitability of an algorithm for their use case from a broader perspective than fairness metrics alone (e.g., questions like what does the algorithm optimize for and in what situations is that (not) a relevant optimization objective? are there scenarios in which you should definitely not use the algorithm? does the algorithm rely on particular empirical assumptions (e.g., assume a particular form of bias in data collection) or normative assumptions (e.g., explicitly address historical inequalities)? etc.).

Issue #951, opened 2021-09-14
DOC Add versionadded and versionchanged to API reference (fairlearn.metrics)


romanlutz:
Related to #855

Issue #855, opened 2021-06-07
Add versionadded and versionchanged to API reference


romanlutz:
Add `.. versionadded` and `.. versionchanged` to all methods (and classes if possible)

```
.. versionadded:: 0.7
          In previous versions only the ``predict`` method was used
            implicitly.
        .. versionchanged:: 0.7
            From version 0.7, 'predict' is deprecated and the default will
            change to 'auto' from v0.10.
```

_Originally posted by @romanlutz in https://github.com/fairlearn/fairlearn/pull/854#r646574960_

matthew-so:
I'd be interested in picking this up.

adrinjalali:
@matthew-so please tackle this issue one file/module at a time. You can figure out when things were added to the library with the help of `git blame` or gitlens on VSCode for instance, and checking the repository's history and changelog.

matthew-so:
Got it; thanks for the advice!

matthew-so:
Since I do not have the time to fully commit to this, it would be preferable if others could tackle sections at a time. For example, I will now be working on fairlearn.metrics.

hildeweerts:
Hi @matthew-so - could you perhaps open separate issues for the different sections? In that way other prospective contributors can easily identify that they can pick it up!

bramreinders97:
Hi, I'd be interested in picking this up. I will start at fairlearn.reductions, considering that @matthew-so is doing metrics right now.

hildeweerts:
Hi @bramreinders97: awesome! Could you open up a new issue in which you refer back to this issue? Thanks!

eskayML:
currently working on **fairlearn.metrics** task

PR #1247, opened 2023-06-27
added version added and version changed 0.7 to fairlearn.metrics

eskayML:
## Description
@romanlutz 
Added all the version added and version changed to fairlearn.metrics according to #855  and #951
## Tests

- [ ] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

### DIFF
--- a/fairlearn/metrics/_annotated_metric_function.py
+++ b/fairlearn/metrics/_annotated_metric_function.py
@@ -25,7 +25,13 @@ class AnnotatedMetricFunction:
     `DataFrame.apply()`. The net result of this is that it is useful
     to have a function wrapper which can be invoked with a DataFrame
     argument, and extract its arguments from that DataFrame.
-
+    .. versionadded:: 0.7
+            In previous versions only the ``predict`` method was used
+            implicitly.
+    .. versionchanged:: 0.7
+            From version 0.7, 'predict' is deprecated and the default will
+            change to 'auto' from v0.10.
+  
     Parameters
     ----------
     func : callable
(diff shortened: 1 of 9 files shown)


fairlearn/metrics/_annotated_metric_function.py
```diff
@@ -25,7 +25,13 @@ class AnnotatedMetricFunction:
...
     to have a function wrapper which can be invoked with a DataFrame
     argument, and extract its arguments from that DataFrame.
-
+    .. versionadded:: 0.7
+            In previous versions only the ``predict`` method was used
+            implicitly.
+    .. versionchanged:: 0.7
+            From version 0.7, 'predict' is deprecated and the default will
+            change to 'auto' from v0.10.
+
```
romanlutz, line 34: @eskayML thanks for finding all the locations where we need to add `versionadded` and `versionchanged`. They need to actually match the class/function, though. So it's about cross-checking with the past versions to see where it was added, and replace 0.7 with that version. If there were no major changes to it since then, `versionchanged` is not needed. If there were major changes (e.g., to the API) then the version in which it changed should be named instead of 0.7 and a description of the change should be added.

Does that make sense?

eskayML: Interesting, so for example the `[_annotated_metric_function.py](https://github.com/fairlearn/fairlearn/blob/release/v0.8.X/fairlearn/metrics/_annotated_metric_function.py)` which looks to have been created in version 8, I'm to use just  the version added alone 

Also I'd love to know if i'm to add the description under the `versionadded` or `versionchanged` just like the example you gave in #855 , I think that might be a little challenging for me.

eskayML: Thanks for clarifying it all.

romanlutz:
Closing this PR since the approach is meant to be different. @eskayML feel free to reopen if you intend to make the required modifications.

PR #1274, opened 2023-07-28
MNT Update main for release v0.9.0

hildeweerts:
Update main for release v0.9.0.

**Do not merge until the release is completed. Otherwise, a new website will be published that points to the new version which doesn’t exist yet.**

## Tests

- [ ] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

### DIFF
--- a/docs/_static/versions.json
+++ b/docs/_static/versions.json
@@ -7,7 +7,7 @@
     {
         "name": "v0.8.0",
         "version": "0.8.0",
-        "url": "https://fairlearn.org/v0.8.0/"
+        "url": "https://fairlearn.org/v0.8/"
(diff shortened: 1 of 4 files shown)


riedgar-ms (approved):

hildeweerts:
As our tests only build the webpage for the current main, we cannot really view whether the more general changes worked on CircleCI, right?

Re. tiny changes, one side effect of having a new version called 0.10.0 is that it's listed after 1.0 instead of 0.9.0:
image
I don't think it's a big issue but it could be slightly confusing.

adrinjalali (approved):

Issue #848, opened 2021-06-04
Test for deprecation warnings


adrinjalali:
Our CI should test and fail for `DeprecationWarning` and `FutureWarning`. This can be done by passing `-Werror::DeprecationWarning -Werror::FutureWarning` to `pytest`.

We can enable this for at least a few of the tests if not all.

riedgar-ms:
Is there any way to restrict this to our direct dependencies? For example, [in a recent build](https://dev.azure.com/responsibleai/fairlearn/_build/results?buildId=12007&view=logs&j=aa2fd561-a338-5537-45ee-49ae22aea958&t=7d222596-41fc-5765-4dd3-df02a80934d4&l=163) there's a deprecation warning on `ansiwrap` but that's being brought in by some other package.

adrinjalali:
I think you can ignore certain warnings: 

https://github.com/scikit-learn/scikit-learn/blob/main/setup.cfg#L20

romanlutz:
Didn't #1171 already address this? From a brief search I found that we're basically configuring all warnings to be errors unless explicitly excluded in the pyproject.toml file as @MiroDudik put it in #1171. So we should be able to just close #1278 , right? @MiroDudik can you confirm and close PR and issue if this is right?

PR #1171, opened 2023-01-11
FIX test warnings

MiroDudik:
Signed-off-by: Miro Dudik <mdudik@gmail.com>

## Description
Fix warnings coming up in build tests.

PR #1278, opened 2023-08-10
MNT deprecation and future warnings

eskayML:
Made an attempt at issue #848

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,6 +49,10 @@ jobs:
           name: "Copy landing_page.js"
           command: cp docs/static_landing_page/js/landing_page.js docs/_build/
 
+      - run:
+          name: "Run tests with warnings as errors"
+          command: pytest -Werror::DeprecationWarning -Werror::FutureWarning
+          
       - persist_to_workspace:
           root: docs/_build
           paths: 


romanlutz:
We'd probably want this in other pipelines, not necessarily the website build.

eskayML:
I have a question

So lets assume i'm to implement this on all pipelines. 

Am I to edit the `setup.cfg` file instead and add something like 

```bash
[pytest]
addopts = -Werror::DeprecationWarning -Werror::FutureWarning
```

romanlutz:
Oooh that's interesting! I hadn't heard of this but that seems perfect. Maybe run it locally to make sure it succeeds. We may need to fox a few things before that can become the rule 🙂 if it does fail please post the output and we can open an issue for that.

eskayML:
This is what i'm having



Screenshot (7)

romanlutz:
That is mostly due to not having pytest matplotlib (mpl) installed. 

Those tests don't currently work on the build agents for reasons I haven't quite figured out yet. Probably something to do with how mpl uses OS-specific tools to create its images.

Is there a way to ignore individual issues like this?

adrinjalali:
Closing since #1171 has resolved the issue. Thanks for the contribution @eskayML . Sorry for this being resolved somewhere else.

Issue #949, opened 2021-09-14
DOC Add versionadded and versionchanged to API reference (fairlearn.preprocessing)


romanlutz:
Related to #855

PR #1279, opened 2023-08-10
add: version added and version changed to fairlearn.preprocessing

eskayML:
This addresses issue #949 , under the thread #855

### DIFF
--- a/fairlearn/preprocessing/_correlation_remover.py
+++ b/fairlearn/preprocessing/_correlation_remover.py
@@ -59,8 +59,6 @@ class CorrelationRemover(BaseEstimator, TransformerMixin):
 
     .. versionadded:: 0.6
 
-    .. versionchanged:: 0.8
-        Added the _more_tags_ method
     """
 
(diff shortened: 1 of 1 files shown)


fairlearn/preprocessing/_correlation_remover.py
```diff
@@ -56,6 +56,11 @@ class CorrelationRemover(BaseEstimator, TransformerMixin):
     Note that the lack of correlation does not imply anything about statistical dependence.
     Therefore, we expect this to be most appropriate as a preprocessing step for
     (generalized) linear models.
+
+    .. versionadded:: 0.6
+
+    .. versionchanged:: 0.8
```
romanlutz, line 62: IMO this isn't a change that is user-facing so we should omit it and just keep versionadded. @adrinjalali  made that change for sklearn compatibility and should probably decide 🙂

adrinjalali: Yes, things which are not _public_ don't have to have a version added or changed tag.

romanlutz (approved):

Issue #1222, opened 2023-03-12
Make it clear that certain metrics, when added to a MetricFrame, need the sample_params argument, and how to fill it out.


tuffstuff9:
#### Describe the issue linked to the documentation

[Here](https://fairlearn.org/main/user_guide/assessment/advanced_metricframe.html) is the section in the docs which gives an 'example' of how sample_params is meant to be used. It simply gives a description which is not entirely clear, considering it took me over 2 hours of searching around and trying different things until I finally got my code to work.

#### Suggest a potential alternative/fix

Something like this is a lot clearer at illustrating what the sample_params does:
![image](https://user-images.githubusercontent.com/57072903/224558009-721d0031-48e5-4dca-9fc8-74f355757a1c.png)

However, it is still not entirely clear to me why I need to have the 'sensitive_features' included in the sample_params in the first place, when it is already included in the MetricFrame. The errors thrown out when you forgo the sample_params led me down the wrong path entirely, as can be seen below. 

![image](https://user-images.githubusercontent.com/57072903/224558103-ccbc32a0-083c-4a9e-bb76-7211d639bf80.png)

This makes me think that I need to include the sensitive_features argument in my one/all of the metrics in the metric_dict, which then says I need a y_true and a y_pred, and once I include those, it gives me this error:

![image](https://user-images.githubusercontent.com/57072903/224558195-c5eeb381-26fd-4dd7-a0e6-789c6a62d605.png)
![image](https://user-images.githubusercontent.com/57072903/224558959-5601ca4f-fa4c-4ad2-80a5-d54d5678cec9.png)

And at this point, one reaches a dead end and goes back to the drawing board. I think these error messages can be improved, but more importantly, I am not sure why the docs don't clearly specify why the metrics with _ratio or _difference (at the very least these are the ones I know of) need a sample_params to be passed into the MetricFrame, and what the nature of the sample_params must be. 

Here is what the docs have to say about the sample_params parameter:

![image](https://user-images.githubusercontent.com/57072903/224558368-55413c22-b0af-478f-939a-c2c2bad20f1d.png)

Perhaps I am missing something, but I am not sure how my above screenshot of a correct sample_params entails "the values being the string-to-array-like dictionaries". Also, I do not see any distinction in the description of the actual metrics regarding whether they need a sample_params to be passed in or not. For example 'false_positive_rate' does not, but 'equalized_odds_ratio' does. 

Thank you for taking this into consideration.

hildeweerts:
[...]

Issue #1231, opened 2023-04-17
DOC Improve metrics documentation


hildeweerts:
#### Describe the issue linked to the documentation

Our documentation currently does not clearly indicate the two main types of fairness assessments in Fairlearn (i.e. `MetricFrame` versus fairness metrics such as `equalized_odds_difference`), which can lead to confusion (see e.g., #1222).

#### Suggest a potential alternative/fix

Some places where we could possibly mention this more explicitly:
* https://fairlearn.org/main/api_reference/index.html#module-fairlearn.metrics
* https://fairlearn.org/main/user_guide/assessment/index.html
* https://fairlearn.org/main/user_guide/assessment/perform_fairness_assessment.html

hildeweerts:
Also related to #1095

alliesaizan:
Assigning myself to this one - I think it makes most sense to either add this information to _Assessment_ or _Performing a Fairness Assessment_. If someone from the community wants to help out, I will offer assistance :). If no one picks it up in a reasonable amount of time, I will propose a path forward and implement it myself.

romanlutz:
It may not be mutually exclusive... clearly, this is a point that confuses people so mentioning it multiple times may be a good thing.

alliesaizan:
> It may not be mutually exclusive... clearly, this is a point that confuses people so mentioning it multiple times may be a good thing.

Agreed, I will mock up a version where this is mentioned in both sections.

alliesaizan:
Resolved by #1284.

PR #1284, opened 2023-09-02
Adding documentation addressing issue 1231

alliesaizan:
## Description
This PR adds information about the distinction between MetricFrame and the predefined fairness metrics to the User Guide. It references open issue #1231 and closed issue #1222. I welcome all recommendations on feedback to the wording. I also added a code snippet that is duplicative of the code in "Common Fairness Metrics" to clarify the documentation, but we can remove it if it does not add anything. Tagging @hildeweerts and @romanlutz (@riedgar-ms I accidentally tagged you as a reviewer but please disregard)

## Tests
- [X] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation
- [X] user guide added or updated

## Screenshots
image
image

### DIFF
--- a/docs/user_guide/assessment/index.rst
+++ b/docs/user_guide/assessment/index.rst
@@ -11,6 +11,13 @@ fairness metrics, such as demographic parity and equalized odds.
 We will show how :class:`MetricFrame` can be used to evaluate the metrics
 identified during the course of a fairness assessment.
 
+Fairlean provides two primary ways of assessing fairness: :class:`MetricFrame`,
+which can be used to perform disaggregated analysis of a particular performance 
+metric (such as accuracy, false positive rate, etc.) across sensitive 
+groups, and a set of predefined fairness metrics that output a single 
+number. In the :ref:`perform_fairness_assessment`, we will dive further into
+each of these types of fairness assessments. 
+
 In the mathematical definitions below, :math:`X` denotes a feature vector 
 used for predictions, :math:`A` will be a single sensitive feature (such as age 
 or race), and :math:`Y` will be the true label.
--- a/docs/user_guide/assessment/perform_fairness_assessment.rst
+++ b/docs/user_guide/assessment/perform_fairness_assessment.rst
@@ -199,6 +199,11 @@ These are accessed through the :attr:`MetricFrame.by_group` property:
 
 All of these values can be checked against the original arrays above.
 
+Note that :class:`MetricFrame` is intended for analyzing the differences 
+between groups with regard to a base metric, and consequently cannot take 
+predefined fairness metrics, such as :func:`demographic_parity_difference`, 
+as input to the `metrics` parameter.
+
 .. _assessment_compare_harms:
 
 Compare quantified harms across the groups
@@ -277,3 +282,46 @@ overall values for the data:
     dtype: float64
 
 In every case, the *largest* difference and *smallest* ratio are returned.
+
+
+.. _assessment_predefined_fairness_metrics:
+
+Predefined fairness metrics
+---------------------------
+
+In addition to the disaggregated analysis of base metrics enabled by
+:class:`MetricFrame`, Fairlearn also provides a set of predefined fairness 
+metrics that output a single score. These metrics take as input 
+`sensitive_features` to compute the maximum difference or ratio between 
+subgroups of a sensitive variable. The predefined fairness metrics offered 
+by Fairlearn are **demographic parity** ratio/difference and **equalized odds** 
+ratio/difference. Note that because these metrics are calculated using 
+aggregations between groups (unlike, say, accuracy, the demographic parity 
+for one subgroup versus another does not make sense), they are meant to be 
+called directly, rather than used within the instantiation of a MetricFrame.
+
+Below, we show an example of calculating demographic parity ratio using the 
+sample data defined above.
+
+.. doctest:: assessment_metrics
+    :options:  +NORMALIZE_WHITESPACE
+
+    >>> from fairlearn.metrics import demographic_parity_ratio
+    >>> print(demographic_parity_ratio(y_true,
+    ...                                y_pred,
+    ...                                sensitive_features=sf_data))
+    0.66666...
+
+Note that to achieve a one-number score for a standard performance metric,
+like false positive rate or selection rate, the user must use the MetricFrame 
+data structure. Under the hood, the fairness assessment metrics also use 
+:class:`MetricFrame` to compute a particular base rate across sensitive 
+groups and subsequently perform an aggregation (the difference or ratio) 
+on the base metric values across groups. For example, 
+:func:`equalized_odds_ratio` uses both the :func:`false_positive_rate` and
+:func:`false_negative_rate` within a MetricFrame on the backend to generate 
+an output.
+
+:ref:`common_fairness_metrics` provides an overview of common metrics used 
+in fairness analyses. For a deep dive into how to extend the capabilities of 
+fairness metrics provided by Fairlearn, review :ref:`custom_fairness_metrics`.


docs/user_guide/assessment/index.rst
```diff
@@ -11,6 +11,13 @@ fairness metrics, such as demographic parity and equalized odds.
 We will show how :class:`MetricFrame` can be used to evaluate the metrics
 identified during the course of a fairness assessment.
 
+Fairlean provides two primary ways of assessing fairness: :class:`MetricFrame`,
+which can be used to perform disaggregated analysis of a particular performance 
+metric (such as accuracy, false positive rate, etc.) across sensitive 
+groups, and a set of predefined fairness metrics that output a single 
+number. In the :ref:`perform_fairness_assessment`, we will dive further into
```
romanlutz, line 18: Maybe "an aggregate number"?

docs/user_guide/assessment/index.rst
```diff
@@ -11,6 +11,13 @@ fairness metrics, such as demographic parity and equalized odds.
 We will show how :class:`MetricFrame` can be used to evaluate the metrics
 identified during the course of a fairness assessment.
 
+Fairlean provides two primary ways of assessing fairness: :class:`MetricFrame`,
+which can be used to perform disaggregated analysis of a particular performance
```
romanlutz, line 15: I'm fine with this. Technically, though, you can get the single number from MetricFrame, too. It's just that if you don't care about by-group values then there's no reason to use MetricFrame

docs/user_guide/assessment/perform_fairness_assessment.rst
```diff
@@ -276,4 +281,47 @@ overall values for the data:
...
+Predefined fairness metrics
+---------------------------
+
+In addition to the disaggregated analysis of base metrics enabled by
+:class:`MetricFrame`, Fairlearn also provides a set of predefined fairness 
+metrics that output a single score. These metrics take as input 
+`sensitive_features` to compute the maximum difference or ratio between 
+subgroups of a sensitive variable. The predefined fairness metrics offered 
+by Fairlearn are **demographic parity** ratio/difference and **equalized odds** 
+ratio/difference. Note that because these metrics are calculated using
```
romanlutz, line 298: Clarification: there's ratio/diff that is max/min or max-min, but some people prefer relative to mean. That's also possible with MetricFrame.

docs/user_guide/assessment/perform_fairness_assessment.rst
```diff
@@ -276,4 +281,47 @@ overall values for the data:
...
+    :options:  +NORMALIZE_WHITESPACE
+
+    >>> from fairlearn.metrics import demographic_parity_ratio
+    >>> print(demographic_parity_ratio(y_true,
+    ...                                y_pred,
+    ...                                sensitive_features=sf_data))
+    0.66666...
+
+Note that to achieve a one-number score for a standard performance metric,
+like false positive rate or selection rate, the user must use the MetricFrame
```
romanlutz, line 316: There's make_derived_metric for this. MetricFrame is only needed for by-group assessments.

docs/user_guide/assessment/index.rst
alliesaizan: Changed from "a single number" to "an aggregate number"

docs/user_guide/assessment/perform_fairness_assessment.rst
alliesaizan: Change has been added

docs/user_guide/assessment/perform_fairness_assessment.rst
alliesaizan: I can add this distinction as a new sentence (that the radio/diff can be max/min or max-min and that MetricFrame also provides this capability), but let me know if there is better phrasing I can use to distinguish between the two.

docs/user_guide/assessment/index.rst
alliesaizan: Added a sentence: " MetricFrame can also be used to output an aggregate number, but the predefined fairness metrics can be used when direct by-group comparison is not necessary." Let me know if that is on the right track.

riedgar-ms: I know it's mentioned down below, but do you want to mention here that the predefined fairness metrics use `MetricFrame` internally?

docs/user_guide/assessment/perform_fairness_assessment.rst
```diff
@@ -276,4 +281,47 @@ overall values for the data:
...
+---------------------------
+
+In addition to the disaggregated analysis of base metrics enabled by
+:class:`MetricFrame`, Fairlearn also provides a set of predefined fairness 
+metrics that output a single score. These metrics take as input 
+`sensitive_features` to compute the maximum difference or ratio between 
+subgroups of a sensitive variable. The predefined fairness metrics offered 
+by Fairlearn are **demographic parity** ratio/difference and **equalized odds** 
+ratio/difference. Note that because these metrics are calculated using 
+aggregations between groups (unlike, say, accuracy, the demographic parity
```
riedgar-ms, line 299: The part in parentheses doesn't quite make sense to me.

alliesaizan: I removed it for now.

docs/user_guide/assessment/index.rst
```diff
@@ -11,6 +11,16 @@ fairness metrics, such as demographic parity and equalized odds.
 We will show how :class:`MetricFrame` can be used to evaluate the metrics
 identified during the course of a fairness assessment.
 
+Fairlean provides two primary ways of assessing fairness: :class:`MetricFrame`,
+which can be used to perform disaggregated analysis of a particular performance 
+metric (such as accuracy, false positive rate, etc.) across sensitive 
+groups, and a set of predefined fairness metrics that output an aggregate 
+number. MetricFrame can also be used to output an aggregate number,
```
riedgar-ms, line 18: Nitpick: should be :class:`MetricFrame`

docs/user_guide/assessment/perform_fairness_assessment.rst
```diff
@@ -199,6 +199,11 @@ These are accessed through the :attr:`MetricFrame.by_group` property:
 
 All of these values can be checked against the original arrays above.
 
+Note that :class:`MetricFrame` is intended for analyzing the differences 
+between groups with regard to a base metric, and consequently cannot take
```
riedgar-ms, line 203: Suggest "...analyzing the disparities between groups when evaluating a base metric, and ....."

My biggest concern is switching 'difference' for 'disparity' here

docs/user_guide/assessment/perform_fairness_assessment.rst
```diff
@@ -276,4 +281,46 @@ overall values for the data:
...
+Note that to achieve a one-number score for a standard performance metric,
+like false positive rate or selection rate, the user must use 
+:func:`make_derived_metric`. Under the hood, the fairness assessment metrics 
+also use :class:`MetricFrame` to compute a particular base rate across 
+sensitive groups and subsequently perform an aggregation (the difference 
+or ratio) on the base metric values across groups. For example, 
+:func:`equalized_odds_ratio` uses both the :func:`false_positive_rate` and
+:func:`false_negative_rate` within a MetricFrame on the backend to generate 
+an output.
+
```
riedgar-ms, line 323: Thought.....would it be helpful to add another code block showing how you get to exactly the same result using `MetricFrame` here?

riedgar-ms (commented):
I'm generally OK with this. The only comment which really animates me is changing 'differences' to 'disparities' because we use 'difference' with a specific meaning when comparing between groups.

docs/user_guide/assessment/index.rst
alliesaizan: Updated!

docs/user_guide/assessment/index.rst
alliesaizan: Sure, I have changed that sentence to say "and a set of predefined fairness metrics that use :class:`MetricFrame` internally to output an aggregate number."

docs/user_guide/assessment/perform_fairness_assessment.rst
alliesaizan: Done

alliesaizan:
> I'm generally OK with this. The only comment which really animates me is changing 'differences' to 'disparities' because we use 'difference' with a specific meaning when comparing between groups.

I made the change suggested above and did not see "differences" referring to disparities elsewhere in the new sections, though please let me know if I missed anything!

docs/user_guide/assessment/perform_fairness_assessment.rst
alliesaizan: Just added an example, but let me know if the language describing the example should change.

riedgar-ms: Perhaps an example which doesn't return '0'? It's great to show how things happen internally, but when both numbers are zero, it's not very convincing (I have a whole separate rant about languages which mix automatic declaration and zero-initialisation).

docs/user_guide/assessment/perform_fairness_assessment.rst
alliesaizan: Added a sentence: "Either the ratio or difference can be calculated as the max, min or max-min between subgroups (:class:`MetricFrame` also has the capability to calculate these metrics)."

docs/user_guide/assessment/perform_fairness_assessment.rst
alliesaizan: Yeah, I think this would entail generating new data. This recommendation fell off my radar, but I can start mocking something up. If there's synthetic data generated elsewhere in the docs that might also work for this, please let me know.

docs/user_guide/assessment/perform_fairness_assessment.rst
```diff
@@ -295,7 +295,10 @@ metrics that output a single score. These metrics take as input
 `sensitive_features` to compute the maximum difference or ratio between 
 subgroups of a sensitive variable. The predefined fairness metrics offered 
 by Fairlearn are **demographic parity** ratio/difference and **equalized odds** 
-ratio/difference. Note that because these metrics are calculated using 
+ratio/difference. Either the ratio or difference can be calculated as the max,
```
romanlutz, line 298: Sorry, I must have been more confusing than helping before. We have
- group_max
- group_min
- ratio
- difference

and for the ratio/difference it can be `between_groups` or `to_overall`, but `to_overall` results in more than 1 number.

alliesaizan: My apologies for the confusion on my end. I replaced the sentence cited above with "The ratio and difference can be calculated `between_groups` or `to_overall`, but `to_overall` results in more than 1 number being returned.  :class:`MetricFrame` can also calculate differences and ratios between groups using these methods. For more information on available method of computing ratios or differences, view the documentation for :meth:`MetricFrame.ratio` and :meth:`MetricFrame.difference`, respectively." Let me know if that is sufficient.

docs/user_guide/assessment/perform_fairness_assessment.rst
```diff
@@ -276,4 +281,75 @@ overall values for the data:
...
+---------------------------
+
+In addition to the disaggregated analysis of base metrics enabled by
+:class:`MetricFrame`, Fairlearn also provides a set of predefined fairness 
+metrics that output a single score. These metrics take as input 
+`sensitive_features` to compute the maximum difference or ratio between 
+subgroups of a sensitive variable. The predefined fairness metrics offered 
+by Fairlearn are **demographic parity** ratio/difference and **equalized odds** 
+ratio/difference. The ratio and difference can be calculated `between_groups` 
+or `to_overall`, but `to_overall` results in more than 1 number being returned.
```
riedgar-ms, line 299: Nitpick: "more than a single value being returned"

But.... when does it do that? I'm confused, because I thought that the `between_groups`/`to_overall` selector just changed the reference point, but that ratio and difference would work as normal.

docs/user_guide/assessment/perform_fairness_assessment.rst
riedgar-ms: It might be simpler to leave this for now, and file an issue to revise the input data. That's going to propagate all over the place, and isn't central to this PR.

docs/user_guide/assessment/index.rst
```diff
@@ -11,6 +11,17 @@ fairness metrics, such as demographic parity and equalized odds.
 We will show how :class:`MetricFrame` can be used to evaluate the metrics
 identified during the course of a fairness assessment.
 
+Fairlean provides two primary ways of assessing fairness: :class:`MetricFrame`,
+which can be used to perform disaggregated analysis of a particular performance 
+metric (such as accuracy, false positive rate, etc.) across sensitive 
+groups, and a set of predefined fairness metrics that use :class:`MetricFrame` 
+internally to output an aggregate number.
+:class:`MetricFrame` can also be used to output an aggregate number,
```
riedgar-ms, line 19: Nitpick: I'd prefer 'value' to 'number' here.

alliesaizan: Done

docs/user_guide/assessment/perform_fairness_assessment.rst
alliesaizan: Made the small change from number to value. @romanlutz can you please clarify what you meant above for @riedgar-ms?

docs/user_guide/assessment/perform_fairness_assessment.rst
alliesaizan: Yeah agreed, I will resolve this comment for now and open an issue sometime this week.

docs/user_guide/assessment/perform_fairness_assessment.rst
```diff
@@ -276,4 +281,75 @@ overall values for the data:
...
+
+Predefined fairness metrics
+---------------------------
+
+In addition to the disaggregated analysis of base metrics enabled by
+:class:`MetricFrame`, Fairlearn also provides a set of predefined fairness 
+metrics that output a single score. These metrics take as input 
+`sensitive_features` to compute the maximum difference or ratio between 
+subgroups of a sensitive variable. The predefined fairness metrics offered 
+by Fairlearn are **demographic parity** ratio/difference and **equalized odds**
```
hildeweerts, line 297: Should we maybe refer to :func:`demographic_parity_difference` and :func:`demographic_parity_ratio` explicitly here?

docs/user_guide/assessment/perform_fairness_assessment.rst
```diff
@@ -276,4 +281,75 @@ overall values for the data:
...
+---------------------------
+
+In addition to the disaggregated analysis of base metrics enabled by
+:class:`MetricFrame`, Fairlearn also provides a set of predefined fairness 
+metrics that output a single score. These metrics take as input 
+`sensitive_features` to compute the maximum difference or ratio between 
+subgroups of a sensitive variable. The predefined fairness metrics offered 
+by Fairlearn are **demographic parity** ratio/difference and **equalized odds** 
+ratio/difference. The ratio and difference can be calculated `between_groups` 
+or `to_overall`, but `to_overall` results in more than 1 value being returned.
```
hildeweerts, line 299: I might be missing something, but multiple values being returned is only true when we have `control_features`, no?

docs/user_guide/assessment/perform_fairness_assessment.rst
```diff
@@ -276,4 +281,75 @@ overall values for the data:
...
+:class:`MetricFrame` can also calculate differences and ratios between groups 
+using these methods. For more information on available method of computing
```
hildeweerts, line 301: The phrasing "using these methods" is a bit confusing for me, as it suggests that `MetricFrame` 'borrows' `'ratio'` and `'difference'` from `equalized_odds_difference` etc. It seems like the parts about `MetricFrame.ratio` and `MetricFrame.difference` are better off in the section about `MetricFrame` itself.

docs/user_guide/assessment/perform_fairness_assessment.rst
```diff
@@ -276,4 +281,75 @@ overall values for the data:
...
+    :options:  +NORMALIZE_WHITESPACE
+
+    >>> from fairlearn.metrics import demographic_parity_ratio
+    >>> print(demographic_parity_ratio(y_true,
+    ...                                y_pred,
+    ...                                sensitive_features=sf_data))
+    0.66666...
+
+Note that to achieve a one-number score for a standard performance metric,
+like false positive rate or selection rate, the user must use
```
hildeweerts, line 321: Maybe phrase this as *can* instead of *must*. How about something like:

"It is also possible to define custom fairness metrics based on any standard performance metric (e.g., the false positive rate or AUC) using :func:`make_derived_metric`."

docs/user_guide/assessment/perform_fairness_assessment.rst
```diff
@@ -199,6 +199,11 @@ These are accessed through the :attr:`MetricFrame.by_group` property:
 
 All of these values can be checked against the original arrays above.
 
+Note that :class:`MetricFrame` is intended for analyzing the disparities
```
hildeweerts, line 202: Shall we put this in a `.. note`? (Although I'm never sure whether those blocks are more noticeable or simply ignored by readers, haha.)

hildeweerts (commented):
Nice! This is already much clearer IMO.

Re-reading the entire text, it feels like "The centerpiece of fairness assessment in Fairlearn are ..." could be the start of a new subheader, moving from more abstract quantification of harms towards the actual implementation in Fairlearn. I wonder if we should restructure it to something like this:

```
- quantify harms
  (with an intro that already hints on the different types of fairness assessment in Fairlearn)
  - disaggregated analysis
  - compare quantified harms across groups
  - predefined fairness metrics
```

WDYT @alliesaizan?

I've also added a few smaller comments :)

docs/user_guide/assessment/perform_fairness_assessment.rst
alliesaizan: Done!

docs/user_guide/assessment/perform_fairness_assessment.rst
alliesaizan: I added this clarification, but let me know if we should remove it.

docs/user_guide/assessment/perform_fairness_assessment.rst
alliesaizan: Added the suggested sentence!

alliesaizan:
> Nice! This is already much clearer IMO.
> 
> Re-reading the entire text, it feels like "The centerpiece of fairness assessment in Fairlearn are ..." could be the start of a new subheader, moving from more abstract quantification of harms towards the actual implementation in Fairlearn. I wonder if we should restructure it to something like this:
> 
> ```
> - quantify harms
>   (with an intro that already hints on the different types of fairness assessment in Fairlearn)
>   - disaggregated analysis
>   - compare quantified harms across groups
>   - predefined fairness metrics
> ```
> 
> WDYT @alliesaizan?
> 
> I've also added a few smaller comments :)

Ah, so we'd just add a section called "Disaggregated Analysis"? That makes sense to me!

hildeweerts:
> Ah, so we'd just add a section called "Disaggregated Analysis"? That makes sense to me!

Exactly! And probably adjust the intro just a little bit so the reader understands the overall structure (in particular, disaggregated analysis versus fairness metrics).

alliesaizan:
> > Ah, so we'd just add a section called "Disaggregated Analysis"? That makes sense to me!
> 
> Exactly! And probably adjust the intro just a little bit so the reader understands the overall structure (in particular, disaggregated analysis versus fairness metrics).

I added the section and the paragraph below:

`Note that by "errors" here, we are referring to the methods we use to assess the performance of the machine learning model overall, for example accuracy or precision in the classification case. We distiniguish these model performance metrics from fairness metrics, which operationalize different definitions of fairness (such as demographic parity or equal opportunity). We will review those metrics in a subseqent section of the User Guide. For more information on fairness metrics, review :ref:`common_fairness_metrics`.`

alliesaizan:
Hi @hildeweerts, @romanlutz, @riedgar-ms, I have addressed most of the outstanding comments. Whenever you get a chance, can you please review and let me know if there are further requested changes?

riedgar-ms (approved):
OK with me, although I've only been dipping in and out of this.

docs/user_guide/assessment/perform_fairness_assessment.rst
```diff
@@ -118,6 +124,16 @@ in our dataset.
 If we found that males were experiencing errors at a much lower rate than
 females and nonbinary persons, we would flag this as a potential fairness harm.
 
+Note that by "errors" here, we are referring to the methods we use to 
+assess the performance of the machine learning model overall, for 
+example accuracy or precision in the classification case. 
+We distiniguish these model performance metrics from fairness metrics, 
+which operationalize different definitions of fairness 
+(such as demographic parity or equal opportunity).
+We will review those metrics in a subseqent section of the User Guide.
```
romanlutz, line 133:
```suggestion
We will review those metrics in a subsequent section of the User Guide. 
```
nit

romanlutz (approved):
Finally got a sec to review it and I'm very happy with it! Thanks for your patience with all our comments @alliesaizan ! Just one typo.

alliesaizan:
> Finally got a sec to review it and I'm very happy with it! Thanks for your patience with all our comments @alliesaizan ! Just one typo.

Committed! thank you :)

docs/user_guide/assessment/perform_fairness_assessment.rst
```diff
@@ -276,4 +299,79 @@ overall values for the data:
...
+:func:`equalized_odds_ratio` uses both the :func:`false_positive_rate` and
+:func:`false_negative_rate` within a :class:`MetricFrame` on the backend 
+to generate an output. As demonstrated below, 
+using :func:`equalized_odds_ratio` and :meth:`MetricFrame.ratio` method 
+produces the same outcome. 
+
+.. doctest:: assessment_metrics
+    :options:  +NORMALIZE_WHITESPACE
+
+    >>> from fairlearn.metrics import equalized_odds_ratio
```
romanlutz, line 357:
```suggestion
    >>> from fairlearn.metrics import equalized_odds_difference
```
This addresses one of the issues in the doctest pipeline.

docs/user_guide/assessment/perform_fairness_assessment.rst
```diff
@@ -276,4 +299,79 @@ overall values for the data:
...
+    ...     'fpr' : false_positive_rate
+    ... }
+    >>> mf = MetricFrame(
+    ...     metrics=my_metrics,
+    ...     y_true=y_true,
+    ...     y_pred=y_pred,
+    ...     sensitive_features=sf_data
+    ... )
+    >>> min(mf.ratio(method="between_groups"))
+    0
```
romanlutz, line 373:
```suggestion
    0.0
```
I believe this was the other issue I saw in the doctest pipeline. You can run that locally as well if you want to check, or just update the PR and check the doctest pipeline output (CTRL+F for "fail")

romanlutz (approved):
Just two suggestions to fix the doctest issues. Is it fair to assume that other build issues are handled by @riedgar-ms 's PRs?

riedgar-ms:
@hildeweerts are you OK with this now? If so, I think it's ready to merge

hildeweerts:
Seemed like we changed the wrong `equalized_odds_`. Let's see if this works.

docs/user_guide/assessment/perform_fairness_assessment.rst
```diff
@@ -276,4 +299,79 @@ overall values for the data:
...
+produces the same outcome. 
+
+.. doctest:: assessment_metrics
+    :options:  +NORMALIZE_WHITESPACE
+
+    >>> from fairlearn.metrics import equalized_odds_ratio
+    >>> print(equalized_odds_ratio(y_true,
+    ...                                y_pred,
+    ...                                sensitive_features=sf_data))
+    0
```
hildeweerts, line 361:
```suggestion
    0.0
```

hildeweerts:
Yay, doc build passed!

codecov fails - but that seems unlikely to be related to this PR? Tagging @riedgar-ms

riedgar-ms:
@hildeweerts Codecov has since passed. This is OK by me to merge.

alliesaizan:
Yay, thank you! I will close the related issue.

PR #1287, opened 2023-09-22
Fix fetch_openml

riedgar-ms:
## Description

It appears that `fetch_openml()` no longer likes `pathlib.Path` objects. I have filed [an issue on SciKit-Learn](https://github.com/scikit-learn/scikit-learn/issues/27447) about this, but in the meantime.....

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

### DIFF
--- a/fairlearn/datasets/_fetch_adult.py
+++ b/fairlearn/datasets/_fetch_adult.py
@@ -5,8 +5,6 @@ import pathlib
 
 from sklearn.datasets import fetch_openml
 
-import fairlearn.utils._compatibility as compat
-
 from ._constants import _DOWNLOAD_DIRECTORY_NAME
 
 
@@ -103,9 +101,9 @@ def fetch_adult(*, cache=True, data_home=None, as_frame=True, return_X_y=False):
 
     return fetch_openml(
         data_id=1590,
-        data_home=data_home,
+        data_home=str(data_home),
         cache=cache,
         as_frame=as_frame,
         return_X_y=return_X_y,
-        **compat._PARSER_KWARG,
+        parser="auto",
     )


requirements.txt
```diff
@@ -1,4 +1,4 @@
 numpy>=1.18.0
 pandas>=0.25.2
-scikit-learn>=0.22.1
+scikit-learn>=1.0.0
```
riedgar-ms, line 3: Thoughts @adrinjalali ? This is related to #1288

adrinjalali: oh yeah, I'd even do 1.1, we're going to release 1.4.

fairlearn/datasets/_fetch_acs_income.py
```diff
@@ -7,8 +7,6 @@
 import pandas as pd
 from sklearn.datasets import fetch_openml
 
-import fairlearn.utils._compatibility as compat
```
adrinjalali, line 10: should we then remove `_PARSER_KWARGS`?

riedgar-ms: Probably.... comments there suggest it was a v1.2 thing, if you're happing bumping up that far.

riedgar-ms: @adrinjalali I've bumped to 1.2 below.... if you'd prefer to stick to 1.1, I'll need to change how compatibility is handled

requirements.txt
```diff
@@ -1,4 +1,4 @@
 numpy>=1.18.0
 pandas>=0.25.2
-scikit-learn>=0.22.1
+scikit-learn>=1.1.0
```
adrinjalali, line 3: have you? O_o

adrinjalali (commented):
I wouldn't mind bumping to 1.2, but you haven't here 😁

requirements.txt
riedgar-ms: I have now :-)

adrinjalali (approved):

PR #1294, opened 2023-09-26
MNT Further pandas fixing

riedgar-ms:
## Description

Further small tweaks to address an onslaught of warnings from `pandas`. Fixes are in both the tests and in the product code.

## Tests

- [ ] no new tests required
- [ ] new tests added
- [x] existing tests adjusted

## Documentation

- [x] no documentation changes needed
- [ ] user guide added or updated
- [ ] API docs added or updated
- [ ] example notebook added or updated

PR #1295, opened 2023-09-26
MNT Fixing documentation

riedgar-ms:
## Description

Applying fixes for the documentation:

- In `plot_new_metrics.py` it appears that `fetch_adult()` is now returning something slightly different.... or pandas `map()` function is now including missing data

Note that this does not address the bounty of `pandas` warnings we're getting right now. Those warning don't kill this build (and should be addressed by #1294 )

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

### DIFF
--- a/examples/plot_new_metrics.py
+++ b/examples/plot_new_metrics.py
@@ -109,11 +109,9 @@ def marriage_transform(m_s_string):
 def occupation_transform(occ_string):
     """Perform some simple manipulations."""
     result = "Small"
-    if isinstance(occ_string, float):
-        # Had missing data
-        assert np.isnan(occ_string)
-        result = "Low"
-    elif occ_string.startswith("Machine"):
+    # The isinstance check is to guard against 'missing'
+    # data marked with NaN
+    if not isinstance(occ_string, float) and occ_string.startswith("Machine"):
         result = "Large"
     return result
 
--- a/examples/plot_new_metrics.py
+++ b/examples/plot_new_metrics.py
@@ -109,7 +109,11 @@ def marriage_transform(m_s_string):
 def occupation_transform(occ_string):
     """Perform some simple manipulations."""
     result = "Small"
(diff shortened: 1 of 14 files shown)


examples/plot_new_metrics.py
```diff
@@ -109,7 +109,11 @@ def marriage_transform(m_s_string):
 def occupation_transform(occ_string):
     """Perform some simple manipulations."""
     result = "Small"
-    if occ_string.startswith("Machine"):
+    if isinstance(occ_string, float):
+        # Had missing data
+        assert np.isnan(occ_string)
+        result = "Low"
```
hildeweerts, line 115: Should this say "Small"?

hildeweerts (commented):
Could this be related to https://github.com/scikit-learn/scikit-learn/pull/26551?

examples/plot_new_metrics.py
riedgar-ms: Yes. I must have read the wrong `fillna()` value below

riedgar-ms: I'll update the logic to correct this.

riedgar-ms:
> Could this be related to [scikit-learn/scikit-learn#26551](https://github.com/scikit-learn/scikit-learn/pull/26551)?

Quite possibly. Alternatively, `DataFrame.map()` may have changed its behaviour, and now always passes 'missing' data to the underlying function.

hildeweerts (approved):
> Alternatively, DataFrame.map() may have changed its behaviour, and now always passes 'missing' data to the underlying function.

tbh this would surprise me, given that `map()` has an explicit argument to control how to deal with missing data and the default value has not changed since previous versions. In any case, changes are approved 👍

riedgar-ms:
I missed that @hildeweerts . So we definitely blame SciKit-Learn, then :-)

PR #1298, opened 2023-09-28
MNT Python 3.11 & LightGBM Update

riedgar-ms:
## Description

Updating the LightGBM testing in `test_otherml`. Since something is preventing the latest `pandas` being pulled, we have to undo some of the changes made recently. However, due to differences between `pandas` 2.0 and 2.1, we also need to suppress some warnings.

Also adds Python 3.11 into the test matrix.

## Tests

- [ ] no new tests required
- [ ] new tests added
- [x] existing tests adjusted

## Documentation

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python: ["3.8", "3.9", "3.10"]
+        python: ["3.8", "3.9", "3.10", "3.11"]
         requirementsPinned: ["false"]
     uses: ./.github/workflows/test-all-deps.yml
     with:
@@ -34,7 +34,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python: ["3.8", "3.9", "3.10"]
+        python: ["3.8", "3.9", "3.10", "3.11"]
     uses: ./.github/workflows/test-minimal-deps.yml
     with:
       os: ${{ matrix.os }}
@@ -45,7 +45,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python: ["3.8"]
+        python: ["3.11"]
         mlpackage: [lightgbm, xgboost, tensorflow, pytorch]
     uses: ./.github/workflows/test-other-ml.yml
     with:
--- a/fairlearn/metrics/_disaggregated_result.py
+++ b/fairlearn/metrics/_disaggregated_result.py
@@ -172,7 +172,7 @@ class DisaggregatedResult:
             elif errors == "coerce":
                 # Fill all impossible columns with NaN before grouping metric frame
                 mf = self.by_group.copy()
-                mf = mf.map(lambda x: x if np.isscalar(x) else np.nan)
+                mf = mf.applymap(lambda x: x if np.isscalar(x) else np.nan)
                 if grouping_function == "min":
                     result = mf.groupby(level=control_feature_names).min()
                 else:
@@ -232,7 +232,7 @@ class DisaggregatedResult:
         mf = self.by_group.copy()
         # Can assume errors='coerce', else error would already have been raised in .group_min
         # Fill all non-scalar values with NaN
-        mf = mf.map(lambda x: x if np.isscalar(x) else np.nan)
+        mf = mf.applymap(lambda x: x if np.isscalar(x) else np.nan)
 
         if control_feature_names is None:
             result = (mf - subtrahend).abs().max()
--- a/fairlearn/metrics/_metric_frame.py
+++ b/fairlearn/metrics/_metric_frame.py
@@ -362,7 +362,10 @@ class MetricFrame:
         self, target: Union[pd.Series, pd.DataFrame]
     ) -> Union[pd.Series, pd.DataFrame]:
         """Convert Nones to NaNs."""
-        result = target.map(lambda x: x if x is not None else np.nan)
+        if isinstance(target, pd.Series):
+            result = target.map(lambda x: x if x is not None else np.nan)
+        else:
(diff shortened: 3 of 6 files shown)


pyproject.toml
```diff
@@ -20,6 +20,7 @@ filterwarnings = [
     "default:You are about to use a dataset with known fairness issues:Warning",
     "default:Auto-removal of overlapping axes is deprecated:Warning",
     "default:DataFrame is highly fragmented:Warning",
+    "default:DataFrame.applymap has been deprecated. Use DataFrame.map instead.:Warning",
```
riedgar-ms, line 23: I have filed #1299 about this

.github/workflows/build-test.yml
```diff
@@ -45,7 +45,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest]
-        python: ["3.8"]
+        python: ["3.11"]
```
riedgar-ms, line 48: I don't believe this is relevant to anything since each creates its own conda environment. However, I'm bumping it on general principles.

test/unit/metrics/test_metricframe_deprecate.py
```diff
@@ -28,7 +28,7 @@ def test_all_positional_arguments():
             skm.accuracy_score, y_true, y_pred, sensitive_features=sf
         )
 
-    assert len(record) == 1
+    # assert len(record) == 1
```
riedgar-ms, line 31: These lengths are troublesome. When `pandas` v2.0 is installed, they are correct (i.e. on Python 3.8). On later Pythons, we might have `pandas` v2.1, and then we will also get warnings about `DataFrame.applymap()` being deprecated.

Since we're going to be removing our own deprecated code soon, I'm taking this simple solution.

.github/workflows/build-test.yml
```diff
@@ -22,7 +22,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python: ["3.8", "3.9", "3.10"]
+        python: ["3.8", "3.9", "3.10", "3.11"]
```
adrinjalali, line 25: why are we not removing `3.8` then?

fairlearn/metrics/_disaggregated_result.py
```diff
@@ -172,7 +172,7 @@ def apply_grouping(
             elif errors == "coerce":
                 # Fill all impossible columns with NaN before grouping metric frame
                 mf = self.by_group.copy()
-                mf = mf.map(lambda x: x if np.isscalar(x) else np.nan)
+                mf = mf.applymap(lambda x: x if np.isscalar(x) else np.nan)
```
adrinjalali, line 175: we're not testing these?

fairlearn/metrics/_metric_frame.py
```diff
@@ -362,7 +362,10 @@ def _none_to_nan(
         self, target: Union[pd.Series, pd.DataFrame]
     ) -> Union[pd.Series, pd.DataFrame]:
         """Convert Nones to NaNs."""
-        result = target.map(lambda x: x if x is not None else np.nan)
+        if isinstance(target, pd.Series):
+            result = target.map(lambda x: x if x is not None else np.nan)
+        else:
+            result = target.applymap(lambda x: x if x is not None else np.nan)
```
adrinjalali, line 368: also worrying that this is not tested

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [tool.black]
 line-length = 88
-target_version = ['py37', 'py38', 'py39', 'py310', 'py311']
+target_version = ['py38', 'py39', 'py310', 'py311']
```
adrinjalali, line 3: remove 3.8?

test/unit/metrics/test_metricframe_deprecate.py
adrinjalali, line 31: I don't understand, why is map vs applymap making this length change?

.github/workflows/build-test.yml
riedgar-ms: We talked about this in the call yesterday (sorry you couldn't make it). My motivation at least was to get pandas to v2.1. However, we do need the otherml tests (they provide the 'real' tests of `adversarial`) and those, somewhere, have a v2.0 pin.

Since Python 3.8 does still have another year of support, I'm not in a rush to drop it from Fairlearn.

test/unit/metrics/test_metricframe_deprecate.py
riedgar-ms: Because when Pandas v2.1 is available, we get extra warnings about `map()` vs `applymap()`

fairlearn/metrics/_disaggregated_result.py
riedgar-ms: According to CodeCov we are:
https://app.codecov.io/gh/fairlearn/fairlearn/pull/1298?src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=fairlearn

test/unit/metrics/test_metricframe_deprecate.py
adrinjalali: I don't understand how that has anything to do with removing this assert.

.github/workflows/build-test.yml
adrinjalali: By the time we do the next release, we'll have python 3.12 there. Do we want to support 5 different python versions?

test/unit/metrics/test_metricframe_deprecate.py
riedgar-ms: Because this `assert` is checking how many warnings occurred during the construction of the `MetricFrame`

```python
 with pytest.warns(FutureWarning) as record:
```

riedgar-ms:
Have come up with an alternative approach in Community Call

Issue #1299, opened 2023-09-29
MNT Deal with pandas 2.0 vs 2.1


riedgar-ms:
In `pandas` v2.1, a new `DataFrame.map()` API was added, while the previous `DataFrame.applymap()` was simultaneously deprecated. Since our builds fail on warnings, this causes issues if the latest `pandas` cannot be install (notably for Python 3.8, but also some of our 'otherml_packages' tests).

PR #1302  squares this particular circle by rewriting the handful of `DataFrame.applymap()` calls to being nested `DataFrame.apply()` ones. This is not going to be particularly performant, but the DataFrames in question should be relatively small.

This issue is to track the nested 'apply' calls, so they can be improved once we drop `pandas` v2.0.

*Note* This has been edited, so some of the discussion below may be obsolete.

adrinjalali:
The issue with not handling this now, is that downstream libraries are getting the `FutureWarning` and their CI fails due to that (I work on one of those libraries)

riedgar-ms:
Sorry.... `FutureWarning` from what? pandas?

adrinjalali:
Yes, the warning is propagated and it's making my CI fail in `skops`.

adrinjalali:
So when would people stop getting the future warnings coming from pandas? Are you saying we'd fix those warnings somehow in the next release, or the one after? If the one after, it'd mean people would get those warnings for quite a long time from our side.

riedgar-ms:
And for extra fun, pandas 2.2 is now out and wreakings its havoc. Some fixes in #1351

TamaraAtanasoska:
Closing because of #1522

PR #1302, opened 2023-10-12
MNT Fixing pandas

riedgar-ms:
## Description

The recent `pandas` 2.1 release has simultaneously introduced a new `DataFrame.map()` API, and deprecated the `DataFrame.applymap()` API which we use in `MetricFrame`. Since our builds fail on warnings, this left us with a conundrum. This PR (temporarily) solves this by changing the `DataFrame.applymap()` calls to a nested `DataFrame.apply()` pattern. This is not particularly great, but this should not be being used in performance-critical portions of `MetricFrame` and can suffice until we decide to require `pandas` be at least version 2.1.

Also make sure that Python 3.7 is fully excised and add 3.11 to the regular builds and setup.

Issue #1299 exists to track the need to tidy all of this up once we drop `pandas` v2.0 support.

## Tests

- [ ] no new tests required
- [ ] new tests added
- [x] existing tests adjusted

## Documentation

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python: ["3.8", "3.9", "3.10"]
+        python: ["3.8", "3.9", "3.10", "3.11"]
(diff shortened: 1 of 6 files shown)


romanlutz (approved):
Thank you!

PR #1304, opened 2023-10-17
MNT Further build fixes

riedgar-ms:
## Description

Taking another swing at getting our builds to Green. There are two changes:

- I saw errors from CodeCov about 'too many uploads for commit.' Since we don't have different code paths based on Python version, I've tweaked the CodeCov step so that it only runs for Python 3.11
- The `test_otherml` directory was still getting warnings. Since those dependencies are not entirely under our control, my solution is to override the `pyproject.toml` and "Treat Warnings as Warnings." I know that this is not ideal, but short of unpicking the exact combination of packages which get installed, it's probably the most robust solution

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -29,6 +29,7 @@ jobs:
       os: ${{ matrix.os }}
       python: ${{ matrix.python }}
       requirementsPinned: ${{ matrix.requirementsPinned }}
+      codeCovPython: "3.11"
   
   minimal-deps:
     strategy:
--- a/.github/workflows/test-all-deps.yml
+++ b/.github/workflows/test-all-deps.yml
@@ -34,8 +34,8 @@ jobs:
       - run: pip install matplotlib
       - run: pytest --ignore=test/install --cov=fairlearn --cov-report=xml test/
       - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v3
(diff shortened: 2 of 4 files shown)


riedgar-ms:
If @fairlearn/fairlearn-maintainers think it would be helpful, I'm happy to file an issue to track the `filterwarnings` override for the 'otherml' tests.

romanlutz (approved):
I agree that treating warnings as warnings isn't ideal, but as long as we have them surfaced as errors in the "main" all-deps workflow I'm fine with it. It looks like you have left that unchanged so I'm in favor. 

For codecov, can we just... display the result and not upload it? What's the benefit of uploading?

riedgar-ms:
Codecov gives us pretty coverage diffs, and also fails the build if we let coverage drop. Neither is essential, but both are nice to have. Have I missed any other benefits @adrinjalali ?

adrinjalali:
> For codecov, can we just... display the result and not upload it? What's the benefit of uploading?

It adds a step to the CI when it fails, So it's nice to see and notice when it fails. I personally heavily rely on it when reviewing code.

adrinjalali (approved):
I don't mind this. Rather to strip down the other ml tests to absolute essentials though.

Issue #1182, opened 2023-01-17
DOC Replace adult/census dataset in "Get Started"


romanlutz:
#### Describe the issue linked to the documentation

We're currently using the adult/census dataset in "Get Started" on the website. As we have agreed on many occasions it's not a particularly good example of a useful application context. This should be replaced with a different dataset.

#### Suggest a potential alternative/fix

Some alternatives that @MiroDudik proposed recently (on the community call and #1177 ):
- the diabetes dataset from https://github.com/fairlearn/talks/tree/main/2021_scipy_tutorial (I think it's now part of fairlearn.datasets, but for the tutorial they added something synthetic) - it needs a lot of context to be explained which makes it less suitable for a quickstart where we can't spend much time (I think the page says 10 minutes), BUT we could just point to the tutorial for a holistic view.
- generate something entirely synthetic with a fairly obvious context. @MiroDudik's idea was this: rolling out a new version of a website happens slowly to avoid having everyone experience downtime if something goes wrong. So a model can select who sees the new version (1) or the old version (0) and we have some characteristics like perhaps location (country?) or browser or such and want to achieve equal risk distribution across groups (risk of having a bad QoS). Thinking about this a bit more, I'm wondering what features the original model without mitigation would use to make its prediction.

MiroDudik:
Some clarifications re scipy tutorial dataset:
* We didn't need to add anything synthetic
* I also don't think that the context would take too long to explain--I think that we could describe things in a paragraph--building on what's already in [scipy tutorial](https://github.com/fairlearn/talks/blob/main/2021_scipy_tutorial/fairness-in-AI-systems-instructors.ipynb), specifically sections: **Dataset and task** and **Measuring fairness-related harms** (just the sentences related to health care scenario).
* As Roman says, we would point to the tutorial for more context / definitions
* One downside with this dataset is that the disparity is between the race group labeled *Unknown* and the other groups; this is a statistically significant disparity, but not something that is brought up in the literature

All of that said, I would be in favor of using that dataset over what we have right now. (Synthetic datasets are a can of worms that would just block this issue...).

shimst3r:
I wouldn’t mind taking care of either. 🤓

romanlutz:
@shimst3r it looks like #1185 is sort of a prerequisite for #1182 and would have to be done first.

Thanks @MiroDudik for the clarifications 🙂

shimst3r:
[...]

PR #1310, opened 2023-10-20
DOC Changing the dataset from Adult to Diabetes Hospital in Getting Started

alliesaizan:
## Description
This PR shifts the dataset used in Getting Started from Adult to Diabetes Hospital (corresponds to issue #1182). All the plots are updated, but save for the dataset swap, most of the evaluation code has not been changed. Instead of gender, the sensitive variable analyzed in this tutorial is race, same as in the SciPy tutorial.

After the Fairlearn community call on 10/19/2023, I ended up going with accuracy as the first evaluation metric (keeping the other evaluation metrics throughout the same) and instead set a lower threshold for predicted probability to trigger a positive class label (0.1, to roughly match the proportion of positive labels in the dataset). 

One issue I'd like to point out is that the label imbalance makes the results for the Mitigation section somewhat unclear. Because Exponentiated Gradient only has the capability to return predicted labels, and not return predicted probabilities, I do not believe I can change the threshold that triggers a positive label. Only around 11% of records have a positive label, so the model never predicts positive outcomes. I'm thinking this might be confusing for newcomers, but I'd like to hear your thoughts.

I am tagging the maintainers that have been involved with this issue, @romanlutz and @MiroDudik.

## Tests
- [X] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation
- [X] example notebook added or updated

## Screenshots
image


docs/quickstart.rst
```diff
@@ -117,35 +135,52 @@ we can evaluate metrics for subgroups within the data as below:
...
+    Caucasian          0.520261
+    Hispanic           0.595528
+    Other              0.561798
+    Unknown            0.591525
     Name: accuracy_score, dtype: float64
 
-Additionally, Fairlearn has lots of other standard metrics built-in, such as
+Note that above, because the outcome variable exhibits label imbalance, we set 
+the threshold for triggering a positive prediction lower 
+(to a probability of 0.1 or greater).
```
romanlutz, line 163: I like that you made this adjustment for imbalance but the number 0.1 kind of falls from the sky...

Did you try a few and decide on that or where is it from? 

I checked if DTC has anything for changing the objective, but all I found was the "balanced" class_weight arg https://scikit-learn.org/stable/modules/generated/sklearn.tree.DecisionTreeClassifier.html
I've never used it but perhaps it could help here?

Adding other @fairlearn/fairlearn-maintainers to chime in if they have ideas.

docs/quickstart.rst
```diff
@@ -77,35 +78,52 @@ In order to run the code samples in the Quickstart tutorial, you need to install
...
-    >>> X = pd.get_dummies(data.data)
-    >>> y_true = (data.target == '>50K') * 1
-    >>> sex = data.data['sex']
-    >>> sex.value_counts()
-    sex
-    Male      32650
-    Female    16192
-    Name: count, dtype: int64
+    >>> data = fetch_diabetes_hospital(as_frame=True)
+    >>> keep_idx = (data.data['gender'] == "Male") | (data.data['gender'] == "Female")
```
romanlutz, line 102: If we're not evaluating by gender we can probably drop this part.

docs/quickstart.rst
```diff
@@ -117,35 +135,52 @@ we can evaluate metrics for subgroups within the data as below:
...
 .. doctest:: quickstart
     :options:  +NORMALIZE_WHITESPACE
 
     >>> from fairlearn.metrics import selection_rate
-    >>> sr = MetricFrame(metrics=selection_rate, y_true=y_true, y_pred=y_pred, sensitive_features=sex)
+    >>> sr = MetricFrame(metrics=selection_rate, y_true=y_test, y_pred=y_pred, \
+    sensitive_features=A_test)
     >>> sr.overall
-    0.1638...
+    0.5124012420895405
```
romanlutz, line 175: Strongly suggest ellipsis as before because it helps doctest succeed when there are minute differences in the 10th digit etc.

This applies to any number with more than 3 beyond the comma

docs/quickstart.rst
```diff
@@ -117,35 +135,52 @@ we can evaluate metrics for subgroups within the data as below:
...
+Fairlearn has many standard metrics built-in, such as
 selection rate, i.e., the percentage of the population which have '1' as
 their label:
 
 .. doctest:: quickstart
     :options:  +NORMALIZE_WHITESPACE
 
     >>> from fairlearn.metrics import selection_rate
-    >>> sr = MetricFrame(metrics=selection_rate, y_true=y_true, y_pred=y_pred, sensitive_features=sex)
+    >>> sr = MetricFrame(metrics=selection_rate, y_true=y_test, y_pred=y_pred, \
```
romanlutz, line 172: Selection rate is probably not the best criterion here. IMO (and thereby up for discussion) FNR is crucial, or perhaps the entire confusion matrix. FNR captures people that would get readmitted but we don't predict them to be, so that's real (medical) harm... 

FPs are the ones that wouldn't have been readmitted but we kept them in the hospital longer than needed so it's a cost but no harm to the person's health presumably.

So I suggest going with FNR/TPR parity or equalized odds (which captures the whole confusion matrix) as fairness criterion for mitigation with balanced accuracy as objective.

docs/quickstart.rst
```diff
@@ -184,18 +219,27 @@ a vastly reduced difference in selection rate:
     >>> constraint = DemographicParity()
     >>> classifier = DecisionTreeClassifier(min_samples_leaf=10, max_depth=4)
     >>> mitigator = ExponentiatedGradient(classifier, constraint)
```
romanlutz, line 221: objective="balanced_accuracy_score"

romanlutz (commented):
It's so good to see this changing to something more meaningful! Thanks @alliesaizan

docs/quickstart.rst
romanlutz: Sorry, that only works for Postprocessing. I don't believe we have a Moment class for balanced accuracy yet, so we @MiroDudik ?

Should be easy to add though... I'll look into it later today.

docs/quickstart.rst
alliesaizan: Yeah, I understand! The proportion of positive labels overall was about 11% so I rounded down, but happy to change it to whatever everyone deems would be a better fit.

docs/quickstart.rst
alliesaizan: Done!

docs/quickstart.rst
alliesaizan: Done!

docs/quickstart.rst
romanlutz: The `ErrorRate` moment with costs of 0.5 for FP and FN each seems promising. @MiroDudik can you confirm?

docs/quickstart.rst
alliesaizan: I updated the portion that calculates the selection rate to instead show the false negative rate. I also added a few sentences to provide context to the example. 

I also updated the mitigation section to have the objective of balanced accuracy with equalized odds as the fairness criterion (unless you would prefer to use error rate as the criterion). I do want to quickly note that in the community call we discussed using balanced accuracy as the evaluation criteria in the initial section of the quickstart guide and decided to use accuracy instead with a lower threshold on the prediction probabilities because [to my understanding] balanced accuracy does not differentiate results by subgroups. But let me know if you want me to mock up what unmitigated balanced accuracy would look like, and we can discuss further here.

docs/quickstart.rst
alliesaizan: Hi @romanlutz and @MiroDudik, any updates on this question?

docs/quickstart.rst
romanlutz: True. Balanced is balanced by 0/1

docs/quickstart.rst
alliesaizan: Can you confirm that ErrorRate is the right moment to use? When I try to use it with Exponentiated Gradient, I am getting `AttributeError: 'ErrorRate' object has no attribute 'default_objective'`

romanlutz: Did you keep the constraint as is and add another arg `objective=ErrorRate(costs={'fp': 0.5, 'fn': 0.5})`? I believe `ErrorRate` wouldn't work as a `constraint` since it doesn't have `default_objective`. 

I suspect you used `ErrorRate` as the `constraint` instead of `DemographicParity`, is that right?

alliesaizan: > Did you keep the constraint as is and add another arg `objective=ErrorRate(costs={'fp': 0.5, 'fn': 0.5})`? I believe `ErrorRate` wouldn't work as a `constraint` since it doesn't have `default_objective`.
> 
> I suspect you used `ErrorRate` as the `constraint` instead of `DemographicParity`, is that right?

Gotcha! I pushed the update.

romanlutz: It still uses DP (ie selection rate parity), and not Equalized Odds (FPR/TPR). As I noted in a different comment earlier that is a problem IMO

alliesaizan: My apologies, it's been a while since I looked at this file so I walked back too many of the changes. I have changed the constraint back to Equalized Odds.

romanlutz: No worries at all! Ah, the numbers look much more interesting now, too 😄 Before, it was all 0s.

docs/quickstart.rst
romanlutz: Is it possible to just mention that it's set based on the percentage of positive labels in the dataset? That way, it doesn't just occur out of nowhere 🙂

alliesaizan: Done! Added a sentence: "We chose this threshold to match the percentage of positive labels in the dataset."

alliesaizan:
Hi @romanlutz, just wanted to ping because all the comments have been addressed. Would you be able to do the approving review?

docs/quickstart.rst
```diff
@@ -167,35 +205,44 @@ Mitigating disparity
...
+    >>> from fairlearn.reductions import ErrorRateParity, EqualizedOdds, ExponentiatedGradient
     >>> np.random.seed(0)  # set seed for consistent results with ExponentiatedGradient
-    >>>
-    >>> constraint = DemographicParity()
+    >>> objective = ErrorRate(costs={'fp': 0.5, 'fn': 0.5})
+    >>> constraint = EqualizedOdds()
     >>> classifier = DecisionTreeClassifier(min_samples_leaf=10, max_depth=4)
-    >>> mitigator = ExponentiatedGradient(classifier, constraint)
-    >>> mitigator.fit(X, y_true, sensitive_features=sex)
+    >>> mitigator = ExponentiatedGradient(classifier, constraint, objective = objective)
```
romanlutz, line 224:
```suggestion
    >>> mitigator = ExponentiatedGradient(classifier, constraint, objective=objective)
```
I know, very nitpicky of me...

docs/quickstart.rst
```diff
@@ -167,35 +205,44 @@ Mitigating disparity
...
 
 .. doctest:: quickstart
     :options:  +NORMALIZE_WHITESPACE
 
-    >>> from fairlearn.reductions import DemographicParity, ExponentiatedGradient
+    >>> from fairlearn.reductions import ErrorRateParity, EqualizedOdds, ExponentiatedGradient
     >>> np.random.seed(0)  # set seed for consistent results with ExponentiatedGradient
-    >>>
-    >>> constraint = DemographicParity()
+    >>> objective = ErrorRate(costs={'fp': 0.5, 'fn': 0.5})
```
romanlutz, line 221: I'd love to have someone other than me cross-checking that this makes sense to them.

romanlutz (approved):
This looks great to me. Left one nitpicky comment and one that others should cross-check.

docs/quickstart.rst
```diff
@@ -156,7 +194,7 @@ Fairlearn also allows us to quickly plot these metrics from the
     :start-after: # Analyze metrics using MetricFrame
```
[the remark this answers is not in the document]
alliesaizan: Yes, I did see that issue. The metrics are imported in plot_quickstart.py, and that should carry through when the file is used in the tutorial. Additionally, in the new tutorial, I don’t believe it uses all the metrics used in the old tutorial. I am not in a place where I can check the documentation in CircleCI now, but I will check it asap and make sure there are no errors.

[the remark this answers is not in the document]
eroell: cool, happy coding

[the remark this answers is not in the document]
alliesaizan: @eroell to address the issue you raised in #1355, the quickstart is not actually pulling plot_quickstart.py from the examples folder - it's pulling from a hidden folder. See below: 
image
Within the plot_quickstart.py in auto_examples, it is importing all the metrics: 
image

In the quickstart in my file, for some reason, the snippet you see is clipped to not show the imports. 

image

However, your question bring to my attention that the images I changed within auto_examples are not getting updated. I'll work on that today and try to get the images pushed up.

MiroDudik (commented):
I've added various suggestions -- I hope they make sense.

docs/quickstart.rst
alliesaizan: Miro suggested `objective = ErrorRate(costs={'fp': 0.1, 'fn': 0.9})`, so let me know if that works for you and I can resolve this comment.

alliesaizan:
@MiroDudik  for some reason, these comments did not load for me until today. My apologies for the delay. In the section you cited, I have removed the sentence "We selected this threshold to match the probability of readmission.". 

I have added the following context:

"In practice, this threshold would be driven by risk or capacity considerations. For this example, we set the threshold based on the risk of readmission. The threshold of 0.1 corresponds to saying that a 10% risk of readmission is viewed as sufficient for referral to a 
post-discharge care program. "

Please let me know if this is along the lines of what you are thinking!

MiroDudik:
@alliesaizan -- no worries, i actually started them when we last spoke, but only finished today, so it's on me. your update sounds good.

alliesaizan:
@MiroDudik regarding the plotting discussion from earlier today, I changed the plots in the examples folder to show the diabetes hospital dataset, and the selection_rate plot now just shows value counts across race (though I did not rename that file)

alliesaizan:
Hi all, after building the docs, the CircleCI doctest keeps failing. It expects different results even when the seed is set and the random state is set in the decision tree classifier. I don't recall how I have solved these issues in the past, but would welcome any input from the other @fairlearn/fairlearn-maintainers

riedgar-ms:
@alliesaizan Have you resynced your branch with `main` recently? I did push a few changes which tried to make CircleCI more reliable recently (specifically #1354).

riedgar-ms:
Although, looking at the output of CircleCI:
```
File "quickstart.rst", line 146, in quickstart
Failed example:
    mf.by_group
Expected:
    Caucasian          76099
    AfricanAmerican    19210
    Unknown             2273
    Hispanic            2037
    Other               1506
    Asian                641
    Name: count, dtype: int64
Got:
    race
    AfricanAmerican    0.530935
    Asian              0.658683
    Caucasian          0.503535
    Hispanic           0.612524
    Other              0.591549
    Unknown            0.574576
    Name: accuracy_score, dtype: float64
```
That looks like some completely different metric is being computed.

```
Expected:
    0.526...
Got:
    0.5257448313811808
```
That could be a misunderstanding of what `...` is doing. The match is "text" not "numeric" so the difference in the third decimal place matters, even though the rounding is correct.

docs/quickstart.rst
```diff
@@ -115,38 +133,59 @@ we can evaluate metrics for subgroups within the data as below:
...
+    Other              0.591549
+    Unknown            0.574576
     Name: accuracy_score, dtype: float64
 
-Additionally, Fairlearn has lots of other standard metrics built-in, such as
-selection rate, i.e., the percentage of the population which have '1' as
-their label:
+Note that above, because the outcome variable exhibits label imbalance, we set 
+the threshold for triggering a positive prediction lower 
+(to a probability of 0.1 or greater).
```
MiroDudik, line 160:
```suggestion
Note that our decision threshold for positive predictions is 0.1.
```

docs/quickstart.rst
```diff
@@ -156,7 +195,7 @@ Fairlearn also allows us to quickly plot these metrics from the
     :start-after: # Analyze metrics using MetricFrame
     :end-before: # Customize plots with ylim
 
-.. figure:: auto_examples/images/sphx_glr_plot_quickstart_001.png
+.. figure:: auto_examples/images/sphx_glr_plot_quickstart_diabetes_001.png
```
MiroDudik, line 198: I don't see anything rendered. I think it should be just `...plot_quickstart_001.png`?

examples/plot_quickstart_selection_rate.py
```diff
@@ -6,20 +6,23 @@
 Selection rates in census dataset
```
MiroDudik, line 6: The filename and the title of the notebook no longer match the content. Either rename the file and change title, or possibly just include the generation of this image in the other notebook.

Also--the exported image gets chopped for me, so it looks like this:
image

alliesaizan: Is it possible that the directory is incorrect? Right now, it's looking in auto_examples, but the images are in build/html/_images

MiroDudik: you might be right. do you want to try to change it?

MiroDudik: i'm not sure why it's not a problem with the current quickstart though... we'll see once it's merged.

examples/plot_quickstart_counts.py
```diff
@@ -0,0 +1,29 @@
+# Copyright (c) Microsoft Corporation and Fairlearn contributors.
+# Licensed under the MIT License.
+
+"""
+=================================
+Selection rates in census dataset
```
MiroDudik, line 6: remember to update this line.

alliesaizan: Done

examples/plot_quickstart_selection_rate.py
alliesaizan: > you might be right. do you want to try to change it?

I can try to pass the examples directory instead - would be more straightforward to update the images there. But yeah, I am also not sure why it's not updating when we push updates. It looks like the auto-examples folder is hidden and untracked by GitHub, so there may be something I am missing with my repository setup on my local machine.

romanlutz: auto_examples is generated from examples, so that's why it's not checked into the repo. It's meant to show up when you run the doc build.

docs/quickstart.rst
alliesaizan: After changing the path to _images/ as we discussed on the call today, the CircleCI artifacts show all the plots rendering.

MiroDudik (approved):

PR #1331, opened 2024-01-13
MetricFrame: docstring update for usage with selection_metric

amitsaha:
## Description

I am aware that the value itself of `y_true` doesn't matter for `selection_rate`. However, as a beginner it confused me when I was going through the examples for `MetricFrame` and the value of `y_true` not matching with the sample data in `sex`.

Hope the change makes sense, worst case my (mis)understanding will be clarified, thanks.

## Tests
- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

## Screenshots

### DIFF
--- a/fairlearn/metrics/_metric_frame.py
+++ b/fairlearn/metrics/_metric_frame.py
@@ -154,7 +154,7 @@ class MetricFrame:
     >>> from fairlearn.metrics import MetricFrame, selection_rate
     >>> from sklearn.metrics import accuracy_score
     >>> import pandas as pd
-    >>> y_true = [1,1,1,1,1,0,0,1,1,0]
+    >>> y_true = [1,1,1,1,1,0,0,0,0,0]
     >>> y_pred = [0,1,1,1,1,0,0,0,1,1]
     >>> sex = ['Female']*5 + ['Male']*5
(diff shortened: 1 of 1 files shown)


romanlutz:
Hi @amitsaha . The true labels have nothing to do with the sensitive features so this is quite intentional. It just so happens that this metric doesn't require the true labels, but for consistency reasons we chose to keep it as an argument. It makes other things easier when you have a consistent API, but I'll spare you the details.

Does that make sense? I'm curious why you thought sensitive features and true labels are identical.

amitsaha:
hi @romanlutz thank you. yes that makes complete sense.

> I'm curious why you thought sensitive features and true labels are identical.

Just my beginner mind, really. I created the PR before i kept digging into the selection rate metric and how it's the sensitive feature that's relevant to it and that the `y_true` has no implication for it.

I guess thinking back, i somehow managed to think to myself that the `sex` was what we were trying to predict - 🧠 🥶 

Closing the PR!

romanlutz:
No worries at all @amitsaha.
If you feel like contributing we're always happy to help you find a task.

amitsaha:
thanks @romanlutz i have been looking at the beginner level issues.

Issue #1332, opened 2024-01-16
Adversarial Mitigation Error


joelthe1:
#### Describe the bug
Unable to run Adversarial Mitigation with a sensitive features vector.

#### Steps/Code to Reproduce
See GitHub Gist: https://gist.github.com/joelthe1/df8b948465741f0cc03f15617f91a609

#### Expected Results
Expected to successfully run training and show predictions.

#### Actual Results
```RuntimeError: The size of tensor a (8) must match the size of tensor b (3) at non-singleton dimension 1```

#### Versions
- Firefox
- Python version: 3.11.3]
- Fairlearn version: 0.9.0

SeanMcCarren:
@joelthe1 Thanks a bunch for this elaborately typed-out issue, and the accompanied gh gist. Please review [this fix](https://github.com/fairlearn/fairlearn/pull/1333) for this issue.

joelthe1:
This looks okay to me. Will raise separate issue if I encounter anything later.

TamaraAtanasoska:
@adrinjalali this issue can be closed.

PR #1333, opened 2024-01-22
MNT Fix bug: FloatTransformer output dimension was set incorrectly.

SeanMcCarren:
## Description
Thanks to [this elaborately typed out issue](https://github.com/fairlearn/fairlearn/issues/1332) we found a bug in Adversarial Debiasing. Considering how big, highly parametrized, and not fully covered this implementation is, I'm surprised this is the first bug :)

The bug is simple. The `FloatTransformer` class's intention is to cast different types of input to a 2D array. But, in a specific case where no transformer is provided as input, and the output is not bi/multinomial, we accidentally set the output shape as the input shape.

## Tests

- [ ] no new tests required
- [ ] new tests added
- [x] existing tests adjusted

## Documentation


hildeweerts:
Many thanks, @SeanMcCarren! This looks good to me - but it seems the tests are failing unrelated to your PR (see #1334). After we fix that and the PR passes all checks I think it should be ready for merging.

fairlearn/adversarial/_preprocessor.py
```diff
@@ -103,11 +103,16 @@ def _check(self, X, dtype=None, init=False):
...
+            if self.input_dim_ > 2:
+                raise ValueError(
+                    "Data can be at most two dimensional, not %d" % self.input_dim_
+                )
         else:
             if X.ndim != self.input_dim_:
                 raise ValueError("Dimension of data is inconsistent with previous call")
         if X.ndim == 1:
             X = X.reshape(-1, 1)
+        assert X.ndim == 2
```
romanlutz, line 115: An if check with raising an exception with a nice error message is probably preferable. Unless this is a case that can't happen at all?

SeanMcCarren: It really shouldn't happen, but I agree!

romanlutz (approved):

joelthe1:
Please consider merging and releasing this fix. Thanks!

fairlearn/adversarial/_preprocessor.py
```diff
@@ -103,11 +103,17 @@ def _check(self, X, dtype=None, init=False):
         )
         if init:
             self.input_dim_ = X.ndim
+            if self.input_dim_ > 2:
+                raise ValueError(
```
riedgar-ms, line 107: Sorry for the delay in reviewing, but is it possible to trigger this error in a test? Same for the `SystemError` below?

SeanMcCarren:
I am surprised this isn't merged yet, I am completely out of the loop of this. I can confirm that the issue raised [here](https://github.com/fairlearn/fairlearn/issues/1360) by [parasurama](https://github.com/parasurama) will also be fixed by this PR.

riedgar-ms (approved):

Issue #1334, opened 2024-01-23
MNT Deprecation warning Pyarrow


hildeweerts:
It looks like our [tests are failing](https://github.com/fairlearn/fairlearn/actions/runs/7608202568/job/20716916294) due to a DeprecationWarning from pandas:

```
ERROR test/install/test_no_ml.py - DeprecationWarning: 
Pyarrow will become a required dependency of pandas in the next major release of pandas (pandas 3.0),
(to allow more performant data types, such as the Arrow string type, and better interoperability with other libraries)
but was not found to be installed on your system.
If this would cause problems for you,
```

PR #1335, opened 2024-01-23
MNT add pyarrow dependency

adrinjalali:
Fixes https://github.com/fairlearn/fairlearn/issues/1334

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,8 @@
 numpy>=1.24.4
 pandas>=2.0.3
+# pyarrow is going to be a mandatory dependency to pandas
+# this explicit dependency is only for the sake removing the DeprecationWarning
+# from pandas in this regard while that happens.
+pyarrow>=15
 scikit-learn>=1.2.1
 scipy>=1.9.3


hildeweerts:
It does look like they're still [gathering feedback](https://github.com/pandas-dev/pandas/issues/54466) before the pyarrow dependency becomes final (but perhaps the actual decision has already been taken?). One major thing to consider is that the pyarrow build is pretty big:

I haven't read the entire thread but a couple of comments stood out to me:

> Including PyArrow would naturally increase the installation size of pandas. For example, installing pandas and PyArrow using pip from wheels, numpy and pandas requires about 70MB, and including PyArrow requires an additional 120MB. An increase of installation size would have negative implication using pandas in space-constrained development or deployment environments such as AWS Lambda.

> Also, if PyArrow is truly going to be a required required dependency in Pandas 3.0, then I don't see the point of the current DeprecationWarning in pandas 2.2.0. All sane package python managers install required dependencies automatically, so users don't need to take any action anyway.

adrinjalali:
Yes, they have already voted and made the decision. I've been watching this for quite some time now. It's happening. My recommendation to people is simply to switch to polars.

romanlutz (approved):

adrinjalali:
oh, I was gonna fix other issues, I'll open a new PR for them then.

Issue #1338, opened 2024-01-24
Issue with UtilityParity indexing


adrinjalali:
We have this code:

```py
        self.pos_basis = pd.DataFrame()
        self.neg_basis = pd.DataFrame()
        self.neg_basis_present = pd.Series(dtype="float64")
        zero_vec = pd.Series(0.0, self.index)
        i = 0

        for e in event_vals:
            # Constraints on the final group are redundant, so they are not
            # included in the basis.
            for g in group_vals[:-1]:
                self.pos_basis[i] = 0 + zero_vec
                self.neg_basis[i] = 0 + zero_vec
                self.pos_basis[i]["+", e, g] = 1
                self.neg_basis[i]["-", e, g] = 1
                self.neg_basis_present.at[i] = True
                i += 1
```

which causes a few issues, due to indexing and copy-on-write, so a better way of writing the same code seems to be:

```py
        self.neg_basis_present = pd.Series(dtype="float64")
        col_count = len(event_vals) * (len(group_vals) - 1)
        self.pos_basis = pd.DataFrame(0.0, index=self.index, columns=range(col_count))
        self.neg_basis = pd.DataFrame(0.0, index=self.index, columns=range(col_count))

        i = 0

        for e in event_vals:
            # Constraints on the final group are redundant, so they are not
            # included in the basis.
            for g in group_vals[:-1]:
                self.pos_basis.loc[("+", e, g), i] = 1
                self.neg_basis.loc[("-", e, g), i] = 1
                self.neg_basis_present.at[i] = True
                i += 1
```

However, the former produces:

```
                         0    1    2    3    4    5    6    7    8    9
sign event   group_id                                                  
+    label=0 0,0       1.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
             1,1       0.0  1.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
             2,2       0.0  0.0  1.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0
     label=1 1,1       0.0  0.0  0.0  0.0  0.0  0.0  1.0  0.0  0.0  0.0
             3,3       0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  1.0  0.0
             4,4       0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  0.0  1.0
[...]

PR #1339, opened 2024-01-24
FIX more pandas issues

adrinjalali:
Trying to fix more issues, but there's an issue I can't figure out. Opened https://github.com/fairlearn/fairlearn/issues/1338 to discuss the issue.


adrinjalali:
@romanlutz this reproduces old code, but I'm not sure if the code is right since some iterations are now skipped, but explicitly rather than implicitly due to writing on copies like the old code.

riedgar-ms:
I believe that this is @MiroDudik 's balliwick ... although once it passes tests, I'd be happy to merge if @MiroDudik  has no comments.

PR #1347, opened 2024-02-06
MNT set requirements pandas>=2.2.0

hildeweerts:
In #1336 we introduced `include_groups=False` to silence a deprecation warning in pandas, but this argument did not exist before pandas 2.2.0, resulting in a bunch of test failures. It seems the easiest solution is to change our min dependency of pandas to 2.2.0.

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 numpy>=1.24.4
-pandas>=2.0.3
+pandas>=2.2.0
 # pyarrow is going to be a mandatory dependency to pandas
 # this explicit dependency is only for the sake removing the DeprecationWarning
 # from pandas in this regard while that happens.


adrinjalali:
Oh no :/

we shouldn't be doing this. This is too strict of a requirement and will break a lot of environments. We should have something in `fixes.py` to handle the two different versions (before and after 2.2)

hildeweerts:
Closed, see #1348

hildeweerts:
ok, I'm glad there's another possible solution that feels less icky, lol.

I'm not sure I will get to this soon so I'm closing this PR in favor of issue #1348 in case someone else can pick it up before then.

Issue #1350, opened 2024-02-09
CI Failing due to scikit-learn 1.4


riedgar-ms:
#### Describe the bug

The latest release of `scikit-learn` (v1.4) is causing issues in the `MetricFrame` tests.

#### Steps/Code to Reproduce

See GitHub Actions

#### Expected Results

Green builds

#### Actual Results

Red builds

#### Screenshots

N/A

#### Versions

Current `main`

PR #1351, opened 2024-02-09
MNT Build fixes

riedgar-ms:
## Description

Various tweaks to fix the builds:

- Temporary upper bound on `scikit-learn`. Issue #1350 for tracking. A full fix will be required in due course
- Tweaks a couple of the grid generator tests to avoid warnings from `pandas`
- Adjust `_disaggregated_result.py` to cope with recently introduced (and deprecated) argument from `pandas`

Both of the `pandas` changes should fall under #1299 

## Tests

- [ ] no new tests required
- [ ] new tests added
- [x] existing tests adjusted

## Documentation

### DIFF
--- a/fairlearn/metrics/_disaggregated_result.py
+++ b/fairlearn/metrics/_disaggregated_result.py
@@ -39,14 +39,22 @@ def extract_unique_classes(
 
 
 def apply_to_dataframe(
-    data: pd.DataFrame, metric_functions: Dict[str, AnnotatedMetricFunction]
+    data: pd.DataFrame,
+    metric_functions: Dict[str, AnnotatedMetricFunction],
+    include_groups: bool = False,
 ) -> pd.Series:
     """Apply metric functions to a DataFrame.
 
     The incoming DataFrame may have been sliced via `groupby()`.
     This function applies each annotated function in turn to the
     supplied DataFrame.
+
+    The include_groups argument is weird. It appears that pandas
+    introduced it as an argument in v2.2, and immediately deprecated
+    it (dependent on when this is being read, may need to adjust):
+    https://pandas.pydata.org/docs/reference/api/pandas.core.groupby.DataFrameGroupBy.apply.html
     """
+    assert not include_groups, "Unexpected change to include_groups"
     values = dict()
     for function_name, metric_function in metric_functions.items():
         values[function_name] = metric_function(data)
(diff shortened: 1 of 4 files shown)


adrinjalali:
If we release with this higher bound, users will end up with the highest scikit-learn and the previous version of fairlearn which doesn't have the higher bound on the dependency.

So I'm okay with this only if we're going to fix the actual issue before the next release, which we should do rather soon.

riedgar-ms:
I would definitely want to release the upper bound again before doing another release. This is _supposed_ to be a quick 'get the builds green' PR (although it appears that `pandas` has other ideas)

adrinjalali:
Releasing with the upper bound doesn't make sense really. As I said, people will end up with the older fairlearn instead. Upper bound only makes sense when every single past release has an upper bound

riedgar-ms:
Sorry, I had too many 'releases' in that sentence, didn't I....

fairlearn/metrics/_disaggregated_result.py
```diff
@@ -39,14 +39,22 @@ def extract_unique_classes(
...
     The incoming DataFrame may have been sliced via `groupby()`.
     This function applies each annotated function in turn to the
     supplied DataFrame.
+
+    The include_groups argument is weird. It appears that pandas
+    introduced it as an argument in v2.2, and immediately deprecated
+    it (dependent on when this is being read, may need to adjust):
+    https://pandas.pydata.org/docs/reference/api/pandas.core.groupby.DataFrameGroupBy.apply.html
     """
+    assert not include_groups, "Unexpected change to include_groups"
```
adrinjalali, line 57: can we avoid assert in our code? I don't understand the motivation behind this line.

riedgar-ms: I found that the `include_groups` argument was being included in the invocation of this function. But per `pandas` docs, it should only ever be False. So if it isn't, something really weird is happening; I just wanted to check on this. I can tweak the message

adrinjalali: but it's not used in this function at all

riedgar-ms: Correct. It's needed to work with pandas<2.2 (which I don't think we're dropping yet?). I've removed the `assert` as requested.

adrinjalali (approved):
at least our tests are now green, let's merge and fix the doc build in a separate PR.

PR #1359, opened 2024-04-25
modified conf.py as per scikit-learn guidelines on adding examples to…

miacodehub:
… API docs

## Tests

- [ ] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation


docs/conf.py
```diff
@@ -173,6 +173,8 @@
 # included in the gallery, but its plot is available for
 # the quickstart
 sphinx_gallery_conf = {
+    "doc_module" : ('fairlearn.fairlearn.datasets', 'fairlearn.fairlearn.metrics', 'fairlearn.fairlearn.postprocessing', 'fairlearn.fairlearn.preprocessing',  'fairlearn.fairlearn.reductions', 'fairlearn.fairlearn.adversarial', 'fairlearn.fairlearn.experimental'),
+    "backreferences_dir" : "backref_dir",
```
adrinjalali, line 177: this is what we have in sklearn though:

```py
    "doc_module": "sklearn",
    "backreferences_dir": os.path.join("modules", "generated"),
```

3.2.1
adrinjalali, line 1: this file seems bogus.

adrinjalali (commented):
I'm not sure what this changes in the rendered docs.

docs/conf.py
miacodehub: the doc_module is supposed to contain the names of the modules for which documentation should be rebuilt. I gave those values to follow the folder structure for the project. I tried with "fairlearn.module_name" and just "fairlearn" but with no results.
The backreferences_dir is supposed to be a place where the generated gallery would be stored.  So it has to be initially empty. The value entered in scikit-learn (modules/generated) doesn't already exist in their project - only the "modules" folder exists. So I believe "generated" gets created when the documentation is rebuilt, although I'm not 100% sure about that.

riedgar-ms:
Sorry, I've been remiss about this.... can you explain what problem this is trying to solve? I'm not being enlightened by the PR description.

docs/conf.py
alliesaizan: I think the right way to specify the doc module is just to specify "fairlearn". I tried building the docs using this conf.py file, and it looks like in the backref_dir, the thumbnails are being generated (see the image below), but they aren't showing up on the API pages. I wonder if that's because of the way that the API reference is now structured (i.e., using autodoc).

image

miacodehub:
> Sorry, I've been remiss about this.... can you explain what problem this is trying to solve? I'm not being enlightened by the PR description.

The goal was to have relevant example notebooks show up at bottom of the respective APIs - like it currently happens in scikit-learn. 
The way to do that was to enable this API example backreferencing in the conf.py file by modifying the sphinx-gallery section. According to the latest update, it appears that the required galleries are being built but not being displayed in the API documentation.

@alliesaizan do let me know if there anything that I can look into regarding this issue. If nothing can be done from my end, do unassign me. Thank you

adrinjalali:
@TamaraAtanasoska this seems to be trying to add a list of examples on the bottom of each API reference page to the class. Similar to this:

image


Could you please have a look and see how it can be done?

TamaraAtanasoska:
Just a comment that I am looking into this next, I started a branch.

alliesaizan:
Solved by #1389

PR #1361, opened 2024-05-09
[Build] Fix issues

riedgar-ms:
## Description

Address a number of issues in the builds

- The `macos-latest` runner is now ARM based and doesn't have older Python versions. Stick with `macos-12` for now
- Fix a variable name which had started upsetting `flake8`
- Upper bound on `scikeras` to at least get the TensorFlow test running again
- Update various GitHub Action versions

## Tests

- [ ] no new tests required
- [ ] new tests added
- [x] existing tests adjusted

## Documentation

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -49,7 +49,7 @@ jobs:
         os: [ubuntu-latest]
         python: ["3.11"]
         mlpackage: [lightgbm, xgboost, tensorflow, pytorch]
-      failfast: false
+      fail-fast: false
     uses: ./.github/workflows/test-other-ml.yml
     with:
       os: ${{ matrix.os }}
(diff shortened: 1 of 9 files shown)


riedgar-ms:
I know that we will want to go back on `macos-latest` and drop the older Python versions fairly soon

.github/workflows/build-test.yml
```diff
@@ -49,6 +49,7 @@ jobs:
         os: [ubuntu-latest]
         python: ["3.11"]
         mlpackage: [lightgbm, xgboost, tensorflow, pytorch]
+      fail-fast: false
```
riedgar-ms, line 52: This allows the otherml packages to complete even if one fails. I think this makes sense, given that each package is likely to fail for its own reasons

romanlutz (approved):

PR #1363, opened 2024-05-23
MNT DOC Fix (some) doc build warnings and add/improve documentation

TamaraAtanasoska:
## Description

This PR fixes a set of `pandas` related warnings that were issued while building the docs, and one related to the `scikit-learn` precision function. It also adds optional information in the development process documentation about installing `pre-commit hooks` and slightly improves the readability of the text in that file. In the documentation contribution guide, information about installing `LightGBM` is added, which affects `MacOS` users. More detailed info below.

8a8ec3a4b89b1a6db459438d94cf17dd350943df

- All the `pandas` warnings were a `SettingWithCopyWarning`. A newer feature that will be included by default in the future releases and can be enabled for all the versions above 1.5 is the [Copy-On-Write](https://pandas.pydata.org/pandas-docs/version/2.0/user_guide/copy_on_write.html) feature. I enable the feature in the files throwing the error. 
While the code itself isn't conceptually wrong and fixing the warning isn't necessary for its proper functioning, clearing out the pandas related warnings before looking into updating the version or the library's replacement is just a step to make sure there is a clean start. Also for me to get to know the library a bit better :). 

086e6ff1a9950a5658c2b8a2631641de8b26e814

- I added information about enabling the `pre-commit` hooks for those who are not familiar, seeing that such a file exists in the library. This might not be desired, please let me know.
- I made small edits to some of the steps of the described development process and added some code formatting.

bb7d915eb979a3813e44c350cf1cc24ca2f47333

- I added information for `MacOS` users that get `Lightgbm OSError, Library not loaded` errors, as additionally installing `LightGBM` is necessary in those cases through `brew`, `conda` or other channels. This isn't the case for all systems for what I've read, but for me it was. I added it as a separate commit to be easily removed if considered unnecessary.

@adrinjalali 

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation


adrinjalali:
```
test/unit/adversarial/test_adversarial_mitigation.py:277: in <module>
    estimator_checks.check_estimator_sparse_data,
E   AttributeError: module 'sklearn.utils.estimator_checks' has no attribute 'check_estimator_sparse_data'
```

This is due to the latest scikit-learn release. We should have an issue / PR to fix the issues regarding the latest release and fix them.

docs/contributor_guide/contributing_documentation.rst
```diff
@@ -16,7 +16,9 @@ To contribute, make sure to install sphinx and its add-ons by running
     python scripts/install_requirements.py --pinned False
 
 in the repository root directory.
-You may also need to `install Pandoc <https://pandoc.org/installing.html>`_.
+You may also need to `install Pandoc <https://pandoc.org/installing.html>`_, and
+if you are a MacOS user in some cases also `Lightgbm
+<https://lightgbm.readthedocs.io/en/stable/Installation-Guide.html>`_.
```
adrinjalali, line 21: why is `install_requirements.py` not handling this then? I think the fix should go there?

docs/contributor_guide/development_process.rst
```diff
@@ -101,24 +101,46 @@ Follow the steps below to create a pull request.
...
    - Neither of the two above: this means issues haven't been scoped out properly yet or are more difficult and likely won't be doable within a day or two. If you're still interested just let us know and we'll figure out a way! Once you find an issue you're interested in comment on the issue with any questions you may have and let us know that you'd like to do it. This helps us avoid duplication and we can help you quicker.
 
 4. Whenever questions come up don't hesitate to reach out (see :ref:`communication`). We're here to help!
 
-5. Clone the Fairlearn repository onto your machine using :code:`git clone https://github.com/fairlearn/fairlearn.git`
+5. Clone the Fairlearn repository onto your machine using
+
+.. code-block::
+
+ git clone https://github.com/fairlearn/fairlearn.git
```
adrinjalali, line 115: Contributors should be cloning from their own fork, not the upstream

similar to: https://scikit-learn.org/dev/developers/contributing.html#how-to-contribute

examples/plot_new_metrics.py
```diff
@@ -66,6 +67,8 @@
 from fairlearn.datasets import fetch_adult
 from fairlearn.metrics import MetricFrame, count, selection_rate
 
+pd.set_option("mode.copy_on_write", True)
```
adrinjalali, line 70: setting this option, basically sets the option for all examples being rendered after this one I think, since they are run in the same process. So we either want to unset at the end of the example, or we should be setting it in the configuration of `sphinx_gallery` or `sphinx` for all examples.

docs/contributor_guide/development_process.rst
TamaraAtanasoska: A great catch, I just copied without rethinking there, as that was the link in the existing docs. I believe I'd like to change a few more things to look more like the scikit-learn version. 

Another thing is the introduction of the extension `sphinx-prompt` that scikit-learn's docs use. It's super convenient, but I didn't want to introduce another dependency just for a few edits. Now that I want to edit a larger block to look like the scikit-learn docs, I believe it might make more sense to include it. Any opinions on that? I could add it this PR together with the improvements.

examples/plot_new_metrics.py
TamaraAtanasoska: Right, thanks. I would like to keep it local to the files for now, as setting this globally impacts the speed of the examples. Setting the `copy_on_write option` is the easiest blanket case to solve this without changing the code or its functionality, but there is another way to go around it using `df.iloc`, in which instead of the drop in place a subgroup of the df is kept. I will either unset or use the other option.

docs/contributor_guide/development_process.rst
adrinjalali: Yeah adding in this PR is fine. Dependencies for our CI / doc build are a no brainer, we can add them.

romanlutz: Why is cloning from your fork better? I personally prefer cloning from the main repo and aliasing all forks by the corresponding usernames. Later in this doc we refer to a fork using the username, too, so everything else would need to be adjusted as well.

Maybe someone can explain the advantage of doing it another way?

TamaraAtanasoska: I also use the process that @adrinjalali suggested, I think it's a bit more straightforward for the users(especially less experienced people). I'm happy to change it all, I will go through all commands anyway as I introduce `sphinx-prompt`.

examples/plot_new_metrics.py
TamaraAtanasoska: I wanted to leave a "clean slate" for benchmarking with polars, but I can always turn it off for that. It seems it will be the default behaviour in the future versions, but it might come with some other improvements, I didn't read about it yet in detail. I think it makes sense though, I'll add it globally.

docs/contributor_guide/development_process.rst
adrinjalali: @romanlutz the canonical way of cloning is to have `origin` as the fork, and `upstream` as the main repo

docs/contributor_guide/contributing_documentation.rst
TamaraAtanasoska: The change reverted in e391f0fd13c3ba6e3890371b19621e7b1d475621.

docs/contributor_guide/development_process.rst
romanlutz: Ok, whatever works for people. Just make sure to change every occurrence then because there may be a few of them in this doc.

examples/plot_new_metrics.py
TamaraAtanasoska: Should we leave it in to allow this kind of behaviour in the future? As it is anyway becoming the default?

docs/conf.py
```diff
@@ -84,6 +85,10 @@
     ),
 }
 
+# Setting pandas option for the examples
+pd.set_option("mode.copy_on_write", True)
+
+
```
adrinjalali, line 91: I think we can remove this.

docs/contributor_guide/development_process.rst
adrinjalali: I think this is still WIP @TamaraAtanasoska ?

TamaraAtanasoska:
@adrinjalali I don't know how one answers to [this comment](https://github.com/fairlearn/fairlearn/pull/1363#discussion_r1624260915), I don't see a direct option 😄 yes the docs need a larger edit, they are the next focus of this week. I can ping when the edits are complete for a final check.

docs/contributor_guide/development_process.rst
adrinjalali: Ah I see. Since this PR is not adding this, just formatting, I'm okay with it then, we can improve in a separate PR.

adrinjalali (approved):

docs/conf.py
TamaraAtanasoska: removed in 6160ac9079c0a8818c656dc241df90dc4a25a9f9

TamaraAtanasoska:
@romanlutz there was a bit of a back and forth with the build, but all checks pass now

PR #1375, opened 2024-06-11
MNT remove lightgbm pin

adrinjalali:
The issue is fixed on the lightgbm side, so let's see if CI agrees.


adrinjalali:
This should be okay @riedgar-ms

riedgar-ms (approved):

PR #1378, opened 2024-06-12
MNT DOC Clear docs build from all warnings

TamaraAtanasoska:
## Description
This PR clears the doc build from all warnings: 
- [x]  WARNING: Underline too long
- [x]  WARNING: py:class reference target not found 
- [x]  WARNING: autosummary: stub file not found
- [x]  WARNING: image file not readable
- [x]  WARNING: undefined label
- [x]  WARNING: Inline interpreted text or phrase reference start-string without end-string.

Because of a conflict between the `napoleon` and `numpydoc` sphinx extensions which prompted the "`py:class reference target not found`"([more here](https://github.com/sphinx-doc/sphinx/issues/7391)), I removed the napoleon extension af9d1f968884732733874f10365368ff6993a230 and rewrote parts of the documentation in numpydoc style in f812ed4b75473b61ad1a2fe37c67c41ea232de6f.

This `numpydoc `class members suppression line in 275a70083b14938c025022f2aebd1ad58d5288c4 solves the `"autosummary: stub file not found"` warning, without changing the existing look of the documentation.

[Reference]( https://github.com/sphinx-gallery/sphinx-gallery/issues/120) for the solution for the`"image file not readable"` warning in 7e971320274bd8a28d90f8613e4b2cd24d2ff17f.

If there are still some warnings on another system, let me know! This makes a clean build for me completely. 

@adrinjalali 

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

- [ ] no documentation changes needed
- [ ] user guide added or updated
- [x] API docs added or updated
- [ ] example notebook added or updated

PR #1382, opened 2024-06-13
MNT Address (not always reproducible) docs build warnings

TamaraAtanasoska:
## Description
PR #1378 addressed all visible warnings after a clean build, but it seems that sometimes a partial build yields more warnings, addressing two small ones here. I will make a subsequent PR that will remove an [older deprecation warning](https://github.com/TamaraAtanasoska/fairlearn/blob/main/fairlearn/postprocessing/_threshold_optimizer.py#L281) that isn't necessary any more. 

I am not able to consistently reproduce the warnings with the same steps, but I saved one of the console outputs and they are pointing to real issues, despite not consistently appearing. If anyone understands why this is so I would love to know 🙏 . After this PR the `FutureWarning` should still appear as it is not addressed, but it doesn't consistently show. I am also surprised how the Notes underline in 821d8dd0f4e3036058d40d781c27a3ce2685f516 was not showing up with all the rest addressed in the previous PR.

@adrinjalali 

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

### DIFF
--- a/fairlearn/reductions/_moments/error_rate.py
+++ b/fairlearn/reductions/_moments/error_rate.py
@@ -85,7 +85,7 @@ class ErrorRate(ClassificationMoment):
         total_fn_cost = np.sum(signed_errors[signed_errors > 0] * self.fn_cost)
         total_fp_cost = np.sum(-signed_errors[signed_errors < 0] * self.fp_cost)
         error_value = (total_fn_cost + total_fp_cost) / self.total_samples
-        error = pd.Series(data=error_value, index=self.index)
+        error = pd.Series(data=error_value, index=self.index).copy()
         self._gamma_descr = str(error)
         return error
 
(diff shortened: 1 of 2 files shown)


fairlearn/reductions/_moments/error_rate.py
```diff
@@ -85,7 +85,7 @@ def gamma(self, predictor) -> pd.Series:
         total_fn_cost = np.sum(signed_errors[signed_errors > 0] * self.fn_cost)
         total_fp_cost = np.sum(-signed_errors[signed_errors < 0] * self.fp_cost)
         error_value = (total_fn_cost + total_fp_cost) / self.total_samples
-        error = pd.Series(data=error_value, index=self.index)
+        error = pd.Series(data=error_value, index=self.index).copy()
```
adrinjalali, line 88: what's the warning here that we're fixing?

TamaraAtanasoska: It is the pandas `SettingWithCopyWarning` again. As it was shown:
```
SettingWithCopyWarning:
A value is trying to be set on a copy of a slice from a DataFrame.
Try using .loc[row_indexer,col_indexer] = value instead
```

adrinjalali: seems like this doesn't break anything, there was a case where @MiroDudik is counting on that behavior to ignore those writes lol

adrinjalali (approved):

Issue #1381, opened 2024-06-12
AdversarialFairnessClassifier has no attribute 'classes__'


haytham918:
So I was running the piece of code from #1373, I just added `scoring="accuracy"` or something else like precision, f1, roc_auc for the grid search, it gives me the error that it has no attribute "classes__"

Code:
```
from sklearn.compose import make_column_transformer, make_column_selector
from sklearn.preprocessing import OneHotEncoder, StandardScaler
from sklearn.impute import SimpleImputer
from sklearn.model_selection import train_test_split
from sklearn.pipeline import Pipeline
from numpy import number
from sklearn.pipeline import make_pipeline
from sklearn.model_selection import GridSearchCV
from fairlearn.adversarial import AdversarialFairnessClassifier
from fairlearn.datasets import fetch_adult

import sklearn

sklearn.set_config(enable_metadata_routing=True)

X, y = fetch_adult(return_X_y=True)
pos_label = y[0]

z = X["sex"]  # In this example, we consider 'sex' the sensitive feature.

ct = make_column_transformer(
    (
        Pipeline(
            [
                ("imputer", SimpleImputer(strategy="mean")),
                ("normalizer", StandardScaler()),
            ]
        ),
        make_column_selector(dtype_include=number),
    ),
    (
        Pipeline(
            [
                ("imputer", SimpleImputer(strategy="most_frequent")),
                ("encoder", OneHotEncoder(drop="if_binary", sparse_output=False)),
            ]
        ),
        make_column_selector(dtype_include="category"),
    ),
)

X_train, X_test, Y_train, Y_test, Z_train, Z_test = train_test_split(
    X, y, z, test_size=0.2, random_state=12345, stratify=y
)

X_prep_train = ct.fit_transform(X_train)  # Only fit on training data!
X_prep_test = ct.transform(X_test)

mitigator = AdversarialFairnessClassifier(
    backend="torch",
    predictor_model=[50, "leaky_relu"],
    adversary_model=[3, "leaky_relu"],
    batch_size=2**8,
    progress_updates=0.5,
    random_state=123,
).set_fit_request(sensitive_features=True)

pipe = make_pipeline(ct, mitigator)

param_grid = {
    "adversarialfairnessclassifier__batch_size": [2**8, 2**6],
}

gs = GridSearchCV(pipe, param_grid, cv=5, scoring="accuracy").fit(
    X_train, Y_train, sensitive_features=Z_train
)
```

Error:
```
warnings.warn(
[...]

PR #1384, opened 2024-06-18
FIX Add classes_ attribute to AdversarialFairnessClassifier

TamaraAtanasoska:
## Description
This PR addresses issue #1381.
Additionally, the new scikit-learn estimator checks are enabled for both `AdversarialFairnessClassifier` and `AdversarialFairnessRegressor`. As a result many checks are failing and they are added under the `xfail` flag. They would be addressed individually in the future.

@adrinjalali 

## Tests

- [ ] no new tests required
- [x] new tests added
- [x] existing tests adjusted

## Documentation

### DIFF
--- a/fairlearn/adversarial/_adversarial_mitigation.py
+++ b/fairlearn/adversarial/_adversarial_mitigation.py
@@ -556,6 +556,9 @@ class _AdversarialFairness(BaseEstimator):
         sensitive_features : array
             Array-like containing the sensitive feature of the
             training data.
+
+        classes : array
+            An array containing the labels of classes.
         """
         X, Y, A = self._validate_input(X, y, sensitive_features, reinitialize=False)
         self.backendEngine_.train_step(X, Y, A)
--- a/fairlearn/adversarial/_adversarial_mitigation.py
+++ b/fairlearn/adversarial/_adversarial_mitigation.py
@@ -541,7 +541,7 @@ class _AdversarialFairness(BaseEstimator):
 
         return self
 
-    def partial_fit(self, X, y, *, sensitive_features=None, classes=None):
+    def partial_fit(self, X, y, *, sensitive_features=None):
         """
         Perform one epoch on given samples and update model.
 
@@ -556,9 +556,6 @@ class _AdversarialFairness(BaseEstimator):
         sensitive_features : array
             Array-like containing the sensitive feature of the
             training data.
-
-        classes : array
-            An array containing the labels of classes.
         """
         X, Y, A = self._validate_input(X, y, sensitive_features, reinitialize=False)
         self.backendEngine_.train_step(X, Y, A)
--- a/fairlearn/adversarial/_adversarial_mitigation.py
+++ b/fairlearn/adversarial/_adversarial_mitigation.py
@@ -541,7 +541,7 @@ class _AdversarialFairness(BaseEstimator):
 
         return self
 
-    def partial_fit(self, X, y, *, sensitive_features=None):
+    def partial_fit(self, X, y, *, sensitive_features=None, classes=None):
         """
         Perform one epoch on given samples and update model.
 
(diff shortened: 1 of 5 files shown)


romanlutz (approved):

TamaraAtanasoska:
@adrinjalali CI is green-ish, openml failed right at the end. Would you mind retriggering just those two last jobs? 

If you don't mind I will do a few test running CI in PR #1385 to check what is the state of our code and the new numpy API compatibility, I'll make new branch with fixes if there are issues.

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -540,7 +541,7 @@ def fit(self, X, y, *, sensitive_features=None):
 
         return self
 
-    def partial_fit(self, X, y, *, sensitive_features=None):
+    def partial_fit(self, X, y, *, sensitive_features=None, classes=None):
```
adrinjalali, line 544: where is this then used?

requirements.txt
```diff
@@ -1,4 +1,4 @@
-numpy>=1.24.4
+numpy~=1.24.4
```
adrinjalali, line 1: O_o, why not less than 2?

fairlearn/adversarial/_adversarial_mitigation.py
TamaraAtanasoska: The estimator checks fail if it isn't there, and I found it the same way in scikit-learn's code. I assumed it must be used somewhere, I am not very familiar with `partial fit `and it use cases 😬

requirements.txt
TamaraAtanasoska: For some reason, CI would still fail with [066218e](https://github.com/fairlearn/fairlearn/pull/1384/commits/066218e7c9fe9bf1138db6f450d8c01277a7517f). This is how I realised we have `numpy` API incompatibilities in the tests. Changing it to [937e09c](https://github.com/fairlearn/fairlearn/pull/1384/commits/937e09c6ff9f39a48727d0afe70a624ae7d5af77) makes CI pass. I was also confused why this is the case, but as I am already working on figuring this out on a more deeper level I though it won't hurt temporarily :)

fairlearn/adversarial/_adversarial_mitigation.py
TamaraAtanasoska: scikit-learn example: https://github.com/scikit-learn/scikit-learn/blob/329a1cf4bd46664fbf81806cde82a27f44b760e0/sklearn/multiclass.py#L394C32-L394C33

requirements.txt
TamaraAtanasoska: I reverted to range [4cbaf64](https://github.com/fairlearn/fairlearn/pull/1384/commits/4cbaf6458c873a9efc302a1a74132d9a8bc9a53e), which should fail only the `tensorflow` `build-test/other-ml` build. That one only seems to pass if we keep the version low as here `numpy~=1.24.4`.

fairlearn/adversarial/_adversarial_mitigation.py
adrinjalali: Yes I know they fail, but it doesn't make sense to take the argument and do nothing with it.

We should either have the failing tests in the xfail list, or actually fix the problem by using this argument.

Making the test pass w/o fixing the issue makes it harder to fix the problem later.

TamaraAtanasoska: Oh I see what you mean, it is better to xfail it until we know if there is an actual use case for it for fairlearn. Makes sense, I'll push a commit now.

adrinjalali: Yes, but you still have this unused argument here. Please remove the argument, to make sure later the relevant test fails.

TamaraAtanasoska:
@adrinjalali all checks have passed now! I removed the lightgbm verbose output as it was making it hard to scroll to see what the doc build failures are in [9fa8bbe](https://github.com/fairlearn/fairlearn/pull/1384/commits/9fa8bbe9ce30f814c6a172a04505610ed1b49e7c), in line with the discussion in [this PR](https://github.com/Microsoft/LightGBM/issues/1157). It is a bonus at the end, few changes.

requirements.txt
```diff
@@ -1,4 +1,4 @@
-numpy>=1.24.4
+numpy>=1.24.4, <2.0.0
```
adrinjalali, line 1: remove this?

adrinjalali (commented):
Otherwise LGTM.

adrinjalali:
I don't get the circle CI issue, rerunning the CI.

TamaraAtanasoska:
> I don't get the circle CI issue, rerunning the CI.

Ah interesting! I just merged the other branch in here, I will let you know when all checks pass.

TamaraAtanasoska:
@adrinjalali all checks have now passed

adrinjalali (approved):
Nice, now you can get to fixing those skipped tests one by one.

Issue #1395, opened 2024-07-09
Always returning a numpy array in FloatTransformer.inverse_transform


TamaraAtanasoska:
The new scikit-learn estimator checks, specifically [check_classifiers_train](https://github.com/scikit-learn/scikit-learn/blob/df25e3747b36bf11d328fe2cc39561a7f933aefd/sklearn/utils/estimator_checks.py#L2416) expect that `Y_pred` always is always a `numpy array` and not `list`. However, `FloatTransformer.inverse_transform` is implemented in a way that allows the input type to be preserved, in the case of `list`, `Dataframe` and `Series` ([here](https://github.com/TamaraAtanasoska/fairlearn/blob/main/fairlearn/adversarial/_preprocessor.py#L189)). Using `FloatTransformer.inverse_transform` [here](https://github.com/TamaraAtanasoska/fairlearn/blob/main/fairlearn/adversarial/_adversarial_mitigation.py#L608) triggers the error. 

Is this behaviour still desired to be kept as it is, in which case we can keep the test under the xfail flag, or I could look into changing this behaviour to always return a `numpy array`?

adrinjalali:
I think the output should really be an array.

adrinjalali:
Unless @SeanMcCarren remembers a reason not to.

TamaraAtanasoska:
this one has been merged and can be closed @adrinjalali

PR #1397, opened 2024-07-16
FIX failing estimator checks: "check_classifiers_train" and "check_regressors_no_decision_function"

TamaraAtanasoska:
## Description
This PR is not big, but it adds multiple non-trivial changes.
- As proposed in #1395, the `check_classifiers_train` always expect a `numpy.array`, while until now the input type was preserved for `list`, `Dataframe` and `Series`. This is now removed, and while those inputs are still allowed, the output will always be a `numpy.array`.
- I removed `_AdversarialFairness.decision_function` of not upholding the expected contract with `_AdversarialFairness.predict` that the estimator checks demand. It's output is preserved in `_raw_predict`, which someone could use if necessary, and it is called by `predict`.
- Because of the lack of a decision function, the `check_regressors_no_decision_function` passes now too so it is removed from under the xfail tag.
- Some minor refactoring on top of those changes. The class and supporting files can be refactored further, I will propose subsequent PRs with those changes at some point.

@adrinjalali 

Added to changelog.

## Tests

- [ ] no new tests required
- [ ] new tests added
- [x] existing tests adjusted

## Documentation

### DIFF
--- a/docs/user_guide/installation_and_version_guide/v0.11.0.rst
+++ b/docs/user_guide/installation_and_version_guide/v0.11.0.rst
@@ -8,3 +8,6 @@ v0.11.0
 
 * Added disparity metric :code:`equal_opportunity_{difference,ratio}` as a frequently used alias
   for :code:`true_positive_rate_{difference,ratio}`.
+
+* Removed :code:`_AdversarialFairness.decision_function` because it did not follow the expected contract with
+  :code:`_AdversarialFairness.predict`.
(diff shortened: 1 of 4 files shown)


adrinjalali (approved):
This LGTM.

If we want to be conservative, we can deprecate `decision_function` before removing it.

cc @fairlearn/fairlearn-maintainers

TamaraAtanasoska:
> This LGTM.
> 
> If we want to be conservative, we can deprecate `decision_function` before removing it.
> 
> cc @fairlearn/fairlearn-maintainers

Happy to do that! I just wanted to mention that the `decision_function` removed is not used in any of the tests or examples, maybe it helps lean towards just removing it instead of deprecating it.

adrinjalali:
Since we don't have any objections, and `decision_function` is almost never called by users, let's merge this. We can always introduce it back as deprecated if it comes up.

PR #1398, opened 2024-07-17
Fix check_fit_non_negative failing estimator check (+3 others)

TamaraAtanasoska:
## Description
This PR fixes the following failing estimator checks:
- `check_fit_non_negative`
-  `check_regressor_data_not_an_array`
- `check_estimators_fit_returns_self`
- `check_classifier_data_not_an_array`

I discovered some of these by looking at the xpassed as side effects to fixing others, so some might overlap with what #1397  has listed. No major changes in this PR.

@adrinjalali 

## Tests

- [ ] no new tests required
- [ ] new tests added
- [x] existing tests adjusted

## Documentation

### DIFF
--- a/fairlearn/adversarial/_adversarial_mitigation.py
+++ b/fairlearn/adversarial/_adversarial_mitigation.py
@@ -258,7 +258,7 @@ class _AdversarialFairness(BaseEstimator):
         self.warm_start = warm_start
         self.random_state = random_state
 
-    def __setup(self, X, Y, A):
+    def __setup(self, X, y, A):
         """
         Initialize the entire model from the parameters and the given data.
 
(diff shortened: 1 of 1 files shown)


fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -445,7 +444,7 @@ def fit(self, X, y, *, sensitive_features=None):
             training data.
         """
         X, Y, A = self._validate_input(X, y, sensitive_features, reinitialize=True)
```
adrinjalali, line 446: I think we shoulddn't introduce a new `Y` here, this seems to need to be 

```suggestion
        X, y, A = self._validate_input(X, y, sensitive_features, reinitialize=True)
```

adrinjalali (commented):
Otherwise LGTM.

Also, merge conflicts.

fairlearn/adversarial/_adversarial_mitigation.py
TamaraAtanasoska: I agree about the consistency! Unfortunately it is everywhere a capital `Y` jus in the function signature a small `y`. I capitalised it in the signature. [6fb9423](https://github.com/fairlearn/fairlearn/pull/1398/commits/6fb9423e1e0608ad75afa771ab78964d60346c8f)

TamaraAtanasoska:
@adrinjalali just pinging to add this to your notifications now after all the updates 🧹

fairlearn/adversarial/_adversarial_mitigation.py
adrinjalali: by convention on the sklearn side we always have `y`, except for historical reasons where it was `Y` before everything became `y`. So I'd go with `y` here and deprecate `Y`.

adrinjalali (approved):

PR #1399, opened 2024-07-23
Fixing failing estimator checks for the AdversarialFairnessRegressor

TamaraAtanasoska:
## Description
This PR fixes several failing estimator checks and makes the `AdversarialFairnessRegressor` more scikit-learn compatible. The `FloatTransfomer._get_type` is completely removed and several other functions minimised. The tests are just minimally edited, to remove some checks that depended on the `FloatTransfomer._get_type` function. 

However, I am treating this proposed PR as a WIP, as the changes are significant and might require a closer look and extra testing.

@adrinjalali thanks for taking a closer look at this!

## Tests

- [ ] no new tests required
- [ ] new tests added
- [x] existing tests adjusted

## Documentation


fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -341,39 +342,7 @@ def __setup(self, X, y, A):
 
         def read_kw(data, kw_or_func, kwname):
```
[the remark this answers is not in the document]
TamaraAtanasoska: Indeed, I agree don't need it. I started to remove it before posting the PR, but decided last minute to leave it for a future refactor. I think it can fit this PR nicely though, the rest can follow later. I will push a commit and ping again.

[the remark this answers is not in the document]
TamaraAtanasoska: addressed in [d52a15c](https://github.com/fairlearn/fairlearn/pull/1399/commits/d52a15ca2597f3504fc37f34bfcae370b5efa92d)

TamaraAtanasoska:
@adrinjalali I updated the test upload according to the latest `codecov` docs in [8546c5b](https://github.com/fairlearn/fairlearn/pull/1399/commits/8546c5b884433244a78c1292fe6099ba4da2936d). Is this update consistent with the previous usage of the service? All checks passed now.

fairlearn/adversarial/_preprocessor.py
```diff
@@ -151,7 +103,7 @@ def fit(self, X, y=None):
             # to the transformer.
             X_temp = self._check(X, init=True)
             self.n_features_in_ = X_temp.shape[0]
-            if _get_type(X_temp, self.dist_assumption) == "category":
+            if type_of_target(X_temp) in ["multiclass", "multilabel-indicator"]:
```
TamaraAtanasoska, line 106: @adrinjalali I am having doubts about this line here. Should it maybe be only `multiclass`? I am looking at the file again now for another test and there is the comment about `multilabel indicator` not needing to be encoded, line 98 above . I initially understood it as in not needing to be one hot encoded, but maybe it doesn't need to be handled at all here? I am not sure.

.github/workflows/test-all-deps.yml
```diff
@@ -44,13 +44,7 @@ jobs:
...
-          env_vars: OS,PYTHON
-          fail_ci_if_error: true
-          files: ./coverage.xml
-          flags: unittests
-          name: codecov-umbrella
-          verbose: true
+      - name: Upload coverage reports to Codecov with GitHub Action
+        uses: codecov/codecov-action@v4.2.0
+        env:
+          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
```
adrinjalali, line 50: the token is not really available via secrets on PRs for forks, that's why we have hardcoded it in our config, and it should stay that way, otherwise we quickly encounter GH rate limits and codecove uploads will fail.

fairlearn/adversarial/_preprocessor.py
adrinjalali: The question is also that this seems to be not tested in the first place. Writing a test for both cases would make it clearer whether we need it or not I guess.

TamaraAtanasoska: I had to review all of it now for another test and I see a few testing gaps, especially since `get_type` is being removed. I might have to think a bit how to replace those tests with something less hardcoded type-wise, but still complete. I have another branch in progress on top of this one with further fixes, but I'll have that wait a bit, I'll focus on this first.

.github/workflows/test-all-deps.yml
TamaraAtanasoska: removed the secrets reference in [73a6af3](https://github.com/fairlearn/fairlearn/pull/1399/commits/73a6af38d8476f5f090d2f6e3d44c5ac15834b87)

.github/workflows/test-minimal-deps.yml
```diff
@@ -39,14 +39,5 @@ jobs:
       - run: pip install .
       - run: pip install pytest pytest-cov
       - run: pytest --cov=fairlearn --cov-report=xml test/install
-      - name: Upload coverage to Codecov
-        uses: codecov/codecov-action@v3
-        if: ${{ (inputs.codeCovPython == inputs.python) && (steps.get_date.outputs.AGE < 172800) }} # 172800 = 48 * 3600
```
riedgar-ms, line 44: I think this conditional is needed, or our test matrix ends up sending too many results to CodeCov. At least for any build which runs regularly.

TamaraAtanasoska: Got it! I don't have experience with the service prior to this, so I wasn't sure how it is exactly used in this context. I will bring it back.

.github/workflows/test-all-deps.yml
```diff
@@ -44,13 +44,6 @@ jobs:
...
+      - name: Upload coverage reports to Codecov with GitHub Action
         if: ${{ (inputs.codeCovPython == inputs.python) && (steps.get_date.outputs.AGE < 172800) }} # 172800 = 48 * 3600
-        uses: codecov/codecov-action@v3
-        with:
-          env_vars: OS,PYTHON
-          fail_ci_if_error: true
-          files: ./coverage.xml
-          flags: unittests
-          name: codecov-umbrella
-          verbose: true
```
adrinjalali, line 56: I'm surprised that none of these are needed / they don't set the correct behavior

adrinjalali (approved):
LGTM!

.github/workflows/test-all-deps.yml
TamaraAtanasoska: would you mind checking if removing them impacts the behaviour on the account since you know what the desired one is? or you are saying that they don't? I changed the GitHub action to a different one so it might mean there is some behaviour included that wasn't there before?

adrinjalali: Basically, these settings need to be either on the codecov's config (codecov.yml) file or in the CI, and right now these settings here are not present in the config file, therefore this is changing behavior.

TamaraAtanasoska: ok, will do! the only one that can't be used at the moment is `path_to_write_report` which is deprecated. Is that one also strictly necessary? found only here https://github.com/fairlearn/fairlearn/pull/1399/files#diff-453070e1c523e30f2e01a1e4715f6ab1e5a4bc94cd0bd73234c8a2952bb33f46L51

adrinjalali: no, remove if deprecated.

TamaraAtanasoska: removed in https://github.com/fairlearn/fairlearn/pull/1399/commits/de2049457e29e4004ebc1a530544320230144d80, we just need a confirmation now that CI passes as before

riedgar-ms (approved):
Thanks for restoring the build conditional!

.github/workflows/test-all-deps.yml
TamaraAtanasoska: CI is green 🥳

adrinjalali (approved):

PR #1405, opened 2024-09-03
MNT Fix errors caused by three estimator checks

TamaraAtanasoska:
## Description
This PR fixes the errors raised by these three scikit-learn estimator checks:
- `check_estimators_pickle`
- `check_estimators_overwrite_params`
- `check_classifiers_classes`

## Tests

- [ ] no new tests required
- [ ] new tests added
- [x] existing tests adjusted

## Documentation

## Screenshots

@adrinjalali

### DIFF
--- a/fairlearn/adversarial/_adversarial_mitigation.py
+++ b/fairlearn/adversarial/_adversarial_mitigation.py
@@ -589,11 +589,32 @@ class _AdversarialFairness(BaseEstimator):
 
     def _validate_input(self, X, y, A, reinitialize=False):
         """
-        Validate the input data and possibly setup this estimator.
+        Validate input data and optionally set up the estimator.
 
-        Important note is that we follow call `__setup` from here, because the
-        setup procedure requires the validated data. If `reintialize` is True,
-        then always call `__setup`.
+        Parameters
+        ----------
+        X : array-like
+            The input features.
+        y : array-like
+            The target values.
+        A : array-like
+            The sensitive features.
+        reinitialize : bool, default=False
+            If True, force reinitialization of the estimator.
+
+        Returns
+        -------
+        X : array-like
+            Validated input features.
+        y : array-like
+            Validated target values.
+        A : array-like
+            Validated sensitive features.
+
+        Notes
+        -----
+        This method calls `__setup` if the estimator is not fitted or if
+        `reinitialize` is True. The setup procedure requires validated data.
         """
         if not self.skip_validation:
             X = self._validate_data(
--- a/fairlearn/adversarial/_adversarial_mitigation.py
+++ b/fairlearn/adversarial/_adversarial_mitigation.py
@@ -413,7 +413,6 @@ class _AdversarialFairness(BaseEstimator):
             training data.
         """
         X, y, A = self._validate_input(X, y, sensitive_features, reinitialize=True)
-        self.classes_ = unique(y)
 
         # Not checked in __setup, because partial_fit may not require it.
         if self.epochs == -1 and self.max_iter == -1:
@@ -647,6 +646,8 @@ class _AdversarialFairness(BaseEstimator):
         if (not is_fitted) or (reinitialize):
             self.__setup(X, y, A)
 
+        self.classes_ = unique(y)
+        y = self._y_transform.transform(y)
         A = self._sf_transform.transform(A)
 
         if not self.skip_validation:
--- a/fairlearn/adversarial/_adversarial_mitigation.py
+++ b/fairlearn/adversarial/_adversarial_mitigation.py
@@ -603,7 +603,7 @@ class _AdversarialFairness(BaseEstimator):
                 dtype=float,
                 allow_nd=True,
             )
-            y = self._validate_data(y, ensure_2d=False)
+            y = self._validate_data(y, dtype=None, ensure_2d=False)
 
             check_consistent_length(X, y)
             check_consistent_length(X, A)
@@ -623,7 +623,6 @@ class _AdversarialFairness(BaseEstimator):
             self.__setup(X, y, A)
 
         if not self.skip_validation:
-            y = self._y_transform.transform(y)
             A = self._sf_transform.transform(A)
 
         if not self.skip_validation:
@@ -971,9 +970,6 @@ class AdversarialFairnessClassifier(_AdversarialFairness, ClassifierMixin):
                 ),
                 "check_classifiers_regression_target": ("the data cannot look continuous."),
                 "check_estimators_partial_fit_n_features": ("number of features cannot change."),
-                "check_classifiers_classes": (
-                    "decision function output must match classifier output."
-                ),
                 "check_supervised_y_2d": "DataConversionWarning not caught.",
             },
             "poor_score": True,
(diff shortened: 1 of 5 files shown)


fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -603,7 +603,7 @@ def _validate_input(self, X, y, A, reinitialize=False):
                 dtype=float,
                 allow_nd=True,
             )
-            y = self._validate_data(y, ensure_2d=False)
+            y = self._validate_data(y, dtype=None, ensure_2d=False)
```
TamaraAtanasoska, line 606: It seems that this here is an issue. In the case of values like `<=50K` which are string objects, there is an issue when converting them to float [when this line is executed](https://github.com/fairlearn/fairlearn/blob/main/fairlearn/adversarial/_adversarial_mitigation.py#L634). I will look into using OneHotEncoder or `LabelEncoder` to transform these types of values somewhere earlier.

test/unit/adversarial/helper.py
```diff
@@ -161,7 +162,7 @@ class initializers:  # noqa: D106
             class GlorotNormal:  # noqa: D106
                 pass
 
-        Model = type("Model", (model_class,), {})
+        Model = Model
```
adrinjalali, line 165: O_o

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -735,9 +739,11 @@ def _set_predictor_function(self):
         elif isinstance(self.predictor_function_, str):
             kw = self.predictor_function_
             if kw == "binary":
-                self.predictor_function_ = lambda pred: (pred >= self.threshold_value).astype(
-                    float
-                )
+
+                def binarization(pred):
```
adrinjalali, line 743: is this used anywhere?

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -619,11 +618,13 @@ def _validate_input(self, X, y, A, reinitialize=False):
...
 
+        if not self.skip_validation:
+            check_consistent_length(X, A)
+
         if (not is_fitted) or (reinitialize):
             self.__setup(X, y, A)
 
         if not self.skip_validation:
-            y = self._y_transform.transform(y)
             A = self._sf_transform.transform(A)
```
adrinjalali, line 628: what does validation have to do with transforming data? confused

test/unit/adversarial/helper.py
TamaraAtanasoska: oh this is definitely questionable. maybe it can be replace with a function so it is more clear for both assignments? it is necessary since Pickle needs to know about `Model` outside of all the nested classes, otherwise it doesn't recognise it. something to do with the nested classes from what I read. but I agree it is not a great way to do things.

fairlearn/adversarial/_adversarial_mitigation.py
TamaraAtanasoska: this is how it was when I found it, do you think it is unnecessary? I can't know why it was added in. I will have to change this when I do the changing of the categorical labels up too, so I can check it out more in detail.

fairlearn/adversarial/_adversarial_mitigation.py
TamaraAtanasoska: I believe there was a test failing, but it has been now many weeks since I wrote this and started this branch and I am not 100% sure :S I will review it all once again.

test/unit/adversarial/helper.py
TamaraAtanasoska: @adrinjalali I need some help in shaping this is a way that is a bit better. I have tried multiple things, but picke rejects everything except defining this value at a top level and using the same object later in the class to assign it to the class attribute. Anything I can do syntax-wise to make this a bit more acceptable? Or do you have some other ideas?

TamaraAtanasoska:
@adrinjalali except the issue mentioned above specifically, I've added a bunch of things that it might be useful looking into. now all tests pass and CI is green. I believe there is more that is clear right now when it comes to the type of distributions we support and most of the estimator checks are passing.

test/unit/adversarial/helper.py
adrinjalali: Can you explain what the purpose of this here is? This is only for testing purposes, so it doesn't matter how dirty it is I'd say.

TamaraAtanasoska: Yes! I don't keep any references to the docs I read at the time when I fixed this (I cherrypicked it from an old branch) so I will try to summarise what I remember.
- The issue: the pickle related estimator checks complain that they can't pickle because they cannot find the `Model` attribute.
- Reading docs I found out that the deep nested classes are an antipattern for pickle, as in it doesn't deal with them well and there are often pickling issues when used. 
- Things I remember trying: rewriting a part of that nested class or reducing the nesting, moving the attribute position around (it had too big implications on the rest of the code and how we test and if I was to make it work I had to rewrite much more), I tried having it twice(top and nested level) assigning from a getter kind of function so it's "nicer" (complained it isn't the same object). 
- The only thing that worked was defining it like this top level and just using it as a reference in the assignment in the nested class. I am also not found of the solution, but I don't know pickle well enough to think of an alternative.

TamaraAtanasoska:
@adrinjalali if you are reviewing this again while I am AFK the next days, I pushed [this branch](https://github.com/TamaraAtanasoska/fairlearn/tree/PROV-remaining-est-checks) that is on top of the one for this PR and solves all the rest of the estimator checks issues. There are some more edits there. Maybe it helps with reviewing this one as it is all connected, and if you think something might be missing (it might be solved there). I won't make a PR from that next one, I'll open a new one and combine some commits, just pushed it for reference.

adrinjalali:
Let's do some pair programming on this when you're back to figure it out.

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -603,10 +625,9 @@ def _validate_input(self, X, y, A, reinitialize=False):
                 dtype=float,
                 allow_nd=True,
             )
-            y = self._validate_data(y, ensure_2d=False)
+            y = self._validate_data(y, dtype=None, ensure_2d=False)
```
adrinjalali, line 628: note for a separate PR, sklearn>=1.6 now has `sklearn.utils.validation.validate_data` instead. So we need to move this into the `fixes.py` and use that when available.

test/unit/adversarial/helper.py
```diff
@@ -23,6 +23,7 @@
 from fairlearn.adversarial._tensorflow_engine import TensorflowEngine
 
 model_class = type("Model", (object,), {})
+Model = type("Model", (model_class,), {})
```
adrinjalali: this whole thing is weird. Talked with @TamaraAtanasoska , and it seems nicer to have a separate CI, one for TF and one for pytorch, instead of mocking these here.

adrinjalali (approved):

test/unit/adversarial/helper.py
TamaraAtanasoska: I have just reverted the changes for this PR that just address the pickling issues, and I'll make a separate PR with the changes.

PR #1416, opened 2024-09-29
Threshold optimizer fit/predict: add batch_size and device options for speed/memory efficiency

ellemcfarlane:
Is there demand for this kind of thing? Without it, for my image data (8k images), it very easily maxes out my RAM and even if not, prediction is slow on cpu. 

I have this working in a messy way (e.g. assumes estimator is torch, etc) for a project, but can maybe clean it up when I have time.


adrinjalali:
There are quite a few changes here that we can separate from one another in separate PRs. But overall, I think it'd be nice to have these optimisations.

TamaraAtanasoska:
@ellemcfarlane are you still interested in finishing this up? maybe as @adrinjalali says we can close this PR and you could propose a few other PR with smaller incremental changes? or this is something you don't see yourself having time for soon? In the second case we could create an issue and link to the closed PR for future reference.

PR #1329, opened 2024-01-03
DOC Added link to raise issue and view source code

Hrittik20:
Issue: #1301

Issue #1301, opened 2023-10-12
DOC add link to raise issue


romanlutz:
#### Describe the issue linked to the documentation

In today's community call, the idea was brought up (by @MiroDudik  or @hildeweerts , I forget...) to add a link on example notebooks to raise an issue. This should make it easy for readers to raise issues without having to navigate to our repo first and then copy-paste links to the correct notebook etc.

#### Suggest a potential alternative/fix

There's no feature in sphinx-gallery that does that automatically BUT I found this is the pytorch docs:

raw.githubusercontent.com/pytorch/tutorials/3e9e8f68cc904f9eb37c862348cc2216611faadd/intermediate_source/ddp_tutorial.rst

```
.. note::
   |edit| View and edit this tutorial in `github <https://github.com/pytorch/tutorials/blob/main/intermediate_source/ddp_tutorial.rst>`__.
```

So this would just be a highlighted box with a link and we'd have to make sure the link is correct. The file names rarely change so it wouldn't be hard to maintain over time, though.

![image](https://github.com/fairlearn/fairlearn/assets/10245648/56fd3d48-407d-492c-bf19-4e86f4dd515c)

Thoughts @fairlearn/fairlearn-maintainers ?

hildeweerts:
Sounds like a fine and not too overengineered solution to me! :)

Hrittik20:
Hello @romanlutz! Just to confirm, do I need to add this to only "Example Notebooks" page?

romanlutz:
It needs to be added to every single notebook I would imagine.

Hrittik20:
@romanlutz So I need to put the links to the appropriate notebooks, right? For example, for contributing example notebook, I need to use this link "https://github.com/fairlearn/fairlearn/blob/main/docs/contributor_guide/contributing_example_notebooks.rst"

romanlutz:
I was thinking more along the lines of linking to the code file in the repo. Once you are there you can hit edit and create a branch with your changes. 

But what you're saying has merit, too. For most people it's probably advisable to first file an issue and ask if the changes are meant to happen instead of doing all the work for a PR.

We could do both...

hildeweerts:
Perhaps we can add two links, 'view source' to the code file and 'open issue' (with a link like https://github.com/fairlearn/fairlearn/issues/new?title=DOC%20Issue%20blabla&body=blabla.)

TamaraAtanasoska:
[...]

PR #1421, opened 2024-10-29
DOC Add links to GitHub and issue raising for every example

TamaraAtanasoska:
## Description
Fixes https://github.com/fairlearn/fairlearn/issues/1301 and closes https://github.com/fairlearn/fairlearn/pull/1329.
I spent some time looking into how I can automate this and make it generate links for each example, but this was a much easier solution for now.

@romanlutz and @hildeweerts please take a look!

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

## Screenshots

Screenshot 2024-10-29 at 14 50 16

Screenshot 2024-10-29 at 14 50 37

### DIFF
--- a/docs/templates/edit-example-link.html
+++ b/docs/templates/edit-example-link.html
@@ -3,7 +3,7 @@
         <a href="https://github.com/fairlearn/fairlearn/edit/main/examples/{{ pagename.split('/')[1]}}.py">
             Edit this example on GitHub
         </a>
-        <a href="https://github.com/fairlearn/fairlearn/issues/new?labels=documentation&template=doc_improvement.md&title=DOC%20Issue%20in%20Example%20Notebook%20{{ pagename.split('/')[1] }}.py>
+        <a href="https://github.com/fairlearn/fairlearn/issues/new?labels=documentation&template=doc_improvement.md&title=DOC%20Issue%20in%20Example%20Notebook%20{{ pagename.split('/')[1] }}.py">
             Report an issue on GitHub
         </a>
     {% endif %}
--- a/docs/templates/edit-example-link.html
+++ b/docs/templates/edit-example-link.html
@@ -1,0 +1,10 @@
+<div class="edit-example-link">
+    {% if pagename.split('/')[-1].startswith('plot_') %}
+        <a href="https://github.com/fairlearn/fairlearn/edit/main/examples/{{ pagename.split('/')[1]}}.py">
+            Edit this example on GitHub
+        </a>
+        <a href="https://github.com/fairlearn/fairlearn/issues/new?assignees=&labels=Documentation&template=doc_improvement.md&title=DOC%20Issue%20in%20{{ pagename.split('/')[1] }}&body=https://github.com/fairlearn/fairlearn/edit/main/examples/{{ pagename.split('/')[1] }}.py">
+            Report an issue on GitHub
+        </a>
+    {% endif %}
+</div>
--- a/docs/templates/edit-example-link.html
+++ b/docs/templates/edit-example-link.html
@@ -3,7 +3,7 @@
         <a href="https://github.com/fairlearn/fairlearn/edit/main/examples/{{ pagename.split('/')[1]}}.py">
             Edit this example on GitHub
         </a>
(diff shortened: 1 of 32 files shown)


hildeweerts:
It's not like we're adding new example notebooks very often, so I think this solution is fine for now.

It would be nice if the created issue follows one of the issue templates, probably the documentation improvement one makes the most sense. I think that should be pretty easy to do by adding something like `labels=Documentation&projects=&template=doc_improvement.md` to each link.

I'm not a huge fan of the double note on top (see screengrab below), but I don't think there's an easy way to resolve that as the first one is generated automatically.

image

TamaraAtanasoska:
@hildeweerts I can edit it to fit the template, that is a great remark.

adrinjalali (commented):
Interesting one.

I guess, why not. I like how in sklearn you just have the buttons at the bottom of each example.

So, I'm kinda, could do, could not do, on this PR. I'll let others decide.

TamaraAtanasoska:
> Interesting one.
> 
> I guess, why not. I like how in sklearn you just have the buttons at the bottom of each example.
> 
> So, I'm kinda, could do, could not do, on this PR. I'll let others decide.

I also like that the right sidebar carries a lot of the buttons and relevant info in scikit-learn

TamaraAtanasoska:
@hildeweerts what do you think about moving this information to the sidebar on the right where the sections are? This way it  will be available as you scroll down all the time.

hildeweerts:
TIL scikit-learn has moved to the pydata theme! @TamaraAtanasoska I like having all the buttons in the sidebar on the right, let's do it :)

TamaraAtanasoska:
I found a better way than doing this manually, however I can't seem to make my setup work with it. I created an issue here: https://github.com/pydata/pydata-sphinx-theme/issues/2027.

TamaraAtanasoska:
@romanlutz I seem to have found a way to do this :) 

Changes:
- there is a link on the right side not that leads to a new issue with the doc template as @hildeweerts proposed
- all the links are on the sidebar now, and there is no need for any notes (consistent with scikit-learn cc @adrinjalali)
- the bottom downloads are hidden as they aren't necessary any more

I really like how it looks now, it is more "clean". 

Screenshot 2024-11-04 at 09 22 03

adrinjalali:
happy to review when things go green again 😁

TamaraAtanasoska:
> happy to review when things go green again 😁

all green, versioning issue :)

TamaraAtanasoska:
@adrinjalali (I replied to the comment a few hours ago, but maybe that doesn't ping :S)

docs/templates/edit-example-link.html
```diff
@@ -0,0 +1,10 @@
+<div class="edit-example-link">
+    {% if pagename.split('/')[-1].startswith('plot_') %}
+        <a href="https://github.com/fairlearn/fairlearn/edit/main/examples/{{ pagename.split('/')[1]}}.py">
+            Edit this example on GitHub
+        </a>
+        <a href="https://github.com/fairlearn/fairlearn/issues/new?assignees=&labels=Documentation&template=doc_improvement.md&title=DOC%20Issue%20in%20{{ pagename.split('/')[1] }}&body=https://github.com/fairlearn/fairlearn/edit/main/examples/{{ pagename.split('/')[1] }}.py">
```
hildeweerts, line 6: Currently, the template isn't used because the body is specified explicitly. I personally prefer having the template over the exact link because it encourages people to write something meaningful (and having the title point to the example name seems sufficient). We also need to use lowercase `documentation`, otherwise it doesn't work. So i'd suggest something like this:
 
```suggestion
        <a href="https://github.com/fairlearn/fairlearn/issues/new?labels=documentation&template=doc_improvement.md&title=DOC%20Issue%20in%20Example%20Notebook%20{{ pagename.split('/')[1] }}.py>
```

Which should result in a link like this: https://github.com/fairlearn/fairlearn/issues/new?labels=documentation&template=doc_improvement.md&title=DOC%20Issue%20in%20Example%20Notebook%20plot_quickstart_counts.py

adrinjalali (approved):

PR #1432, opened 2024-11-06
feat: add literals for compare method and error handling

taharallouche:
## Context
This, imho, can be a small QoL improvement for the users whose IDEs might list the valid options when entering these arguments. It can also make the code more readable (no need to go through the doc string to figure out the valid options).

Please ignore if it doesn't match the repo's conventions (I don't want to be stepping over any toes here :sweat_smile:). More generally, if you prefer that I open issues to discuss the changes before opening PRs I'd be very happy to do it :pray: 

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

### DIFF
--- a/fairlearn/metrics/_disaggregated_result.py
+++ b/fairlearn/metrics/_disaggregated_result.py
@@ -2,7 +2,7 @@
 # Licensed under the MIT License.
 
 import logging
-from typing import Dict, List, Optional, Union
+from typing import Dict, List, Literal, Optional, Union
 
 import numpy as np
 import pandas as pd
@@ -105,7 +105,7 @@ class DisaggregatedResult:
         self,
         grouping_function: str,
         control_feature_names: Optional[List[str]] = None,
-        errors: str = "raise",
+        errors: Literal["raise", "coerce"] = "raise",
     ) -> Union[pd.Series, pd.DataFrame]:
         """Compute mins or maxes.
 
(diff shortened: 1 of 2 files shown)


fairlearn/metrics/_disaggregated_result.py
```diff
@@ -105,7 +108,7 @@ def apply_grouping(
         self,
         grouping_function: str,
         control_feature_names: Optional[List[str]] = None,
-        errors: str = "raise",
+        errors: ErrorHandler = "raise",
```
adrinjalali, line 111:
```suggestion
        errors: "raise" | "coerce" = "raise",
```

adrinjalali (commented):
I'd rather list them in every instance, instead of creating a typehint kinda thing. It makes reading the code harder. I'd be happy with the change then.

fairlearn/metrics/_disaggregated_result.py
taharallouche: Hey @adrinjalali , did you mean `errors: Literal["raise, "coerce"] = "raise"` ? I don't think  `errors: "raise" | "coerce" = "raise"` is a valid syntax, is it ? :thinking:

adrinjalali: It's valid python syntax, but mypy and pylance complain. So I'm okay with using `Literal` here. pity.

fairlearn/metrics/_metric_frame.py
```diff
@@ -814,7 +814,7 @@ def difference_ci(
         return value
 
     def ratio(
-        self, method: str = "between_groups", errors: str = "coerce"
+        self, method: Literal["between_groups", "to_overall"] = "between_groups", errors: Literal["raise", "coerce"] = "coerce"
```
adrinjalali, line 817: these lines are too long now. I think `black` should be handling these right.

taharallouche: My bad, I haven't initialized the pre-commit hooks :eyes:

adrinjalali:
@taharallouche note that there's no need to force push here, we squash and merge anyway.

adrinjalali (approved):

PR #1435, opened 2024-11-09
fix: type hints within metrics

taharallouche:
Fixed some type hints in the metrics package.

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

### DIFF
--- a/fairlearn/metrics/_annotated_metric_function.py
+++ b/fairlearn/metrics/_annotated_metric_function.py
@@ -1,9 +1,10 @@
 # Copyright (c) Fairlearn contributors.
 # Licensed under the MIT License.
+from __future__ import annotations
 
 import logging
 import warnings
-from typing import Callable, Dict, List, Optional
+from typing import Callable
 
 import numpy as np
 import pandas as pd
@@ -30,12 +31,12 @@ class AnnotatedMetricFunction:
     ----------
     func : callable
         The metric function we wish to invoke
-    name: str
+    name: str | None
         An optional string defining the name of the function
-    positional_argument_names: List[str]
+    positional_argument_names: list[str] | None
         The column names to be extracted and passed as positional arguments
         when invoking the function
-    kw_argument_mapping: Dict[str, str]
+    kw_argument_mapping: dict[str, str] | None
         The column names which are to be passed as keyword arguments
         when invoking the function. Since the DataFrame column names may
         not match the function's argument names, this is a dictionary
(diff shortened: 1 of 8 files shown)


fairlearn/metrics/_disaggregated_result.py
```diff
@@ -183,7 +183,7 @@ def apply_grouping(
 
     def difference(
         self,
-        control_feature_names: List[str],
+        control_feature_names: Optional[List[str]],
```
romanlutz, line 186: Shouldn't this have a default of `None`?
Also, it's missing a docstring section 😮

taharallouche: Indeed :eyes: I added both (I actually copied the argument's docstring from `.apply_grouping`, which is the method that uses it inside `.difference` and `.ratio`)

TamaraAtanasoska (approved):
LGTM, but I will leave it to @romanlutz to check it out again after commenting and merge.

fairlearn/metrics/_annotated_metric_function.py
```diff
@@ -48,8 +48,8 @@ def __init__(
         *,
         func: Callable,
         name: Optional[str],
-        positional_argument_names: List[str] = None,
-        kw_argument_mapping: Dict[str, str] = None,
+        positional_argument_names: Optional[List[str]] = None,
+        kw_argument_mapping: Optional[Dict[str, str]] = None,
```
adrinjalali, line 52: I prefer this:

```suggestion
        positional_argument_names: List[str] | None = None,
        kw_argument_mapping: Dict[str, str] | None = None,
```

you might need a `from __future__ import annotations` at the start of the file

adrinjalali (commented):
Same with other `Optional` usages

fairlearn/metrics/_annotated_metric_function.py
taharallouche: Oh that would be great actually ! :ok_hand: 

I can by the same occasion get rid of the `Dict`, `List` type hints and use plain standard classes `dict`, `list` etc .. (only for the metrics module for now to limit the scope of this PR :sweat_smile: )

taharallouche: @adrinjalali Done :exploding_head:

adrinjalali (approved):
much nicer!

PR #1437, opened 2024-11-13
Raise user-friendly error on missing ci parameters

taharallouche:
## Description
Hi again !  this small MR aims to improve the user experience a bit by hiding the metric frames internals more.

Basically, when the user tries to access a confidence-intervals-related attribute of a `MetricFrame` object which was not initialized with the confidence-intervals-related parameters (`n_boot` , `ci_quantiles`), she gets a `KeyError` exposing the internal `result_cache` object, which imho is not user-friendly (example in the figure below).
image

If you agree with these changes, please let me know, I'll go ahead and add associated unit tests :pray: 

## Tests

- [ ] no new tests required
- [x] new tests added
- [ ] existing tests adjusted

## Documentation

### DIFF
--- a/fairlearn/metrics/_metric_frame.py
+++ b/fairlearn/metrics/_metric_frame.py
@@ -506,6 +506,9 @@ class MetricFrame:
         The elements of the list are indexed by the `ci_quantiles` array supplied
         to the constructor.
         """
+        if self._ci_quantiles is None:
+            raise ValueError(_BOOTSTRAP_NEED_N_AND_CI)
+
         return self._result_cache["overall_ci"]
 
     @property
@@ -550,6 +553,9 @@ class MetricFrame:
         The elements of the list are indexed by the `ci_quantiles` array supplied
         to the constructor.
         """
+        if self._ci_quantiles is None:
+            raise ValueError(_BOOTSTRAP_NEED_N_AND_CI)
+
         return self._result_cache["by_group_ci"]
 
     @property
@@ -686,6 +692,9 @@ class MetricFrame:
         Unlike :meth:`MetricFrame.group_max` there is no :code:`errors` parameter, because
         a bootstrapped :class:`MetricFrame` requires all the metrics to return scalars.
         """
+        if self._ci_quantiles is None:
+            raise ValueError(_BOOTSTRAP_NEED_N_AND_CI)
+
         value = self._result_cache["group_max_ci"]
         return value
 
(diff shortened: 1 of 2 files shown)


fairlearn/metrics/_metric_frame.py
```diff
@@ -506,6 +506,9 @@ def overall_ci(
         The elements of the list are indexed by the `ci_quantiles` array supplied
         to the constructor.
         """
+        if self._ci_quantiles is None:
```
taharallouche, line 509: I'm using this as a proxy to check if the [ci-related parameters were not `None`](https://github.com/fairlearn/fairlearn/blob/main/fairlearn/metrics/_metric_frame.py#L284-L304). I can also just check if the `"overall_ci"` key is in the `_result_cache` dictionnary. I don't have any strong preference

adrinjalali: I think it would actually be nice to store the oroginal parameters of the metric frame in the object, and check them directly

taharallouche: Done @adrinjalali :+1: 

However the pipeline keeps failing when trying to [fetch the dataset from openml](https://github.com/fairlearn/fairlearn/blob/main/test/unit/adversarial/test_preprocessing.py#L36). 

Is the test known to be flaky ?

hildeweerts: yes that test is unfortunately very much known to be flaky :(

taharallouche: Thank you @hildeweerts it's all good now :ok_hand:

adrinjalali (commented):
Otherwise LGTM.

fairlearn/metrics/_metric_frame.py
```diff
@@ -281,14 +285,15 @@ def __init__(
         self._populate_results(result)
 
         # Handle bootstrapping
-        self._ci_quantiles = None
+        self._ci_quantiles = ci_quantiles
+        self._n_boot = n_boot
```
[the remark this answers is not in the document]
taharallouche: @adrinjalali imho it would safer to keep them private in this case: 

if they are public attributes it might signify that the user is allowed to change them. However if the user changes them after the `__init__` , during which the `result_cache` has been computed, it would  leave the object in an inconsistent state: `self.n_boot`, and `self.ci_quantiles` do not match the computed confidence intervals.

wdyt ?

adrinjalali (approved):
Thanks then!

PR #1438, opened 2024-11-14
fix: correlation remover docs

taharallouche:
## Description
Hi @TamaraAtanasoska @hildeweerts , I’ve updated the mathematical formulation in the doc to better match the optimization problem that the `CorrelationRemover` class is actually solving. WDYT ? :pray: 

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

## Screenshots [OUTDATED]
image

### DIFF
--- a/docs/user_guide/mitigation/preprocessing.rst
+++ b/docs/user_guide/mitigation/preprocessing.rst
@@ -22,44 +22,45 @@ while details from the original data are retained as much as possible (as measur
 by the least-squares error). The user can control the level of projection via the
 :code:`alpha` parameter.
 
-In mathematical terms, assume we have the original dataset :math:`X`, which
-contains a set of **sensitive attributes** denoted by :math:`S` and a set of
-**non-sensitive attributes** denoted by :math:`Z`. The goal is to remove
-correlations between the sensitive attributes and the non-sensitive attributes.
+In mathematical terms, assume we have the original dataset :math:`\mathbf{X}`, which
+contains a set of **sensitive features** denoted by :math:`\mathbf{S}` and a set of
+**non-sensitive features** denoted by :math:`\mathbf{Z}`. The goal is to remove
+correlations between the sensitive features and the non-sensitive features.
 
 Let :math:`m_s` and :math:`m_{ns}` denote the number of sensitive and non-sensitive
 features, respectively.
-Let :math:`\bar{S}` represent the mean of the sensitive attributes, *i.e.*,
-:math:`\bar{S} = (\bar{s}_1, \dots, \bar{s}_{m_s})`, where
+Let :math:`\bar{\mathbf{s}}` represent the mean of the sensitive features, *i.e.*,
+:math:`\bar{\mathbf{s}} = [\bar{s}_1, \dots, \bar{s}_{m_s}]^\top`, where
 :math:`\bar{s}_j` is the mean of the :math:`j\text{-th}` sensitive feature.
 
-For each non-sensitive feature :math:`z_j` in :math:`Z`,
-we compute an optimal weight vector :math:`w_j^* \in \mathbb{R}^{m_s}` that minimizes the
+For each non-sensitive feature :math:`\mathbf{z}_j\in\mathbb{R}^n`, where :math:`j=1,\dotsc,m_{ns}`,
+we compute an optimal weight vector :math:`\mathbf{w}_j^* \in \mathbb{R}^{m_s}` that minimizes the
 following least squares objective:
 
 .. math::
 
-    \min _{w} \| z_j - (S - \bar{S}) w \|_2^2
+    \min _{\mathbf{w}} \| \mathbf{z}_j - (\mathbf{S} - \bar{\mathbf{s}}) \mathbf{w} \|_2^2
 
-In other words, :math:`w_j^*` is the solution to a linear regression problem where we project
-:math:`z_j` onto the centered sensitive attributes. The weight matrix
-:math:`W^* = (w_1^*, \dots, w_{m_{ns}}^*)` is thus obtained by solving this regression
+In other words, :math:`\mathbf{w}_j^*` is the solution to a linear regression problem where we project
+:math:`\mathbf{z}_j` onto the centered sensitive features. The weight matrix
+:math:`\mathbf{W}^* = (\mathbf{w}_1^*, \dots, \mathbf{w}_{m_{ns}}^*)` is thus obtained by solving this regression
 for each non-sensitive feature.
 
-Once we have the optimal weight matrix :math:`W^*`, we compute the **residual
-non-sensitive attributes** :math:`Z^*` as follows:
+Once we have the optimal weight matrix :math:`\mathbf{W}^*`, we compute the **residual
+non-sensitive features** :math:`\mathbf{Z}^*` as follows:
 
 .. math::
 
-    Z^* = Z - (S - \bar{S}) W^*
+    \mathbf{Z}^* = \mathbf{Z} - (\mathbf{S}-\mathbf{1}_n\times\bar{\mathbf{s}}^\top) \mathbf{W}^*
 
+where :math:`\mathbf{1}_n` is a the all-one vector in :math:`\mathbb{R}^n`.
 
-The columns in :math:`S` will be dropped from the dataset :math:`X`, and :math:`Z^*` will replace
-the original non-sensitive features :math:`Z`, but the hyper parameter :math:`\alpha`
+The columns in :math:`\mathbf{S}` will be dropped from the dataset :math:`\mathbf{X}`, and :math:`\mathbf{Z}^*` will replace
+the original non-sensitive features :math:`\mathbf{Z}`, but the hyper parameter :math:`\alpha`
 does allow you to tweak the amount of filtering that gets applied:
 
 .. math::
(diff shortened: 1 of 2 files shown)


hildeweerts:
Many thanks for thinking this through and making a PR @taharallouche. This formulation follows the code more closely and is hopefully easier to understand.

To further enhance clarity, it might be useful to briefly explain some of the variables before you introduce the mathematical formulation (e.g. $W$ the vector of coefficients of the LR model, $\bar{S}$ average of $S$, and $Z^*$ the transformed non-sensitive features) and possibly add a sentence in between $min_W$ and $\texttt{let}  Z^*$ to make clear these are two separate steps.

@MiroDudik @koaning were involved in the original PR of this method and may have thoughts!

taharallouche:
@hildeweerts I hope these changes address your suggestions. I had to go more into details to accurately describe what [numpy's `lstsq`](https://numpy.org/doc/stable/reference/generated/numpy.linalg.lstsq.html) is doing under the hood (solving the least squares column-wise).

I'll be happy to make any other changes (including rolling back to a higher-level description :sweat_smile: ).

docs/user_guide/mitigation/preprocessing.rst
```diff
@@ -20,23 +20,43 @@ Sensitive features can be correlated with non-sensitive features in the dataset.
...
-:code:`alpha` parameter. In mathematical terms, assume we have the original dataset
-:math:`X` which contains a set of sensitive attributes :math:`S` and a set of
-non-sensitive attributes :math:`Z`. The removal of correlation is then
-described as:
+:code:`alpha` parameter.
+
+In mathematical terms, assume we have the original dataset :math:`X`, which
+contains a set of **sensitive attributes** denoted by :math:`S` and a set of
+**non-sensitive attributes** denoted by :math:`Z`. The goal is to remove
+correlations between the sensitive attributes and the non-sensitive attributes.
```
MiroDudik, line 28:
```suggestion
contains a set of **sensitive features** denoted by :math:`S` and a set of
**non-sensitive features** denoted by :math:`Z`. The goal is to remove
correlations between the sensitive and the non-sensitive features.
```

docs/user_guide/mitigation/preprocessing.rst
```diff
@@ -20,23 +20,43 @@ Sensitive features can be correlated with non-sensitive features in the dataset.
...
+
+Let :math:`m_s` and :math:`m_{ns}` denote the number of sensitive and non-sensitive
+features, respectively.
+Let :math:`\bar{S}` represent the mean of the sensitive attributes, *i.e.*,
```
MiroDudik, line 32: Let's use "sensitive features" and "non-sensitive features" rather than attributes.

The feature mean is a vector, not a matrix. It might be better to write it as `\bar{\mathbf{s}}`, and it would be probably more standard if it was a column vector (although its dimension is the number of columns of the matrix S).

For the vectors, I'd go with the boldface, e.g., `\mathbf{s}` (but the same would apply to all other vectors in this PR). For consistency, I think that we should do the same for the matrices and write `\mathbf{S}` and `\mathbf{Z}`. What do you think?

docs/user_guide/mitigation/preprocessing.rst
```diff
@@ -20,23 +20,43 @@ Sensitive features can be correlated with non-sensitive features in the dataset.
...
+:math:`\bar{S} = (\bar{s}_1, \dots, \bar{s}_{m_s})`, where
+:math:`\bar{s}_j` is the mean of the :math:`j\text{-th}` sensitive feature.
+
+For each non-sensitive feature :math:`z_j` in :math:`Z`,
```
MiroDudik, line 36: Maybe we can say something like `\mathbf{z}_j\in\mathbb{R}^n` where `j=1,\dotsc,m_{ns}` to drive home the dimension of `\mathbf{z}_j`.

docs/user_guide/mitigation/preprocessing.rst
```diff
@@ -20,23 +20,43 @@ Sensitive features can be correlated with non-sensitive features in the dataset.
...
+
+In mathematical terms, assume we have the original dataset :math:`X`, which
+contains a set of **sensitive attributes** denoted by :math:`S` and a set of
+**non-sensitive attributes** denoted by :math:`Z`. The goal is to remove
+correlations between the sensitive attributes and the non-sensitive attributes.
+
+Let :math:`m_s` and :math:`m_{ns}` denote the number of sensitive and non-sensitive
+features, respectively.
+Let :math:`\bar{S}` represent the mean of the sensitive attributes, *i.e.*,
+:math:`\bar{S} = (\bar{s}_1, \dots, \bar{s}_{m_s})`, where
```
MiroDudik, line 33: If we go with the column vector for `\bar{\mathbf{s}}`, then we may want to write it as `[\bar{s}_,1, \dots, \bar{s}_{m_s}]^\top`.

docs/user_guide/mitigation/preprocessing.rst
```diff
@@ -20,23 +20,43 @@ Sensitive features can be correlated with non-sensitive features in the dataset.
...
-        \frac{1}{n} \sum_{i=1}^{n} \mathbf{z}_{i}\left(\mathbf{s}_{i}-\overline{\mathbf{s}}
-        \right)^{T}=\mathbf{0}
-
-The solution to this problem is found by centering sensitive features, fitting a
-linear regression model to the non-sensitive features and reporting the residual.
-The columns in :math:`S` will be dropped from the dataset :math:`X`.
-The amount of correlation that is removed can be controlled using the
-:code:`alpha` parameter. This is described as follows:
+
+    \min _{w} \| z_j - (S - \bar{S}) w \|_2^2
```
MiroDudik, line 42: In numpy, this works thanks to broadcasting, but here we have a dimension mismatch, so we would need to write `\mathbf{S}-\mathbf{1}\bar{\mathbf{s}}^\top` where `\mathbf{1}` is the all-ones vector in `\mathbb{R}^n`.

MiroDudik (commented):
Some clarity suggestions to incorporate.

taharallouche:
Hi @MiroDudik thanks for the review ! I believe I addressed all your suggestions (I just took the liberty to use `\mathbf{1}_n` to emphasize that it is a vector of `n` ones, all the rest follows the review) :ok_hand: 

PS: the flaky test is breaking the CI pipeline, again :pensive:

TamaraAtanasoska:
> Hi @MiroDudik thanks for the review ! I believe I addressed all your suggestions (I just took the liberty to use `\mathbf{1}_n` to emphasize that it is a vector of `n` ones, all the rest follows the review) 👌
> 
> PS: the flaky test is breaking the CI pipeline, again 😔

hey yes unfortunately those problems are known :( For a while it was better as the server response improved on the OpenML side and I implemented caching, but I might need to check how persistent it is as it gets deleted after a while. One thing I can suggest is to push an empty commit with `git commit --allow-empty -m "Trigger Build"` and then a push. There were times when the server response is so bad it took me several tries, I couldn't get to the caching as it just won't download. It is much better than before atm, but if it becomes unmanageable again we can think of doing something more.

fairlearn/preprocessing/_correlation_remover.py
```diff
@@ -32,28 +32,51 @@ class CorrelationRemover(BaseEstimator, TransformerMixin):
     Notes
     -----
     This method will change the original dataset by removing all correlation with sensitive
-    values. To describe that mathematically, let's assume in the original dataset :math:`X`
-    we've got a set of sensitive attributes :math:`S` and a set of non-sensitive attributes
-    :math:`Z`. Mathematically this method will be solving the following problem.
+    values. To describe that mathematically, let's assume in the original dataset
+    :math:`\mathbf{X}` we've got a set of sensitive attributes :math:`\mathbf{S}` and a set of
+    non-sensitive attributes :math:`\mathbf{Z}`.
```
MiroDudik, line 37:
```suggestion
    features. To describe that mathematically, let's assume in the original dataset
    :math:`\mathbf{X}` we've got a set of sensitive features :math:`\mathbf{S}` and a set of
    non-sensitive features :math:`\mathbf{Z}`.
```

MiroDudik, line 37:
```suggestion
    features.
```

fairlearn/preprocessing/_correlation_remover.py
```diff
@@ -32,28 +32,51 @@ class CorrelationRemover(BaseEstimator, TransformerMixin):
...
+
+    In mathematical terms, assume we have the original dataset :math:`\mathbf{X}`, which
```
MiroDudik, line 39:
```suggestion
    features. In mathematical terms, assume we have the original dataset :math:`\mathbf{X}`, which
```

MiroDudik (commented):
This looks good. I just caught some leftover text that should have been deleted in the doc (see the suggested edit).

MiroDudik (approved):
This looks good from my side--I just noted one leftover text that should have been deleted in the doc (see the suggested change). Once that is incorporated this is good to go.

taharallouche:
> This looks good from my side--I just noted one leftover text that should have been deleted in the doc (see the suggested change). Once that is incorporated this is good to go.

Thanks for spotting that :pray: It's fixed now :ok_hand:

Issue #1419, opened 2024-10-24
MNT Adding statistical property/integrity tests for the adversarial mitigation module


TamaraAtanasoska:
The [adversarial mitigation module](https://github.com/fairlearn/fairlearn/tree/main/fairlearn/adversarial) has an acceptable level of test coverage. However, additional tests inspecting the statistical properties of the mitigation results are necessary to ensure statistical integrity of the results. 

Ideally whoever takes this issue would be looking into inspecting the results produced by the current methods, look for spaces for improvement or inconsistencies and propose tests that would make sure the future changes and development maintain those statistical properties.

adrinjalali:
I think generally speaking, we can add tests to see if mitigation techniques (not just adversarial), indeed improve on the metrics that they're optimising for, and if on train/test datasets they perform better. We can use these tests as a sort of a common test for all our methods then.

TamaraAtanasoska:
Sounds like a great idea, will add some result integrity and not to complex to start with.

DresdenGman:
I would like to extend this with a focused adversarial-mitigation integrity test. I reviewed the current suite and found a useful existing pattern in test/unit/preprocessing/test_preprocessing_fairness.py: compare a mitigated model with a LogisticRegression baseline on the same data, then assert improvement in the target fairness metric. The adversarial tests currently cover setup and execution, but I could not find one that checks whether training actually improves a fairness metric.

My proposed first scope is deliberately small:

- add one synthetic, fixed-seed dataset with a controlled group/feature correlation;
- train AdversarialFairnessClassifier with the torch backend and demographic_parity constraint;
- compare demographic_parity_difference against a LogisticRegression baseline on the same fixed split;
- assert improvement with a documented loose margin;
- fit twice with the same random_state and assert identical predictions, covering the existing reproducibility concern;
- keep it to one new unit-test file, no downloads and no production/API changes.

I would first calibrate the seed, margin, and runtime against the supported torch version so the test stays deterministic across CI platforms. If you would prefer the statistical form already used by the preprocessing test, I can instead factor/reuse its seeded bootstrap plus one-sided Wilcoxon pattern. Does the smaller deterministic margin test sound like the right first extension, or would you prefer the bootstrap form from the outset?

DresdenGman:
Follow-up after calibrating the proposed Torch integrity test:

[...]

PR #1439, opened 2024-11-16
Fix correlation remover none case

taharallouche:
## Description
Hi again ! This PR is all about the `CorrelationRemover`, and it basically does three things:
- Fix the default `None` `sensitive_feature_ids` value, which yields a `TypeError: 'NoneType' object is not iterable`.
- Adds an input validation checking the compatibility of `sensitive_feature_ids` and `X` , it has the same motivation as #1437, namely hiding the internals of our class (here it's the `lookup`) from the user.
- I refactored the existing tests, added tests for the `alpha` parameter and the new features that this PR introduces.

In the next PR, I'm planning to give #1419 a try on this mitigation method.

## Tests

- [x] no new tests required
- [x] new tests added
- [x] existing tests adjusted

## Documentation

## Screenshots
This is the exception raised when using the `None` default value
image

This the exception raised when some `sensitive_feature_ids` are not found in `X`
image

### DIFF
--- a/fairlearn/preprocessing/_correlation_remover.py
+++ b/fairlearn/preprocessing/_correlation_remover.py
@@ -86,6 +86,7 @@ class CorrelationRemover(BaseEstimator, TransformerMixin):
 
     def fit(self, X, y=None):
         """Learn the projection required to make the dataset uncorrelated with sensitive columns."""  # noqa: E501
+        self._check_sensitive_features_in_X(X)
         self._create_lookup(X)
         X = self._validate_data(X)
         X_use, X_sensitive = self._split_X(X)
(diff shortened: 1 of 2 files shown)


fairlearn/preprocessing/_correlation_remover.py
```diff
@@ -61,7 +63,7 @@ class CorrelationRemover(BaseEstimator, TransformerMixin):
 
     """
 
-    def __init__(self, *, sensitive_feature_ids=None, alpha=1):
+    def __init__(self, *, sensitive_feature_ids: Iterable = (), alpha: float = 1):
```
taharallouche, line 66: At first I tried to leave the `None` unchanged and mutate it to an empty list within the `__init__`, but the `sklearn` compatibility check didn't like that :sweat_smile: 

So I put an empty `tuple` (for immutability) and allowed any `Iterable` instead of restricting the argument to be a list (I double checked that the code can indeed handle any `Iterable`).

fairlearn/preprocessing/_correlation_remover.py
```diff
@@ -113,11 +116,26 @@ def transform(self, X):
...
+
     def _more_tags(self):
         return {
             "_xfail_checks": {
                 "check_transformer_data_not_an_array": (
                     "this estimator only accepts pandas dataframes or numpy ndarray as input."
                 ),
+                "check_estimators_empty_data_messages": (
+                    "this estimator raises on missing sensitive features in data."
+                ),
```
taharallouche, line 139: The `Estimator` can handle empty data fine. But we still throw an exception in case the data is empty and the `sensitive_feature_ids` are not.

fairlearn/preprocessing/_correlation_remover.py
```diff
@@ -113,11 +116,26 @@ def transform(self, X):
         X_filtered = np.atleast_2d(X_filtered)
         return self.alpha * X_filtered + (1 - self.alpha) * X_use
 
+    def _check_sensitive_features_in_X(self, X) -> None:
+        """Check if the sensitive features are in X."""
+        if isinstance(X, pd.DataFrame):
+            missing_columns = [c for c in self.sensitive_feature_ids if c not in X.columns]
+        else:
+            if X.ndim == 1:
```
adrinjalali, line 124: I'm surprised this doesn't fail since at this point we cannot be sure that X is an array since it's not processed.

taharallouche: Indeed I only handled pandas and numpy objects, but that was mainly because I saw it would [break earlier](https://github.com/fairlearn/fairlearn/blob/main/fairlearn/preprocessing/_correlation_remover.py#L82) if the input has another type

PR #1442, opened 2024-11-23
test: correlation remover enhances fairness metrics

taharallouche:
## Description
Hi @TamaraAtanasoska 

This is my first attempt to address #1419. Even though this PR only covers the `CorrelationRemover` and the `adult` dataset, the test function itself (and the `FairnessComparisonData` class) is agnostic to both the mitigation method and the dataset it is applied to, which can hopefully make it easily extendible. 

## Tests

- [ ] no new tests required
- [x] new tests added
- [ ] existing tests adjusted

## Documentation

### DIFF
--- a/test/unit/preprocessing/linear_dep_remover/test_correlation_remover_efficiency.py
+++ b/test/unit/preprocessing/linear_dep_remover/test_correlation_remover_efficiency.py
@@ -75,47 +75,50 @@ def _get_adult_fairness_comparison_data() -> FairnessComparisonData:
     )
 
 
-ADULT_FAIRNESS_COMPARISON_DATA = _get_adult_fairness_comparison_data()
-
-
 @pytest.mark.parametrize(
-    "fairness_metric",
-    [
-        demographic_parity_difference,
-        equal_opportunity_difference,
-        equalized_odds_difference,
-    ],
+    "fairness_metrics",
+    [[demographic_parity_difference, equal_opportunity_difference, equalized_odds_difference]],
 )
-@pytest.mark.parametrize("data", (ADULT_FAIRNESS_COMPARISON_DATA,))
+@pytest.mark.parametrize("fairness_comparison_data_loader", [_get_adult_fairness_comparison_data])
 def test_correlation_remover_mitigates_bias(
-    data: FairnessComparisonData, fairness_metric: Callable
+    fairness_comparison_data_loader: Callable[[], FairnessComparisonData],
+    fairness_metrics: list[Callable],
 ) -> None:
+    data = fairness_comparison_data_loader()
+
     bootstrap_iterations = 30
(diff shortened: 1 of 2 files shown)


test/unit/preprocessing/linear_dep_remover/test_correlation_remover_efficiency.py
```diff
@@ -0,0 +1,121 @@
...
+
+    return FairnessComparisonData(
+        sensitive_feature=X_test[senstive_feature_id],
+        y_true=y_test,
+        y_pred_without_mitigation=estimator_with_correlation.predict(X_test),
+        y_pred_with_mitigation=estimator_without_correlation.predict(X_test_transformed),
+    )
+
+
+ADULT_FAIRNESS_COMPARISON_DATA = _get_adult_fairness_comparison_data()
```
adrinjalali, line 78: I don't think we should be fitting in the get method here. In effect, this runs the `fit` during `import` time, which I think is not a good idea. I'd be happy with the function to return the data which is then to be used inside the test methods.

adrinjalali (commented):
Otherwise this is quite neat!

test/unit/preprocessing/linear_dep_remover/test_correlation_remover_efficiency.py
```diff
@@ -0,0 +1,124 @@
...
+
+            fairness_metrics_values_with_mitigation[i, iteration] = metric_with_mitigation
+            fairness_metrics_values_without_mitigation[i, iteration] = metric_without_mitigation
+
+    for metric in range(len(fairness_metrics)):
+        _, p_value = wilcoxon(
+            fairness_metrics_values_without_mitigation[metric],
+            fairness_metrics_values_with_mitigation[metric],
+            alternative="greater",
+        )
```
taharallouche, line 122: Should I rename `fairness_metrics` to `unfairness_metrics` or is it clear enough ?

It's kinda confusing that I'm asserting that the "fairness" metrics without mitigation are greater than those with mitigation :thinking:

test/unit/preprocessing/linear_dep_remover/test_correlation_remover_efficiency.py
taharallouche: Thanks @adrinjalali that's a great catch indeed :100: 

What I did to fix it is:
- pass the data preparation function to the test function, so that the fit (of both the transformer and the estimators) happens within it. It also keeps the test extendible: one just has to write a new function that returns a `FairnessComparisonData` and add it to the test parameters.
- to avoid unnecessarily repeating the fitting, I now compare all the metrics in one go. The test is still easily extendible in that regard also.

adrinjalali (approved):
Nice, and it runs under 3s on my machine (we should keep an eye on the test runtimes)

PR #1445, opened 2024-11-27
Plot thresold optimizer weighted objective

taharallouche:
## Description
Hello, this PR is mainly about the user guide for the `ThresholdOptimizer` with constraints other than equalized-odds. It has two parts:
- The first one consists of small fixes to the docs (commented out code line, extra space, missing double-quotation...).
- The second one is more interesting: I added the curve of the objective weighted by the size of the groups to the plot so that the quantity we are maximizing is displayed, and the user can understand why the solution's vertical line happen to be at that exact spot.

> To ensure that the fairness constraint is upheld [ThresholdOptimizer](https://fairlearn.org/v0.11/api_reference/generated/fairlearn.postprocessing.ThresholdOptimizer.html#fairlearn.postprocessing.ThresholdOptimizer) simultaneously walks along all the curves while monitoring the objective value, and picks the spot on the x-axis with the optimal objective weighed by the sizes of the groups corresponding to each sensitive feature value.

## Tests

- [ ] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

By the way I would have expected that [these tests](https://github.com/fairlearn/fairlearn/blob/main/test/unit/postprocessing/test_plots.py#L67-77) would fail, but they don't, and I can't find the corresponding snapshot files anywhere :eyes: 
 
## Documentation

## Screenshots
Before:
image

After:
image

### DIFF
--- a/docs/user_guide/mitigation/postprocessing.rst
+++ b/docs/user_guide/mitigation/postprocessing.rst
@@ -93,7 +93,6 @@ true positive rate parity as the fairness constraint.
     ... )
     >>> categorical_transformer = Pipeline(
     ...     [
-    ...         # ("impute", SimpleImputer(strategy="most_frequent")),
     ...         ("ohe", OneHotEncoder(handle_unknown="ignore")),
     ...     ]
     ... )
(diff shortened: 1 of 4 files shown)


TamaraAtanasoska (approved):

TamaraAtanasoska:
I love these kinds of change ideas, thanks @taharallouche! I will look into the tests today and why they aren't performing as expected. I will leave someone to maybe look over from a conceptual perspective if this for some reason I don't see atm would be undesirable. cc @adrinjalali maybe?

TamaraAtanasoska:
> Nice!
> 
> The tests might not be failing cause matplotlib or pytest_mpl might not be installed and then the tests are skipped?

Unfortunately the libraries being present doesn't affect the tests, but I will find wha is up if anything and make a PR if there is something to be fixed :)

PR #1448, opened 2024-11-30
feat: vectorize interpolate curve

taharallouche:
## Description

This PR focuses on improvements to the `ThresholdOptimizer` and includes two main types of changes:

1. **Readability Enhancements**:
   - Added and fixed type hints.
   - Renamed variables for better clarity.
   - Updated what seems to be outdated documentation.

2. **Vectorization of the `_interpolate_curve` Function**:
   - From a very rough benchmark, this seems to speedup the function significanty (~ x10 times).
   - Imho, it simplifies understanding the algorithm.

## Tests

- [x] no new tests required
- [ ] new tests added
- [x] existing tests adjusted

## Documentation


fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -415,7 +424,7 @@ def _threshold_optimization_for_simple_constraints(self, sensitive_features, lab
             # Determine probability of current sensitive feature group based on data.
             p_sensitive_feature_value = len(group) / n
 
-            roc_convex_hull = _tradeoff_curve(
+            metrics_curve_convex_hull = _tradeoff_curve(
```
taharallouche, line 427: This is not always the roc_curve's convex hull. 

It only coincides with the roc_curve if y is "true_positive_rate" and x is "false_positive_rate".

fairlearn/postprocessing/_tradeoff_curve_utilities.py
```diff
@@ -95,21 +103,20 @@ def _tradeoff_curve(
...
     Parameters
     ----------
     points_sorted : pandas.DataFrame
-        Points represented as rows with 'x' and 'y' columns, sorted by 'x'.
+        Points represented as rows with 'x' and 'y' columns, sorted lexicographically by 'x' and 'y'.
 
     Returns
     -------
-    result : pandas.DataFrame
+    result : list
```
taharallouche, line 119:
```suggestion
    result : pandas.DataFrame
```

fairlearn/postprocessing/_tradeoff_curve_utilities.py
```diff
@@ -139,17 +146,16 @@ def _filter_points_to_get_convex_hull(points_sorted):
...
 def _interpolate_curve(
     data: pd.DataFrame, x_col: str, y_col: str, content_col: str, x_grid: NDArray
-):
+) -> pd.DataFrame:
     """Interpolates the DataFrame in `data` along the values in `x_grid`.
 
-    Assumes: (1) data[y_col] is convex and non-decreasing in data[x_col]
-             (2) min and max in x_grid are below/above min and max in data[x_col]
-             (3) data is indexed 0,...,len(data)
+    Assumes: (1) data[y_col] is concave in data[x_col]
```
taharallouche, line 157: The function should be concave, it is the shape made by the curve with the x-axis that is convex, if I'm not mistaken

fairlearn/postprocessing/_tradeoff_curve_utilities.py
```diff
@@ -139,17 +146,16 @@ def _filter_points_to_get_convex_hull(points_sorted):
...
+             (2) min and max in x_grid are above and below min and max in data[x_col], respectively
```
taharallouche, line 158: You can check on the tests that if you set the upper-limit in x_grid is greater than the biggest score on the x-axis, it breaks.

fairlearn/postprocessing/_tradeoff_curve_utilities.py
```diff
@@ -169,50 +175,44 @@ def _interpolate_curve(
...
-                x_col: x,
-                y_col: y,
-                P0_KEY: p0,
-                content_col_0: data_transpose[i][content_col],
-                P1_KEY: p1,
-                content_col_1: data_transpose[i + 1][content_col],
-            }
-        )
+    indices = np.searchsorted(x_values, x_grid, side="right") - 1
+    indices = np.clip(indices, 0, len(x_values) - 2)
```
taharallouche, line 186: To handle the last element in `x_grid` (if it's equal to the biggest element in `x_values`)

fairlearn/postprocessing/_tradeoff_curve_utilities.py
```diff
@@ -260,11 +264,6 @@ def _calculate_tradeoff_points(
                 i += 1
             threshold = (threshold + scores[i]) / 2
 
-        # For the ROC curve we calculate points (x, y), where x represents
```
taharallouche, line 263: We are not always dealing with the ROC curve here.

fairlearn/postprocessing/_tradeoff_curve_utilities.py
```diff
@@ -139,17 +146,16 @@ def _filter_points_to_get_convex_hull(points_sorted):
...
+    return pd.DataFrame(selected)[[X_KEY, Y_KEY, OPERATION_KEY]]
 
 
 def _interpolate_curve(
     data: pd.DataFrame, x_col: str, y_col: str, content_col: str, x_grid: NDArray
-):
+) -> pd.DataFrame:
     """Interpolates the DataFrame in `data` along the values in `x_grid`.
 
-    Assumes: (1) data[y_col] is convex and non-decreasing in data[x_col]
```
taharallouche, line 150: It do not necessarily to be non-decreasing (see the [curve on the user-guide ](https://fairlearn.org/v0.11/user_guide/mitigation/postprocessing.html)for instance)

fairlearn/postprocessing/_tradeoff_curve_utilities.py
```diff
@@ -169,50 +175,47 @@ def _interpolate_curve(
...
-                y_col: y,
-                P0_KEY: p0,
-                content_col_0: data_transpose[i][content_col],
-                P1_KEY: p1,
-                content_col_1: data_transpose[i + 1][content_col],
-            }
-        )
+    indices = np.searchsorted(x_values, x_grid, side="right") - 1
+    indices = np.clip(indices, 0, len(x_values) - 2)
+    indices[1:-1] = np.where(
```
taharallouche, line 187: I only had to add this operation to not break the current tests. In principle the algorithm works without it, with a slightly different output that doesn't affect the behaviour of the `InterpolatedThresholder`.


Here's a detailed description of the change that break some tests (but does not affect the final results of the thresholding):

The difference only concerns cases where interpolation is not required:
- **Current Code (Main Branch), and this PR's code**:
  - If `x` (in `x_grid`) equals a value in `x_metric` at index `i`, the "interpolated" operations set `p0=0` (corresponding to the threshold at index i-1) and `p1=1` (corresponding to threshold at index i).
  - Exception: When `x` in `x_grid` coincides with `x_metric[0]` and `x_metric[1] > x_metric[0]`, it is the other way around.

- **This PR's code if we remove the `np.where`**:
  - If `x` (in `x_grid`) equals a value in `x_metric` at index `i`, the operations set `p0=1` (corresponding to threshold at index `i`) and `p1=0` (corresponding to the threshold at index `i+1`).
  - Exception: When `x` in `x_grid` coincides with the last element in `x_metric`, it is the other way around.

Notive how this does not affect the final output since the threshold with probability 1 remains unchanged, and the threshold with probability 0 does not matter. The only test expectations that would need to be updated are for test cases where there is no interpolation of operations/thresholds and where `x_metric[1] > x_metric[0]`.

adrinjalali (commented):
I like the PR overall, but the vectorised code is not as readable as the existing code. Could you please add enough inline comments as to explain what each line does? Like why `searchsorted() - 1`, etc.

fairlearn/postprocessing/_constants.py
```diff
@@ -6,6 +6,11 @@
 SENSITIVE_FEATURE_KEY = "sensitive_feature"
 P0_KEY = "p0"
 P1_KEY = "p1"
+OPERATION_KEY = "operation"
+OPERATION0_KEY = "operation0"
+OPERATION1_KEY = "operation1"
+X_KEY = "x"
+Y_KEY = "y"
```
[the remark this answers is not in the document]
taharallouche: No problem I can remove them :+1: 

My idea here was not to increase the readability per se, but to "centralize" these constant plain strings that are used across functions and files in a common place (more like for consistency).

fairlearn/postprocessing/_tradeoff_curve_utilities.py
taharallouche: @adrinjalali what do you think about this ? Should I keep this `np.where` just to keep the test expectation strictly unchanged (more details in my comment above)?

taharallouche:
If the new function `_get_interpolation_indices`  is good enough, I can add some unit tests specifically for it, which might also be helpful for anyone that would try to understand the details of the algorithm and its edge cases

taharallouche:
ping @adrinjalali (sorry I am not sure if you're busy or waiting for me to ping to you after addressing the comments)

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -454,12 +458,12 @@ def _threshold_optimization_for_simple_constraints(self, sensitive_features, lab
...
+            best_interpolation = self._tradeoff_curve[sensitive_feature_value].iloc[i_best]
             interpolation_dict[sensitive_feature_value] = Bunch(
-                p0=best_interpolation.p0,
-                operation0=best_interpolation.operation0,
-                p1=best_interpolation.p1,
-                operation1=best_interpolation.operation1,
+                p0=best_interpolation[P0_KEY],
+                operation0=best_interpolation["operation0"],
+                p1=best_interpolation[P1_KEY],
+                operation1=best_interpolation["operation1"],
```
adrinjalali, line 466:
```suggestion
                p0=best_interpolation.p0,
                operation0=best_interpolation.operation0,
                p1=best_interpolation.p1,
                operation1=best_interpolation.operation1,
```

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -31,6 +31,8 @@
     BASE_ESTIMATOR_NOT_FITTED_WARNING,
     LABEL_KEY,
     OUTPUT_SEPARATOR,
+    P0_KEY,
+    P1_KEY,
```
adrinjalali, line 35: I'd just remove these

adrinjalali (commented):
nits, otherwise LGTM

adrinjalali (approved):
LGTM. I'd let @TamaraAtanasoska have a look and merge if she's happy with this.

TamaraAtanasoska (approved):
Engineering-wise this looks good and it looks to me like it is not moving away from the original implementation, however @romanlutz wrote this many years ago and I'd love a last yes from him if he finds the time. Let's give him a bit of time :) @romanlutz pls merge if you are in agreement.

romanlutz (approved):
I sadly don't have time to dig into this deeper right now, but from a high-level point of view the optimizations/simplifications seem sensible. I would have loved to check the details, but don't want to hold this back either. Did you check examples for both types of plots (ROC plots and accuracy plots)? I'd hope they both still work as before.

fairlearn/postprocessing/_tradeoff_curve_utilities.py
taharallouche: That's what I thought too :sweat_smile:

taharallouche:
Thanks for checking this nonetheless @romanlutz I appreciate it :pray: 

I addressed your review above, and I did run a couple of manual tests to double-check that the curves are unchanged compared to `main` for different constraints :+1:

TamaraAtanasoska:
> Would this need a changelog entry ?

Yes, let's note the vectorisation/speedup. I believe it is ready to merge otherwise?

taharallouche:
> > Would this need a changelog entry ?
> 
> Yes, let's note the vectorisation/speedup. I believe it is ready to merge otherwise? 

Hi @TamaraAtanasoska I added the change log, and afaik this is ready to go

TamaraAtanasoska:
@taharallouche I'll take a look!

TamaraAtanasoska:
@riedgar-ms how does one tell codecov that these lines that it is marking [here](https://app.codecov.io/gh/fairlearn/fairlearn/pull/1448/indirect-changes) don't need to have test coverage? All of it is just part of a different sklearn version, and the tags processing is the `_fixes.py` copied from `sklearn-compat` where it has been tested. I don't have a login for the service, and not so much experience with it either.

taharallouche:
> @riedgar-ms how does one tell codecov that these lines that it is marking [here](https://app.codecov.io/gh/fairlearn/fairlearn/pull/1448/indirect-changes) don't need to have test coverage? All of it is just part of a different sklearn version, and the tags processing is the `_fixes.py` copied from `sklearn-compat` where it has been tested. I don't have a login for the service, and not so much experience with it either.

Hi @TamaraAtanasoska, I believe it's achievable by changing the codecov.yml accordingly (we can add an 'ignore' field)

TamaraAtanasoska:
@taharallouche as far as I look at the online info it seems it is only for whole files, I'd definitely not want that for the adversarial code it complains about :S maybe just the fixes file being ignored it should work to have it pass? Would you give it a try? thinking if the sklearn compatibility fixes should maybe renamed as such and moved to a file as other fixes might as well need tests. But that is for another PR out of scope here :)

TamaraAtanasoska:
@taharallouche `# pragma: no cover` seems like a great solution for this PR. Merging :)

PR #1450, opened 2024-12-03
fix: bootstrapping with missing sampled value

### LINKED ISSUE #1441: MetricFrame bootstrapping breaks if not all the sensitive feature's values are sampled
#### Describe the bug
When the data fed to a `MetricFrame` (with `n_boot` and `ci_quantiles` not set to `None`) is sampled for bootstraping, there's no guarantee that all the values the sensitive features takes will be represented in the sampled data.

If it happens that in some samples, some values are left out, the `MetricFrame` object instantiation breaks.

The probability of this occurring would depend on the distribution of the sensitive features in the data, and it would be more likely to happen if some "minority" classes are present in the data.

#### Steps/Code to Reproduce
Example:
```python
y_true = [0, 0, 1, 1, 1, 1, 1, 1, 1, 1, 0]
y_pred = [0, 0, 1, 1, 1, 1, 1, 1, 1, 1, 0]
sensitive_features = ["a", "b", "c", "a", "b", "c", "a", "b", "c", "a", "b"]

random_state = np.random.RandomState(1234)


mf = MetricFrame(
    metrics={"selection_rate": selection_rate},
    y_true=y_true,
    y_pred=y_pred,
    sensitive_features=sensitive_features,
    ci_quantiles=[0.1, 0.5, 0.9],
    n_boot=3,
    random_state=random_state,
)
```

#### Expected Results
The MetricFrame object is created.

#### Actual Results
Here's the error stack:
```
  File "/home/allouche/repos/fairlearn/fairlearn/metrics/experimental_bootstrap.py", line 21, in <module>
    mf = MetricFrame(
         ^^^^^^^^^^^^
  File "/home/allouche/repos/fairlearn/fairlearn/metrics/_metric_frame.py", line 309, in __init__
    self._populate_results_ci(_bootstrap_samples, ci_quantiles)
  File "/home/allouche/repos/fairlearn/fairlearn/metrics/_metric_frame.py", line 406, in _populate_results_ci
    result_group = calculate_pandas_quantiles(
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/allouche/repos/fairlearn/fairlearn/metrics/_bootstrap.py", line 137, in calculate_pandas_quantiles
    result = _calc_dataframe_quantiles(quantiles=quantiles, samples=bootstrap_samples)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/allouche/repos/fairlearn/fairlearn/metrics/_bootstrap.py", line 112, in _calc_dataframe_quantiles
    assert all(s.index == samples[0].index), "Sanity check shape of bootstrap sample"
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/allouche/repos/fairlearn/.venv/lib/python3.12/site-packages/pandas/core/ops/common.py", line 76, in new_method
    return method(self, other)
           ^^^^^^^^^^^^^^^^^^^
  File "/home/allouche/repos/fairlearn/.venv/lib/python3.12/site-packages/pandas/core/arraylike.py", line 40, in __eq__
    return self._cmp_method(other, operator.eq)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/allouche/repos/fairlearn/.venv/lib/python3.12/site-packages/pandas/core/indexes/base.py", line 7185, in _cmp_method
    raise ValueError("Lengths must match to compare")
ValueError: Lengths must match to compare
```

Tracing back the cause of the error, I see that one of the samples doesn't include any row with a "b" sensitive feature:
```
    y_true  y_pred sensitive_feature_0
0        1       1                   a
1        1       1                   c
2        0       0                   a
3        1       1                   a
4        0       0                   a
5        1       1                   c
6        0       0                   a
7        1       1                   c
8        1       1                   c
9        1       1                   c
10       1       1                   c
```

#### Fix Suggestion
I would suggest that instead of bootstrapping the whole dataset, we bootstrap each group within the sensitive feature alone, and we concatenate the results. So basically, doing "stratified sampling".

This fix would break reproducibility though (with the same data and random seed, a `fairlearn` user that has an older version would get different results).

#### Versions
- OS: Ubuntu 24.04.1 LTS
- Python version: 3.12.3
- Fairlearn version: https://github.com/fairlearn/fairlearn/commit/8ce6e35b822929c30d4a1ab732f3c0fc3a0b384b

taharallouche:
Fixes #1441 (Notice, that I do not implement the stratified sampling here, as it would be a new feature that could go intro a dedicated PR)

## Description
Hello this my attempt to implement the [suggested](https://github.com/fairlearn/fairlearn/issues/1441#issuecomment-2504665116) solution to fix bootstrapping:

> When the bootstrap replicate does not contain any elements from a given group, it should return np.nan, and we should use np.nanquantile to get the values of various *_ci quantities. This will yield valid quantiles (under the same assumptions as the standard bootstrap). I don't think that we should allow substituting some other default value (like 0 or 1) in this case, because that is likely to lead to wrong quantiles. This approach would be also backward compatible re. random seeds. One possible downside is that *_ci might sometimes still include np.nan.

NB: With this implementation, if a a group happens to be missing from **all the samples**, it will simply not appear in the `ci` attributes. If we absolutely want to make it appear with a `NaN` , I can think of a couple of ways to achieve it (for instance, aligning the groups of each `*_ci` attributes with those of the corresponding "non-ci" attribute of the metric frame).

## Tests

- [ ] no new tests required
- [x] new tests added
- [ ] existing tests adjusted

## Documentation

### DIFF
--- a/fairlearn/metrics/_bootstrap.py
+++ b/fairlearn/metrics/_bootstrap.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import logging
+from functools import reduce
 from typing import List
 
 import numpy as np
@@ -106,13 +107,15 @@ def _calc_series_quantiles(*, quantiles: list[float], samples: list[pd.Series]) 
 def _calc_dataframe_quantiles(
     *, quantiles: list[float], samples: list[pd.DataFrame]
 ) -> list[pd.DataFrame]:
+    samples = _align_sample_indices(samples)
+
     for s in samples:
         assert isinstance(s, pd.DataFrame)
         assert all(s.columns == samples[0].columns)
         assert all(s.index == samples[0].index), "Sanity check shape of bootstrap sample"
 
     try:
-        result_np = np.quantile(samples, q=quantiles, axis=0)
+        result_np = np.nanquantile(samples, q=quantiles, axis=0)
     except ValueError as ve:
         raise ValueError(BOOTSTRAP_QUANTILE_ERROR) from ve
 
@@ -127,8 +130,34 @@ def _calc_dataframe_quantiles(
     return result
 
 
+def _align_sample_indices(samples: list[pd.DataFrame]) -> list[pd.DataFrame]:
+    """
+    Align the indices of the bootstrapped DataFrames, so that all the combinations of sensitive
+    and control feature values are present in all the samples.
+
+    This is achieved by reindexing them to a common outer union of all indices, filling the missing
+    combinations with NaNs.
+
+    If a combination is missing from all the samples, it won't appear in the common index.
+
+    Parameters
+    ----------
+    samples : list[pd.DataFrame]
+        A list of pandas DataFrames to be aligned.
+
+    Returns
+    -------
+    list[pd.DataFrame]
+        A list of pandas DataFrames with aligned indices.
+    """
+    all_indices = [sample.index for sample in samples]
+    outer_common_index = reduce(lambda x, y: x.union(y), all_indices)
+    samples = [sample.reindex(outer_common_index) for sample in samples]
+    return samples
+
+
 def calculate_pandas_quantiles(
-    quantiles: list[float], bootstrap_samples: list[pd.Series | pd.DataFrame]
+    quantiles: list[float], bootstrap_samples: list[pd.Series] | list[pd.DataFrame]
 ) -> list[pd.Series] | list[pd.DataFrame]:
     """Calculate quantiles for a list of pandas objects."""
     if isinstance(bootstrap_samples[0], pd.Series):
--- a/fairlearn/metrics/_bootstrap.py
+++ b/fairlearn/metrics/_bootstrap.py
@@ -63,12 +63,14 @@ def generate_bootstrap_samples(
     assert n_samples >= 1
     if random_state is None:
         generator = np.random.default_rng()
-        rs = generator.integers(low=0, high=np.iinfo(np.uint32).max, size=n_samples)
+        rs = generator.integers(low=0, high=np.iinfo(np.uint32).max, size=n_samples, dtype="int64")
     elif isinstance(random_state, np.random.RandomState):
-        rs = random_state.randint(low=0, high=np.iinfo(np.uint32).max, size=n_samples)
+        rs = random_state.randint(
+            low=0, high=np.iinfo(np.uint32).max, size=n_samples, dtype="int64"
+        )
     elif isinstance(random_state, int):
         generator = np.random.default_rng(seed=random_state)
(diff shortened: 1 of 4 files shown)


taharallouche:
The use of [high=np.iinfo(np.uint32).max](https://github.com/fairlearn/fairlearn/blob/main/fairlearn/metrics/_bootstrap.py#L65) seem to cause an issue on Windows, I'll try to debug

fairlearn/metrics/_bootstrap.py
```diff
@@ -62,12 +63,14 @@ def generate_bootstrap_samples(
     assert n_samples >= 1
     if random_state is None:
         generator = np.random.default_rng()
-        rs = generator.integers(low=0, high=np.iinfo(np.uint32).max, size=n_samples)
+        rs = generator.integers(low=0, high=np.iinfo(np.uint32).max, size=n_samples, dtype="int64")
```
taharallouche, line 66:
```suggestion
        rs = generator.integers(low=0, high=np.iinfo(np.uint32).max, size=n_samples, dtype=np.uint32)
```

test/unit/metrics/test_metricframe_bootstrap.py
```diff
@@ -576,3 +576,23 @@ def test_ci_properties_raise_on_uninitialized_bootstrap_params(
...
+    with does_not_raise():
+        MetricFrame(
+            metrics=count,
+            y_true=y_true,
+            y_pred=y_pred,
+            sensitive_features=sensitive_features,
+            ci_quantiles=[0.1, 0.5, 0.9],
+            n_boot=3,
+            random_state=random_state,
+        )
```
riedgar-ms, line 598: Can we check anything else? I'm sort of concerned that if the PRNG changes on us, then this will work because the above seed _won't_ produce a sample with a missing 'b' value.

riedgar-ms: Or only have a single 'b' in the sensitive features, which should make it much easier to hit the case of 'b' being absent

taharallouche: Done @riedgar-ms :pray: I double checked that the new input and random state do yield a bootstrapped sample that is missing "b".

riedgar-ms (approved):
I'm happy with this (after all, tests have only been added, not modified). Can another maintainer give it a look too please?

Issue #1446, opened 2024-11-28
MNT Make sure fairlearn is scikit-learn 1.6 compatible


TamaraAtanasoska:
`scikit-learn` has a new release candidate for `version 1.6` which introduces some API changes. There are some already some minor marked changes to be made, but a thorough check is necessary. 

I've been working on this with @adrinjalali for `skops` ([PR](https://github.com/skops-dev/skops/pull/447)), so I will also go through the issues that might show up here. The biggest challenge for now for all libraries that are doing the compatibility changes is to stay compatible with the older versions too. 

There is now also [sklearn-compat](https://github.com/glemaitre/sklearn-compat) to help with these efforts.

PR #1451, opened 2024-12-03
MNT scikit-learn 1.6 compatibility

TamaraAtanasoska:
## Description
This PR fixes https://github.com/fairlearn/fairlearn/issues/1446.
Pinging @adrinjalali for a review 🙏 .

@fairlearn/fairlearn-maintainers it would be great if we could soon make a release with these changes, as soon as `scikit-learn 1.6` is available.

Locally all tests for both versions pass, if CI comes up with anything else I will fix it asap.

## Tests

- [ ] no new tests required
- [ ] new tests added
- [x] existing tests adjusted

## Documentation

### DIFF
--- a/fairlearn/preprocessing/_correlation_remover.py
+++ b/fairlearn/preprocessing/_correlation_remover.py
@@ -2,15 +2,15 @@
 # Licensed under the MIT License.
 
 from collections.abc import Iterable
+from test.unit.fixes import validate_data
 
 import numpy as np
 import pandas as pd
 from sklearn.base import BaseEstimator, TransformerMixin
-from sklearn.utils import check_array
 from sklearn.utils.validation import check_is_fitted
 
 
-class CorrelationRemover(BaseEstimator, TransformerMixin):
+class CorrelationRemover(TransformerMixin, BaseEstimator):
     r"""
     A component that filters out sensitive correlations in a dataset.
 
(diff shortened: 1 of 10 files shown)


test/unit/fixes.py
```diff
@@ -2,12 +2,76 @@
...
     # Handle change of message for sklearn
     if sklearn.__version__ < "0.23.0":
         expected_message = "bad input shape"
     else:
-        expected_message = "y should be a 1d array"
+        expected_message = "Data must be 1-dimensional"
     return expected_message
+
+
+def validate_data(estimator, X, **kwargs):
```
adrinjalali, line 17: I think these new ones are better off under a `fairlearn/utils/_fixes.py` file instead of here.

fairlearn/preprocessing/_correlation_remover.py
```diff
@@ -2,15 +2,15 @@
...
 
 import numpy as np
 import pandas as pd
 from sklearn.base import BaseEstimator, TransformerMixin
-from sklearn.utils import check_array
 from sklearn.utils.validation import check_is_fitted
 
 
-class CorrelationRemover(BaseEstimator, TransformerMixin):
+class CorrelationRemover(TransformerMixin, BaseEstimator):
```
adrinjalali, line 13: I'm quite happy with the test that I wrote which is making us fix these 😁

test/unit/fixes.py
```diff
@@ -2,12 +2,76 @@
...
+    try:
+        from sklearn.utils.validation import validate_data
+
+        return validate_data(estimator, X, **kwargs)
+    except ImportError:
+        # sklearn version < 1.6
+        return estimator._validate_data(X, **kwargs)
+
+
+def _check_array(X, **kwargs):
```
adrinjalali, line 50: why call it `_check_array`. I'd call it `check_array` so that only the `import` statement in other places is changed and not the actual code.

test/unit/preprocessing/linear_dep_remover/test_sklearn_compat.py
```diff
@@ -7,10 +7,31 @@
...
+
+
+def check_extended(estimator, check, expected_failed_checks=None):
+    try:
+        check_estimator(
+            estimator, expected_failed_checks=EXPECTED_FAILED_CHECKS[estimator.__class__.__name__]
+        )
+    except TypeError:
+        # sklearn version < 1.6
+        check(estimator)
```
adrinjalali, line 33: `check_estimator` runs the whole common test suite, not only a single test. We need to have a fix for `parametrize_with_checks` and not `check_estimator`

adrinjalali (commented):
writing a nice wrapper for `parametrize_with_checks` is not trivial, I need to add it to `sklearn_compat`

test/unit/fixes.py
TamaraAtanasoska: yeah this was def a questionable decision I agree, I wanted to leave obvious breadcrumbs for others as this is a very used function, but I can leave a comment at import level

test/unit/preprocessing/linear_dep_remover/test_sklearn_compat.py
TamaraAtanasoska: just for transparency if anyone else peaks here: I misunderstood @adrinjalali when we discussed this change, it needs a bigger implementation. they will work on it in `sklearn_compat` and then we can take an example and implement it for our use case.

TamaraAtanasoska:
@adrinjalali this is ready for a review again with the tests fix

fairlearn/preprocessing/_correlation_remover.py
```diff
@@ -145,16 +159,21 @@ def _check_sensitive_features_in_X(self, X) -> None:
...
                 ),
-                "check_estimators_empty_data_messages": (
-                    "this estimator raises on missing sensitive features in data."
-                ),
             }
         }
+
+    def __sklearn_tags__(self):
+        tags = super().__sklearn_tags__()
+        return tags
```
adrinjalali, line 179: With the new `parametrize_with_checks` you don't need to have this information here anymore so they can be removed

adrinjalali (commented):
Otherwise LGTM.

You might wanna add a changelog entry as well.

fairlearn/preprocessing/_correlation_remover.py
TamaraAtanasoska: I didn't not read with such attention the example code and that is on ME :D done in https://github.com/fairlearn/fairlearn/pull/1451/commits/25e647dfca52f9453daee75cde933aedf7762ef8

test/unit/preprocessing/linear_dep_remover/test_sklearn_compat.py
```diff
@@ -7,16 +7,25 @@
...
 from fairlearn.preprocessing import CorrelationRemover
+from fairlearn.utils._fixes import parametrize_with_checks
+
+EXPECTED_FAILED_CHECKS = {
+    "CorrelationRemover": {
+        "check_transformer_data_not_an_array": (
+            "this estimator only accepts pandas dataframes or numpy ndarray as input."
+        ),
+    },
+}
```
adrinjalali, line 20: sklearn compatibility tests and these dicts can probablay be moved to a single file, but a different PR.

TamaraAtanasoska: I have a new PR today when I remove this one, and the next one I should remove the remaining so we might not need to :) but the adapted code should stay there anyway for the future. thanks for reviewing and merging today 🥳

PR #1452, opened 2024-12-05
MNT Fix CorrelationRemover failing estimator checks

TamaraAtanasoska:
## Description
This PR removes resolves the failing estimator checks in the `Correlation Remover`.
The class could use some additional refactoring, added to my list for the future.

ping @adrinjalali this is what I mentioned in https://github.com/fairlearn/fairlearn/pull/1451

## Tests

- [ ] no new tests required
- [ ] new tests added
- [x] existing tests adjusted

## Documentation

### DIFF
--- a/fairlearn/preprocessing/_correlation_remover.py
+++ b/fairlearn/preprocessing/_correlation_remover.py
@@ -100,6 +100,7 @@ class CorrelationRemover(TransformerMixin, BaseEstimator):
             self.lookup_ = {c: i for i, c in enumerate(X.columns)}
             return X.values
         # correctly handle a 1d input
+        X = validate_data(self, X, ensure_2d=False, ensure_min_samples=0)
         if len(X.shape) == 1:
             return {0: 0}
         self.lookup_ = {i: i for i in range(X.shape[1])}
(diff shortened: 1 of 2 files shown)


fairlearn/preprocessing/_correlation_remover.py
```diff
@@ -155,6 +156,7 @@ def _check_sensitive_features_in_X(self, X) -> None:
         if isinstance(X, pd.DataFrame):
             missing_columns = [c for c in self.sensitive_feature_ids if c not in X.columns]
         else:
+            X = validate_data(self, X, ensure_2d=False, ensure_min_samples=0)
             if X.ndim == 1:
                 return
```
adrinjalali, line 161: so we don't really support 1D data, do we?

TamaraAtanasoska: I thought we do we just don't check for missing columns then? Is that reading wrong?

adrinjalali: we have `if X.ndim ==  1: return`, so we don't really support it?

TamaraAtanasoska: it still gets fitted(fit?! not sure the english use there :D) and passed to the `_create_lookup` which has a part where it says `#correctly handle a 1D input`([here](https://github.com/fairlearn/fairlearn/pull/1452/files#diff-aae694603fed4bca3298249f11ff7580b7d681165771c823ee59d392170f38fbL102)), so it is supported, just no need to check for missing columns. My checks are just for the the `_NotAnArray` case from the `estimator_checks`, otherwise it all worked well.

adrinjalali (approved):

Issue #1246, opened 2023-06-20
ENH relaxed fairness constraint fulfillment via postprocessing (currently only supports strict fulfillment)


AndreFCruz:
Hi fairlearn maintainers,

We’ve implemented a relaxation for the common postprocessing (threshold optimization) fairness method, that essentially enables optimizing for some loss (e.g., accuracy) with a maximum difference between groups’ TPRs or FPRs. 

The implementation is self-contained and available at https://github.com/andrefcruz/error-parity

There are also several example notebooks in the “examples” folder.

What would be a road map to have this included into fairlearn?

I guess the main options would be:
1. Changing the `ThresholdOptimizer` class;
	- Add an optional argument `tolerance: float = None`, and whenever that argument is passed the relaxed implementation would be used?
	- If no tolerance is passed, the previous thresholding implementation would be used (no need to run the LP solver).
2. Adding a new `RelaxedThresholdOptimizer` class;
	- Advantages: makes it clear that the backend and algorithm is different than `ThresholdOptimizer` (as the relaxation implies solving a linear program).
	- Disadvantages: potentially creates clutter in the API.

Other thoughts/questions:
- If going with option (1), do you think it'd be a simple if/else branch that falls back to the original implementation whenever `tolerance=None` (the default)?
- Currently only a relaxation on equalized odds (equal group-specific TPR and FPR) has been implemented. This is of course much less than the current ThresholdOptimizer class functionality; and the remaining functionality could be added in the near future.
	- Does this change the calculus between using the same `ThresholdOptimizer` class or a separate `RelaxedThresholdOptimizer`? Or can we just throw `NotImplementedError` whenever a specific unimplemented combination of kwargs is requested?

Thanks and all the best,
André

---
#### Is your feature request related to a problem? Please describe.

Currently it is only possible to achieve exact equality between group-specific TPR or FPR (or both). Having a relaxed constraint fulfillment can enable direct head-to-head comparison between different fair ML methods at the same level of fairness. Additionally, relaxed constraint fulfillment also enables us to map the whole fairness-accuracy Pareto front for a single given classifier.

#### Describe the solution you'd like

The solution for relaxed equalized odds has been implemented at https://github.com/AndreFCruz/error-parity/blob/main/error_parity/cvxpy_utils.py#L159

[...]

PR #1453, opened 2024-12-06
feat: relaxed threshold optimizer for simple constraints

taharallouche:
## Description
Hello again !

This is a small MR that aims to allow for a relaxed optimization of the threshold for simple constraints.

This is complementary to the work that has started in #1246 and #1248, and which only allows tolerance for `equalized odds`. The main difference though is that this does very minimal changes to the API, and doesn't require adding any new dependencies (no LP to solve).

Implementation-wise, I leveraged the fact that the `x_metric` values stored in the interpolated curves are the same (by construction) and are sorted in an ascending order for all the sensitive feature groups. \
The algorithm is a combination of a sliding window and a double-ended queue which makes it quite efficient ($O(n x m)$, where n is `x_grid` size and m is the number of sensitive groups).

This obviously lacking examples and user-guide etc.. But I wanted to first get some feedback from the maintainers before putting more effort into documentation. 

## Tests

- [ ] no new tests required
- [x] new tests added
- [ ] existing tests adjusted

## Documentation

## Screenshots
These are the different plots I get when running the current [user-guide example](https://fairlearn.org/v0.11/user_guide/mitigation/postprocessing.html) with different levels of `tol`.

### `tol = None (or 0)`
image

### `tol = 0.05`
Screenshot from 2024-12-05 17-25-19

### `tol = 0.5`
Screenshot from 2024-12-05 17-24-34

### DIFF
--- a/fairlearn/postprocessing/_plotting.py
+++ b/fairlearn/postprocessing/_plotting.py
@@ -132,15 +132,32 @@ def plot_threshold_optimizer(threshold_optimizer: ThresholdOptimizer, ax=None, s
             "$P[\\hat{Y}=1|Y=1]$",
         )
     else:
-        _plot_overall_tradeoff_curve(ax, threshold_optimizer._overall_tradeoff_curve)
-        _plot_solution(
-            ax,
-            threshold_optimizer._x_best,
-            None,
-            "solution",
-            threshold_optimizer.x_metric_,
-            threshold_optimizer.y_metric_,
-        )
+        if hasattr(threshold_optimizer, "_overall_tradeoff_curve"):
+            _plot_overall_tradeoff_curve(ax, threshold_optimizer._overall_tradeoff_curve)
+            _plot_solution(
+                ax,
+                threshold_optimizer._x_best,
+                None,
+                "solution",
+                threshold_optimizer.x_metric_,
+                threshold_optimizer.y_metric_,
+            )
+        else:
+            _plot_solution_with_tol(threshold_optimizer, ax)
 
     if show_plot:
         plt.show()
+
+
+def _plot_solution_with_tol(threshold_optimizer: ThresholdOptimizer, ax):
+    for sensitive_feature_value, x_best in threshold_optimizer._x_best_per_group.items():
+        color = _get_debug_color(sensitive_feature_value)
+        ax.axvline(
+            x=x_best,
+            label=f"constraint (sensitive feature = {sensitive_feature_value})",
+            ls="--",
+            c=color,
+        )
+        ax.legend()
+        ax.set_xlabel(threshold_optimizer.x_metric_)
+        ax.set_ylabel(threshold_optimizer.y_metric_)
--- a/fairlearn/postprocessing/_relaxed_constraints.py
+++ b/fairlearn/postprocessing/_relaxed_constraints.py
@@ -1,0 +1,73 @@
+from __future__ import annotations
+
+from collections import deque
+from typing import Iterable
+
+import pandas as pd
+
+
+def maximize_objective_with_tolerance(
+    dataframes: Iterable[pd.DataFrame],
+    weights: Iterable[float],
+    tol: float,
+    x_col: str = "x",
+    y_col: str = "y",
+) -> tuple[list[int], float]:
+    """
+    Finds the indices that maximize the sum of weighted `y` values across multiple DataFrames
+    under the constraint that the difference between the maximum and minimum `x` values
+    in the selected indices is at most `tol`.
+
+    Args:
+        dataframes (Iterable[pd.DataFrame]): An Iterable of DataFrames. Each DataFrame must have:
+                                         - An `x` column (sorted in ascending order).
+                                         - A `y` column.
+        weights (Iterable[float]): An Iterable of weights for each DataFrame `y` column.
+        tol (float): The maximum allowed range of `x` values for the selected indices.
+
+    Returns:
+        list[int], float: - A list of selected indices (one per DataFrame).
+                          - The maximum sum of weighted `y` values under the constraint.
+    """
+    # Extract `x` values (all DataFrames must share the same `x` values)
+    x_values = dataframes[0][x_col].values
+    n = len(x_values)
+
+    # Extract weighted `y` columns
+    y_columns = [df[y_col].values * weight for df, weight in zip(dataframes, weights)]
+    m = len(y_columns)
+
+    # Initialize a deque for each DataFrame to store indices in descending order of y values
+    deques = [deque() for _ in range(m)]
+
+    # Variables to track the best solution
+    best_indices = []
+    max_sum = float("-inf")
+
+    # Sliding window start pointer
+    start = 0
+
+    # Traverse with the end pointer
+    for end in range(n):
+        # Add the current `end` index to the deques, maintaining descending order of y values
+        for k in range(m):
+            while deques[k] and y_columns[k][deques[k][-1]] <= y_columns[k][end]:
+                deques[k].pop()
+            deques[k].append(end)
+
+        # Adjust the window to ensure the range of x values is within tol
+        while x_values[end] - x_values[start] > tol:
+            for k in range(m):
+                if deques[k] and deques[k][0] <= start:
+                    deques[k].popleft()
+            start += 1
+
+        # Calculate the sum of the maximum weighted `y` values in the current window
+        current_sum = sum(y_columns[k][deques[k][0]] for k in range(m))
+
+        # Update the best solution if the current sum is larger
+        if current_sum >= max_sum:
+            max_sum = current_sum
+            best_indices = [deques[k][0] for k in range(m)]
+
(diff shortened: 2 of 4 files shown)


fairlearn/postprocessing/_plotting.py
```diff
@@ -132,15 +132,32 @@ def plot_threshold_optimizer(threshold_optimizer: ThresholdOptimizer, ax=None, s
...
-        _plot_overall_tradeoff_curve(ax, threshold_optimizer._overall_tradeoff_curve)
-        _plot_solution(
-            ax,
-            threshold_optimizer._x_best,
-            None,
-            "solution",
-            threshold_optimizer.x_metric_,
-            threshold_optimizer.y_metric_,
-        )
+        if hasattr(threshold_optimizer, "_overall_tradeoff_curve"):
```
taharallouche, line 135: If it has an overall_tradeoff_curve, it means that there is no tolerance, so I use the vanilla plotting function

TamaraAtanasoska:
Hi @taharallouche, dropping by without are review at the moment to say that we'd like to do a release this week so the changes making `fairlearn` compatible with `scikit-learn 1.6` are out there in time. The work you've been doing will be a part of it. Could you make a small section about all the PRs you've done with a little summary for them as part of this PR?

- [here is where to make them now](https://github.com/fairlearn/fairlearn/blob/main/docs/user_guide/installation_and_version_guide/v0.12.0.rst)
- [example from last time](https://github.com/fairlearn/fairlearn/blob/main/docs/user_guide/installation_and_version_guide/v0.11.0.rst)

Thank you! 

pinging @fairlearn/fairlearn-maintainers for help with reviewing this PR

adrinjalali:
This is gonna take me a while to review. We could release w/o this one and include this in the next release.

TamaraAtanasoska:
> This is gonna take me a while to review. We could release w/o this one and include this in the next release.

makes sense, I agree. @taharallouche would you maybe consider what I [outlined in this comment ](https://github.com/fairlearn/fairlearn/pull/1453#issuecomment-2527599065)and push it to the other remaining open PR then?

TamaraAtanasoska:
> Hi @TamaraAtanasoska 👋 ok I'll add the relevent change logs (probably tomorrow, in the afternoon).
> 
> I'll open a dedicated PR to update the rst if that's okay

Great, thank you! I plan start doing the release on Wednesday morning, CET time (where I am) 😊 . If you don't make it, I will add this info for you, no worries. Ping on Discord if anything.

taharallouche:
NOTE TO FUTURE SELF: 
- [ ] Since the ROC curve is non-decreasing, check if the same sliding window idea can be leveraged to handle `equalized-odds` as well.

MiroDudik:
I'll need to double-check the logic of the algorithm to verify correctness. I agree that for the case of simple constraints, we should be able to just move the sliding window along the x axis, while monitoring the best y value for each of the groups (by a priority queue with an inverse-index lookup to support deletions as points leave the sliding window?).

More generally, I also coded up this reference implementation using reductions--it is quite slow, but I used it to double check the correctness of the error-parity package with equalized odds:
[example-postprocess.py.gz](https://github.com/user-attachments/files/18215043/example-postprocess.py.gz)

By the way, it seems that [error-parity](https://github.com/socialfoundations/error-parity) now also supports some of the simple constraints, but if your sliding window solution works, that would be a much simpler way to do it.

taharallouche:
Hi @MiroDudik  !

> move the sliding window along the x axis, while monitoring the best y value for each of the groups (by a priority queue with an inverse-index lookup to support deletions as points leave the sliding window?)

Exactly that's the idea.

> I also coded up this reference implementation using reductions--it is quite slow, but I used it to double check the correctness of the error-parity package with equalized odds:
[example-postprocess.py.gz](https://github.com/user-attachments/files/18215043/example-postprocess.py.gz)

Great this would be very helpful for me too to double check the method ! 

I'm currently experimenting with your script, and I noticed that the `ThresholdOptimizerExpGrad` considers the `difference_bound` as a `to_overall` difference, whereas the sliding-window method as I implemented it considers the tolerance as `between_groups`. 

I'm pretty sure I can make it work for `to_overall` as well. I have a first draft of what it would look like. It's giving me the same results as the `ThresholdOptimizerExpGrad` for now (and is around x10 faster on the `acs_income` data).

I will dig deeper into this, clean up the code that handles both `to_overall` and `between_groups`, and get back to you with a small report summarizing the results (how it compares to the reductions approach in matters of correctness and speed).

> it seems that [error-parity](https://github.com/socialfoundations/error-parity) now also supports some of the simple constraints, but if your sliding window solution works, that would be a much simpler way to do it.

According to the documentation, the relaxation only works for `equalized-odds` if I'm not mistaken, I'll take a closer look.

by the way, I'm also thinking the same sliding window technique can work for `equalized-odds` as well given that the ROC curve is increasing. I'll need to check if this intuition is correct though.

taharallouche:
Hi again @MiroDudik 

I implemented the tolerance to handle `to_overall` as well (it was just a tiny bit trickier than `between_groups`) (see [my previous comment](https://github.com/fairlearn/fairlearn/pull/1453#issuecomment-2558062669))

I also run some comparisons with the reductions approach using the script you sent me. It seems that for different datasets, tolerance levels, and constraints (demographic parity and true positive rate parity), the sliding window yields optimal results and is faster.

Here's an html reporting the results for all the experiments with the notebook to reproduce these results. 
[report_and_notebook.zip](https://github.com/user-attachments/files/18228719/report_and_notebook.zip)

MiroDudik (commented):
I have just checked `_relaxed_constraints.py` and `_threshold_optimizer.py`.

I have verified the correctness for the case "between_groups". For the case "to_overall", it is actually not clear to me whether the approach is correct.

@taharallouche : if you see a way to prove correctness for "to_overall", let me know. if we don't have a proof of correctness, I would be in favor of implementing only "between_groups" version (we could leave "to_overall" version for future?).

Also, it would be good to add some documentation describing the algorithm. Maybe something along the lines: At each point in time, there's a sliding window for the values of "x" metric, and that you keep indices corresponding to the decreasing values of "y" metric--where the first index is the largest y value within the window, the next is the largest after that etc., and then briefly arguing that this property is easily maintained as you move the sliding window, and that this property allows you to compute the largest value of the optimized metric subject to constraints.

taharallouche:
Thanks for taking the time to go through this @MiroDudik !

It will take me some time to formalize the `to_overall` algorithm and try and draft a proof, but I will give it a shot. For now I removed the associated code from this PR so that we can move forward with it.

I will add a `matplotlib` snapshot test for the new plotting functionality and extend the post processing user-guide.

TamaraAtanasoska:
Hi @MiroDudik, pinging for a look here before we merge it, seems like @taharallouche has wrapped this up 🥳

fairlearn/postprocessing/_relaxed_constraints.py
```diff
@@ -0,0 +1,103 @@
...
+    best_indices: list[int] = []
+    max_objective = float("-inf")
+    min_violation = float("inf")
+
+    # Sliding window start pointer
+    start = 0
+
+    # Traverse with the end pointer
+    for end in range(n):
+        # Add the current `end` index to the deques, maintaining descending order of y values
```
MiroDudik, line 71:
```suggestion
        # Add the current `end` index to the deques, maintaining descending order of y values
        # INVARIANT(start, end-1) holds
```

fairlearn/postprocessing/_relaxed_constraints.py
```diff
@@ -0,0 +1,103 @@
...
+    """
+    # Extract `x` values (all DataFrames must share the same `x` values)
+    x_values = dataframes[0][x_col].values
+    n = len(x_values)
+
+    # Extract weighted `y` columns
+    y_columns = [df[y_col].values * weight for df, weight in zip(dataframes, weights)]
+    m = len(y_columns)
+
+    # Initialize a deque for each DataFrame to store indices in descending order of y values
```
MiroDudik, line 58:
```suggestion
    # Initialize a deque for each DataFrame to store indices in descending order of y values.
    # Throughout the execution we will enforce the following INVARIANT(start, end):
    #     Let I = {start, start+1, ... , end} be the set of sliding window indices.
    #     Then deque[k] contain indices i_1, i_2, ... , i_N in I such that:
    #         * i_1 < i_2 < ... < i_N
    #         * i_1 = argmax_{i in I} y_columns[k][i]
    #         * i_j = argmax_{i in I, and i > i_{j-1}} y_columns[k][i]
    #     This data structure supports an efficient increase of start and end by 1, and an
    #     efficient retrieval of argmax of y_columns[k][i] across i in I.
```

MiroDudik (commented):
I have checked the correctness of the algorithm. Looks good! I have just added one comment to clarify what exactly is represented in the deques.

With the comment incorporated this is good to be merged. This is great work!!!

TamaraAtanasoska:
awesome! @taharallouche ping and I'll merge :)

PR #1457, opened 2024-12-09
MNT Prep fairlearn for release and add a changelog requirement to PRs

TamaraAtanasoska:
## Description
This PR adds a changelog entry check like `scikit-learn`'s, when a user affecting change is introduced. 

cc @adrinjalali, what do you think? I think this will make the release process a bit easier.

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

### DIFF
--- a/.github/workflows/check-changelog.yml
+++ b/.github/workflows/check-changelog.yml
@@ -24,7 +24,7 @@ jobs:
           set -xe
           changed_files=$(git diff --name-only origin/main)
           # Changelog should be updated only if tests have been modified
-          if [[ ! "$changed_files" =~ tests ]]
+          if [[ ! "$changed_files" =~ test/* ]]
           then
             exit 0
           fi
--- a/.github/workflows/check-changelog.yml
+++ b/.github/workflows/check-changelog.yml
@@ -1,35 +1,68 @@
-#copied configuration from scikit-learn
-
 name: Check Changelog
 # This check makes sure that the changelog is properly updated
 # when a PR introduces a change in a test file.
 # To bypass this check, label the PR with "No Changelog Needed".
 on:
   pull_request:
-    types: [opened, synchronize, labeled, unlabeled]
+    types: [opened, edited, labeled, unlabeled, synchronize]
 
 jobs:
   check:
     name: A reviewer will let you know if it is required or can be bypassed
     runs-on: ubuntu-latest
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'No Changelog Needed') == 0 }}
     steps:
+      - name: Get PR number and milestone
+        run: |
+          echo "PR_NUMBER=${{ github.event.pull_request.number }}" >> $GITHUB_ENV
+          echo "TAGGED_MILESTONE=${{ github.event.pull_request.milestone.title }}" >> $GITHUB_ENV
       - uses: actions/checkout@v4
         with:
           fetch-depth: '0'
-      - name: Check if tests have changed
-        id: tests_changed
+      - name: Check the changelog entry
         run: |
           set -xe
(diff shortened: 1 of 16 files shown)


.github/workflows/check-changelog.yml
```diff
@@ -0,0 +1,35 @@
...
+          changed_files=$(git diff --name-only origin/main)
+          # Changelog should be updated only if tests have been modified
+          if [[ "$changed_files" =~ tests ]]
+          then
+            echo "check_changelog=true" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Check changelog entry
+        if: steps.tests_changed.outputs.check_changelog == 'true'
+        uses: scientific-python/action-towncrier-changelog@v1
```
adrinjalali, line 32: this doesn't work for us cause we don't use `towncrier`, which requires a separate file for each changelog entry. You can check scikit-learn's bot from before `towncrier` was a thing. If you look at the `1.4` branch, you'll find it.

TamaraAtanasoska: oooh got it, I will make it work with the older config then. I thought it will all fit together somehow. I will check it out and try to make it work the old way then.

TamaraAtanasoska: (hopefully) fixed in https://github.com/fairlearn/fairlearn/pull/1457/commits/263cd8dd1f3f1df0c5848374f3d659ad6ee1e7fd. could you check it out again? I see a doctest failure not related to my PR, I will inspect now.

adrinjalali:
Cool, you can now test this PR by changing some python files and see if it fails.

.github/workflows/check-changelog.yml
TamaraAtanasoska: fixed the doctest, it was a `numpy` version related error interestingly. I was behind a version

TamaraAtanasoska:
@taharallouche this new PR changes the formatting of the changelog slightly, I hope we get to merge it before the release and have the changelong in that form.

TamaraAtanasoska:
@adrinjalali oops 

Screenshot 2024-12-09 at 17 57 16

adrinjalali:
if you do a `pytest -l ...` it gives better info (local vars). But it seems the issue is due to wrong ordering of inheritance, i.e. `BaseClassifier, ClassifierMixin` instead of `ClassifierMixin, BaseEstimator`

test_othermlpackages/test_pytorch.py
```diff
@@ -84,24 +84,28 @@ def get_loss(self, y_pred, y_true, X, *args, **kwargs):
     return net
 
 
+@pytest.mark.skip(reason="external library errors because of new scikit-learn release")
```
adrinjalali, line 87: I don't think we should completely skip them. They should be skipped only if the latest sklearn is installed, with a nice comment that once they support it we put them back

TamaraAtanasoska:
@adrinjalali this is now passing. the tests containing the objects that have tags issues are skipped.

test_othermlpackages/test_pytorch.py
TamaraAtanasoska: ooh I agree, makes sense, brb with a commit

adrinjalali (approved):
It seems like before the release the codebase wasn't tested against the latest release? We really need to fix the CI (and use pixi for it maybe)

TamaraAtanasoska:
> It seems like before the release the codebase wasn't tested against the latest release? We really need to fix the CI (and use pixi for it maybe)

It seems so!

TamaraAtanasoska:
@adrinjalali all green! solves also a python3.8 issue that could've been a problem for the release tomorrow (just updates pip, check last commit).

Issue #1456, opened 2024-12-09
DOC Fix the copy output of the first link in the "Contributing Documentation" section


TamaraAtanasoska:
In the[ "Contributing Documentation" section](https://fairlearn.org/v0.11/contributor_guide/contributing_documentation.html), clicking on the copy button of the first link copies additional `.css` code that shouldn't be there.

```
span.prompt1:before {
  content: "$ ";
}
python scripts/install_requirements.py --pinned False
```

The fix would include going over the contents of that `.rst` document, checking if there is anything that could cause this, and fixing it. Ideally, the person working on this issue would check if this problem is present elsewhere in the generated documentation as well.

The copy button is part of the `sphinx_copybutton` extension.

Thank you @audreyfeldroy for finding this issue during the PyLadiesCon sprint!

TamaraAtanasoska:
Hi @anhpnv, no problem, thanks for taking a look! Feel free to make a PR and tag me (and the issue) :) I will take review it and comment on the PR.

anhpnv:
Hi @TamaraAtanasoska, I'm already create a Pull Request, could you please review it for me 😄

TamaraAtanasoska:
@anhpnv thanks for your patience, I was sick last week. Merged! This issue can be closed now 🥳

PR #1460, opened 2024-12-11
Fix:Add regex promptcopybutton, switch prompt to code-block

anhpnv:
## Description

General summary change:
1. Add `copybutton_prompt_text` and `copybutton_prompt_is_regexp` attribute in conf.py
2. Switch the `.. prompt::` directive to `..code-block::`
3. Add static `$` in paragraph

Mention issue: https://github.com/fairlearn/fairlearn/issues/1456

## Tests

- [ ] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

## Screenshots
The result changed from
 `
span.prompt1:before {
  content: "$ ";
}
python scripts/install_requirements.py --pinned False
` 

to only 

`python scripts/install_requirements.py --pinned False` 

image
@TamaraAtanasoska can you please review my pull request 🙏 🚀

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -70,6 +70,9 @@ extensions = [
     "sphinx_issues",
 ]
 
+copybutton_prompt_text = r"\$\s+"
+copybutton_prompt_is_regexp = True
+
 source_suffix = [".rst"]
 
 # sphinx-issues config
--- a/docs/contributor_guide/contributing_documentation.rst
+++ b/docs/contributor_guide/contributing_documentation.rst
@@ -11,9 +11,9 @@ check the documentation of the version of the package that they are using.
 
 To contribute, make sure to install sphinx and its add-ons by running
 
-.. prompt:: bash
+.. code-block:: bash
 
-    python scripts/install_requirements.py --pinned False
+    $ python scripts/install_requirements.py --pinned False
 
 in the repository root directory.
 You may also need to `install Pandoc <https://pandoc.org/installing.html>`_.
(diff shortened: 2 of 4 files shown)


TamaraAtanasoska:
hi @anhpnv, what happens if you just replace the first `..prompt::` directive with the code you suggest and not the rest? As the rest are working fine on the page at the moment, will the problem be isolated and fixed then, or it gets propagated?

anhpnv:
> hi @anhpnv, what happens if you just replace the first `..prompt::` directive with the code you suggest and not the rest? As the rest are working fine on the page at the moment, will the problem be isolated and fixed then, or it gets propagated?

Hi @TamaraAtanasoska thanks for your reply, but yes, it's will propagated.
If i replace the first `..prompt::` to `.. code-block::`, the problem will present in the next `.. prompt::` too.
I checked the code of `sphinx_prompt` plugins twice, they still continue add `span.prompt1:before { content: "$ "; }` in your code.
I've tested as you suggest above, it still have a bug in the next `.. prompt::`, the copy still return

`
span.prompt1:before {
  content: "$ ";
}
pip install matplotlib>=3.2.1`

image

TamaraAtanasoska:
Cool thanks for checking @anhpnv! I will merge this after the release that is underway at this very moment.

anhpnv:
Okay, thank you so much @TamaraAtanasoska 😁😁

riedgar-ms:
@TamaraAtanasoska I just had a look at the doc build, and the copy button was working as expected, without the extra CSS code.

riedgar-ms (approved):
Seems to work for me

anhpnv:
> Seems to work for me

Yes, refer to this [documentation](https://sphinx-copybutton.readthedocs.io/en/latest/use.html#using-regexp-prompt-identifiers), if you want to change another symbol prompt like `>>>`, `C: \>`, etc..  you can add regex only in `copybutton_prompt_text` variable like this:
```python
copybutton_prompt_text = r"\$\s+|\>>>\s+"
```
I tested in my localhost and it's work 
image

TamaraAtanasoska (approved):

PR #1467, opened 2024-12-12
Add missing imports to quickstart

alliesaizan:
## Description

A very small edit to the quickstart.rst file, adding a note about the missing imports to help users who try to copy and run the quickstart on their machines.

## Tests

- [ ] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation


romanlutz:
But the metrics imports are in this file... I don't think I understand. This code wouldn't run if it was missing imports.

alliesaizan:
> But the metrics imports are in this file... I don't think I understand. This code wouldn't run if it was missing imports.

Some of them are - like accuracy_score. But false_positive_rate, false_negative_rate, and selection_rate were imported in the quickstart,py file. I realized after the fact too that zero_div_precision_score is created in that file. if nothing needs to be done, I'll close this and it'll be my error.

romanlutz:
> > But the metrics imports are in this file... I don't think I understand. This code wouldn't run if it was missing imports.
> 
> Some of them are - like accuracy_score. But false_positive_rate, false_negative_rate, and selection_rate were imported in the quickstart,py file. I realized after the fact too that zero_div_precision_score is created in that file. if nothing needs to be done though, I'll close this and it'll be my error.

Ah, I misunderstood the problem! You're absolutely right, of course. There's a `literalinclude` that does this. Somehow I missed this the first time I looked at it (on my phone). Apologies!

I would suggest adding the missing imports instead (?) or perhaps in the meantime someone came up with a way to have doctest show plots? The reason this exists (and I hated the solution when I created it...) was that with normal code snippets the text output is shown but the plots were not. Hopefully, in the few years since someone provided that functionality. @adrinjalali  may know...

alliesaizan:
> > > But the metrics imports are in this file... I don't think I understand. This code wouldn't run if it was missing imports.
> > 
> > 
> > Some of them are - like accuracy_score. But false_positive_rate, false_negative_rate, and selection_rate were imported in the quickstart,py file. I realized after the fact too that zero_div_precision_score is created in that file. if nothing needs to be done though, I'll close this and it'll be my error.
> 
> Ah, I misunderstood the problem! You're absolutely right, of course. There's a `literalinclude` that does this. Somehow I missed this the first time I looked at it (on my phone). Apologies!
> 
> I would suggest adding the missing imports instead (?) or perhaps in the meantime someone came up with a way to have doctest show plots? The reason this exists (and I hated the solution when I created it...) was that with normal code snippets the text output is shown but the plots were not. Hopefully, in the few years since someone provided that functionality. @adrinjalali may know...

I am not sure if the functionality to make doctest show plots exists. But I like your workaround. Easy enough, I will add import statements and remove the note!

adrinjalali (commented):
This section is referring to the quick start example, which includes all the imports. We shouldn't have to write any code here. The example is clickable and the user can see the whole code there.

TamaraAtanasoska:
I have to say that I am really confused by the issue discussed here, but I wanted to ask what your original intention behind the code you wrote was there @romanlutz? There have been some new additions when it comes to the plots and `sphinx/doctest`, it would be nice if you have minute to make an issue about what would the ideal situation look like there (without a workaround)? Someone could pick that up, at least research what options are there.

romanlutz:
> I have to say that I am really confused by the issue discussed here, but I wanted to ask what your original intention behind the code you wrote was there @romanlutz? There have been some new additions when it comes to the plots and `sphinx/doctest`, it would be nice if you have minute to make an issue about what would the ideal situation look like there (without a workaround)? Someone could pick that up, at least research what options are there. 

Sphinx-gallery has all we need. You can run code, it will show output and even plots. However, not all our docs are Sphinx-gallery pages. The rest are just restructured text with some occasional doctest blocks. They show the text output, but not the plots. That is/was the issue. To work around that, I added the example, literalinclude of the code, and imported the resulting plot. I don't think there's a simpler way at this point unless we switch to Jupyter-book where everything is code (like Sphinx-gallery) and one doesn't have to make workarounds like this.

alliesaizan:
@adrinjalali and @romanlutz I am happy to close this PR if it doesn't address the actual issue. Sounds like users are either able to get what they need from the Quickstart as it is or we may need to consider switching to a new format. Please let me know.

romanlutz:
> @adrinjalali and @romanlutz I am happy to close this PR if it doesn't address the actual issue. Sounds like users are either able to get what they need from the Quickstart as it is or we may need to consider switching to a new format. Please let me know.

I think it addresses the problem it was meant for. I misunderstood at first and thought there must be a different solution. I would recommend merging it as is.

docs/quickstart.rst
```diff
@@ -186,7 +186,14 @@ the model does not predict that outcome.
...
 Fairlearn also allows us to quickly plot these metrics from the
-:class:`fairlearn.metrics.MetricFrame`
+:class:`fairlearn.metrics.MetricFrame`. 
+
+.. doctest:: quickstart
+    :options:  +NORMALIZE_WHITESPACE
+
+    >>> from fairlearn.metrics import false_positive_rate, selection_rate, count
+    >>> from functools import partial
+    >>> zero_div_precision_score = partial(precision_score, zero_division=0)
```
adrinjalali, line 196: I don't think these should be included here, since they might go out of sync with the example if the example changes.

The code snippet bellow doesn't run as is anyway. It requires `y_true` and `y_pred` and the rest of the script to run to work. We can instead add an explicit note that this is a snippet from the example.

docs/quickstart.rst
```diff
@@ -186,7 +186,14 @@ the model does not predict that outcome.
     Name: false_negative_rate, dtype: float64
 
 Fairlearn also allows us to quickly plot these metrics from the
-:class:`fairlearn.metrics.MetricFrame`
+:class:`fairlearn.metrics.MetricFrame`.
```
adrinjalali, line 189:
```suggestion
:class:`fairlearn.metrics.MetricFrame`. For example, this is from our
example :ref:`plt_quick_start`:
```

we then need to add

```
.. _plt_quick_start
```

on top of the example for the `ref` link to find it.

TamaraAtanasoska:
@alliesaizan is this still something you are interested in finishing? it seems there are just two tiny suggestions trailing.

docs/quickstart.rst
alliesaizan: Removed

alliesaizan:
> @alliesaizan is this still something you are interested in finishing? it seems there are just two tiny suggestions trailing.

Thank you for the reminder on this! It totally slipped my mind. I just committed the last two small changes requested by Adrin, so if the tests pass should be OK to merge.

PR #1468, opened 2024-12-13
Enhance exponentiated gradient algorithm readibility 

taharallouche:
## Description
Hello, just for the sake of readability, it helped me (and my IDE) a lot to make these changes, so I was thinking they might be useful for other readers of the `reduction` module:
- Fixed the `dtype`  of a couple of pandas `Series`: it's not a big deal in python as pandas handle it automatically, but it is misleading for the reader.
- Added some type annotations.
- Fixed some docstrings (mostly missing a possible `None` option for one argument, or a second return value).
- I added an (abstract?) method `default_objective` for the `Moment` (abstract?) class since [the code assumes](https://github.com/fairlearn/fairlearn/blob/main/fairlearn/reductions/_exponentiated_gradient/_lagrangian.py#L79) its instances implement such a method.

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation


taharallouche:
Why didn't the changelog check fail ? :thinking:

taharallouche:
Converting this to draft because I can add a couple of type hints in there

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -66,11 +69,11 @@ def __init__(
...
         y,
         estimator,
-        constraints,
-        B,
-        objective=None,
-        opt_lambda=True,
-        sample_weight_name="sample_weight",
+        constraints: Moment,
+        B: float,
+        objective: Moment | None = None,
```
romanlutz, line 74: theoretically, this could be `Optional[Moment] = None` if you want to import `Optional` from `typing`. Just a suggestion...

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -140,7 +146,7 @@ def _eval(self, Q, lambda_vec):
             L_high = error + self.B * max_constraint
         return L, L_high, gamma, error
 
-    def eval_gap(self, Q, lambda_hat, nu):
+    def eval_gap(self, Q, lambda_hat, nu) -> "_GapResult":
```
romanlutz, line 149: This should work without quotes since you're importing annotations, right? Maybe I'm missing something...

romanlutz (commented):
Nice, I like this!

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
taharallouche: You're absolutely right :+1:

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
taharallouche: I think I prefer to stick with the new python syntax :sweat_smile: no strong preference though

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -153,10 +154,19 @@ def fit(self, X, y, **kwargs):
...
+
+        # Stores the duality gap of the Exponentiated Gradient (EG) at each iteration
+        gaps_EG: list[float] = []
+
+        # Stores the best gap at each iteration among the ones EG and Linear Programming (LP) yield
+        gaps: list[float] = []
+
+        # Stores the best-reponse estimators' weights at each iteration
+        # The appended weights come either from (EG) or (LP), depending on their respective gaps
+        Qs: list[pd.Series] = []
```
taharallouche, line 169: I very rarely use type hints outside of function signatures, but when initialization containers for a complex algorithm I think it's helpful for the reader to know what each variable will look like

taharallouche:
I believe I covered all the type hints within the algorithmic part of the module (I didn't tackle the Moments) :slightly_smiling_face:

riedgar-ms:
I've no objection to this - I assume it makes `mypy` a little less unhappy :-)

@romanlutz, did  you have any other comments?

romanlutz (approved):

Issue #898, opened 2021-07-14
Bug in mitigation using Fairlearn on Azure AutoML models


MG-Microsoft:
#### Describe the bug

`GridSearch` is being too restrictive when validating its inputs, rejecting data which it should be able to handle

#### Steps/Code to Reproduce

```python
from sklearn.model_selection import train_test_split
from fairlearn.reductions import GridSearch
from fairlearn.reductions import DemographicParity, ErrorRate
from fairlearn.metrics import MetricFrame, selection_rate, count
from sklearn.preprocessing import LabelEncoder, StandardScaler
from sklearn import metrics as skm
import pandas as pd


from sklearn.datasets import fetch_openml
data = fetch_openml(data_id=1590, as_frame=True)
X_raw = data.data
Y = (data.target == '>50K') * 1


A = X_raw["sex"]
X = X_raw.drop(labels=['sex'], axis=1)

le = LabelEncoder()
Y = le.fit_transform(Y)

X_train, X_test, Y_train, Y_test, A_train, A_test = train_test_split(X,
                                                                     Y,
                                                                     A,
                                                                     test_size=0.2,
                                                                     random_state=0,
                                                                     stratify=Y)

# Work around indexing bug
X_train = X_train.reset_index(drop=True)
A_train = A_train.reset_index(drop=True)
X_test = X_test.reset_index(drop=True)
A_test = A_test.reset_index(drop=True)



# The following shows that LGBMClassifier can handle this dataset
# without the scaling and one-hot encoding used in other examples
from lightgbm import LGBMClassifier

unmitigated_predictor = LGBMClassifier()
unmitigated_predictor.fit(X_train, Y_train)

metric_frame = MetricFrame(metrics={"accuracy": skm.accuracy_score,
                                    "selection_rate": selection_rate,
                                    "count": count},
                           sensitive_features=A_test,
                           y_true=Y_test,
                           y_pred=unmitigated_predictor.predict(X_test))
print(metric_frame.overall)
print(metric_frame.by_group)

# However a GridSearch fails:
sweep = GridSearch(LGBMClassifier(),
                   constraints=DemographicParity(),
                   grid_size=71)

sweep.fit(X_train, Y_train,
          sensitive_features=A_train)

```

#### Expected Results

That the `sweep.fit()` call works

#### Actual Results

```
[...]

PR #1470, opened 2024-12-14
Fix #1352: Allow X with ndims greater than 2

### LINKED ISSUE #1352: mitigation methods with image/multi-dimension data
I am trying to use the fairlearn bias mitigation methods with multi-dimension data like images. To my understanding, the postprocessing methods like ThresholdOptimizer, should not be constraint by the type of classifiers. However, when I try to implement the ThresholdOptimizer with input X (shape:[BS, C, W, H]), it keeps giving me error like: 

> ValueError: Found array with dim 4. None expected <= 2.

Which, if I understand correctly, suggests that X should only have a dim <=2.

I try to wrapped the clf model based on sklearn estimator, but it does not help as well.

I was wondering, how could I solve this issue?

------ 
version
scikit-learn         1.3.0

------

The error is coming from this line
`if not allow_nd and array.ndim >= 3:
raise ValueError(
"Found array with dim %d. %s expected <= 2."
  % (array.ndim, estimator_name)
  )`
in sklearn/utils/validation.py:953.
I belive that in fairlearn/utils/_input_validation.py:89:
`X, y = check_X_y(X, y, dtype=None, force_all_finite=False)`, we should add `allow_nd = True`.

taharallouche:
## Description
Hello, this PR fixes #1352.

Just like #898, this issue shows that the call to check_array is unnecessarily too restrictive.

This is a low hanging fruit that would hopefully make the library usable for more applications :chart_with_upwards_trend:

I'll add a changelog entry if the fix is accepted :pray: 

-------

PS: the error message that `check_array` outputs is not user-friendly in absence of an `estimator` argument:
> ValueError: Found array with dim 4. None expected <= 2.

I'll try to open a PR to fix it upstream :sweat_smile: 

## Tests

- [ ] no new tests required
- [x] new tests added
- [x] existing tests adjusted

## Documentation


fairlearn/utils/_input_validation.py
```diff
@@ -125,12 +125,7 @@ def _validate_and_reformat_input(
     else:
         result_y = pd.Series(dtype="float64")
 
-    return (
-        pd.DataFrame(X),
-        result_y,
-        sensitive_features,
-        control_features,
-    )
+    return (X, result_y, sensitive_features, control_features)
```
taharallouche, line 128: `X` can no longer be converted to a `DataFrame`, because it's no longer guaranteed to be tabular.

I double checked that the returned `X` of this function is never used in the code-base so this change is safe.

TamaraAtanasoska:
I just pushed to this file recently for the first when I was making fairlearn `sklearn 1.6` compatible and it is definitely in a need for refactoring 🙈. 

Generally LGTM, checking some things around the `Dataframe` type change and how it relates to something I am working on right now, and I'll do an official review in the next two days.

fairlearn/utils/_input_validation.py
TamaraAtanasoska: I am wondering if it is safer for now to preserve the type? If a dataframe was entered then it is also returned again. It  will make it safer to merge and an easy fix if (more like when) we have to play with the internal pandas representations.

taharallouche:
Hi @TamaraAtanasoska, I put back the `DataFrame` type, added a test case to ensure `ExponentiatedGradient` can handle multi-dimensional data (which it doesn't on `main`), and added a changelog entry.

TamaraAtanasoska (approved):
Thanks @taharallouche LGTM! I can't merge it atm because of the persisting OpenML outage, we'd have to wait a bit.

PR #1471, opened 2024-12-15
Tidy annotated metric functions construction

### LINKED ISSUE #1004: Error when using MetricFrame with sample_params
Hi, I am using `MetricFrame` to generate evaluations with a set of built-in metrics with extra parameters. The parameters are passed to `MetricFrame` using `sample_params`. I found that for built-in metrics with `sample_weight` and `pos_label` as extra parameters, `sample_weight` is always working well, but when `pos_label` was passed with a not-None value, it produced IndexError from fairlearn/metrics/_function_container.py. I've tried all *_rate metrics imported from fairlearn.metrics, like `false_positive_rate`, and also similar metrics imported from sklearn.metrics, like `recall_score`. All behaved similarly. 

Here is the minimal code to reproduce, and I am using fairlearn 0.7.0

```
import fairlearn.metrics as flm
y_true = [0, 1, 1, 1, 1, 0, 1, 0, 1, 0, 0, 0, 1, 1, 1, 1]
y_pred = [0, 0, 1, 0, 1, 1, 1, 0, 0, 1, 1, 1, 1, 0, 0, 1]
gp = ['d', 'a', 'c', 'b', 'b', 'c', 'c', 'c',
                         'b', 'd', 'c', 'a', 'b', 'd', 'c', 'c']
sample_weight = [0.1]*len(y_true) 
pos_label = 1
print(flm.false_positive_rate(y_true, y_pred, sample_weight, pos_label))
gm = MetricFrame(metrics=flm.false_positive_rate,
                             y_true=y_true,
                             y_pred=y_pred,
                             sensitive_features=gp,
                sample_params={'sample_weight': sample_weight, 'pos_label': pos_label})
print(gm.overall)
```

The above code will print 0.666666 for the direct computation by calling the metric function, and throw error when using in MetricFrame. The last part of the error message is: 
```
/home/cdsw/.local/lib/python3.6/site-packages/fairlearn/metrics/_function_container.py in generate_sample_params_for_mask(self, mask)
     82         curr_sample_params = dict()
     83         for name, value in self.sample_params_.items():
---> 84             curr_sample_params[name] = value[mask]
     85         return curr_sample_params
     86 

IndexError: too many indices for array: array is 0-dimensional, but 1 were indexed
```
and the screenshot is:
image
image

When setting pos_label to None, both computation will work and print out the same result. 

Thanks in advance for any suggestion and help.

taharallouche:
## Description
This PR renames some variables related to the `AnnotatedMetricFunction` construction for readibility. It also simplifies the associated conditional statesments.

NOTE: This initially aimed to fix #1004, by allowing to pass kwargs to the metrics via `sample_params` (when adding them as a column to the dataframe is impossible/doesn't make sense), which is to my understanding the solution that was suggested in the issue's comment section.

However, it was agreed that it's probably not a good idea. Instead, we will close the issue and point to the [documentation](https://fairlearn.org/main/user_guide/assessment/advanced_metricframe.html#scalar-arguments) guiding the user on how to use `partial`.

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation


TamaraAtanasoska:
@riedgar-ms do you maybe have time to review this one? I think it will benefit from your perspective.

riedgar-ms:
I've no objection to the renaming going on inside the MetricFrame code (the names made sense to me at the time, but that was a while ago), but for scalar parameters, I'd rather encourage users to go with `functools.partial` (as we [currently do in the docs](https://fairlearn.org/main/user_guide/assessment/advanced_metricframe.html#scalar-arguments) ). As you note @taharallouche , the point of the name `sample_params` is that they're per-sample.

I'd suggest doing the renaming, but closing #1004  with a link to that doc page.

taharallouche:
I accidentaly found out that, in case  the `sample_params` can not be broadcasted to fit as a column into `all_data` (eg, the `sample_param` is a list of 100 element while `sensitive_features` is a list of 200 values), the error message is not user-friendly: it doesn't point the user to the root cause.

I will open a follow-up issue to improve it , and I might tackle it myself subsequently.

taharallouche:
Hi @riedgar-ms @TamaraAtanasoska I reverted the kwargs forwarding and only kept the renaming and tidy changes, as agreed.

The CI is still failing because of OpenML.

riedgar-ms (approved):
So long as the tests pass, I'm happy with this. Could you update the PR title and description though, please?

Issue #1490, opened 2025-01-30
MNT Minimal numpy version when requirements --pinned true and python 3.12 and 3.13


TamaraAtanasoska:
In the case of `requirements --pinned true`, the script installs the minimal version of `numpy` (1.24.4), which isn't compatible with `python=3.12 `and `python=3.13` ([see nightly](https://github.com/fairlearn/fairlearn/actions/runs/13048138880)).

`python=3.12` requires at lest `numpy=1.26.0`, and `python=3.13` requires `numpy=2.1.0`. 2.1.0 also drops support for `python=3.9`.

I see two options to go forward:
- for now, increase the minimum version to `numpy=1.26.0`, but that still means it won't work for `python=3.13`.
- change the script to allow some flexibility for different `python` versions

@riedgar-ms, what do you think? What is the idea behind `--fixed`?

riedgar-ms:
The idea of the 'fixed' build was to pin to our lower bounds, to make sure we really worked with them. If we need to bump the lower bound upwards, that's fine.

TamaraAtanasoska:
thanks @riedgar-ms! However bumping all the way up to 2.1.0 for `numpy` is a bit too much, I can make a PR with `numpy` bumped to 1.26. What would be a good practice here for this newest `python` version? We just support it sometime in the future? Should I leave it in the nightly?

riedgar-ms:
What are scikit-learn doing around the numpy-v2 change? We've generally followed them in the past. It looks like Python 3.9 is still in official support until the end of this year?

TamaraAtanasoska:
@riedgar-ms yes and we should definitely support it too, I agree. I can add a special case for now for `python=3.13` with the fitting `numpy` version, which only the nightly will run in our case.

PR #1491, opened 2025-01-30
MNT Fix failing CI for the --pinned requirements case for python 3.12 and 3.13

TamaraAtanasoska:
## Description
Fixes https://github.com/fairlearn/fairlearn/issues/1490.

@riedgar-ms we can maybe move the discussion here and add to the PR whatever else might be necessary.

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation


TamaraAtanasoska:
@riedgar-ms would this make sense to you? Commit : https://github.com/fairlearn/fairlearn/pull/1491/commits/112ba4496bea1889745a86b9f2bddd375c7b277e 
It seems scikit-learn has multiple configs in the pyproject.toml, but this would do for our case at the moment. 
Unfortunately pytorch still doesn't have 3.13 support, but we will know when they do as the nightly will turn green as they are on their way to support it https://github.com/pytorch/pytorch/issues/130249

riedgar-ms (approved):
OK by me. @adrinjalali , do you have any other suggestions?

TamaraAtanasoska:
@adrinjalali make sense what you say! I wanted to quickly patch CI so it is green while I can't work on it and have something green during the two weeks, catering to the limitations of the code in https://github.com/fairlearn/fairlearn/blob/main/scripts/install_requirements.py#L48 (thus specific version). I have some sprint prep leftover week this week and the pre-event, but we should definitely work more fundamentally on this soon. Let's leave this PR open here this week I will ping with something better in a bit.

scripts/install_requirements.py
```diff
@@ -42,7 +50,10 @@ def _build_argument_parser():
 
 
 def _process_line(src_line):
-    return src_line.replace(">=", "==")
+    if python_version in _PINNED_REQUIREMENTS and src_line.startswith("numpy"):
+        return _PINNED_REQUIREMENTS[python_version]
+    else:
+        return src_line.replace(">=", "==")
```
TamaraAtanasoska, line 56: @riedgar-ms and @adrinjalali how about this as a transitional solution before we move towards something more sustainable? I don't have the time to invest at the moment, and I would really like to get rid of the red in CI as it makes the daily monitoring tedious. I started by implementing a more elaborate solution creating a new flag to get the python version input from the CI conf files, enabling us also more convenient testing, but I it felt like an overkill for something we want to change. I also created more flexible dictionaries and code that can be modified easily to support more packages than `numpy`, but at the moment we don't need that either and we might not in the near future. It is really just for the `--pinned` case which isn't something a user will use.

scripts/install_requirements.py
```diff
@@ -11,12 +11,20 @@
...
+
 _REQUIREMENTS_STEMS = ["requirements", "requirements-dev"]
 
 _INSERTION_FIXED = "-fixed"
 
 _REQUIREMENTS_EXTENSION = "txt"
 
+# Configuration for the --pinned argument including explicit package versions
+# that are minimally required for the latest Python versions.
+_PINNED_REQUIREMENTS = {
```
riedgar-ms, line 24: Small point, but if this is just for `numpy`, should the name reflect that?

riedgar-ms (approved):

scripts/install_requirements.py
TamaraAtanasoska: yes true, it is from the initial code where it was all more general and more customisable. we could change it, but we can also leave so we could add something(and of course change the other line where numpy is explicit with it then). I think soon `torch` will become an issue for `3.13`, I am following their thread. what do you think?

Issue #1475, opened 2024-12-19
DOC Identify and list components of the API that could benefit from short examples in the documentation


TamaraAtanasoska:
#### Describe the issue linked to the documentation
The [Fairlearn API](https://fairlearn.org/v0.12/api_reference/index.html) still has many of it's components not covered by short examples that show their functionality in the [User Guide](https://fairlearn.org/v0.12/user_guide/index.html). 

#### Suggest a potential alternative/fix
Adding more examples of example usage of the different components of the API in the User Guide.
Some of the examples can be added within the existing documents, but the User Guide can also be expanded. 

**A big step forward would be go to through the API reference and identify which components aren't already in the User Guide and create a list as part of this issue.** 

While not directly merging anything, this issue provides great value for the project. The benefit for the contributor is getting familiar with the API in detail and easily picking up subsequent tasks. 

An example of such a short example in the user guide of a PR we merged recently: 
- [Place in the docs where it resides](https://fairlearn.org/v0.12/user_guide/mitigation/reductions.html#exponentiated-gradient)
- The PR discussion around it: https://github.com/fairlearn/fairlearn/pull/1353

Feel free to ping me with any questions! You can also go ahead and open a PR about adding an example that is missing.

StefanieSenger:
I am working on adding an example for `ErrorRate`.

TamaraAtanasoska:
Hi @StefanieSenger, are there maybe some functions/classes you came upon searching what to do that you decided not to go into and you could list here? It is ok if not, maybe we can leave a breadcrumb for the next person. Thanks!

alliesaizan:
Hi @TamaraAtanasoska, this issue is pretty similar to #987 opened by @hildeweerts. Would it make sense to consolidate these two issues?

TamaraAtanasoska:
yes they definitely can be consolidated into a larger project. both the tasks are still relevant though, as a list with all the missing information make it more tangible for the newcomers to the project.

alliesaizan:
> yes they definitely can be consolidated into a larger project. both the tasks are still relevant though, as a list with all the missing information make it more tangible for the newcomers to the project.

Understood

3Paradox:
Hi, I noticed make_derived_metric doesn't have an example in the User Guide. Ran into this while building a churn model. Adding one with transform='difference' might be useful. Happy to work on this if no one else is on it!

3Paradox:
[...]

PR #1502, opened 2025-02-12
DOC add example for using ErrorRate

### LINKED ISSUE #1496: ErrorRate.load_data does not use what is passed as argument for control_features param
The signature of `ErrorRate.load_data` is: `load_data(self, X, y, *, sensitive_features, control_features=None)`.

The `control_features` param is however not used / ignored inside the method. Should this param be deprecated or removed?

StefanieSenger:
## Description
This PR suggests a way to add examples on how to use `ErrorRate` in the UserGuide. Would love to receive feedback, especially on the second example, which I stole from the quickstart and where I was less eloquent than I would like to be, because I am not sure about why we would want to use it like this from the practical point of view.

I have also removed the `control_features` param from `ErrorRate.load_data()`, which was unused and which we had discussed in the linked issue.

towards #1475
closes #1496


riedgar-ms:
Thanks for looking at this; tagging @MiroDudik  who knows most about reductions. It appears that the changes to the API need propagating a little further?

fairlearn/reductions/_moments/error_rate.py
```diff
@@ -60,28 +62,49 @@ def __init__(self, *, costs=None):
...
+        """Load the specified data into the object.
+
+        Parameters
+        ----------
+        X : array of shape (n,p)
+            The feature array
+        y : :class:`pandas.Series`
+            The label vector
+        sensitive_features : :class:`pandas.Series`
+            The sensitive feature vector
```
taharallouche, line 75: I don't think that these two arguments here have to be a `pandas.Series`.

`y_train` and `sf_train`, output of `_validate_and_reformat_input`, will be series though

docs/user_guide/mitigation/reductions.rst
```diff
@@ -351,6 +351,67 @@ and false positive rate.
...
+    >>> classifier = LogisticRegression().fit(X, y)
+    >>> costs = {"fp":0.1, "fn":0.9}
+    >>> errorrate = ErrorRate(costs=costs)
+    >>> errorrate.load_data(X, y, sensitive_features=sensitive_features)
+    >>> errorrate.gamma(classifier.predict)
+    all    0.139
+    dtype: float64
+
+
+Using ErrorRate as part of a reductions approach for fairness mitigation is equivalent
```
taharallouche, line 386:
```suggestion
Using :class:`ErrorRate` as part of a reductions approach for fairness mitigation is equivalent
```

fairlearn/reductions/_moments/error_rate.py
```diff
@@ -60,28 +62,49 @@ def __init__(self, *, costs=None):
         else:
             raise ValueError(_MESSAGE_BAD_COSTS)
 
-    def load_data(self, X, y, *, sensitive_features, control_features=None):
-        """Load the specified data into the object."""
+    def load_data(self, X, y, *, sensitive_features):
```
taharallouche, line 65: IMHO, we need to mention this removal in the changelog.

Even though the argument was not supposed to be used, some users might pass `control_features=None` to stick to the method's signature for instance, and we are breaking their code (`TypeError: unexpected argument ...`)

StefanieSenger (commented):
Thank you for your feed back, @taharallouche and @riedgar-ms!

I have given the issue concerning the `control_features` and the failing tests in test_control_features.py some thought. Seems like `_Lagrangian` puts `ErrorRate` as its default objective whenever a `UtilityParity` constraint is used and the user hasn't set any objective themselves.

I would suggest to add a helper function to filter kwargs needed to load data to the objective object in `_Lagrangian.__init__()`, but please see this only as a suggestion. I am happy to reverse that or do it differently, when it fits better to the overall design decisions for fairlearn.

taharallouche:
> I would suggest to add a helper function to filter kwargs needed to load data to the objective object in _Lagrangian.__init__(), but please see this only as a suggestion. I am happy to reverse that or do it differently, when it fits better to the overall design decisions for fairlearn.

I gave it a thought and I can't find a better solution that wouldn't require redesigning the whole module :+1: 


--------

For the record, I tried another solution, namely:
- change the `Lagrangian`'s `init` signature, adding a two new arguments `sensitive_features` and `control_features` which default to `None`, and removing `**kwargs`.
- forward both of them to the constraints' `load_data`, and only `sensitive_features` to the objective's `load_data`

but it doesn't work because some constraints don't support `control_features` (`BoundedGroupLoss` for instance) :disappointed:

MiroDudik:
Hi folks. Joining a conversation a bit late, but let me first clarify a few things and then make a suggestion.

* `load_data` method is meant to be parallel to the `fit` method in estimators; we made the decision to use `X,y,**kwargs` provided to `fit()` in `ExponentiatedGradient` in a pass-through manner, so there shouldn't really be any logic (including validation or standardization) related to them until the `Moments` class uses them
* I agree that listing `control_features` and `sensitive_features` in the signature of `load_data()` in `ErrorRate` is weird if they're not being used (we might have similar issues in other moments classes?)
  * one solution would be to change the signatures of `load_data` methods to something like `load_data(X,y,*,<additional used arguments>, **kwargs)` where `**kwargs` would ensure that the method works also if a bunch of extra arguments that it doesn't use are passed to it 
  * another solution would be to use introspection in the `Lagrangian` constructor, that is `inspect.signature(obj.load_data)` and `inspect.signature(constraints.load_data)` and only pass on the `kwargs` that are listed as keyword arguments in the corresponding `load_data` method
* `gamma` actually returns a "vector of moments"
  * when the `Moment` object is used as a constraint, then "`gamma[j]`&leq;0 for all j" is used as the set of constraints (in that sense it captures constraint violation)
  * when the `Moment` object is used as an objective, then `gamma[0]` is used as the objective
* `load_data` and `gamma` are actually not designed for the end user; these are methods that need to be implemented by anybody who wants to introduce a new objective or a new set of constraints--they are implementing an internal API to support `Lagrangian` and `ExponentiatedGradient`. That said--we don't do the best job documenting this internal API (but the audience for this documentation are package developers/contributors, not necessarily package users)

My suggestion:
* I think that in this PR, we should just focus on adding an example of `ErrorRate` and fixing some related documentation issues (e.g., we could correct the description of `gamma` in the files that are touched)
* we can create a separate issue, where we discuss if we want to tweak the internal API around `load_data`

StefanieSenger:
I have now removed the removing of `control_features` from this PR and it can be dealt with separately.

> gamma actually returns a "vector of moments"
> 
>     when the Moment object is used as a constraint, then "gamma[j]≤0 for all j" is used as the set of constraints (in that sense it captures constraint violation)
>     when the Moment object is used as an objective, then gamma[0] is used as the objective

Thank you for the explanations, @MiroDudik. I have used this almost literally on the docstring.

StefanieSenger:
Note to myself: Read API conventions part in https://fairlearn.org/main/contributor_guide/contributing_code.html to learn about how to use Reductions and re-work added example.

docs/user_guide/mitigation/reductions.rst
```diff
@@ -351,6 +351,70 @@ and false positive rate.
...
 
+
+.. _error_rate:
+
+Error Rate
+~~~~~~~~~~
+
+We can use the :class:`ErrorRate` either to measure performance of a trained model that does or
+does not take sensitive features into account, or as an addition to the
+objective function during the training of a fairness aware model.
```
MiroDudik, line 362:
```suggestion
does not take sensitive features into account, or as an
objective function during the training of a fairness aware model.
```

MiroDudik (commented):
Generally looks good to me, except for one small editorial change. (I don't need to review once the change is implemented.)

MiroDudik:
> I have now removed the removing of `control_features` from this PR and it can be dealt with separately.
> 
> > gamma actually returns a "vector of moments"
> > ```
> > when the Moment object is used as a constraint, then "gamma[j]≤0 for all j" is used as the set of constraints (in that sense it captures constraint violation)
> > when the Moment object is used as an objective, then gamma[0] is used as the objective
> > ```
> 
> Thank you for the explanations, @MiroDudik. I have used this almost literally on the docstring. I'm still wondering, how and where `gamma()` would be used once it's calculated, though.

Yeah. The changes that you made look good to me.

I think of `gamma` is mainly an internal API. What you wrote about the standalone use of `ErrorRate` is totally correct, but it's more of a side-effect than the intended use case. The main use case for all `Moment` subclasses is to provide a structure for the input arguments for reductions.

TamaraAtanasoska:
thanks @MiroDudik, @StefanieSenger feel free to ping me to merge when you feel ready!

StefanieSenger:
Thanks for looking through this PR, @MiroDudik and @TamaraAtanasoska!

I've had a superficial glimpse into reductions and I am happy that I have contributed something that might help others. I would have loved to also explain how to _really_ use it, but I don't have the bandwidth to figure this out right now. So, since you are both happy, I'm fine with the state of the PR. Thanks for supporting this. :)

TamaraAtanasoska (approved):

Issue #1104, opened 2022-06-09
Fix logo and colors for dark mode on the website


adrinjalali:
The new version of the theme used in https://github.com/fairlearn/fairlearn/pull/1103 supports dark mode, but our logo and the colors could use some adjustments.

To fix this issue we need a logo for dark mode, and maybe some settings for colors in that mode.

xref: https://github.com/fairlearn/fairlearn/pull/1103#issuecomment-1151287033

hildeweerts:
For whoever will work on this, note that text in a `<dt>` element is rendered in dark gray, which I pretty much unreadable in dark mode.

A screenshot of selected text within a `<dt>` element: https://fairlearn.org/main/user_guide/datasets/acs_income.html
![image](https://user-images.githubusercontent.com/24417440/178430525-e4b00633-af9b-45a9-a724-4bce8287d6db.png)

Issue #1493, opened 2025-01-30
Add night mode on the website's static landing page


taharallouche:
### Description

The documentation generated by sphinx has a night/dark mode. We do not have that option on the homepage yet.
It would be nice to add it, and to default to the user preference on the browser itself.

PR #1507, opened 2025-02-15
Static Landing Page Lightmode

mosalah16:
## Description
This PR solves #1493 

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

## Screenshots
image
image
image


taharallouche:
The output looks good to me. I also tested that it follows my browser light-mode settings.

taharallouche (approved):

romanlutz (approved):
Lgtm 

If I really want to be picky I'd say the logo and some icons don't stand out well against black background but I don't think it's a huge deal.

Thank you!

mosalah16:
Hi, and thank you for replying!
I've made 2 changes:
- Brightened the logo in dark mode.
- Made "Fairlearn can be installed with pip from PyPI as follows" clearer for dark mode

Please don't hesitate to give me further suggestions :D
pre
post

TamaraAtanasoska:
I had to recently update these icons for the LinkedIn addition, and there is an alternative icon for the small ones in the social row that come up on hoover and are white, thus more noticeable. I wonder if it would make sense to invert them in dark mode, the white version default, the blue appearing on hoover. The problem with noticing them would be then less prominent and they would fit the colour scheme.

mosalah16:
> I had to recently update these icons for the LinkedIn addition, and there is an alternative icon for the small ones in the social row that come up on hoover and are white, thus more noticeable. I wonder if it would make sense to invert them in dark mode, the white version default, the blue appearing on hoover. The problem with noticing them would be then less prominent and they would fit the colour scheme.

Hmmm, inverting both default and hover makes the icons hover looks off 

image
image

so I removed the icon hover in dark mode and I think this looks better 

image

and a small change to make "Fairlearn can be installed with pip from PyPI as follows" close to "pip install fairlearn" without changing the original html

TamaraAtanasoska:
@mosalah16 thanks, I like your idea! I found this old issue about the same topic, which we should also close when we merge this PR: https://github.com/fairlearn/fairlearn/issues/1104
The log is mentioned in that PR and the `<dt>` element. The logo looks good enough for me at the moment. We can always add future improvements to the individual elements as well.

taharallouche:
Hi @TamaraAtanasoska I think that, although this PR is related to #1104, it doesn't resolve it, since the issue is about the automatically generated part of the website, while this only handles the static landing page.

This being said, I'd recommend merging this one, and @mosalah16 will probably be able to fix #1104 in a follow-up PR :smile:

TamaraAtanasoska:
@taharallouche ah you are right, agreed 👍  I am going through the issues today for sprint reasons one more time and I noticed this other one the process. would be cool if we could solve that one too, but no pressure @mosalah16, thank you for your contribution!

Issue #1522, opened 2025-02-25
MNT Replace pandas with narwhals


TamaraAtanasoska:
To improve the maintenance effort and offer users options, we have decided to replace `pandas` with `narwhals` in the project, with the exception of the `examples`.

The `narwhals` team has already done [a previous exploration](https://github.com/narwhals-dev/narwhals/issues/204) of the feasibility of the topic, which is now a bit older, but still relevant. There are few files they looked into and some of that work could be transfered.

They also have a vibrant [discord community](https://discord.gg/V3PqtB4VA4) and will help if we come across a type of expression we need and `narwhals` still doesn't support cc @MarcoGorelli.

Files to be converted:

- [ ] ./metrics/_bootstrap.py
- [ ] ./metrics/_disaggregated_result.py
- [ ] ./metrics/_plotter.py
- [ ] ./metrics/_group_feature.py
- [ ] ./metrics/_annotated_metric_function.py
- [ ] ./metrics/_metric_frame.py
- [ ] ./metrics/_metric_frame.py
- [ ] ./utils/_input_validation.py
- [ ] ./postprocessing/_threshold_optimizer.py
- [ ] ./postprocessing/_plotting.py
- [ ] ./postprocessing/_tradeoff_curve_utilities.py
- [ ] ./preprocessing/_correlation_remover.py
- [ ] ./reductions/_exponentiated_gradient/_lagrangian.py
- [ ] ./reductions/_exponentiated_gradient/exponentiated_gradient.py
- [ ] ./reductions/_grid_search/grid_search.py
- [ ] ./reductions/_grid_search/_grid_generator.py
- [ ] ./reductions/_moments/moment.py
- [ ] ./reductions/_moments/error_rate.py
- [ ] ./reductions/_moments/bounded_group_loss.py
- [ ] ./reductions/_moments/utility_parity.py

Additionally, we would have to make sure that the utils functions work with the `polars` objects, as well as include appropriate testing.

I have created the `narwhals-integration` branch so multiple people can work on the issue before we merge the branch into main. 

I am interested in taking some of this work on when time permits in the next few weeks, as well as making sure we are keeping up with the development form `main` in the integrations branch, consulting with the `narwhals` folks in case of issues and seeing it through in time for release this summer 🌞.
@StefanieSenger has expressed interest in joining the project(thanks!).

@taharallouche your perspective and skills would be put to great use here, if you have any capacity to join forces in the near future, especially as a reviewer 🙏 

[...]

PR #1525, opened 2025-02-26
MNT Narwhalify preprocessing module

FBruzzesi:
## Description

First attempt to introduce narwhals for preprocessing module. Related to #1522 

## Tests

- [ ] no new tests required
- [x] new tests added
- [x] existing tests adjusted

## Documentation


test/unit/preprocessing/linear_dep_remover/test_sklearn_compat.py
```diff
@@ -6,6 +6,7 @@
 
 import numpy as np
 import pandas as pd
+import polars as pl
```
FBruzzesi, line 9: Should we test with pyarrow as well?

fairlearn/preprocessing/_correlation_remover.py
```diff
@@ -3,8 +3,8 @@
 
 from collections.abc import Iterable
 
+import narwhals.stable.v1 as nw
```
StefanieSenger, line 6: I find this [Perfect backwards compatibility policy](https://narwhals-dev.github.io/narwhals/backcompat/) extremely interesting! 

But I am not sure if I perfectly understand it: Does it imply that all the `narwhals.stable.v1` versions also keep relying on the same versions from their dependencies? I am asking because I have the impression that `polars` is developing and changing very fast. So if fairlearn mixes functions from `narwhals.stable.v1` and the versioned narwhals (I mean functions and classes outside of the `stable` directory) (and possibly later from `narwhals.stable.v2` as well), which polars version could be used for testing/could be recommended to users to use?

StefanieSenger (commented):
Hi @FBruzzesi,

just adding some comments/questions, that interest me while I am getting myself into this new topic. I hope this is generally helpful for others as well.

fairlearn/preprocessing/_correlation_remover.py
```diff
@@ -96,9 +96,10 @@ def _split_X(self, X):
 
     def _create_lookup(self, X):
         """Create a lookup to handle column names correctly."""
-        if isinstance(X, pd.DataFrame):
+        X = nw.from_native(X, pass_through=True, eager_only=True)
+        if isinstance(X, nw.DataFrame):
```
StefanieSenger, line 100: If `X` comes as a numpy array, shouldn't it stay a numpy array?

I would assume that `X`'s type should be handled in `validate_data()` instead of here and turned into a narwhals DataFrame only if it was a pandas DataFrame before.

fairlearn/preprocessing/_correlation_remover.py
FBruzzesi: > I find this [Perfect backwards compatibility policy](https://narwhals-dev.github.io/narwhals/backcompat/) extremely interesting!

That's great to hear! I am sure Marco agrees with the fact that this feature is not discussed enough!

> Does it imply that all the narwhals.stable.v1 versions also keep relying on the same versions from their dependencies?

Not really, we have some [minimal versions](https://github.com/narwhals-dev/narwhals/blob/3103b20afc1a523137f74003478c77a9debc2781/narwhals/utils.py#L470-L481) of the libraries we support, but this has nothing to do with narwhals stable api.

It's almost quite the oppositive, Narwhals will take care of some incompatibilities across different versions of the same library by remapping to polars output.

Now coming to the narwhals stable api, this means that by importing narwhals as such, you will (almost) never incur in backward incompatible changes in the narwhals api. As an example, we recently removed `LazyFrame.gather_every` method in the main namespace (due to future no guarantee in row order), but kept it in the stable v1 version ([PR](https://github.com/narwhals-dev/narwhals/pull/2103)).

Please let me know if this is still unclear :)

fairlearn/preprocessing/_correlation_remover.py
FBruzzesi: > If X comes as a numpy array, shouldn't it stay a numpy array?

Yes, that's what `pass_through=True` does - if the object passed in is not convertible to a narwhals DataFrame, then it is returned as it is and therefore the condition in the next line will be false

fairlearn/preprocessing/_correlation_remover.py
MarcoGorelli: Thanks both!

Just to try to further clarify and respond to

> I am asking because I have the impression that polars is developing and changing very fast

The idea is that even if Polars' API changes, we'll keep `narwhals.stable.v1` stable and handle the Polars differences inside there

test/unit/preprocessing/linear_dep_remover/test_sklearn_compat.py
FBruzzesi: I implemented different dataframe constructors similarly to how we do it in narwhals itself, but also how plotly now runs its test suite (see [plotly.py/tests/test_optional/test_px/conftest.py](https://github.com/plotly/plotly.py/blob/main/tests/test_optional/test_px/conftest.py))

FBruzzesi:
Hey @StefanieSenger, CI failures seem unrelated.

Let me know your thoughts on the conftest and test methodology 🤞🏼

StefanieSenger:
Hi @FBruzzesi, the linting issues are unrelated and had been fixed in #1530. So if you merge main, everything except for the Changelog test should pass.

test/conftest.py
```diff
@@ -0,0 +1,60 @@
+from importlib.util import find_spec
+
+import pandas as pd
+import pytest
+from narwhals.typing import IntoDataFrame
+from narwhals.utils import parse_version
+
+
+def pandas_constructor(obj) -> IntoDataFrame:
```
taharallouche, line 9: Wouldn't `pd.DataFrame` be a more precise return value type hint here ? (same remark for the next two `pandas` constructor functions)

taharallouche (approved):
@FBruzzesi Thanks for the great PR ! It looks great to me :pray: 

I only left one nitpicky remark.


[EDIT]: Could you also add an entry into the [changelog](https://github.com/fairlearn/fairlearn/blob/main/docs/user_guide/installation_and_version_guide/v0.13.0.rst) please ?

FBruzzesi:
> @FBruzzesi Thanks for the great PR ! It looks great to me 🙏
> 
> I only left one nitpicky remark.
> 
> [EDIT]: Could you also add an entry into the [changelog](https://github.com/fairlearn/fairlearn/blob/main/docs/user_guide/installation_and_version_guide/v0.13.0.rst) please ?

That's such great news @taharallouche 🙌🏼
Just pushed the changes!

Looking forward to merge this one, especially so that conftest is setup with the constructors to use across the entire codebase 🚀

PR #1526, opened 2025-03-06
MNT Narwhalify reductions.ErrorRate

StefanieSenger:
Towards #1522

PR to narwhalify `reductions.ErrorRate`. 

It is now ready for review. I have added some comments to facilitate the review process.

## Tests

- [ ] no new tests required
- [x] new tests added
- [ ] existing tests adjusted

## Documentation

- [x] API docs added or updated (no API is affected actually, but internal docs have been slightly updated)


fairlearn/reductions/_moments/moment.py
```diff
@@ -30,26 +31,29 @@ class Moment:
...
-        if sensitive_features is not None:
-            assert isinstance(sensitive_features, pd.Series)
+        X = nw.from_native(X, pass_through=True, eager_only=True)
+        # TODO: check if y can be a DataFrame too; it is not documented here, but don't some subclasses support multioutput?
+        y = nw.from_native(y,  pass_through=True, series_only=True)
+        sensitive_features = nw.from_native(sensitive_features, pass_through=True, series_only=True)
         self.X = X
         self._y = y
-        self.tags = pd.DataFrame({_LABEL: y})
+        self.tags = nw.from_dict({_LABEL: y}) # TODO: I don't need to specify a backend here? It picks whatever is passed?
```
StefanieSenger, line 54: I now understand that backend would be "pandas" for now, since fairlearn has a dependency on pandas. Once it's removed, the backend would dynamically depend on whatever the user uses as a backend, I think.

fairlearn/reductions/_moments/error_rate.py
```diff
@@ -63,6 +64,11 @@ def __init__(self, *, costs: dict[Literal["fp", "fn"], float] | None = None):
 
     def load_data(self, X, y, *, sensitive_features, control_features=None) -> None:
         """Load the specified data into the object."""
+        X = nw.from_native(X, pass_through=True, eager_only=True)
+        y = nw.from_native(y, pass_through=True, eager_only=True)
+        sensitive_features = nw.from_native(sensitive_features, pass_through=True, eager_only=True)
+        if control_features:
+            control_features = nw.from_native(control_features, pass_through=True, eager_only=True)
```
StefanieSenger, line 71: TODO: check if this condition is needed. pass_through=True might also pass through None.

test/unit/reductions/moments/test_moments_error_rate.py
```diff
@@ -22,3 +29,52 @@ def test_bad_costs(bad_costs):
     with pytest.raises(ValueError) as execInfo:
         _ = ErrorRate(costs=bad_costs)
         assert _MESSAGE_BAD_COSTS in execInfo.value.args[0]
+
+
+# TODO: at the moment only tests compatibility between pandas and polars; should also 
+# test pyarrow
+# TODO: add control_features to tests
```
StefanieSenger, line 36: There is no need to narwhalify or test the control_features, because they are not used inside ErrorRate anyways, see https://github.com/fairlearn/fairlearn/issues/1496.

fairlearn/reductions/_moments/moment.py
```diff
@@ -30,26 +31,33 @@ class Moment:
...
+        sensitive_features : :class:`pandas.Series` or :class:`polars.Series`
             The sensitive feature vector (default None)
         """
         assert self.data_loaded is False, "data can be loaded only once"
-        if sensitive_features is not None:
-            assert isinstance(sensitive_features, pd.Series)
+        if sensitive_features is None:
+            raise ValueError("The 'sensitive_features' parameter must be provided.")
         self.X = X
         self._y = y
```
StefanieSenger, line 51: Learning: Make sure that object attributes are the original type that the user has passed and the narwhalification happens afterwards. If dataframes are narwhalified too early, we loose flexibility.

In this particular case, the following error occured for polars dataframes only, because the narwhalified polars was passed to LogisticRegression.predict() instead of the original one:

<details>
    def test_error_rate_narwhals_compatible():
        """Test that ErrorRate is compatible with any types passed via narwhals, that
        returns are the same for all the methods and that the input types are returned."""
    
        # make up data for testing
        rng = np.random.default_rng(42)
        X, y = make_classification(n_features=10, class_sep=0.1, random_state=42)
        X[:, -1] = rng.integers(0, 2, size=(X.shape[0],))
        sensitive_features = X[:, -1]
        X_pd, y_pd, sensitive_features_pd = pd.DataFrame(X), pd.Series(y), pd.Series(sensitive_features)
        X_pl, y_pl, sensitive_features_pl = pl.DataFrame(X), pl.Series(y), pl.Series(sensitive_features)
        costs = {"fp":0.1, "fn":0.9}
    
        # we cannot check the method on the same instance, because data cannot be loaded
        # twice to the same instance)
        errorrate_np = ErrorRate(costs=costs)
        errorrate_pd = ErrorRate(costs=costs)
        errorrate_pl = ErrorRate(costs=costs)
    
        # check `ErrorRate.load_data()` works with all input types
        errorrate_np.load_data(X, y, sensitive_features=sensitive_features)
        errorrate_pd.load_data(X_pd, y_pd, sensitive_features=sensitive_features_pd)
        errorrate_pl.load_data(X_pl, y_pl, sensitive_features=sensitive_features_pl)
    
        # check `ErrorRate.gamma()` returns similar returns for all input types
        classifier = LogisticRegression().fit(X, y)
        error_np = errorrate_np.gamma(classifier.predict)
        assert isinstance(error_np, pd.Series)
    
        classifier = LogisticRegression().fit(X_pd, y_pd)
        error_pd = errorrate_pd.gamma(classifier.predict)
        assert isinstance(error_pd, pd.Series)
    
        # TODO: do we need with sklearn.config_context(transform_output='polars') here?
        classifier = LogisticRegression().fit(X_pl, y_pl)
>       error_pl = errorrate_pl.gamma(classifier.predict) # TODO: need some feature_name handling in the polars case

X          = array([[-0.11808479,  0.45970566, -0.03800853,  0.36803315,  0.60600995,
        -1.55662917,  1.75479418,  1.69645637...3273, -0.79302842, -0.36001065,  0.79166269,
         1.15811087,  0.62834551, -0.89725437,  0.62411982,  0.        ]])
X_pd       =            0         1         2         3         4         5         6         7         8    9
0  -0.118085  0.4597... 0.875483 -0.714533 -0.793028 -0.360011  0.791663  1.158111  0.628346 -0.897254  0.624120  0.0

[100 rows x 10 columns]
X_pl       = shape: (100, 10)
┌───────────┬───────────┬───────────┬───────────┬───┬───────────┬───────────┬───────────┬──────────┐
...412   ┆ 0.0      │
└───────────┴───────────┴───────────┴───────────┴───┴───────────┴───────────┴───────────┴──────────┘
classifier = LogisticRegression()
costs      = {'fn': 0.9, 'fp': 0.1}
error_np   = 0    0.139
Name: error, dtype: float64
error_pd   = 0    0.139
Name: error, dtype: float64
errorrate_np = <fairlearn.reductions._moments.error_rate.ErrorRate object at 0x7a219d21ff50>
errorrate_pd = <fairlearn.reductions._moments.error_rate.ErrorRate object at 0x7a219d21fec0>
errorrate_pl = <fairlearn.reductions._moments.error_rate.ErrorRate object at 0x7a219d21fe90>
rng        = Generator(PCG64) at 0x7A219D258580
sensitive_features = array([0., 1., 1., 0., 0., 1., 0., 1., 0., 0., 1., 1., 1., 1., 1., 1., 1.,
       0., 1., 0., 1., 0., 0., 1., 1., 1., ...0., 0., 1., 0., 0., 0., 1., 1., 1., 0., 0., 1., 1.,
       1., 0., 0., 1., 1., 0., 1., 1., 0., 1., 0., 0., 1., 1., 0.])
sensitive_features_pd = 0     0.0
1     1.0
2     1.0
3     0.0
4     0.0
     ... 
95    0.0
96    0.0
97    1.0
98    1.0
99    0.0
Length: 100, dtype: float64
sensitive_features_pl = shape: (100,)
Series: '' [f64]
[
	0.0
	1.0
	1.0
	0.0
	0.0
	…
	0.0
	0.0
	1.0
	1.0
	0.0
]
y          = array([1, 1, 0, 0, 0, 1, 0, 0, 0, 0, 1, 0, 0, 0, 0, 1, 1, 0, 0, 1, 1, 0,
       0, 0, 0, 0, 0, 1, 1, 1, 1, 0, 0, 0, 0,...1,
       0, 0, 0, 0, 1, 1, 1, 1, 1, 0, 1, 0, 1, 0, 1, 0, 0, 0, 1, 1, 0, 1,
       0, 1, 1, 0, 1, 1, 1, 0, 1, 1, 1, 0])
y_pd       = 0     1
1     1
2     0
3     0
4     0
     ..
95    0
96    1
97    1
98    1
99    0
Length: 100, dtype: int64
y_pl       = shape: (100,)
Series: '' [i64]
[
	1
	1
	0
	0
	0
	…
	0
	1
	1
	1
	0
]

test/unit/reductions/moments/test_moments_error_rate.py:72: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
fairlearn/reductions/_moments/error_rate.py:89: in gamma
    pred = predictor(self.X)
        predictor  = <bound method LinearClassifierMixin.predict of LogisticRegression()>
        self       = <fairlearn.reductions._moments.error_rate.ErrorRate object at 0x7a219d21fe90>
../../../.pyenv/versions/3.12.2/envs/fairlearn_dev/lib/python3.12/site-packages/sklearn/linear_model/_base.py:374: in predict
    scores = self.decision_function(X)
        X          = ┌────────────────────────────────────────────────────────────────────────────────────────────────────┐
|              ...───┴──────────┘|
└────────────────────────────────────────────────────────────────────────────────────────────────────┘
        _          = False
        self       = LogisticRegression()
        xp         = <sklearn.utils._array_api._NumPyAPIWrapper object at 0x7a21a996be30>
../../../.pyenv/versions/3.12.2/envs/fairlearn_dev/lib/python3.12/site-packages/sklearn/linear_model/_base.py:351: in decision_function
    X = validate_data(self, X, accept_sparse="csr", reset=False)
        X          = ┌────────────────────────────────────────────────────────────────────────────────────────────────────┐
|              ...───┴──────────┘|
└────────────────────────────────────────────────────────────────────────────────────────────────────┘
        _          = False
        self       = LogisticRegression()
        xp         = <sklearn.utils._array_api._NumPyAPIWrapper object at 0x7a21a996be30>
../../../.pyenv/versions/3.12.2/envs/fairlearn_dev/lib/python3.12/site-packages/sklearn/utils/validation.py:2919: in validate_data
    _check_feature_names(_estimator, X, reset=reset)
        X          = ┌────────────────────────────────────────────────────────────────────────────────────────────────────┐
|              ...───┴──────────┘|
└────────────────────────────────────────────────────────────────────────────────────────────────────┘
        _estimator = LogisticRegression()
        check_params = {'accept_sparse': 'csr'}
        reset      = False
        skip_check_array = False
        validate_separately = False
        y          = 'no_validation'
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

estimator = LogisticRegression()
X = ┌────────────────────────────────────────────────────────────────────────────────────────────────────┐
|              ...───┴──────────┘|
└────────────────────────────────────────────────────────────────────────────────────────────────────┘

    def _check_feature_names(estimator, X, *, reset):
        """Set or check the `feature_names_in_` attribute of an estimator.
    
        .. versionadded:: 1.0
    
        .. versionchanged:: 1.6
            Moved from :class:`~sklearn.base.BaseEstimator` to
            :mod:`sklearn.utils.validation`.
    
        Parameters
        ----------
        estimator : estimator instance
            The estimator to validate the input for.
    
        X : {ndarray, dataframe} of shape (n_samples, n_features)
            The input samples.
    
        reset : bool
            Whether to reset the `feature_names_in_` attribute.
            If False, the input will be checked for consistency with
            feature names of data provided when reset was last True.
            .. note::
               It is recommended to call `reset=True` in `fit` and in the first
               call to `partial_fit`. All other methods that validate `X`
               should set `reset=False`.
        """
    
        if reset:
            feature_names_in = _get_feature_names(X)
            if feature_names_in is not None:
                estimator.feature_names_in_ = feature_names_in
            elif hasattr(estimator, "feature_names_in_"):
                # Delete the attribute when the estimator is fitted on a new dataset
                # that has no feature names.
                delattr(estimator, "feature_names_in_")
            return
    
        fitted_feature_names = getattr(estimator, "feature_names_in_", None)
        X_feature_names = _get_feature_names(X)
    
        if fitted_feature_names is None and X_feature_names is None:
            # no feature names seen in fit and in X
            return
    
        if X_feature_names is not None and fitted_feature_names is None:
            warnings.warn(
                f"X has feature names, but {estimator.__class__.__name__} was fitted "
                "without feature names"
            )
            return
    
        if X_feature_names is None and fitted_feature_names is not None:
>           warnings.warn(
                "X does not have valid feature names, but"
                f" {estimator.__class__.__name__} was fitted with feature names"
            )
E           UserWarning: X does not have valid feature names, but LogisticRegression was fitted with feature names
</details>

fairlearn/reductions/_moments/error_rate.py
```diff
@@ -79,21 +80,31 @@ def index(self) -> pd.Index:
...
+        # TODO: at the moment, native_namespace only depends on the user-passed type
+        # of X; but what should happen if the user pass inputs from several
+        # dataframe libraries?
+        if isinstance(self.X, np.ndarray):
+            # TODO (when dependency from pandas is removed): always return the type that
+            # the user has passed; for now: if user has passed np.array for X, still
+            # return a pd.Series as before
+            error = nw.new_series(name="error", values=error_value, native_namespace=pd)
+        else:
+            error = nw.new_series(name="error", values=[error_value], native_namespace=nw.get_native_namespace(self.X))
```
StefanieSenger, line 105: Passing values as a list (`values=[error_value]`) is necessary, because a series natively polars cannot be constructed from a numeric type. Commit [b46e3f3](https://github.com/fairlearn/fairlearn/pull/1526/commits/b46e3f3e62bab87f52da46049e842a8f2bef0f9c)

Otherwise this TypeError:
<details>
    def test_error_rate_narwhals_compatible():
        """Test that ErrorRate is compatible with any types passed via narwhals, that
        returns are the same for all the methods and that the input types are returned."""
    
        # make up data for testing
        rng = np.random.default_rng(42)
        X, y = make_classification(n_features=10, class_sep=0.1, random_state=42)
        X[:, -1] = rng.integers(0, 2, size=(X.shape[0],))
        sensitive_features = X[:, -1]
        X_pd, y_pd, sensitive_features_pd = pd.DataFrame(X), pd.Series(y), pd.Series(sensitive_features)
        X_pl, y_pl, sensitive_features_pl = pl.DataFrame(X), pl.Series(y), pl.Series(sensitive_features)
        costs = {"fp":0.1, "fn":0.9}
    
        # we cannot check the method on the same instance, because data cannot be loaded
        # twice to the same instance)
        errorrate_np = ErrorRate(costs=costs)
        errorrate_pd = ErrorRate(costs=costs)
        errorrate_pl = ErrorRate(costs=costs)
    
        # check `ErrorRate.load_data()` works with all input types
        errorrate_np.load_data(X, y, sensitive_features=sensitive_features)
        errorrate_pd.load_data(X_pd, y_pd, sensitive_features=sensitive_features_pd)
        errorrate_pl.load_data(X_pl, y_pl, sensitive_features=sensitive_features_pl)
    
        # check `ErrorRate.gamma()` returns similar returns for all input types
        classifier = LogisticRegression().fit(X, y)
        error_np = errorrate_np.gamma(classifier.predict)
        assert isinstance(error_np, pd.Series)
    
        classifier = LogisticRegression().fit(X_pd, y_pd)
        error_pd = errorrate_pd.gamma(classifier.predict)
        assert isinstance(error_pd, pd.Series)
    
        classifier = LogisticRegression().fit(X_pl, y_pl)
>       error_pl = errorrate_pl.gamma(classifier.predict)

test/unit/reductions/moments/test_moments_error_rate.py:71: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
fairlearn/reductions/_moments/error_rate.py:105: in gamma
    error = nw.new_series(name="error", values=error_value, native_namespace=nw.get_native_namespace(self.X))
../../../.pyenv/versions/3.12.2/envs/fairlearn_dev/lib/python3.12/site-packages/narwhals/stable/v1/__init__.py:2185: in new_series
    _new_series_impl(
../../../.pyenv/versions/3.12.2/envs/fairlearn_dev/lib/python3.12/site-packages/narwhals/functions.py:255: in _new_series_impl
    native_series = native_namespace.Series(name=name, values=values, dtype=dtype_pl)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <[AttributeError("'NoneType' object has no attribute 'as_str'") raised in repr()] Series object at 0x75467a917440>, name = 'error'
values = np.float64(0.139), dtype = None

    def __init__(
        self,
        name: str | ArrayLike | None = None,
        values: ArrayLike | None = None,
        dtype: PolarsDataType | None = None,
        *,
        strict: bool = True,
        nan_to_null: bool = False,
    ) -> None:
        # If 'Unknown' treat as None to trigger type inference
        if dtype == Unknown:
            dtype = None
        elif dtype is not None and not is_polars_dtype(dtype):
            dtype = parse_into_dtype(dtype)
    
        # Handle case where values are passed as the first argument
        original_name: str | None = None
        if name is None:
            name = ""
        elif isinstance(name, str):
            original_name = name
        else:
            if values is None:
                values = name
                name = ""
            else:
                msg = "Series name must be a string"
                raise TypeError(msg)
    
        if isinstance(values, Sequence):
            self._s = sequence_to_pyseries(
                name,
                values,
                dtype=dtype,
                strict=strict,
                nan_to_null=nan_to_null,
            )
    
        elif values is None:
            self._s = sequence_to_pyseries(name, [], dtype=dtype)
    
        elif _check_for_numpy(values) and isinstance(values, np.ndarray):
            self._s = numpy_to_pyseries(
                name, values, strict=strict, nan_to_null=nan_to_null
            )
            if values.dtype.type in [np.datetime64, np.timedelta64]:
                # cast to appropriate dtype, handling NaT values
                input_dtype = _resolve_temporal_dtype(None, values.dtype)
                dtype = _resolve_temporal_dtype(dtype, values.dtype)
                if dtype is not None:
                    self._s = (
                        # `values.dtype` has already been validated in
                        # `numpy_to_pyseries`, so `input_dtype` can't be `None`
                        self.cast(input_dtype, strict=False)  # type: ignore[arg-type]
                        .cast(dtype)
                        .scatter(np.argwhere(np.isnat(values)).flatten(), None)
                        ._s
                    )
                    return
    
            if dtype is not None:
                self._s = self.cast(dtype, strict=strict)._s
    
        elif _check_for_pyarrow(values) and isinstance(
            values, (pa.Array, pa.ChunkedArray)
        ):
            self._s = arrow_to_pyseries(name, values, dtype=dtype, strict=strict)
    
        elif _check_for_pandas(values) and isinstance(
            values, (pd.Series, pd.Index, pd.DatetimeIndex)
        ):
            self._s = pandas_to_pyseries(name, values, dtype=dtype, strict=strict)
    
        elif not hasattr(values, "__arrow_c_stream__") and _is_generator(values):
            self._s = iterable_to_pyseries(name, values, dtype=dtype, strict=strict)
    
        elif isinstance(values, Series):
            self._s = series_to_pyseries(
                original_name, values, dtype=dtype, strict=strict
            )
    
        elif isinstance(values, pl.DataFrame):
            self._s = dataframe_to_pyseries(
                original_name, values, dtype=dtype, strict=strict
            )
    
        elif hasattr(values, "__arrow_c_array__"):
            self._s = PySeries.from_arrow_c_array(values)
    
        elif hasattr(values, "__arrow_c_stream__"):
            self._s = PySeries.from_arrow_c_stream(values)
    
        else:
            msg = (
                f"Series constructor called with unsupported type {type(values).__name__!r}"
                " for the `values` parameter"
            )
>           raise TypeError(msg)
E           TypeError: Series constructor called with unsupported type 'float64' for the `values` parameter
</details>

fairlearn/reductions/_moments/moment.py
```diff
@@ -30,26 +31,40 @@ class Moment:
...
-        sensitive_features : :class:`pandas.Series`
+        # TODO: check if np.arrays or lists are also supported for sensitive_features; some subclasses do allow this I think
+        sensitive_features : Series object supported by narwhals
             The sensitive feature vector (default None)
         """
         assert self.data_loaded is False, "data can be loaded only once"
-        if sensitive_features is not None:
-            assert isinstance(sensitive_features, pd.Series)
+        if sensitive_features is None:
+            raise ValueError("The 'sensitive_features' parameter must be provided.")
```
StefanieSenger, line 56: TODO: Document on this change in the PR.

fairlearn/reductions/_moments/error_rate.py
```diff
@@ -70,39 +71,59 @@ def load_data(self, X, y, *, sensitive_features, control_features=None) -> None:
...
-    def index(self) -> pd.Index:
+    def index(self) -> List[str]:
         """Return the index listing the constraints."""
         return self._index
 
-    def gamma(self, predictor: Callable) -> pd.Series:
+    def gamma(self, predictor: Callable) -> nw.typing.IntoSeries:
         """Return the gamma values for the given predictor."""
         pred = predictor(self.X)
+        pred = nw.from_native(pred, pass_through=True, eager_only=True)
```
StefanieSenger, line 90: TODO: document in PR, why did I not make a narwhals df from X first?

fairlearn/reductions/_moments/moment.py
```diff
@@ -30,26 +31,37 @@ class Moment:
...
+        y : numpy.ndarray, or Series object supported by narwhals, list
             The label vector
-        sensitive_features : :class:`pandas.Series`
+        # TODO: check if np.arrays or lists are also supported for sensitive_features; some subclasses do allow this I think
+        sensitive_features : Series object supported by narwhals
             The sensitive feature vector (default None)
         """
         assert self.data_loaded is False, "data can be loaded only once"
-        if sensitive_features is not None:
-            assert isinstance(sensitive_features, pd.Series)
```
StefanieSenger, line 47: `sensitive_features` don't need to be `pd.Series` anymore.

Explanation: `sensitive_features` are passed from `_validate_and_reformat_input()` (via `ErrorRate.load_data()`) in which case currently it is passed as a pandas Series. But they can also be passed directly into `Moment.load_data()`, in which case they can be either None or a narwhals compatible series.

test/unit/reductions/moments/test_moments_error_rate.py
```diff
@@ -22,3 +27,54 @@ def test_bad_costs(bad_costs):
...
+def test_error_rate_narwhals_compatible():
+    """Test that ErrorRate is compatible with any types passed via narwhals, that
+    returns are the same for all the methods and that the input types are returned."""
+
+    # make up data for testing
+    rng = np.random.default_rng(42)
+    X, y = make_classification(n_features=10, class_sep=0.1, random_state=42)
+    X[:, -1] = rng.integers(0, 2, size=(X.shape[0],))
+    sensitive_features = X[:, -1]
+    X_pd, y_pd, sensitive_features_pd = (
```
StefanieSenger, line 44: TODO: check if narwhals has utility to make types for testing.

test/unit/reductions/moments/test_moments_error_rate.py
```diff
@@ -22,3 +27,54 @@ def test_bad_costs(bad_costs):
...
+    errorrate_pd = ErrorRate(costs=costs)
+    errorrate_pl = ErrorRate(costs=costs)
+
+    # check `ErrorRate.load_data()` works with all input types
+    errorrate_np.load_data(X, y, sensitive_features=sensitive_features)
+    errorrate_pd.load_data(X_pd, y_pd, sensitive_features=sensitive_features_pd)
+    errorrate_pl.load_data(X_pl, y_pl, sensitive_features=sensitive_features_pl)
+
+    # check `ErrorRate.gamma()` returns similar returns for all input types
+    classifier = LogisticRegression().fit(X, y)
```
StefanieSenger, line 68: TODO: replace by histgradientboosting

fairlearn/reductions/_moments/moment.py
```diff
@@ -30,26 +31,35 @@ class Moment:
...
         self.X = X
         self._y = y
-        self.tags = pd.DataFrame({_LABEL: y})
-        if sensitive_features is not None:
-            self.tags[_GROUP_ID] = sensitive_features
+        # TODO (when dependency from pandas is being removed): Dynamically change
+        # backends for y, sensitive_features and self.tags to the user's backend:
+        y = nw.new_series(name="y", values=y, native_namespace=pd)
```
StefanieSenger, line 57: TODO: why not from_native

fairlearn/reductions/_moments/error_rate.py
```diff
@@ -62,47 +63,77 @@ def __init__(self, *, costs: dict[Literal["fp", "fn"], float] | None = None):
...
         if isinstance(pred, np.ndarray):
             # TensorFlow is returning an (n,1) array, which results
             # in the subtraction in the 'error =' line generating an
             # (n,n) array
             pred = np.squeeze(pred)
         signed_errors = self.tags[_LABEL] - pred
-        total_fn_cost = np.sum(signed_errors[signed_errors > 0] * self.fn_cost)
-        total_fp_cost = np.sum(-signed_errors[signed_errors < 0] * self.fp_cost)
+        total_fn_cost = (signed_errors.filter(signed_errors > 0) * self.fn_cost).sum()
+        total_fp_cost = (signed_errors.filter(signed_errors < 0) * self.fp_cost * -1).sum()
```
StefanieSenger, line 108: TODO: check if they are the same

fairlearn/reductions/_moments/error_rate.py
```diff
@@ -62,47 +63,77 @@ def __init__(self, *, costs: dict[Literal["fp", "fn"], float] | None = None):
...
+        # (because then sf_train likely comes as a narwhals series, while now it comes
+        # as a pandas series):
+        sf_train = nw.from_native(sf_train, pass_through=True, eager_only=True)
+        # The following uses X so that the estimators get X untouched
         super().load_data(X, y_train, sensitive_features=sf_train)
         self._index = [_ALL]
 
     @property
-    def index(self) -> pd.Index:
+    def index(self) -> List[str]:
```
StefanieSenger, line 93: @TamaraAtanasoska: note on reminding you to check indexing

fairlearn/reductions/_moments/error_rate.py
```diff
@@ -62,47 +63,77 @@ def __init__(self, *, costs: dict[Literal["fp", "fn"], float] | None = None):
...
         signed_errors = self.tags[_LABEL] - pred
-        total_fn_cost = np.sum(signed_errors[signed_errors > 0] * self.fn_cost)
-        total_fp_cost = np.sum(-signed_errors[signed_errors < 0] * self.fp_cost)
+        total_fn_cost = (signed_errors.filter(signed_errors > 0) * self.fn_cost).sum()
+        total_fp_cost = (signed_errors.filter(signed_errors < 0) * self.fp_cost * -1).sum()
         error_value = (total_fn_cost + total_fp_cost) / self.total_samples
-        error = pd.Series(data=error_value, index=self.index)
+        # TODO: at the moment, native_namespace only depends on the user-passed type
+        # of X; but what should happen if the user pass inputs from several
+        # dataframe libraries?
```
StefanieSenger, line 112: TODO: raise error "Input should be of  same name namespace / type" in input validation, and add test

test/unit/reductions/moments/test_moments_error_rate.py
```diff
@@ -22,3 +28,65 @@ def test_bad_costs(bad_costs):
...
+    error_pd = errorrate_pd.gamma(classifier.predict)
+    assert isinstance(error_pd, pd.Series)
+
+    classifier = HistGradientBoostingClassifier().fit(X_pl, y_pl)
+    error_pl = errorrate_pl.gamma(classifier.predict)
+    assert isinstance(error_pl, pl.Series)
+
+    classifier = HistGradientBoostingClassifier().fit(X_pa, y_pa)
+    error_pa = errorrate_pa.gamma(classifier.predict)
+    assert isinstance(error_pa, (pa.Array, pa.ChunkedArray))
```
StefanieSenger, line 88: For reviewers:

`assert isinstance(error_pa, (pa.Array, pa.ChunkedArray))`

It seems that under circumstances `error_pa` can be a `pa.ChunkedArray` type instead of a `pa.Array` type. This is for memory efficiency, but I didn't get myself very far into this topic and I am not aware of what these circumstances are.

Is it alright to test for both types?

test/unit/reductions/moments/test_moments_error_rate.py
```diff
@@ -22,3 +28,64 @@ def test_bad_costs(bad_costs):
     with pytest.raises(ValueError) as execInfo:
         _ = ErrorRate(costs=bad_costs)
         assert _MESSAGE_BAD_COSTS in execInfo.value.args[0]
+
+
+def test_error_rate_narwhals_compatible():
+    """Test that `ErrorRate` is compatible with several dataframe backends via narwhals,
+    that return values are the same across different backends and that they come in
+    types from the dataframe backend used."""
```
StefanieSenger, line 36: For reviewers:

I had originally intended to parametrise this test, but since we want to compare the results at the end, that doesn't seem to make so much sense to me anymore.

Then I tried to manage the repetitions (especially in lines 60-71) by writing loops, but honestly, that's as long as it is now, but without the simplicity that comes with writing things up line by line.

Also, I didn't find a utility in narwhals to generate types for testing, which would maybe make this test less wordy, but in case there is one, I'd love to be pointed to that.

taharallouche (commented):
Thanks a lot for the great work @StefanieSenger 

I will be going slowly over this one, I left one small remark for now :pray:

romanlutz:
## Index Preservation in gamma() - Options for Discussion

Following up on @taharallouche's [comment](https://github.com/fairlearn/fairlearn/pull/1526#discussion_r2103203514) about the index being lost for non-pandas backends:

**The Problem:** `gamma()` returns a pandas Series with index `["all"]` that optimization algorithms rely on. Narwhals doesn't support pandas-style row indexes (Polars/PyArrow don't have them), so non-pandas backends lose this information.

### Options

| Option | Approach | Pros | Cons |
|--------|----------|------|------|
| **A** | Return a 2-column DataFrame `(index_col, value_col)` for all backends | Uniform API; index preserved | Breaking change for pandas users expecting Series |
| **B** | Keep pandas return as-is; return DataFrame for non-pandas | Backwards compatible for pandas | Inconsistent return types across backends |
| **C** | Always return pandas Series (convert non-pandas internally) | Uniform; fully backwards compatible | Loses "same-backend-out" goal; forces pandas in return path |
| **D** | Return a dict/namedtuple `{"index": [...], "values": [...]}` | Simple; no dataframe dependency | Unconventional; breaks existing code |

### Recommendation

**Option C** seems safest: `gamma()` is consumed by fairlearn's internal optimization algorithms (ExponentiatedGradient, GridSearch), not user-facing code. Those algorithms expect pandas Series with a specific index structure. The "same-backend-out" goal is valuable for user-facing APIs but less critical for internal algorithm plumbing.

Would appreciate @fairlearn/fairlearn-maintainers input on preferred direction before implementing.

adrinjalali:
I haven't looked in detail on this part in a long time, but I generally find relying on pandas indexing to this level quite brittle, and I rather have all the info on the dataframe itself, which would make using narwhals quite straight forward. I wouldn't mind a breaking change on this approach whenever it's exposed to the user, especially since it'll enable us being a lot more flexible with backend dataframes.

romanlutz:
> I haven't looked in detail on this part in a long time, but I generally find relying on pandas indexing to this level quite brittle, and I rather have all the info on the dataframe itself, which would make using narwhals quite straight forward. I wouldn't mind a breaking change on this approach whenever it's exposed to the user, especially since it'll enable us being a lot more flexible with backend dataframes.

Thanks for chiming in. This is very helpful. I haven't had too much exposure to narwhals to develop an opinion. I'll try this and report back 👍🏻

romanlutz:
Thanks @adrinjalali, that direction (put the constraint info on the dataframe as columns rather than relying on the pandas index) makes sense to me. I dug into how `gamma()`'s index is actually consumed to scope what a full move would entail. Sharing findings so we can agree on where to draw the line for this PR.

### How the index is consumed today
`gamma()` returns a value whose **index carries meaning** (`"all"` for `ErrorRate`; a `MultiIndex` of `(sign, event, group)` for `UtilityParity`). Two consumers use it, and **neither is narwhalified** — both are pure pandas and lean on index alignment:
| Usage | Location | Relies on |
|-------|----------|-----------|
| `self.obj.gamma(Q).iloc[0]` | `_lagrangian` (135, 249), `grid_search` (208) | **positional** `.iloc[0]` (objective path) |
| `gamma - self.constraints.bound()` | `_lagrangian` (144, 146, 175) | **label-aligned** subtraction |
| `h_gamma.dot(lambda_vec)` | `_lagrangian` (251, 254) | **label-aligned** dot product |
| `self.gammas[h_idx] = h_gamma` | `_lagrangian` (267) | column assignment aligned by index |
| `lambda_vec = pd.Series(x, self.constraints.index)` | `_lagrangian` (198) | constraint index labels |
| `lambda_vec[_ALL]` | `error_rate.signed_weights` (156) | **label-based** access `lambda_vec["all"]` |
| `pos_basis` / `neg_basis` / `project_lambda` | `utility_parity` (209-220) | `MultiIndex` membership tests |

### Two observations
1. When `ErrorRate` is used as the **objective** (the common case), its gamma index is accessed **positionally** (`.iloc[0]`) — the `"all"` label is never actually needed there.
2. The current polars/pyarrow return isn't consumable end-to-end yet: feeding a native polars series into `ExponentiatedGradient`/`GridSearch` would hit `.iloc`/index-alignment and fail. The new unit test only calls `gamma()` in isolation and checks types — it never runs the optimizer on non-pandas data.
### Scope of the full change
Doing this "properly" means putting identifiers in columns for **all** moments (`ErrorRate`, `UtilityParity` + 4 subclasses, `BoundedGroupLoss`) and rewriting the two optimization engines (`_lagrangian`, `_grid_search`, `_grid_generator`) plus `bound()` / `index()` / `project_lambda()` / `signed_weights()` to join on identifier columns instead of pandas alignment. That's effectively narwhalifying the whole reductions pipeline — bigger than this PR.
### Proposed split
- **This PR:** Make `ErrorRate.gamma()` return a small **DataFrame** with an explicit identifier column + value column (instead of a Series-with-index), and update the objective-path consumers (`.iloc[0]` → select the value column) so nothing breaks. This establishes the backend-agnostic pattern with minimal blast radius, since `ErrorRate` is objective-only in practice.
- **Follow-up PR(s):** Apply the same column-based pattern to `UtilityParity` / `BoundedGroupLoss` and fully de-index `_lagrangian` / `_grid_search` / `_grid_generator`.Does this split sound right, or would you rather tackle the optimizer de-indexing together with `ErrorRate` in one go? 

cc @taharallouche @StefanieSenger

adrinjalali:
I quite like the summary, in the sense that it makes me realize "why are we returning all that heterogeneous data in a single dataframe in the first place" 😅

I think it'd make sense to return a dict or a Bunch or even a data class with a bunch of attributes each of which a single value or a dataframe.

romanlutz:
Good point. Thinking through this more... for a single moment, gamma() isn't actually heterogeneous: it's a homogeneous vector of float violations. What's heterogeneous is the index labels, i.e., the (sign, event, group) tuple crammed into a MultiIndex alongside the value. So the cleanup you're pointing at is "stop encoding structured metadata in the index," which I'm fully on board with.

The one hard constraint to design around: the optimizer does vectorized numeric math on the gamma vector,  gamma - bound(), gamma.dot(lambda_vec), and it stacks them into a frame (self.gammas[i] = gamma) then does self.gammas.transpose().dot(lambda_vec). So whatever shape we pick, the numeric violations need to stay an aligned array/frame; only the identifiers should move out of the index.

That leaves two concrete options:

A. Long-format dataframe — identifier columns (sign, event, group) + a value column:
```
gamma() -> DataFrame
    sign  event  group   value
     +     all   grp_0    0.12
     -     all   grp_1    0.07
```


Backend-agnostic, tidy, single object. The optimizer selects the `value` column for its math and joins on identifiers instead of relying on index alignment.

**B. Bunch / dataclass** — separate the metadata from the numbers explicitly:


``` 
Gamma(labels=<frame of identifiers>, values=<array/series>)
```


Cleanest conceptual separation, but it's a new type to thread through `bound()`, `project_lambda()`, `lambda_vec`, and both optimizers so they line up.

My take: the Bunch/dataclass shape you're describing maps most naturally onto the **collection** the optimizer assembles per predictor (`error` scalar + `gamma` vector + `bound` vector + `lambda` vector), rather than a single `gamma()` call — that's arguably where a structured container adds the most clarity.

We should probably design the Bunch/dataclass container in a way that actually rewrites `_lagrangian` / `_grid_search`, since that's where it has to interoperate. I guess it could also be done as dataframe for this PR and then we redesign in a follow-up but it depends on what people prefer.

cc @taharallouche @StefanieSenger @adrinjalali

adrinjalali:
Yeah I think a dataclass is the cleanest option here, albeit a more substantial refactor. But it should make the code a lot easier to read.

PR #1527, opened 2025-03-11
MNT Refactor _validate_and_reformat_input

StefanieSenger:
This PR suggests both a simplification of `_validate_and_reformat_input` concerning the handling of `y` and an error handling using fairlearn errors, instead of relying on other libraries such as pandas or scikit-learn.

Trigger for the PR: In the context of #1522, this removes the necessity to type check for dataframe types and backends (this part of the code was dependent on pandas and I need it not to be for #1526). 
This function relied heavily on raising error messages from scikit-learn and pandas, but with introducing narwhals types, these narwhals errors might raise first, so it is better to raise fairlearn errors for input validation first, to have something to rely on in tests.
Apart from this, it looked very complicated and could be simplified.

Refactoring logic:
1. Top level check if we expect y.
2. Deal with the cases where y is invalid (None, empty, wrong shape, non-binary labels) and raise fairlearn error messages in these cases, to not depend on pandas or scikit-learn error messages and have stability.

I have also added a test.
And I have removed a file with an old fix used for testing in utils, that referred to a very old scikit_learn version that is not supported anymore.

(There would be much(!) more work to be done, to fully remove pandas dependency from `_validate_and_reformat_input` and prepare it for narwhals but that's probably for later and I would suggest to do this separately to make it easier reviewable.)

## Tests

- [ ] no new tests required
- [x] new tests added
- [ ] existing tests adjusted

## Documentation

- [x] API docs added or updated (no API is affected actually, but internal docs have been slightly updated)


StefanieSenger:
I am not sure if a changelog entry is necessary. The only change in behaviour is the different error message (error type remains equal):

Description of change in behaviour:
Originally, `raise ValueError(_MESSAGE_Y_NONE)` was present in the code, but it was unused and an error from scikit-learn's `check_array` was raised when `expect_y` was set to `True`, but an empty list, empty array or empty pandas series or dataframe was passed. In this PR, fairlearn's `_MESSAGE_Y_NONE` is raised instead.

TamaraAtanasoska:
Thanks @StefanieSenger! I believe no changelog entry is necessary for this one.

test/unit/reductions/grid_search/test_grid_search_arguments.py
```diff
@@ -319,9 +315,8 @@ def test_Y_df_bad_columns(self, transformX, transformA, A_two_dim):
         X, Y, A = _quick_data(A_two_dim)
 
         Y_two_col_df = pd.DataFrame({"a": Y, "b": Y})
-        with pytest.raises(ValueError) as execInfo:
+        with pytest.raises(ValueError, match="Found input variables with inconsistent"):
```
StefanieSenger, line 318: Comment on `match="Found input variables with inconsistent"`:

With the changes, this test now errors with a different scikit-learn error (indicating shape mismatch during cross-validation). The previous error was coming from scikit-learn as well. They both indicate to the same problems in the message, so I don't think it makes much of a difference?

Ideally, we would catch the ErrorType and display one of fairlearn's errors instead. Though I think it's not existing yet (I only found `_MESSAGE_X_SENSITIVE_ROWS`, but in this case here, it is that y and sensitive features have a different shape, I think. I could make a new one, but there are many other things that I also see are not so convenient with the error messages and I don't want to go overboard within this PR, which is already taking care of more than one thing.

test/unit/postprocessing/test_threshold_optimization.py
```diff
@@ -90,17 +90,14 @@ def test_none_input_data(X, y, sensitive_features, constraints):
...
     elif X is None:
-        with pytest.raises(ValueError) as exception:
+        with pytest.raises(ValueError, match="Expected 2D array, got scalar array instead"):
             adjusted_predictor.fit(X, y, sensitive_features=sensitive_features)
-        assert "Expected 2D array, got scalar array instead" in str(exception.value)
     elif sensitive_features is None:
-        with pytest.raises(ValueError) as exception:
+        with pytest.raises(ValueError, match=re.escape(_MESSAGE_SENSITIVE_FEATURES_NONE)):
             adjusted_predictor.fit(X, y, sensitive_features=sensitive_features)
-        assert str(exception.value) == _MESSAGE_SENSITIVE_FEATURES_NONE
```
StefanieSenger, line 103: Concering lines 96-100:

This is unrelated (sorry for that), but I couldn't hold back to update this, since the change in lines 93-94 (which is related to the PR) is in the same test and this way everything looks the same.

fairlearn/utils/_input_validation.py
```diff
@@ -77,25 +77,25 @@ def _validate_and_reformat_input(
...
-        raise ValueError(_MESSAGE_Y_NONE)
+        y = check_array(y.reshape(-1), ensure_2d=False, dtype="numeric", ensure_all_finite=False)
 
     result_X = check_array(X, dtype=None, ensure_all_finite=False, allow_nd=True)
     if isinstance(X, pd.DataFrame):
         result_X = pd.DataFrame(result_X)
 
+    if (y is not None) and y.shape[0] != result_X.shape[0]:
+        raise ValueError(_MESSAGE_X_Y_ROWS)
+
```
StefanieSenger, line 98: Adding this is necessary if we want to pass narwhals `y` and `X` though here. Otherwise, narwhals raises an error depending on the backend of y and it's easier not to be dependent on the error messages of all the backends on the fairlearn tests. See this [issue](https://github.com/narwhals-dev/narwhals/issues/2197) where I have passed the traceback I had gotten when I had passed `y` as a narwhals series.

StefanieSenger:
I have further worked on this and added a few more errors raising from fairlearn. Only `ValueErrors` from other libraries have been substituted by `ValueErrors` from fairlearn. Surely, these `ValueErrors` could be neater, but maybe this could be a new issue someone could work on.

This was all inspired from preparing this function to pass narwhals DataFrames and Series. Passing narwhals types makes it raise errors differently (and how exactly also depends on the backend used). This is why I think it's better to raise fairlearn errors before this happens. I have spend a few days in clutter before coming to this conclusion.

Status on `_validate_and_reformat_input`: it looks like it can now accept and deal with narwhals types (but still returns pandas types only, as this is the historical behaviour). I am not sure if it is fully narwhalified, since this was not my focus, but it works well for assisting `ErrorRate`.

fairlearn/utils/_input_validation.py
```diff
@@ -77,20 +77,18 @@ def _validate_and_reformat_input(
...
-    elif expect_y:
-        raise ValueError(_MESSAGE_Y_NONE)
+    if expect_y:
+        if y is None:
+            raise ValueError(_MESSAGE_Y_NONE + f", got y={y}.")
+        y = np.asarray(y)
+        if y.size == 0:
+            raise ValueError(_MESSAGE_Y_NONE + f", got y={y}.")
+        if y.ndim > 2 or (y.ndim == 2 and y.shape[1] != 1):
+            raise ValueError("`y` must be of shape (n,) or (n,1), got y of shape=({y.shape}).")
```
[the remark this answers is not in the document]
TamaraAtanasoska: thanks for catching that! I find it easier to read if the exact case we want the data to look like is violated, something like  `if not (y.ndim == 2 and y.shape[1] == 1) or y.ndim > 2` but this a nitpick and not a necessary change.

fairlearn/utils/_input_validation.py
TamaraAtanasoska: makes sense, so awesome they solved it so quick!

test/unit/postprocessing/test_threshold_optimization.py
```diff
@@ -90,17 +90,14 @@ def test_none_input_data(X, y, sensitive_features, constraints):
     )
 
     if y is None:
-        with pytest.raises(ValueError) as exception:
+        with pytest.raises(ValueError, match=re.escape("Must supply y, got y=None.")):
```
TamaraAtanasoska, line 93: would use this here `_MESSAGE_Y_NONE` just for consistency

test/unit/postprocessing/test_threshold_optimization.py
```diff
@@ -157,11 +154,13 @@ def test_threshold_optimization_different_input_lengths(data_X_y_sf, constraints
     inconsistent_exception_messages = {
         "inconsistent_sklearn": "Found input variables with inconsistent numbers of samples",
         "inconsistent_pandas": "All arrays must be of the same length",
+        "inconsistent_fairlearn": "X and y must have same number of rows",
     }
 
     empty_exception_messages = {
         "empty_sklearn": "Found array with 0 sample(s) (shape=(0,)) while a minimum of 1 is required.",
         "empty_pandas": "Found array with 0 sample",
+        "empty_fairlearn": "Must supply y",
```
TamaraAtanasoska, line 163: same here. not sure if any of the other ones have corresponding constants, would be nice to exchange them if we've missed that in the past

test/unit/postprocessing/test_threshold_optimization.py
```diff
@@ -503,15 +502,14 @@ def test_predict_different_argument_lengths(data_X_y_sf, constraints):
     )
 
     with pytest.raises(
-        ValueError, match="Found input variables with inconsistent numbers of samples"
+        ValueError,
+        match=re.escape("Found input variables with inconsistent numbers of samples: [20, 19]"),
```
TamaraAtanasoska, line 506: is the `re.escape` in this case necessary? isn't the string matched as is in full? same for the uses above where this PR introduces it, thanks for double checking that.

test/unit/postprocessing/test_threshold_optimization.py
```diff
@@ -503,15 +502,14 @@ def test_predict_different_argument_lengths(data_X_y_sf, constraints):
...
     ):
         adjusted_predictor.predict(
             data_X_y_sf.X, sensitive_features=data_X_y_sf.sensitive_features[:-1]
         )
 
-    with pytest.raises(
-        ValueError, match="Found input variables with inconsistent numbers of samples"
-    ):
+    with pytest.raises(ValueError, match="X and y must have same number of rows"):
```
TamaraAtanasoska, line 512: could this be replaced by `_MESSAGE_X_Y_ROWS`?

test/unit/reductions/grid_search/test_grid_search_arguments.py
```diff
@@ -37,7 +37,6 @@
 )
 from fairlearn.utils._input_validation import (
     _LABELS_NOT_0_1_ERROR_MESSAGE,
-    _MESSAGE_Y_NONE,
```
TamaraAtanasoska, line 40: I think I'd keep this for consistency as above, as it would be easy to change if we need to refactor as it is found across many files. How does `scikit-learn` deal with the constants for errors? same for the re here too, we should introduce it only if we need it.

TamaraAtanasoska (approved):
Thanks @StefanieSenger! A few style-related changes, no comments on the functionality. Otherwise LGTM. @taharallouche got some time to share what you think? I'd appreciate a second glance. It seems that it will be ok to remove the type preservation we mentioned here too: https://github.com/fairlearn/fairlearn/pull/1470#discussion_r1913456484

test/unit/postprocessing/test_threshold_optimization.py
StefanieSenger: Yes, all the `re.escape()` are necessary. I had only added them in response to failing tests. In this case, it's because of the square brackets.

test/unit/reductions/grid_search/test_grid_search_arguments.py
StefanieSenger: > I think I'd keep this for consistency as above, as it would be easy to change if we need to refactor as it is found across many files. 

I agree.

> How does scikit-learn deal with the constants for errors? 

In scikit-learn error messages aren't stored in constants, because the errors are all every specific and nothing gets reused. Keeping the constants for fairlearn suggests to contributors that these should stay more generic and prevents them from developing into more specialised messages. It is certainly out of scope of this PR to make decisions about that, and thus I understand it is better to keep to the conventions here.

> same for the re here too, we should introduce it only if we need it.

Yes, I agree there should be only re.escape if it is needed. Here, all the re.escape are necessary because the error messages contain characters that need to be escaped.

fairlearn/utils/_input_validation.py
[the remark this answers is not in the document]
StefanieSenger: I've re-factored into `if not (y.ndim == 1 or (y.ndim == 2 and y.shape[1] == 1))`.
Previously it was `if (y.ndim == 2 and y.shape[1] != 1) or y.ndim > 2`.

I am not sure which is easier to read and I don't have a preference.

fairlearn/utils/_input_validation.py
StefanieSenger: Oh yes, but since narwhals would still throw different errors for different backends, it will be easier to maintain if fairlearn tries to avoid these being triggered still and the fairlearn error is used in the tests.

StefanieSenger (commented):
Thanks for your review and your comments, @TamaraAtanasoska! :)
I have implemented all your suggestions.

test/unit/reductions/grid_search/test_grid_search_arguments.py
TamaraAtanasoska: Agreed with the explanation! We could potentially reformulate the message to not need the brackets which is what I had initially in mind with the comment, but as we (hopefully) expand the testing this might come up more often in other error messages, so the added import can be justified from this perspective. All good from my side then!

TamaraAtanasoska:
Thanks @StefanieSenger! LGTM from my side @taharallouche, since you last were on this side of the codebase I will leave it to you to take another glance when you have the time if something crucial is missed that I haven't seen functionality-wise, maybe some formatting usage I didn't anticipate, feel free to just merge if not.

fairlearn/utils/_input_validation.py
```diff
@@ -57,7 +57,7 @@ def _validate_and_reformat_input(
     X : numpy.ndarray, pandas.DataFrame
         The feature matrix
     y : numpy.ndarray, pandas.DataFrame, pandas.Series, or list
```
taharallouche, line 59: It can be `None` too. This wasn't introduced in this PR though

taharallouche (commented):
Thanks a lot for handling this @StefanieSenger :pray: I only have a couple of minor comments.

fairlearn/utils/_input_validation.py
```diff
@@ -16,7 +16,7 @@
 _KW_CONTROL_FEATURES = "control_features"
 
 _MESSAGE_X_NONE = "Must supply X"
-_MESSAGE_Y_NONE = "Must supply y"
+_MESSAGE_Y_NONE = "Must supply y, got y={y}."
```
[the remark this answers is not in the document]
StefanieSenger: Oh, you're right. Good that you've spotted this.

fairlearn/utils/_input_validation.py
```diff
@@ -57,7 +57,7 @@ def _validate_and_reformat_input(
     X : numpy.ndarray, pandas.DataFrame
         The feature matrix
     y : numpy.ndarray, pandas.DataFrame, pandas.Series, or list
-        The label vector
+        The label vector. Must be of shape (n,) or (n,1).
     expect_y : bool
         If True y needs to be provided, otherwise ignores the argument; default True
```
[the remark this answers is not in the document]
StefanieSenger: Oh yes, I have added this. That's not new in this PR though, scikit-learn's `check_array` would have risen before if an empty type was provided.

StefanieSenger:
Thanks a lot for your feedback, @taharallouche! :heart: 
I have changed the docstring to explain better the functions behaviour and added f-strings to the error messages.

fairlearn/utils/_input_validation.py
```diff
@@ -77,25 +77,25 @@ def _validate_and_reformat_input(
...
-        # Using an adapted version of check_array to avoid a warning in sklearn version < 1.6
-        y = check_array(y, ensure_2d=False, dtype="numeric", ensure_all_finite=False)
+    if expect_y:
+        if y is None:
+            raise ValueError(_MESSAGE_Y_NONE)
+        y = np.asarray(y)
+        if y.size == 0:
+            raise ValueError(_MESSAGE_Y_NONE)
+        if not (y.ndim == 1 or (y.ndim == 2 and y.shape[1] == 1)):
+            raise ValueError("`y` must be of shape (n,) or (n,1), got y of shape=({y.shape}).")
```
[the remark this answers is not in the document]
TamaraAtanasoska: 🤦‍♀️ I was too focused on content. second reviewers FTW!

TamaraAtanasoska:
congrats on your first merged PR @StefanieSenger 🥳!

Issue #1480, opened 2025-01-15
MNT Improving the adversarial module 


TamaraAtanasoska:
During the refactoring to make the adversarial module `scikit-learn` compliant, a few issues became known:
- the documentation is not fully up to date
- the [preprocessor](https://github.com/fairlearn/fairlearn/blob/main/fairlearn/adversarial/_preprocessor.py) can be very likely replaced with a scikit-learn [ColumnTransformer](https://scikit-learn.org/1.5/modules/generated/sklearn.compose.ColumnTransformer.html). Towards this effort there were big parts of that file that were removed, but the goal would be the remove it fully.
- Improving the testing, or adding tests that make sure that the module mitigates harm/improves over a baseline, and produces the results that the paper outlines.
- [this is a related issue](https://github.com/fairlearn/fairlearn/issues/1417), postponed for now while the needed changes are integrated into `keras`
- the code in the `_AdversarialMitigation` class can be further refactored and reduced.

TamaraAtanasoska:
@taharallouche as you mentioned the adversarial module in today's meeting, I put here a few things that I noticed could use improvement in the module while rushing through it to make it scikit-learn compatible, just as info. I suppose each of these can be its own issue if someone takes any on.

@adrinjalali do you maybe have something to add? This is all I have saved in my list. You reviewed all the PRs back then, maybe there is something else I am not remembering?

adrinjalali:
Can't remember anything on top of these, sounds good.

StefanieSenger:
Summing up our discussion that @TamaraAtanasoska and I had today about #1529:

- the code for `FloatTransformer` needs refactoring in multiple ways (work has already been done since it was added in #973) and the general direction (discussed ihere) is to remove `FloatTransformer` and substitute it's usage in AdversarialMitigation by scikit-learn's `ColumnTransformer` on the long term; or only by OneHotEncoder as a first step

Special points:
 - `type_of_target(X)` at it's terminology referring to targets are a bit misleading here
- passing `transformer` as a string in `FloatTransformer.__init__()` and then checking this string to raise an error in `fit()` if the wrong type was used, is tautologic: instead it should properly check the actual data to be transformed if it is of correct shape
[...]

PR #1529, opened 2025-03-15
MNT remove condition in FloatTransformer

StefanieSenger:
Started as a typo fix in check for error to raise (scikit-learn compatibility) in FloatTransformer.
But now I think, the check that contained the typo is not needed.


fairlearn/adversarial/_preprocessor.py
```diff
@@ -96,7 +96,7 @@ def fit(self, X, y=None):
                 self.n_features_out_ = sum(
                     len(cat) if len(cat) != 2 else 1 for cat in self.transform_.categories_
                 )
-            if self.inferred_type_ in ["continous-multioutut", "multiclass-multioutput"]:
+            if self.inferred_type_ in ["continous-multioutput", "multiclass-multioutput"]:
```
romanlutz, line 99: Continuous is also misspelled

StefanieSenger:
Thanks for the hint with the other typo, @romanlutz. Now this check should work.

I have also fixed some ruff warnings that had shown up in the CI:

```
warning: Invalid rule code provided to `# noqa` at fairlearn/metrics/_metric_frame.py:363: B902
warning: Invalid rule code provided to `# noqa` at fairlearn/metrics/_metric_frame.py:388: B902
warning: Invalid rule code provided to `# noqa` at fairlearn/show_versions.py:7: RST304
```

In contrast to `4 | from ._base_metrics import true_positive_rate  # noqa: F401; noqa: F401`, that had been fixed with #1530, these did not error but only raised a warning of being invalid and hence I think these can be removed. It passed my local ruff test.

StefanieSenger:
After fixing the typos properly, there were 2 errors, [in CI](https://github.com/fairlearn/fairlearn/actions/runs/13881279789/job/38840400863?pr=1529) and also locally:
```
FAILED test/unit/adversarial/test_adversarial_mitigation.py::test_valid_input_data_types[data2-continuous] - ValueError: Multioutput not supported
FAILED test/unit/adversarial/test_adversarial_mitigation.py::test_valid_input_data_types[data3-multiclass] - ValueError: Multioutput not supported
```

I think `FloatTransformer` wants to support having several columns for `X` and this is why I have removed this check. Please let me know what you think, @TamaraAtanasoska.

romanlutz:
Thanks @StefanieSenger ! I don't have much knowledge of this part of the code. Maybe someone else from @fairlearn/fairlearn-maintainers can take a look?

test/unit/adversarial/helper.py
```diff
@@ -198,14 +198,8 @@ def set_seed(seed):
 Cat[np.arange(rows), np.random.choice([i for i in range(cols)], size=(rows,))] = 1.0
 Cat1d = Cat[:, 0]
 Cont2d = np.random.rand(rows, cols)
-Cont1d = np.random.rand(
-    rows,
-)
-MultiClass2d, _ = make_classification(
-    random_state=42, n_classes=3, n_clusters_per_class=1, n_samples=rows, n_features=cols
-)
-MultiClass1d = np.random.choice([0.0, 1.0, 0.2], size=(rows,))
```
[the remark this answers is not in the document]
TamaraAtanasoska: this looks good thanks!

StefanieSenger:
Summing up our discussion that @TamaraAtanasoska and I had today about it:

- the code for `FloatTransformer` is broken on many ends (has always been (#973)) and the general direction (discussed in this issue #1480) is to remove `FloatTransformer` and substitute it's usage in AdversarialMitigation by scikit-learn's `ColumnTransformer` on the long term
- the dealing with the `X` in `FloatTransformer` has nothing to do with `type_of_target(X)` (which is misused and also it's terminology is misused --> it should actually distinguish between categorical and continuous data, not between multioutput or, multiclass or binary)
- passing `transformer` as a string in `FloatTransformer.__init__()` and then checking this string to raise an error in `fit()` if the wrong type was used, is tautologic (instead it should check the actual data to be transformed)
- `self.n_features_in_, self.n_features_out_ = X.shape[0], X.shape[1]` does not contain useful information about the number of features before and after `transform()` was called on it

My suggestions:
- I would suggest to allow any 2d data to be passed as `X` for now, not matter what is passed as string to `transform()`, since this is dealt with by one-hot-encoding in the
- on the mid term I would suggest to remove `FloatTransformer` from AdversarialMitigation and use OneHotEncoder instead (which is the only usecase of the former) and could be a step towards implementing support for ColumnTransformer afterwards

test/unit/adversarial/helper.py
```diff
@@ -201,10 +200,7 @@ def set_seed(seed):
 Cont1d = np.random.rand(
     rows,
 )
-MultiClass2d, _ = make_classification(
-    random_state=42, n_classes=3, n_clusters_per_class=1, n_samples=rows, n_features=cols
-)
-MultiClass1d = np.random.choice([0.0, 1.0, 0.2], size=(rows,))
+MultiClass2d = np.random.choice([0.0, 1.0, 2.0], size=(rows, cols))
```
StefanieSenger, line 203: This is the data to be used in the test 

```python
@pytest.mark.parametrize(
    "data, valid_choice",
    [
        (Bin2d, "binary"),
        (Cat, "binary"),
        (Cont2d, "continuous"),
        (MultiClass2d, "multiclass"),
    ],
)
def test_valid_input_data_types(data, valid_choice):
    """Test if the model processes the right data types"""
    prep = FloatTransformer(transformer=valid_choice)
    prep.fit(data)
    assert prep.n_features_in_ == data.shape[0]
    assert prep.n_features_out_ == data.shape[1]
```

in `test/unit/adversarial/test_adversarial_mitigation.py`.

I suggest to allow any 2d data to be passed as `X` for now, not matter what is passed as string for `valid_choice`.

fairlearn/adversarial/_preprocessor.py
```diff
@@ -96,8 +96,6 @@ def fit(self, X, y=None):
                 self.n_features_out_ = sum(
                     len(cat) if len(cat) != 2 else 1 for cat in self.transform_.categories_
                 )
-            if self.inferred_type_ in ["continous-multioutut", "multiclass-multioutput"]:
-                raise ValueError("Multioutput not supported")
         return self
```
TamaraAtanasoska, line 99: @hildeweerts I'd love some input here. I will need to read through the whole discussion [here](https://github.com/fairlearn/fairlearn/pull/973) again in more detail, but it seemed that the `multiclass-multioutput` case needed some extra implementation, thus was not supported as indicated in this [now deleted comment from the initial implementation](https://github.com/fairlearn/fairlearn/pull/973/files#diff-4532ad2fce46b0bb35763ffbb293ec1b71eb0306c6842a37a6989d9507e44392R254). Why does that particular case need a special treatment from a statistical perspective? I remember discussing this with @adrinjalali in person and we paired all the multioutput cases together then and hid them under the error of `"Not implemented"`, but @StefanieSenger questioning this and showing that the tests pass if those kind of input is allowed made me think if we made the right decision. 

There is a lot more work to be done on the module, some outlined in https://github.com/fairlearn/fairlearn/issues/1480, so for the scope of this PR I think we can stick to this question without opening more general refactoring. Rereading of the paper and making sure the necessary logic moves from the `FloatTransformer` to the `_AdversarialMitigation` class when it will be replaced with `ColumnTransformer` is necessary, as I am not sure if we are fully reproducing the paper at the moment, some of this is known as outlined in https://github.com/fairlearn/fairlearn/issues/1419. 

It is possible that there is no quick answer to this and we just need to patch to fix the typo and focus on larger refactoring in another PR when we have more informed opinions by digging more in detail into the paper and the logic initially implemented.

StefanieSenger: Discussing with @TamaraAtanasoska, I learned that `FloatTransformer`'s behaviour is tightly linked to the    AdversarialFairnessRegressor and AdversarialFairnessClassifier, too tightly properly. 

So, while `FloatTransformer` should be able to transform 2d data (X), if the AdversarialFairness* estimators cannot deal with y being multioutput, I think it should be dealt with there. So if this check has to rise due to not being supported, then maybe it could be moved there?

adrinjalali: Yes, I agree with this assessment.

hildeweerts: @TamaraAtanasoska (Caveats: I'm not a deep learning girlie and had limited time to look into this.)

The `multiclass-multioutput` scenario requires aggregating the losses of a set of `multiclass` problems (e.g. through summing or averaging the losses of the individual multiclass problems). This is not a _super_ common scenario (IIRC none of our other techniques support multioutput classification). From a quick skim, I don't think we are testing for multioutput `A` and `y`, so this could be one explanation of the tests still passing (might have missed something though). 

However, we allow for `continuous-multioutput` for the adversary (aiming to predict `A`), seemingly without having to do anything special (the loss function is just `MeanSquaredError`). From what I can tell, we might actually deal with the multioutput scenario natively too: [the number of nodes in the output layer depends on the number of features in the output variable (i.e. the number of variables in `Y` for the predictor and `A` for the adversary)](https://github.com/fairlearn/fairlearn/blob/35c3d7fa056b552ef8bd8f71c34f0a504f6d6542/fairlearn/adversarial/_backend_engine.py#L98C66-L98C76) and the used loss functions seem to support this directly (e.g. `torch.nn.MSELoss(reduction="mean")`). I might be missing something here, too.

I expect @MiroDudik @SeanMcCarren will have more thoughts, as they have a lot more experience with deep learning models and were most involved with the implementation of this module. 

(Relatedly, I'm not entirely sure why we are [using `CategoricalCrossentropy` in case of `multilabel-indicator`](https://github.com/fairlearn/fairlearn/blob/35c3d7fa056b552ef8bd8f71c34f0a504f6d6542/fairlearn/adversarial/_tensorflow_engine.py#L109), it seems binary cross entropy loss (averaged or summed across labels) would make more sense in that scenario, as multilabel classification assumed all labels are independent of each other, while I believe cross entropy relies on softmax activation? I'm not sure about the exact implementations in pytorch/tensorflow though)

TamaraAtanasoska:
@fairlearn/fairlearn-maintainers anyone to weigh in on this additionally? It seems we could merge this as is? I find it really important to solve the typo issue, and as there is no additional input, this is a good enough estimation? 

I've been raising multiple times the question of the integrity of the module and if we are really giving users what we promise with it, it could be that if it is difficult to maintain and we can't spend enough time to rewrite some of its parts that we could sunset the feature. Maybe a topic for the next Unconference.

adrinjalali (approved):

adrinjalali:
Not sure if we need a changelog here, but I'm happy for you to merge this as is @TamaraAtanasoska

PR #1531, opened 2025-03-18
MNT remove control_features passing into ErrorRate.load_data

StefanieSenger:
## Description
This PR suggests to remove the control_features param from `ErrorRate.load_data()`, which was unused and which we had discussed in the linked issue and PR:

closes https://github.com/fairlearn/fairlearn/issues/1496
discussed in #1502

## Tests

- [ ] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation


TamaraAtanasoska:
@taharallouche any chance you could help with this one? I know you participated in the original PR, maybe you can help move this forward? thanks 🙏

fairlearn/utils/_input_validation.py
```diff
@@ -168,3 +169,10 @@ def _join_names(names: Sequence[str]) -> str:
         )
 
     return np.array([_join_names(row) for row in feature_columns.astype(str)])
+
+
+def _filter_kwargs(func, kwargs):
+    """Helper function to filter out kwargs that are not accepted by `func`."""
+    sig = inspect.signature(func)
+    valid_params = set(sig.parameters.keys())
+    return {k: v for k, v in kwargs.items() if k in valid_params}
```
taharallouche, line 178: This seems good enough for the time being, given `ErrorRate.load_data`'s signature. However I believe it doesn't account for the possibility of `func` having `*args` and/or `**kwargs` in its signature (which is not the case now).

This being said, I'm ok with merging it as it is, I would just ask for some unit tests for this specific  function to cover such edge cases since it's now in our toolbox.

Also, less importantly, I think we can move it to `utils._common`, rather than `utils._input_validation`

taharallouche (commented):
Thanks for tackling this @StefanieSenger :pray: 

the TLDR of my comment:
- can you please add some unit tests to `_filter_kwargs`  for different signatures please ?
- we can move it to `utils._common`

PR #1533, opened 2025-03-25
MNT Narwhals in `utils._input_validations`

EdAbati:
## Description

Towards #1522 
(thanks @StefanieSenger for your refactoring PR, made this work way easier :) )

This PR makes the `utils _input_validations.py` narwhals compatible. Since all the rest doesn't expect `narwhals` yet, I convert everything to native before the return. This should be updated in the future :) 

FYI @TamaraAtanasoska I made the PR to `narwhals-integration` that is behind `main` (but branched of `main`): that's why you see some unrelated commits. We should update `narwhals-integration` to see a cleaner diff.

## Tests

- [ ] no new tests required
- [x] new tests added
- [ ] existing tests adjusted

## Documentation

## Screenshots

Tagging the narwhalifying experts @FBruzzesi @MarcoGorelli for suggestions on how to improve this


test/conftest.py
```diff
@@ -0,0 +1,62 @@
...
+    return pd.DataFrame(obj)  # type: ignore[no-any-return]
+
+
+# TODO: sklearn.utils.validation.check_array does not support it
+# def pandas_nullable_constructor(obj) -> IntoDataFrame:
+#     return pd.DataFrame(obj).convert_dtypes(dtype_backend="numpy_nullable")  # type: ignore[no-any-return]
+
+# TODO: sklearn.utils.validation.check_array does not support it
+# def pandas_pyarrow_constructor(obj) -> IntoDataFrame:
+#     return pd.DataFrame(obj).convert_dtypes(dtype_backend="pyarrow")  # type: ignore[no-any-return]
```
EdAbati, line 18: This is an interesing one.

To reproduce:
```python
import pandas as pd
from sklearn.utils.validation import check_array

df = pd.DataFrame({"alpha": ["a", "a", "b"], "beta": [1, 2, 1]}).convert_dtypes(dtype_backend="pyarrow")

check_array(df, dtype=None)
```

Did you encounter this already @Fbruzzesi? Maybe in `scikit-lego`?

FBruzzesi: I believe scikit-lego was very early days of Narwhals, and we are not testing with different pandas `dtype_backend` 🙈

fairlearn/utils/_input_validation.py
```diff
@@ -38,96 +46,103 @@
...
+
 def _validate_and_reformat_input(
-    X,
-    y=None,
-    expect_y=True,
-    expect_sensitive_features=True,
-    enforce_binary_labels=False,
-    **kwargs,
-):
+    X: np.ndarray | IntoFrameT,
```
StefanieSenger, line 66: Does fairlearn expect their users to know what `IntoFrameT` is?

Maybe it could help to link this to a part in the docs that explains what this type can all be.

I was natively thinking something like `X : numpy.ndarray, or DataFrame object supported by narwhals`, but I see that it's quite long.

TamaraAtanasoska:
Hi @EdAbati, just acknowledging today that I saw this PR! I think that the frequency of my engagement here is not enough to keep everything up to date on the spin-off branch fully unfortunately, I was expecting a different situation these weeks but here we are 😄. We have a maintainer meeting tomorrow, I might discuss to move the merges directly into `main`, as we are very likely going to postpone the patch release that we planned for. Not sure how fast we will move, but it would be nice to wrap up the first stable edition of the integration by the summer.

fairlearn/utils/_narwhals_compat.py
```diff
@@ -0,0 +1,50 @@
...
+        "No supported dataframe backend found. Please install either pandas or polars."
+    )
+
+
+def get_native_namespace_or_default(
+    X: np.ndarray | IntoFrame, y: np.ndarray | IntoSeries | IntoFrame | list | None
+) -> Any:
+    narwhals_objects = tuple(o for o in (X, y) if is_into_dataframe(o) or is_into_series(o))
+    if narwhals_objects:
+        return nw.get_native_namespace(*narwhals_objects)
```
StefanieSenger, line 39: Does it make sense to add raising an error of not all the inputs are from the same native namespace?

fairlearn/utils/_narwhals_compat.py
```diff
@@ -0,0 +1,50 @@
...
+    if get_pandas():
+        return Implementation.PANDAS
+    raise ImportError(
+        "No supported dataframe backend found. Please install either pandas or polars."
+    )
+
+
+def get_native_namespace_or_default(
+    X: np.ndarray | IntoFrame, y: np.ndarray | IntoSeries | IntoFrame | list | None
+) -> Any:
```
StefanieSenger, line 36: I think this could account for `sensitive_features` and `control_features` as well?

If I see this correctly, then whatever gets passed could be passed as `*args` right away, so there is not need to pass it nicely and then make it onto a tuple again.

fairlearn/utils/_narwhals_compat.py
MarcoGorelli: Yup, `get_native_namespace` raises an error:

https://narwhals-dev.github.io/narwhals/api-reference/narwhals/#narwhals.get_native_namespace

> Multiple objects can be passed positionally, in which case they must all have the same native namespace (else an error is raised).

EdAbati:
No problem at all @TamaraAtanasoska :)

I changed the base branch to `main` to make the review easier then.
Let me know if I should switch back to the integration one

fairlearn/utils/_narwhals_compat.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def all_to_native(*args) -> tuple:
+    """Convert all inputs to native objects if they are narwhals objects.
+
+    Temporary function to use while narwhals is not fully integrated.
+    """
+    return tuple(
+        nw.to_native(arg) if is_into_dataframe(arg) or is_into_series(arg) else arg for arg in args
+    )
```
FBruzzesi, line 50:
```suggestion
    return tuple(nw.to_native(arg, pass_through=True) for arg in args)
```

fairlearn/utils/_narwhals_compat.py
FBruzzesi: It might be worth to make that more visible in the docstring - I definitly had to search for such info

fairlearn/utils/_input_validation.py
FBruzzesi: Agree, I think it's fine for the annotation itself, but the docstring below can be left more generic/human readable

FBruzzesi (commented):
@EdAbati I found some time to skim through this again! Left ~a couple of comments~ actually one relevant comment and it does look 🔥 😁

EdAbati:
Apologies for the delay, but I finally found time to address the comments! 😄

Thank you for the review everyone, please let me know if there is anything else that doesn't look right

fairlearn/metrics/_plot_model_comparison.py
```diff
@@ -119,6 +120,8 @@ def plot_model_comparison(
         zeros((len(y_true), 1)),  # Dummy values for X
         y=y_true,
         sensitive_features=sensitive_features,
+        # TODO: remove this once we supports all narwhals inputs in this module
+        default_backend=nw.Implementation.PANDAS,
```
EdAbati, line 124: I had to add this in places were `narwhals` is not yet supported. Since we import `polars` in the tests, it becomes the default backend.

test/conftest.py
romanlutz: The root cause is that _validate_and_reformat_input routes X through check_array, which materializes to NumPy — so a string[pyarrow] (or nullable) column can't be cast and it blows up. It's not really fixable as long as validation goes through check_array; the proper fix is to validate via narwhals instead, which is out of scope here.

After merging main, the shared test/conftest.py now also parametrizes over the nullable/pyarrow dtype backends (other modules rely on them). To keep that coverage without breaking the utils tests, I added a local test/unit/utils/conftest.py that narrows constructor to the backends check_array actually supports (pandas, polars, native pyarrow table). I can open a follow-up issue to track proper support for pandas extension-dtype backends in input validation. Wdyt?

fairlearn/metrics/_plot_model_comparison.py
romanlutz: This is fine as a temporary workaround and the TODO makes the intent clear, so no objection to keeping it for this PR. The only reason it's needed is the .values calls right after, which are pandas-only. If we swap those for np.asarray(y_true) / np.asarray(sensitive_features) (works on both pandas and polars Series), the forced backend goes away entirely. Happy to make that small change here or leave it for the follow-up that removes these temporary shims — your call.

fairlearn/utils/_narwhals_compat.py
```diff
@@ -0,0 +1,77 @@
...
+    """
+    if get_polars():
+        return Implementation.POLARS
+    if get_pandas():
+        return Implementation.PANDAS
+    if get_pyarrow():
+        return Implementation.PYARROW
+    raise ImportError(
+        "No supported dataframe backend found. Please install either polars, pandas or pyarrow."
+    )
```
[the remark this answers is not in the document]
romanlutz: One thing to weigh: this default only kicks in when inputs aren't dataframes/series (e.g. NumPy) and no explicit backend is passed — it decides what type comes back out. With polars first, a user passing NumPy arrays gets polars Series back whenever polars is installed, which is a behavior shift for our historically pandas-centric users (and it's also what makes the default_backend=PANDAS workaround in the other thread necessary in tests).

My inclination is pandas > polars > pyarrow for backward compatibility, unless we explicitly want to push polars as the default output. Since it's a user-facing policy call, this feels like a good one to confirm with @fairlearn/fairlearn-maintainers @FBruzzesi, thoughts? I am fine with it either way.

PR #1534, opened 2025-03-25
DOC Fix sphinx-gallery text overflow and trailing doc warnings/errors

TamaraAtanasoska:
## Description
This PR fixes a few things:

- The issue pictured on the images down concerning text-overflow in the `sphinx-gallery` examples. The simplest solution was to reduce the length of the title, as the larger strings don't get automatically truncated and wouldn't be properly displayed if cut off manually. 

- An issue with the `MeanLoss` typing.

- Errors caused by referencing the same paper in both a class and a function. Not sure why this shows up, I will open an issue to ask. Right now I have resorted to just referring to the paper hardcoded in the two problematic functions as I couldn't find a better solution. The PR also adds the bibliography to the templates rather than the individual files, removing the repetition from a few places.

- Updates the `sphinx` logo path to match the `pydata-sphinx-template` settings as it wasn't showing after a build.

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

## Screenshots

The original:
Screenshot 2025-03-25 at 17 00 11
Changed:
Screenshot 2025-03-25 at 17 02 59


TamaraAtanasoska:
@romanlutz do you have time to look at this PR I made recently? It is mostly deletions and moving to templates and changing a few tiny things to rid the doc build of errors completely again.

docs/api_reference/index.rst
```diff
@@ -122,6 +122,7 @@ API Docs
    BoundedGroupLoss
    SquareLoss
    ZeroOneLoss
+   MeanLoss
```
romanlutz, line 125: Looks like these were perhaps sorted alphabetically at some point and then... not?

docs/templates/function.rst
```diff
@@ -6,6 +6,8 @@
 
 .. _sphx_glr_backreferences_{{ module }}.{{ objname }}:
 
+.. footbibliography::
```
romanlutz, line 9: Wow this is great!

fairlearn/reductions/_moments/utility_parity.py
```diff
@@ -136,7 +136,7 @@ def load_data(
         This adds a column `event` to the `tags` field.
 
         The `utilities` is a 2-d array which corresponds to g(X,A,Y,h(X))
-        from :footcite:t:`agarwal2018reductions`.
+        from Agarwal et al. (2018).
```
romanlutz, line 139: why not cite with reference?

romanlutz (approved):

fairlearn/reductions/_moments/utility_parity.py
hildeweerts: I had the same question, but then I noticed @TamaraAtanasoska already explained it in the PR description: 

> "Errors caused by referencing the same paper in both a class and a function. Not sure why this shows up, I will open an issue to ask. Right now I have resorted to just referring to the paper hardcoded in the two problematic functions as I couldn't find a better solution."

hildeweerts:
@TamaraAtanasoska can you open an issue re. the problem with double references in Fairlearn too, just so we don't forget?

fairlearn/reductions/_moments/utility_parity.py
TamaraAtanasoska: @hildeweerts I will open an issue directly with either the extension or `sphinx`, I need to check what is the best address for it. Maybe I am missing some conf setting or some other way to reference that won't clash, just reading the documentation wasn't telling me anything.

Issue #1069, opened 2022-04-07
Adding references to ref.bib file


alliesaizan:
#### Describe the issue linked to the documentation

Right now, references are scattered around the User Guide as needed for different sections. and not all articles cited are present in the bibtex file. This issue piggybacks on #1048 and #1056. 

#### Suggest a potential alternative/fix

Someone should look through the User Guide and aggregate all missing references into the _refs.bib_ file, and then replace references throughout with the new citation style. This fix was also suggested by @MiroDudik in PR #1032, which has good guidance for implementation.

Michael-AI-ML:
I hope to help out with this documentation. Perhaps I need access to view the _refs.bib_ file to help.

Nobody4591:
Hi @hildeweerts. Can I start working on this issue?

romanlutz:
@Nobody4591 I think it's fair to say that nobody is actively working on it. Go ahead!

Michael-AI-ML:
Sorry team. I took this on right when I was in the midst of some job interviews. Subsequent to that, I got a new job so I was unable to devote the time I had planned. Thanks @Nobody4591 for taking this on. My apologies to the rest of the FairLearn Team. Cheers

Nobody4591:
No problem and congratulations for the job 🎉

romanlutz:
Congratulations @Michael-AI-ML ! If there's some other way you'd like to stay involved that works with your job, let's talk. No pressure, though! Nothing to apologize for, life happens and this is a very positive piece of news. Thanks for sharing 🙂

romanlutz:
Am I right in assuming that `boston_housing_dataset.rst` is the only remaining file with such references? I searched for `[#1]` which should be sufficient (?)

hildeweerts:
@romanlutz Considering footnotes, that should be it. But on some pages of our documentation, we also use direct links to research papers rather than proper references, e.g. [in the contributor guide](https://github.com/fairlearn/fairlearn/blob/main/docs/contributor_guide/contributing_example_notebooks.rst).

romanlutz:
Perhaps we should create a list of all remaining ones.
- [ ] https://github.com/fairlearn/fairlearn/blob/main/docs/contributor_guide/contributing_example_notebooks.rst
- [ ] quickstart.rst has a link to the "Adult" dataset. We want the quickstart tutorial revamped completely anyway, so perhaps it doesn't matter so much...
- [ ] further_resources.rst
[...]

PR #1538, opened 2025-04-08
DOC Added citations to contributing example notebooks and update refs.bib

Mariam-ke:
## Description
Fixes issue #1069 on aggregating references to the refs.bib file @TamaraAtanasoska 

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

### DIFF
--- a/docs/contributor_guide/contributing_example_notebooks.rst
+++ b/docs/contributor_guide/contributing_example_notebooks.rst
@@ -10,19 +10,15 @@ A good example notebook exhibits the following attributes:
 
 1. **Deployment context**: Describes a real deployment context, not just a
    dataset.
-2. **Real harms**: Focuses on real harms to real people.
-   See `Blodget et al. (2020) <https://arxiv.org/abs/2005.14050>`_.
+2. **Real harms**: Focuses on real harms to real people :footcite:`blodgett2020languagepower`.
 3. **Sociotechnical**: Models the Fairlearn team's value that fairness is a
    sociotechnical challenge.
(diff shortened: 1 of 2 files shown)


TamaraAtanasoska:
Hi @Mariam-ke thanks for the PR! The build is failing because it cannot download a dataset from the database online where it is usually available, it has nothing to do with what you did at the moment. This makes it though difficult for me to check the doc build unless I build it locally :) it looks good, I will restart the build tomorrow morning, this usually passes after a while on its own. 

One thing I noticed is that your linting settings might have reformatted the whole `ref.bib` file, but it would be best to avoid that as it makes it easier to review and it prevents a new linting rule that our linters will not maintain. Could you please just have the added change with the new references there? Thank you!

TamaraAtanasoska (changes requested):
Hi @Mariam-ke, the build now passes as expected, and I've verified that the changes are correct. Let me know when you update the `ref.bib` file as explained in the previous comment and we can merge :)

Mariam-ke:
Hi @TamaraAtanasoska I've restored the original formatting of refs.bib from upstream/main and only added the missing citation entries to avoid reformatting noise. I also made sure the pre-commit hooks pass cleanly.

Let me know if anything else needs adjusting and if it works now.

TamaraAtanasoska (approved):

TamaraAtanasoska:
congrats on merging your second PR @Mariam-ke!

Mariam-ke:
Thank you!!!

PR #1539, opened 2025-04-09
FIX Bug in the callback validation of AdversarialFairnessClassifier

### LINKED ISSUE #1537: validate() Outputting Zeroes for AdversarialClassifier DP_Diff, Accuracy, and Selection Rate

TamaraAtanasoska:
## Description
Fixes #1537.

There is a larger issue that causes this problem, as currently data validation and transformation are combined in one step. This gets the data in the validation function and one-hot encodes it before continuing with the fitting, and we then have a very different representation in the middle step to compare with the predictions, which go through the inverse transform in before the `predict` function outputs them. 

I am not sure of the computational impact of having to do this transformation each time. It would definitely be more optimal to break down the `_validate_input` function and refactor parts of it, but this fixes the bug for now.

There was a more intricate way of dealing with the `y` types and conversations before we simplified that, and I am a bit uncomfortable with the code quality/integrity state of the whole module as it is, and if it works as intended. I was really new to the project when I worked on making the module more `sklearn` compatible, and I remember a lot of issues, but it was difficult for me to make sense of it all without depending heavily on reviews at the time. 

There are multiple issues to work on the module in more detail at the moment:
- #1480
- #1419
- https://github.com/fairlearn/fairlearn/pull/1529#discussion_r1998613908

This is the output atm, early stopping is working again too: 
```
Output from /Users/tamara/FOSS/fairlearn/examples/plot_adversarial_fine_tuning.py
DP diff: 0.0003, accuracy: 0.6146, selection_rate: 0.5991
DP diff: 0.0000, accuracy: 0.7964, selection_rate: 0.0520
DP diff: 0.0035, accuracy: 0.8088, selection_rate: 0.0904
Output from /Users/tamara/FOSS/fairlearn/examples/plot_adversarial_fine_tuning.py
DP diff: 0.0334, accuracy: 0.8061, selection_rate: 0.9093
        accuracy  selection_rate
sex
Female  0.910637        0.931664
Male    0.754399        0.898240
```

cc @hildeweerts for review

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation


TamaraAtanasoska:
@hildeweerts again green after the doc build was fixed :) merging this will really help with the memory issues that https://github.com/fairlearn/fairlearn/pull/1568 solved

hildeweerts (approved):

Issue #1055, opened 2022-04-01
Nightly requirements-fixed builds are failing


romanlutz:
The title says it all. We need to look into that and get it fixed.

romanlutz:
#1183 makes them run on GitHub workflows via `all-deps` with the `requirementsPinned` arg set to `"true"`. I suspect they'll still fail and I have some guess as to why: 

Python 3.10 and scipy 1.4.1 don't work together since it wasn't built for that Python version. So we can either deactivate the workflow for that python version OR make the minimum requirements dynamic based on the Python version.

adrinjalali:
minimum requirement should probably install the minimum python as well ;)

TamaraAtanasoska:
oops seems that this is the exact same issue we have again in https://github.com/fairlearn/fairlearn/pull/1491

I interpreted the minimal requirements as using the minimum required package versions for each python version, as the code was written in a way that this was implied. If the minimum requirements need only the minimum python version then it would make sense for it to be a separate CI run. I did something else there today more in line with my first interpretation.

riedgar-ms:
I think we originally had pinned requirements for each supported Python version, but.... if it would make things simpler (and minimise our number of builds-which-fail-randomly), then just doing pinned requirements for the minimum supported Python version does make sense too.

TamaraAtanasoska:
@riedgar-ms thanks, it seems I got the idea right then. would you mind looking at the PR I tagged you in? it is a different take, but maybe it is ok for now?

TamaraAtanasoska:
closed with #1491

PR #1542, opened 2025-04-11
CI Update pinned requirements

TamaraAtanasoska:
## Description
Updated minimum package versions for `python` `3.12` and `3.13`. 
Commit 5753a732ab7843d508758f7ffed7febe55de949b needs to be reverted before merge, as it is just for the testing purposes of this PR to verify the local passes. 

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation


TamaraAtanasoska:
Hey @adrinjalali, thanks again for wanting to look over this again.
I added all the extra checks so we can see all the green out of my machine in https://github.com/fairlearn/fairlearn/commit/5753a732ab7843d508758f7ffed7febe55de949b including all the `--pinned` versions that aren't part of a normal PR and run just in the `nightly` as well as `python 3.13` checks. We need to remove this before merging. Side effect is knowing now that all our tests pass for `3.13` 😄. 

Small unrelated addition in [Update pyproject.toml black config](https://github.com/fairlearn/fairlearn/pull/1542/commits/9634d760eb2e8076450f4d7efdde225d774db805), I realised we haven't added `py312` there when poking around.

This is still a very temporary solution, and none of this is in any way anything that would concern an user as it is just for our `nightly`. 

Unfortunately the doc build is not passing because of flaky OpenML file, it will be back soon.

adrinjalali:
I don't think it's necessary to test minimal versions with every single python version. Testing with the lowest python version should be enough. You can do a `pip freeze` command in your CI to check the version of all dependencies after you install the `fairlearn` package in CI, to check what's installed. That should pretty much check all issues.

TamaraAtanasoska:
this isn't the initial idea cc this discussion here about the same issue earlier https://github.com/fairlearn/fairlearn/issues/1055#issuecomment-2790027176

TamaraAtanasoska:
if we decided to test only the minimal python version, the `--pinned` function needs to be removed and there should be an explicit separate conf for that as it is unnecessary. we might pause this until there is an agreement then, as it keeps circling back. if there is no consensus we can close this PR and I can reopen a new one in a few weeks, or anyone else that wants to pick this up in the meantime. we can also revert the last two in that case.

adrinjalali (approved):
> If you mean the testing CI specifically, that we don't need to do at all, it is just to prove it is green for me as I had many envs locally and it was difficult to keep track.

The CI needs a revamp anyway. I don't wanna picky here. You asked me to do an actual check and that's what I did. 😅  

If having a passing CI is the main thing we care about, sure. Feel free to merge when you're happy with it.

TamaraAtanasoska:
no, I mean I agree with you @adrinjalali, here is my initial question a few months ago if we really need it, https://github.com/fairlearn/fairlearn/issues/1490#issue-2820329663, but I felt like there is no consensus to remove it and I didn't wan us here to make that decision. but tbh let's just do that, nothing speaks strictly against it. I will remove the `--pinned` option and make a `requirements-min.txt` and add it to run during the nightly only. It will make the nightly much faster too since we won't check every python version twice.

TamaraAtanasoska:
@adrinjalali, please take another look :) I tested the workflow in this PR with https://github.com/fairlearn/fairlearn/pull/1542/commits/06d947255673b9c7ca2bc6cd3f1a0721e5a31ddf, works as expected: 
- https://github.com/fairlearn/fairlearn/actions/runs/14423204018/job/40448376254?pr=1542
- https://github.com/fairlearn/fairlearn/actions/runs/14423204018/job/40448376252?pr=1542
- https://github.com/fairlearn/fairlearn/actions/runs/14423204018/job/40448376251?pr=1542

Screenshot 2025-04-12 at 22 57 06

Now the workflow is only triggered in the nightly of course, removed in https://github.com/fairlearn/fairlearn/pull/1542/commits/a862bdd42baf69aad0a5b32247f06378d558fd75.

TamaraAtanasoska:
@adrinjalali thanks! seems to be some `narwhals` vs `pandas` stuff that were implemented in the meantime, I'll find a solution and ping again

TamaraAtanasoska:
@adrinjalali had a quick chat with @FBruzzesi, went to pick up a kid, was back in a few hours, there was a fix (thanks @FBruzzesi!) :D green and ready to merge! I find it all very slow now. I will work on optimising the doc build a bit, some examples are too slow. The rest hopefully will be fixed when someone has time to redo CI properly 🤞

PR #1546, opened 2025-04-17
MNT: Make `GroupFeature` dataframe agnostic

FBruzzesi:
## Description

Make `GroupFeature` dataframe agnostic via narwhals.

Opening as draft as I have a few questions/remarks (see code comments)

Related to #1522 

## Tests

- [ ] no new tests required
- [ ] new tests added
- [x] existing tests adjusted

## Documentation

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -40,6 +40,11 @@ filterwarnings = [
     "ignore:.*parameter is deprecated and will be removed in Pillow 13.*:DeprecationWarning"
 ]
 
+markers = [
+    "narwhals: tests that involve dataframe logic backed by Narwhals (deselect with '-m \"not narwhals\"')",
+    "mpl_image_compare: tests that require 'pytest-mpl' to be installed"
+]
+
 [tool.ruff]
 # Exclude a variety of commonly ignored directories.
 exclude = [
--- a/test/unit/metrics/test_annotated_metric_function.py
+++ b/test/unit/metrics/test_annotated_metric_function.py
@@ -9,6 +9,8 @@ from sklearn.metrics import accuracy_score, recall_score, roc_auc_score
 
 from fairlearn.metrics._annotated_metric_function import AnnotatedMetricFunction
 
+pytestmark = pytest.mark.narwhals
+
 
 def test_constructor_unnamed():
     fc = AnnotatedMetricFunction(func=recall_score, name=None)
--- a/test/unit/metrics/test_metricframe_aggregates.py
+++ b/test/unit/metrics/test_metricframe_aggregates.py
@@ -25,22 +25,43 @@ metric = [
 
 class Test1m1sf0cf:
     # Single metric supplied as callable
+
+    @classmethod
+    def setup_class(cls):
+        results = {}
+        for metric_fn in metric:
+            target = metrics.MetricFrame(
+                metrics=metric_fn, y_true=y_t, y_pred=y_p, sensitive_features=g_4
+            )
+
+            assert target.control_levels is None
+            assert isinstance(target.sensitive_levels, list)
+            assert target.sensitive_levels == ["sensitive_feature_0"]
+
+            overall = metric_fn(y_t, y_p)
+            mask_p = g_4 == "pp"
+            mask_q = g_4 == "q"
+            metric_p = metric_fn(y_t[mask_p], y_p[mask_p])
+            metric_q = metric_fn(y_t[mask_q], y_p[mask_q])
+
+            mfn = metric_fn.__name__
+
+            results[metric_fn.__name__] = {
+                "target": target,
+                "overall": overall,
+                "metric_p": metric_p,
+                "metric_q": metric_q,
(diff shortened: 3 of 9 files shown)


fairlearn/metrics/_group_feature.py
```diff
@@ -47,18 +47,21 @@ class GroupFeature:
 
     def __init__(self, base_name: str, feature_vector, index: int, name: str | None = None):
         """Help with the metrics."""
-        self.classes_ = np.unique(feature_vector)
-        self.raw_feature_ = feature_vector
+        nw_feature_vector = nw.from_native(feature_vector, pass_through=True, allow_series=True)
+        is_nw_series = isinstance(nw_feature_vector, nw.Series)
+
+        self.classes_ = np.unique(nw_feature_vector)
```
FBruzzesi, line 53: I noticed that `self.classes_` is set but unused. If commented out, the only failing tests are those that specifically test for the attribute value

fairlearn/metrics/_group_feature.py
```diff
@@ -47,18 +47,21 @@ class GroupFeature:
...
+        self.raw_feature_ = (
+            list(nw_feature_vector) if not is_nw_series else nw_feature_vector.to_list()
+        )
```
FBruzzesi, line 56: Since `raw_feature_` is _always_ casted to list in `MetricFrame`, I moved that directly here so that for those inputs that are converted to narwhals Series we can use the native `to_list` methods

fairlearn/metrics/_group_feature.py
```diff
@@ -47,18 +47,21 @@ class GroupFeature:
...
         if name is not None:
             self.name_ = name
-        elif isinstance(feature_vector, pd.Series):
-            if feature_vector.name is not None:
-                if isinstance(feature_vector.name, str):
-                    self.name_ = feature_vector.name
+        elif is_nw_series:
+            if (name_ := nw_feature_vector.name) is not None:
+                if isinstance(name_, str):
+                    self.name_ = name_
```
FBruzzesi, line 64: Polars does not allow `None` as series name, but when unnamed, then you get an empty string, hence that's what `self.name_` would be set to.

I didn't put too much thought into this, I just wanted to highlight the behavior

fairlearn/metrics/_metric_frame.py
```diff
@@ -259,10 +259,10 @@ def __init__(
 
         # Add sensitive and conditional features to all_data
         for sf in sf_list:
-            all_data[sf.name_] = list(sf.raw_feature_)
+            all_data[sf.name_] = sf.raw_feature_
         if cf_list is not None:
             for cf in cf_list:
-                all_data[cf.name_] = list(cf.raw_feature_)
+                all_data[cf.name_] = cf.raw_feature_
```
FBruzzesi, line 265: Related to [previous comment](https://github.com/fairlearn/fairlearn/pull/1546/files#r2049609262)

fairlearn/metrics/_group_feature.py
riedgar-ms: It's entirely possible that it was simply there because I thought I'd use it. Since `GroupFeature` is not public-facing, I've not objection to removing it if unused.

FBruzzesi: Thanks for the feedback and the information @riedgar-ms - I removed it and adjusted the tests that were checking for such attribute in https://github.com/fairlearn/fairlearn/pull/1546/commits/5e001b016b76e6c7c151a5ffb5aee9216c890622 and https://github.com/fairlearn/fairlearn/pull/1546/commits/131a084fc8bf428411511ab3a5709f4f721d89ff

It should be easy and straightforward to revert those changes in case you change your mind.

A compromise is to have it as a property, so that the computation of unique values is triggered _only_ if and when required. Let me know if this option is preferable

fairlearn/metrics/_group_feature.py
taharallouche: Ok looking at the caller code of this class (namely, the  `_process_features` method), I think this will not be a problem (at least it it will not create a new problem that doesn't exist with the current code if any).

[Out of the scope of this PR ofc] I also noticed that there is room for refactoring there (which can simplify this function as a side effect, e.g. GroupFeature is never called with a `name` argument that is not `None`) ...

docs/user_guide/installation_and_version_guide/v0.14.0.rst
```diff
@@ -11,3 +11,4 @@ Other improvements
 ------------------
 
 * Removed support for python 3.9 and added support for python 3.13: :pr:`1592` by :user:`Tamara Atanasoska <tamaraatanasoska>`.
+* Make :code:`GroupFeature` class dataframe agnostic through Narwhals: :pr:`1546` by :user:`Francesco Bruzzesi <FBruzzesi>`.
```
taharallouche: Since the class is not user facing, I don't think we need this.

PR #1560, opened 2025-07-07
MNT Ignore DeprecationWarning for scipy future incompatibility

parulgupta1004:
## Description
CI was failing due to unit-test failure with scipy-1.16.0 due to deprecation of [disp and iprint](https://docs.scipy.org/doc/scipy/release/1.16.0-notes.html#deprecated-features) used while trying to fit logistic regression model using scikit-learn. This PR should temporarily resolve the failure by suppressing 'DeprecationWarning' in unittest. Once the newer version of sklearn resolve the warnings, this can be removed.
Discussed the solution with @TamaraAtanasoska on discord server to go ahead with this approach for now.

There are more issues with tests but are not related to this change and can be fixed separately.

## Tests
- [X] no new tests required
- Tested using 'python -m pytest -s ./test/unit'

## Documentation

## Screenshots
Not Applicable


pyproject.toml
```diff
@@ -31,6 +31,9 @@ filterwarnings = [
 
     # Warnings in test_othermlpackages
     "default:The LGBMClassifier or classes from which it inherits use `_get_tags` and `_more_tags`:FutureWarning",
+
+    # Warnings in scipy.optimize due to incompatibility with sklearn
```
TamaraAtanasoska, line 35: Does it appear just for the #all-deps errors? If so, I would place it right under the rest from that category. Otherwise we can keep it as it is here, if it is appearing in more builds.

PR #1597, opened 2025-10-22
tests: Add Narwhals pytest marker

### LINKED ISSUE #1555: TEST Add pytest marker for narwhals
# Description

This might look like an odd request, but please bear with me.
In Narwhals we take some effort to make sure that what we release/merge does not end up breaking downstream libraries that use Narwhals. We run tests of these libraries as part of our usual CI with Narwhals installing from the PR codebase (see [downstream_tests.yaml](https://github.com/narwhals-dev/narwhals/blob/main/.github/workflows/downstream_tests.yml)).

For some libraries, these test might take a while and we end up running only specific tests (see [darts example](https://github.com/narwhals-dev/narwhals/blob/63c8e4771a1df4e0bfeea5559c303a4a447d5cc2/.github/workflows/downstream_tests.yml#L534-L536)).
In a way this is a hack which might break easily in (at least) the following two scenarios:

- A test is moved -> breaks our CI (happened with plotly)
- A new test is added and we don't run it -> silent "failure"

I would consider Fairlearn to be among the libraries which tests can be quite long in their entirety. 

---

In light of the progressive Narwhals adoption in Fairlearn, I am here to ask what should we test, that also runs in reasonable time.

Would the minimal-deps test suffice? Will there be in the future parts of the code outside of what's tested with minimal-deps workflow that use Narwhals, either directly or indirectly?

One idea would be to add a `pytestmark = pytest.mark.narwhals` at the top of the test files that run with Narwhals.
In this way we simply run `pytest test -m narwhals` to collect all and only these tests, without the need to change it over time.

Let me know if this is a viable idea.

---

Narwhals issue to track it: [narwhals#2603](https://github.com/narwhals-dev/narwhals/issues/2603)

FBruzzesi:
## Description

Closes #1555

## Tests

- [ ] no new tests required
- [ ] new tests added
- [x] existing tests adjusted

## Documentation

## Screenshots

It's possible to run `pytest test/unit -m narwhals` to execute only the relevant tests:

image


taharallouche:
> @taharallouche do you have any insight here? I am not able to reproduce there failures locally independently of which python and library versions I use 😭

Unfortunately this is a recurrent failure where the pipeline fails to pull some data from the OpenMl API 🫠 It is not related to these changes

FBruzzesi:
> Unfortunately this is a recurrent failure where the pipeline fails to pull some data from the OpenMl API 🫠 It is not related to these changes

Oh I see! Maybe there is a way to use the github runners caches to store the datasets there so that you can avoid the API call  completely?
I tried locally, the total size is <25Mb with one dataset (acs_income) which is 22Mb. Not sure if you want to have that in the repo tests, 25Mb seems on the high end to me, but github caches might work. I can try to put a PR if this is of interest 🙏🏼

taharallouche:
> > Unfortunately this is a recurrent failure where the pipeline fails to pull some data from the OpenMl API 🫠 It is not related to these changes
> 
> Oh I see! Maybe there is a way to use the github runners caches to store the datasets there so that you can avoid the API call completely? I tried locally, the total size is <25Mb with one dataset (acs_income) which is 22Mb. Not sure if you want to have that in the repo tests, 25Mb seems on the high end to me, but github caches might work. I can try to put a PR if this is of interest 🙏🏼

Whenever you have time to try this please do, that would be very nice and would benefit all the parties (less pings for openml, less failures on our side) ! @FBruzzesi :pray:

TamaraAtanasoska:
Unfortunately we can't download and serve the datasets legally because of licensing, at least that is how it was explained to me. When we had the big OpenML outage last year that left most of us not being able to work, including scikit-learn, I was involved then, some projects downloaded and served caches on their own, but it is a bit of a grey zone . We have some caching implemented in the actions themselves, but it might not work.

PR #1600, opened 2025-10-25
MNT avoid sys module patching leak

taharallouche:
## Description
We noticed that the side effects of the adversarial backend patch [leaked](https://github.com/fairlearn/fairlearn/pull/1596#discussion_r2462646077) to other modules, which is highly undesirable.

This obviously wouldn't replace getting rid of the mock in favour of real instances, but at least makes the side effects contained to the relevent tests.

Technically, for all tests besides the scikit-learn checks only require adding a pytest fixture to handle the sys modules patch. For the estimator checks, we need an extra context manager to do that when instanciating the mitigator instance at parametrization time. 

## Tests

- [x] no new tests required
- [ ] new tests added
- [x] existing tests adjusted

## Documentation


taharallouche:
@TamaraAtanasoska This is basically what I had in mind

riedgar-ms (approved):

PR #1079, opened 2022-04-29
ENH Add "adversarial debiasing" 2nd PR

SeanMcCarren:
A clean slate PR of #973 so feedback is cleared. 
- Base class is made private.
- Subclass API is more digestible as we removed arbitrary choices of loss functions and preprocessors from this description.

PR #1617, opened 2026-01-18
Fixed input dimension calculation for >2D arrays in BackendEngine

MiloszDev:
Fixed input dimension calculation for >2D arrays in BackendEngine

## Description
This PR fixes a bug in `fairlearn.adversarial._backend_engine.py` where multi-dimensional inputs, like images, weren’t being handled correctly. The number of input features is now calculated using `numpy.prod(X.shape[1:])`, so all dimensions are counted and the neural network is initialized with the correct input size. Before, only the first non-batch dimension was considered, which could lead to wrong assumptions about the input shape.

## Tests

- [ ] no new tests required
- [x] new tests added
- [ ] existing tests adjusted

## Documentation

### DIFF
--- a/fairlearn/adversarial/_backend_engine.py
+++ b/fairlearn/adversarial/_backend_engine.py
@@ -1,7 +1,7 @@
 # Copyright (c) Fairlearn contributors.
 # Licensed under the MIT License.
 
-from numpy import ndarray
+from numpy import ndarray, prod
 from sklearn.utils import shuffle
 
 from ._constants import (
@@ -33,7 +33,7 @@ class BackendEngine:
         """
         self.base = base
 
-        n_X_features = X.shape[1]  # FIXME: what if X.ndim > 2?
+        n_X_features = int(prod(X.shape[1:]))
         n_Y_features = base._y_transform.n_features_out_
         n_A_features = base._sf_transform.n_features_out_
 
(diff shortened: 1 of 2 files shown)


TamaraAtanasoska:
I unfortunately do not remember any more what the conclusion was, but I remember this coming up in conversations with @adrinjalali almost two years ago, but we left it as is. I wouldn't change these without reading the paper again and looking through the whole implementation, but I understand that is a lot of work.

romanlutz:
Potentially relevant context may be:

Issue #1656 — Created recently (June 2026 by me as I was trying to look for cleanup suggestions with AI) — specifically tracks this X.shape[1] bug and suggests either:
- Add explicit shape check (error when X.ndim != 2), or
- Flatten/accept higher-dimensional inputs intentionally
The original PR #1079 (Oct 2022) added the adversarial module with the FIXME comment already present, suggesting the author knew this was a gap but deferred it.

Issue #1480 has a comment from [@StefanieSenger](https://github.com/StefanieSenger) noting:
> self.n_features_in_, self.n_features_out_ = X.shape[0], X.shape[1] stores the number of rows and columns... but these are not the correct values

My recommendation is to close this PR and then proceed with the explicit shape check I recommended in #1656 until someone wants to do the more extensive fix.

TamaraAtanasoska:
@romanlutz I think that is a good idea! I'll close it. thanks @MiloszDev for bring this up again.

PR #1629, opened 2026-02-27
MAINT: Port project to pyproject.toml

tiran:
## Description

Port packaging metadata from `setup.py` + `MANIFEST.in` + `requirements.txt` into a single `pyproject.toml`. This is a metadata-only refactor — the build backend stays on `setuptools`, the release process is unchanged, and runtime/dev dependency lists are unaffected.

### What changed

- `pyproject.toml` now contains the project metadata (name, description, classifiers, runtime dependencies, Python requirement, etc.) using the PEP 621 schema.
- `setup.py`, `MANIFEST.in`, and `requirements.txt` are removed; their behavior is preserved in `pyproject.toml`:
  - Runtime dependencies → `[project.dependencies]`
  - Test exclusion (formerly via `find_packages(exclude=...)` in `setup.py`) → `[tool.setuptools.packages.find]`
  - Inclusion of `requirements*.txt` in the sdist (formerly in `MANIFEST.in`, see #1628) is no longer needed since runtime deps are now in `pyproject.toml` itself.

### What deliberately did NOT change

- **Build backend.** Still `setuptools.build_meta`. An earlier revision of this PR proposed switching to `hatchling`, but that change is orthogonal to consolidating metadata into `pyproject.toml` and was reverted to keep this PR minimal.
- **Versioning.** `__version__` remains a static string in `fairlearn/__init__.py`. An earlier revision used `hatch-vcs` to derive the version from `git describe`, but Fairlearn's release tags live on `release/v<x.y>.X` branches and are never merged back to `main`, so `git describe` on `main` only sees `v0.12.0` and produced wheels versioned `0.12.1.dev111+…` instead of `0.15.0.dev0`. Static versioning preserves the existing release workflow ([docs/contributor_guide/release.rst](https://github.com/fairlearn/fairlearn/blob/main/docs/contributor_guide/release.rst)) exactly — a release PR still bumps the literal in `__init__.py`.
- **The classifier list, minus the `License ::` entry.** `setuptools>=78` rejects the legacy `License :: OSI Approved :: MIT License` classifier when a PEP 639 SPDX `license = "MIT"` expression is also present, so the classifier was dropped (the SPDX expression replaces it).

### Caller cleanups

A few scripts and docs still referenced the removed files:

- `Makefile`: `setup.py build_ext -i` → `pip install -e .` (Fairlearn has no C extensions, so the old `build_ext` was a no-op left over from the setuptools era).
- `scripts/install_requirements.py`: no longer references the removed `requirements.txt`. Runtime deps now come from the existing `pip install .` step in each CI workflow; this script handles `requirements-dev.txt` (default) and `requirements-min.txt` (`--min`).
- `scripts/build_wheels.py` / `scripts/process_readme.py`: a new `_get_fairlearn_version()` helper in `scripts/_utils.py` reads `__version__` statically from `fairlearn/__init__.py` via AST parsing, removing the up-front `pip install -e .` step in `build_wheels.py` that existed only so it could `import fairlearn` to read the version.
- `docs/contributor_guide/development_process.rst`: the "Requirements Files" section is rewritten to describe `pyproject.toml` as the source of runtime deps; the broken links to `setup.py` and `requirements.txt` are removed.

### Verified locally

- `python -m build` produces `fairlearn-0.15.0.dev0-py3-none-any.whl` + sdist with the correct version and no `test/` files bundled.
- `pip install .` and `pip install -e .` both succeed and import correctly.
- `python scripts/build_wheels.py --version-filename version.txt` succeeds without a prior `pip install`.
- `python scripts/install_requirements.py` installs `requirements-dev.txt`.
- `python scripts/install_requirements.py --min` (on Python 3.11) installs `requirements-min.txt`, and a subsequent `pip install .` resolves cleanly with the pinned minimums (numpy 1.24.4, scipy 1.9.3, scikit-learn 1.6.0).

## Tests
- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation
- [x] contributor guide updated (`development_process.rst` — "Requirements Files" section)


romanlutz:
Quick note on why I reverted the `hatchling` + `hatch-vcs` parts of this PR (and kept just the `setup.py` → `pyproject.toml` metadata consolidation on the `setuptools` backend):

Our release-branch naming is `release/v<x.y>.X` — the literal `X` is a placeholder for the patch version, and the actual `v<x.y.z>` tags are placed on those release branches, **not** on `main`. For example, `v0.13.0` lives on `release/v0.13.X` and `v0.14.0` lives on `release/v0.14.X`; neither tag is reachable from `main`'s history. The newest tag `main` can see is `v0.12.0`.

`hatch-vcs` derives the version via `git describe`, which only walks ancestor commits. When I built this branch as it was originally proposed (with `hatch-vcs`), the resulting wheel was versioned `0.12.1.dev111+ga43ed90da` instead of `0.15.0.dev0` — i.e., it would silently ship the wrong version to PyPI from any post-`v0.12.0` build on `main`.

We could fix that by either changing the release-tagging workflow (tag on `main` too, or fast-forward-merge release branches back) or by configuring a `fallback-version` in hatch-vcs — both are reasonable but neither is required by this PR, and either would also answer @riedgar-ms's earlier question about release-workflow impact. Keeping `__version__` as a static literal in `fairlearn/__init__.py` means zero changes to the release process documented in [`docs/contributor_guide/release.rst`](https://github.com/fairlearn/fairlearn/blob/main/docs/contributor_guide/release.rst).

Thanks for the original PR @tiran — happy to revisit the `hatchling` + VCS-versioning piece separately once we've decided how (or whether) to change the release-branch tagging convention.

romanlutz:
Waiting for #1666

romanlutz (approved):

tiran:
Excellent! Thanks for updating my PR and landing this change, @romanlutz 

About `hatchling`:
In the past two years, we had several issues with breaking changes in `setuptools` as well as broken releases of `setuptools`. Setuptools was created 20 years ago and suffers from backwards compatibility issues.

Thanks to PEP 517, the Python ecosystem has a multitude of build backends. We have moved most of our projects to `hatchling`. `hatchling` is a modern, versatile build system. Flit and `flit_core` are great choices for simple, pure Python projects, too. Flit's documentation https://flit.pypa.io/en/stable/rationale.html explains very well why package maintainers should consider alternatives.

`flit_core` and `hatchling` can read a `__version__` string from a Python module.
- https://hatch.pypa.io/1.12/version/#pyprojecttoml
- https://flit.pypa.io/en/stable/pyproject_toml.html#project-metadata

PR #1632, opened 2026-03-09
FIX deepcopy Moment constraints in ExponentiatedGradient and GridSearch (#1210)

### LINKED ISSUE #1210: data can be loaded only once
When I try to run the GridSearch twice or ExponentiatedGradient after GridSearch, the constraints returns the following error.

`AssertionError: data can be loaded only once`

Full stack trace is:

```
File ~\OneDrive - EY\fairness2\FSRM-shEYzam-repos\shazamlib\group_fairness.py:313, in GroupFairness.reduction_grid_search(self, base_model)
    311 # train model 
    312 print ('Training base model on specified constraints..')
--> 313 model_gridsearch.fit(self.X_train, self.y_train, sensitive_features=self.S_train)
    314 self.inprocess['model'] = model_gridsearch
    316 # make predictions

File ~\Anaconda3\envs\sheyzam-fairness-env\lib\site-packages\fairlearn\reductions\_grid_search\grid_search.py:143, in GridSearch.fit(self, X, y, **kwargs)
    141 # Prep the parity constraints and objective
    142 logger.debug("Preparing constraints and objective")
--> 143 self.constraints.load_data(X, y, **kwargs)
    144 objective = self.constraints.default_objective()
    145 objective.load_data(X, y, **kwargs)

File ~\Anaconda3\envs\sheyzam-fairness-env\lib\site-packages\fairlearn\reductions\_moments\utility_parity.py:333, in DemographicParity.load_data(self, X, y, sensitive_features, control_features)
    331 base_event = pd.Series(data=_ALL, index=y_train.index)
    332 event = _merge_event_and_control_columns(base_event, cf_train)
--> 333 super().load_data(X, y_train, event=event, sensitive_features=sf_train)

File ~\Anaconda3\envs\sheyzam-fairness-env\lib\site-packages\fairlearn\reductions\_moments\utility_parity.py:146, in UtilityParity.load_data(self, X, y, sensitive_features, event, utilities)
    123 def load_data(
    124     self,
    125     X,
   (...)
    130     utilities=None,
    131 ):
    132     """Load the specified data into this object.
    133 
    134     This adds a column `event` to the `tags` field.
   (...)
    144 
    145     """
--> 146     super().load_data(X, y, sensitive_features=sensitive_features)
    147     self.tags[_EVENT] = event
    148     if utilities is None:

File ~\Anaconda3\envs\sheyzam-fairness-env\lib\site-packages\fairlearn\reductions\_moments\moment.py:42, in Moment.load_data(self, X, y, sensitive_features)
     30 def load_data(self, X, y: pd.Series, *, sensitive_features: pd.Series = None):
     31     """Load a set of data for use by this object.
     32 
     33     Parameters
   (...)
     40         The sensitive feature vector (default None)
     41     """
---> 42     assert self.data_loaded is False, "data can be loaded only once"
     43     if sensitive_features is not None:
     44         assert isinstance(sensitive_features, pd.Series)

AssertionError: data can be loaded only once
```

BALOGUN-DAVID:
## Summary

`GridSearch.fit` and the `_Lagrangian` used by `ExponentiatedGradient.fit` previously passed the user-supplied `constraints` object directly to `load_data()`. Because `Moment.load_data` enforces an "only loaded once" assertion, calling `fit()` twice with the same constraints instance — or sharing one instance between estimators — raised `AssertionError: data can be loaded only once`.

Closes #1210.

## Change

Both estimators now `copy.deepcopy(constraints)` before calling `load_data()`. The user's original object is left untouched, and the loaded copy is exposed as the fitted attribute `constraints_` (sklearn convention) so users can still introspect post-fit constraint state via `eg.constraints_.gamma(predictor)`, `gs.constraints_._y_as_series`, etc.

`deepcopy` (rather than `copy.copy`) is intentional: custom `Moment` subclasses can hold mutable constructor state that a shallow copy would still share. An inline comment in both files records this.

## Files

* `fairlearn/reductions/_exponentiated_gradient/_lagrangian.py` — deepcopy `constraints` in `_Lagrangian.__init__`.
* `fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py` — set `self.constraints_ = lagrangian.constraints` post-fit; route the two remaining internal accesses (`_y_as_series`, `bound()`) through the lagrangian copy instead of `self.constraints`. Defensive `.copy()` on `self.lambda_vecs_`.
* `fairlearn/reductions/_grid_search/grid_search.py` — deepcopy `self.constraints` into a local `constraints`, set `self.constraints_ = constraints`, route all internal accesses through the local.
* `test/unit/reductions/exponentiated_gradient/test_exponentiatedgradient_arguments.py` — add `test_constraints_reused_across_multiple_fits` regression test; update the existing `test_input_X_unchanged` deepcopy mock to a `side_effect` that only intercepts deepcopy of the mocked estimator and lets constraints deepcopy fall through.
* `test/unit/reductions/grid_search/test_grid_search_arguments.py` — add matching `test_constraints_reused_across_multiple_fits` regression test.
* `docs/user_guide/installation_and_version_guide/v0.15.0.rst` — changelog entry.

## Tests

* New regression tests assert that the same `DemographicParity()` instance can be passed to `fit()` twice, that the original object remains unloaded (`not constraints.data_loaded`), and that `est.constraints_` is the loaded copy.
* Full `test/unit/reductions/` suite passes locally (3807 passed, 320 skipped, 888 deselected).


fairlearn/reductions/_grid_search/grid_search.py
```diff
@@ -204,11 +208,14 @@ def predict_fct(X):
...
             self.predictors_.append(current_estimator)
-            self.lambda_vecs_[i] = lambda_vec
+            lambda_vecs_dict[i] = lambda_vec
             self.objectives_.append(objective.gamma(predict_fct).iloc[0])
-            self.gammas_[i] = self.constraints.gamma(predict_fct)
+            gammas_dict[i] = constraints.gamma(predict_fct)
             self.oracle_execution_times_.append(oracle_call_execution_time)
 
+        self.lambda_vecs_ = pd.DataFrame(lambda_vecs_dict, dtype=np.float64)
+        self.gammas_ = pd.DataFrame(gammas_dict)
```
immu4989: Missing dtype=np.float64 for parity with lambda_vecs_ two lines up.

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
```diff
@@ -170,8 +170,8 @@ def fit(self, X, y, **kwargs):
 
             # set lambdas for every constraint
             lambda_vec = B * np.exp(theta) / (1 + np.exp(theta).sum())
-            self.lambda_vecs_EG_[t] = lambda_vec
-            lambda_EG = self.lambda_vecs_EG_.mean(axis=1)
+            lambda_vecs_EG_dict[t] = lambda_vec
+            lambda_EG = pd.DataFrame(lambda_vecs_EG_dict).mean(axis=1)
```
immu4989: This conflicts with the lambda_cumsum running-sum on main. See main comment §1.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -78,7 +79,10 @@ def __init__(
         sample_weight_name: str = "sample_weight",
         **kwargs,
     ):
-        self.constraints = constraints
+        # Deepcopy constraints so the user's original object is not mutated,
+        # allowing the same constraints instance to be reused across multiple
+        # calls to fit() or shared between different mitigators.
+        self.constraints = copy.deepcopy(constraints)
```
immu4989, line 87: Have you considered shallow copy.copy? deepcopy on a populated constraint can be expensive , see main comment §5.

immu4989 (changes requested):
Thanks @BALOGUN-DAVID : this is a real bug and the deepcopy approach is the right shape of fix. A few substantive concerns and one merge-conflict heads-up before I'd be comfortable approving.

1. Stale base : conflicts with lambda_cumsum optimization on main (blocking)
The diff against _lagrangian.py / exponentiated_gradient.py is based on a pre-#1170 main. Main now uses an incremental running sum:


lambda_cumsum = pd.Series(0.0, dtype="float64", index=lagrangian.constraints.index)
for t in range(0, self.max_iter):
    ...
    lambda_cumsum += lambda_vec
    lambda_EG = lambda_cumsum / (t + 1)
The PR replaces this with lambda_vecs_EG_dict[t] = lambda_vec; lambda_EG = pd.DataFrame(lambda_vecs_EG_dict).mean(axis=1), which rebuilds the full DataFrame on every iteration — O(T²) work instead of O(T). When you rebase, the right resolution is to keep main's running-sum approach for lambda_EG and only use the lambda_vecs_EG_dict for the post-loop self.lambda_vecs_EG_ = pd.DataFrame(lambda_vecs_EG_dict) assignment (which is the actual fragmentation fix). Same pattern likely applies to lambda_vecs_LP_dict.

2. Missing dedicated regression test for #1210 (blocking)
The PR closes #1210 ("data can be loaded only once") but there's no new test that exercises the regression case directly. A reviewer / future maintainer would have to read the full diff to confirm the fix holds. Suggested addition under test/unit/reductions/exponentiated_gradient/:


def test_constraints_reused_across_multiple_fits():
    """Same constraints instance must be reusable for multiple fit() calls (#1210)."""
    X, y, A = _make_data()  # any deterministic synthetic data
    constraints = DemographicParity()
    eg = ExponentiatedGradient(LogisticRegression(random_state=42), constraints, max_iter=5)
    eg.fit(X, y, sensitive_features=A)
    eg.fit(X, y, sensitive_features=A)  # would raise on main pre-fix
    # additionally: user's constraints object must remain unmodified
    assert not constraints.data_loaded
Same shape needed for GridSearch.fit.

3. Behavior change in eg.constraints / gs.constraints after fit (worth a note in CHANGELOG, possibly a deprecation)
Pre-fix, self.constraints on the fitted estimator pointed to a loaded moment that users could introspect (eg.constraints.gamma(predictor), eg.constraints._y_as_series, etc.). Post-fix, self.constraints is the user's untouched instance and the loaded copy lives only inside lagrangian (or as a local constraints in GridSearch.fit).

The internal call sites are correctly updated in this PR (e.g., lagrangian.constraints._y_as_series at line 184, lagrangian.constraints.bound() at line 249), but external users who relied on eg.constraints being loaded post-fit will see a silent behavior change. Two options:

(preferred) also store the loaded copy on the estimator, e.g. self.constraints_ = lagrangian.constraints (note the trailing underscore — sklearn convention for fitted attributes), so users have a documented way to introspect the post-fit constraint state.
At minimum, document the change in the v0.14.0 release notes.
4. dtype inconsistency in GridSearch.gammas_ (non-blocking)
Line 215 (new): self.gammas_ = pd.DataFrame(gammas_dict) — no dtype. Compare with line 214: self.lambda_vecs_ = pd.DataFrame(lambda_vecs_dict, dtype=np.float64) which is explicit. Pre-fix code had self.gammas_ = pd.DataFrame(dtype=np.float64). Suggest preserving the explicit dtype for consistency with the lambda_vecs construction and with the pre-fix behavior.

5. deepcopy vs shallow copy (clarification, not blocking)
Have you measured the cost of deepcopy on a populated constraint? Constraints can hold pandas DataFrames (tags, prob_group_event, pos_basis, neg_basis, U) that are O(n_samples) — for large X, a full deepcopy on every fit() call is non-trivial memory + CPU. If the only mutation that needs guarding is data_loaded flag flip + setting attributes that are themselves replaced (not mutated in place) on each load, a shallow copy.copy may suffice. Worth a comment in the code documenting why deepcopy is needed if it's strictly required.

6. Test refactor (nit, non blocking)
The new mock_deepcopy side effect at test_input_X_unchanged relies on obj is estimator identity. Works for the current test but couples the test tightly to the assumption that estimator is the only object whose deepcopy needs special handling. If a future change adds another deepcopy site, this mock could silently mishandle it. A short comment in the test explaining the intent ("only intercept deepcopy of the estimator; let constraint deepcopy fall through") would help future maintainers.

Smaller positives worth calling out
The dict accumulation pattern for _pmf_predict.pred, lambda_vecs_*, and gammas_dict is the correct fix for the pandas PerformanceWarning , clean idiom.
Defensive copy self.lambda_vecs_ = lagrangian.lambdas.copy() is a nice touch for caller safety.
Comment block in _lagrangian.py:80-82 explaining why the deepcopy exists is exactly the kind of context that helps future readers , please retain it after rebase.
Happy to re review once the rebase is in. If you'd like, I can sketch the rebase resolution for the lambda_cumsum conflict in a follow up comment.

fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
BALOGUN-DAVID: Kept deepcopy intentionally to isolate custom moments with mutable constructor state; added/retained code comment explaining why. Focused regression tests pass.

fairlearn/reductions/_exponentiated_gradient/exponentiated_gradient.py
BALOGUN-DAVID: Resolved during rebase by keeping main’s lambda_cumsum O(T) running sum while retaining dict accumulation only for final lambda_vecs_EG_ construction

fairlearn/reductions/_grid_search/grid_search.py
BALOGUN-DAVID: Fixed by restoring dtype=np.float64

BALOGUN-DAVID:
Thanks for the detailed review.

I rebased/merged against current `main` and resolved the conflicts by keeping the `lambda_cumsum` running-sum optimization. The EG loop now stores `lambda_vec` values in `lambda_vecs_EG_dict` for the final `lambda_vecs_EG_` DataFrame, but computes `lambda_EG` incrementally with `lambda_cumsum / (t + 1)`.

I also restored `dtype=np.float64` for `GridSearch.gammas_`, added focused regression coverage for reusing the same constraints instance across multiple `fit()` calls in both `ExponentiatedGradient` and `GridSearch`, and exposed the loaded fitted constraints as `constraints_` while leaving the user-provided `constraints` object unloaded.

On `deepcopy` vs `copy.copy`: I kept `deepcopy` intentionally for stronger isolation, especially for custom moments with mutable constructor state. A shallow copy may work for many built-in moments today, but it would still share nested mutable attributes. I added/retained code comments to make that tradeoff explicit.

Focused tests pass locally:
`python -m pytest test\unit\reductions\exponentiated_gradient\test_exponentiatedgradient_arguments.py::test_constraints_reused_across_multiple_fits test\unit\reductions\grid_search\test_grid_search_arguments.py::test_constraints_reused_across_multiple_fits`

romanlutz (approved):

PR #1634, opened 2026-03-26
MNT: Remove redundant type_of_target call in adversarial preprocessor

Vaibhav-sa30:
Fixes the `# FIXME` in `fairlearn/adversarial/_preprocessor.py` about `type_of_target` being expensive and called repeatedly.

## Description
Removes the redundant `type_of_target(X)` call inside `FloatTransformer.transform`. The call existed only as a defensive check that the input distribution at `transform` time matched the one inferred at `fit` time, but downstream `OneHotEncoder` / `check_array` already raise on shape and category mismatches with clearer errors.

## Problem
Profiling on large datasets (100M+ rows) showed `type_of_target` taking ~0.5s per call. In a typical fit + predict loop the same array was hitting `type_of_target` multiple times (once in `__setup`, once in `FloatTransformer.fit`, and again in every `FloatTransformer.transform` — and `transform` is called from `_validate_input`, which runs on every `fit`, `partial_fit`, and `predict`).

## Approaches considered

| Approach | Outcome |
| :--- | :--- |
| `id(X)` cache | Fast but unsafe — Python can reuse ids after GC |
| `X.tobytes()` hash | Correct but slower than the call it was meant to avoid (~1.1s on 100M rows) |
| `(X.ctypes.data, shape, dtype, strides)` cache | Fast but unsafe: raw memory addresses can be reused, the underlying arrays are mutable (in-place mutation produces a silently stale result), and the key crashes on non-numpy inputs like `list` and `pandas.Series` (regressed 17 existing tests) |
| **Remove the redundant call (this PR)** | **Eliminates the hotspot with no cache, no invalidation hazards, and no API change** |

Per @taharallouche's review feedback, the cache-based approaches all had soundness issues that outweighed their performance benefit. The `transform`-time check was the only call site that actually re-ran `type_of_target` on the same data, so removing it directly is both safer and equivalent in net effect to a correct cache.

## Tests
- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

Full adversarial test suite passes (287 passed, 5 skipped, 6 xfailed, 6 xpassed) locally.

## Documentation


fairlearn/adversarial/_preprocessor.py
```diff
@@ -7,8 +7,30 @@
...
-# FIXME: memoize type_of_target. It is quite expensive and called repeatedly.
+
+# FIXED: memoize type_of_target. It is quite expensive and called repeatedly.
+
+_TYPE_OF_TARGET_CACHE = OrderedDict()
+_MAX_CACHE_SIZE = 10 # We limit the cache to the 10 most recent arrays to avoid memory leaks
+
+def _memoized_type_of_target(X):
+    # Robust and Fast Key
+    key = (X.ctypes.data, X.shape, X.dtype, X.strides)
```
taharallouche, line 20: I think relying on `X.ctypes.data` is inherently unsafe for the same reason as relying on `id(X)`, because it's the raw memory address, which can be freed by the garbage collector and used to store other data.

Also the input arrays we are dealing with are mutable, so I don't think it's safe yo use this as a cache key.

taharallouche (commented):
Thanks for thinking this through @Vaibhav-sa30 but I think there are some issues with the current approach.

I'm personally more inclined towards simply removing the `transform`'s call to `type_of_target` if it proves to be significantly slowing down the pipeline. I don't know what other maintainers think

romanlutz:
Thanks @taharallouche — I dug into this and fully agree with your direction. A few empirical notes that back up your concern:

- The `X.ctypes.data` cache key crashes on any non-numpy input (`list`, `pandas.Series`), so the previous version regressed 17 existing tests in `test/unit/adversarial/` (270 pass / 17 fail vs. 287 pass / 0 fail on `main`).
- The mutability concern you raised reproduces cleanly: mutating a cached array in place via `arr[:] = ...` (binary -> multiclass) returns the stale `""binary""` from the cache while `type_of_target` would return `""multiclass""` — a silent correctness bug.
- Memory-address reuse after GC is reachable too, as you noted.

I went with your ""just remove the `transform` call"" approach over any form of caching (per-instance `id(X)` keying has the same GC/mutability hazards). The downstream `OneHotEncoder` / `check_array` already raise on shape and category mismatches with clearer errors, and no existing test depends on the `""Unknown label type""` path, so removing the check is safe. I took the liberty of pushing the change directly to the branch (3de62ac) since `maintainerCanModify` is on — @Vaibhav-sa30, hope that's OK, please yell if you'd rather rework it yourself.

Tests are back to 287 passed / 0 failed locally.

romanlutz (approved):

taharallouche (approved):

Issue #1190, opened 2023-01-20
DOC simplify plot_grid_search_census example


romanlutz:
#### Describe the issue linked to the documentation

The example at https://fairlearn.org/main/auto_examples/plot_grid_search_census.html is outright painful to read at the end, starting after
```
predictors = sweep.predictors_
```
and ending with the plot. We now have `plot_model_comparison` that can take care of all that. The filtering for non-dominated may require a little extra code, but we shouldn't have to mess around with moments like
```
error = ErrorRate()
error.load_data(X_train), pd.Series(Y_train), sensitive_features=A_train)
disparity = DemographicParity()
disparity.load_data(X_train), pd.Series(Y_train), sensitive_features=A_train)
errors.append(error.gamma(classifier)[0])
disparities.append(disparity.gamma(classifier).max()
```

#### Suggest a potential alternative/fix

Use 
```
accuracy_score
```
and
```
make_derived_metric(metric=selection_rate, transform="difference")
```
to get the error and selection rate differences (called disparities above). Use `plot_model_comparison` rather than creating the plot from scratch.

MiroDudik:
don't use `make_derived_metric`, but instead use `demographic_parity_difference`.

romanlutz:
Sure. They are equivalent, though, right? I'm guessing you're pointing it out to make people aware of the shorthand?

MiroDudik:
i think it is simpler to use existing metrics when they are available. it's also more concise--more obvious what's going on.

CheerathAniketh:
Working on this in #1635.

TemidayoA:
Picked this up in #1682 — it applies the simplification from the earlier attempt (#1635) plus the review comment that was left unaddressed there (moving plot_model_comparison up so the Pareto-front discussion follows the plot). Verified the example runs end-to-end. Happy to be assigned.

PR #1635, opened 2026-04-05
DOC simplify plot_grid_search_census example

CheerathAniketh:
Simplifies the grid search census example as suggested in #1190.

Replaces the verbose manual error/disparity loop (using `ErrorRate` and 
`DemographicParity.load_data`) with `accuracy_score` and 
`demographic_parity_difference`. Removes the manual matplotlib scatter 
plot since `plot_model_comparison` already handles visualization.

@romanlutz

### DIFF
--- a/examples/plot_grid_search_census.py
+++ b/examples/plot_grid_search_census.py
@@ -44,13 +44,6 @@ from fairlearn.datasets import fetch_adult
 from fairlearn.metrics import (
     MetricFrame,
     count,
-    plot_model_comparison,
-    selection_rate,
-    selection_rate_difference,
-)
-from fairlearn.metrics import (
-    MetricFrame,
-    count,
     demographic_parity_difference,
     plot_model_comparison,
     selection_rate,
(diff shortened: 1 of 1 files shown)


examples/plot_grid_search_census.py
```diff
@@ -225,15 +219,6 @@ def classifier(X):
...
 
-x = [metric_frame.overall["accuracy"] for metric_frame in metric_frames.values()]
-y = [metric_frame.difference()["selection_rate"] for metric_frame in metric_frames.values()]
-keys = list(metric_frames.keys())
-plt.scatter(x, y)
-for i in range(len(x)):
-    plt.annotate(keys[i], (x[i] + 0.0003, y[i]))
-plt.xlabel("accuracy")
-plt.ylabel("selection rate difference")
-
```
taharallouche, line 236: I agree it's a good idea to remove this ad-hoc plotting section.

However, I think we should move the last section "Comparing models easily" here  before "We see a Pareto front forming".

So we just need to add a sentence here "Fairlearn provides functionality to compare models easily" with the associated code cell using `plot_model_comparison`, before interpreting the plot ("We see a  Pareto front forming ...")

taharallouche (commented):
Thanks a lot @CheerathAniketh I just have one small remark, otherwise this is a great simplification of the doc and is good to be merge !

PR #1641, opened 2026-04-27
docs(examples): add encrypted-metrics example using fairlearn-fhe

BAder82t:
Adds one sphinx-gallery example at `examples/plot_encrypted_metrics.py` showing how to compute the canonical Fairlearn fairness metrics on encrypted predictions via the third-party [`fairlearn-fhe`](https://pypi.org/project/fairlearn-fhe/) package (Apache-2.0; depends on `fairlearn>=0.10`).

## What it does

1. Trains a logistic-regression baseline on `fetch_adult`.
2. Computes `demographic_parity_difference`, `equalized_odds_difference`, and `equal_opportunity_difference` in plaintext (Fairlearn).
3. Computes the same three metrics on `encrypt(ctx, y_pred)` via `fairlearn_fhe.metrics`.
4. Asserts the two agree within `1e-3`.

## Verified locally

Runs end-to-end via `python examples/plot_encrypted_metrics.py`. Plaintext vs encrypted verdicts agree at `~1e-7` abs error.

## Optional dependency

If `fairlearn-fhe` is not installed, the script prints a skip message and exits `0` so sphinx-gallery builds stay green without adding it to dev-requirements.

## Notes

- New file only.
- The script lives at the top level of `examples/` to match the existing `plot_*.py` gallery layout (the contributor guide also mentions `examples/notebooks/`, but the actual gallery in this repo is flat).

### DIFF
--- a/examples/plot_encrypted_metrics.py
+++ b/examples/plot_encrypted_metrics.py
@@ -26,22 +26,22 @@ Computing Fairlearn metrics on encrypted predictions
 #
 # This example requires ``fairlearn-fhe`` (which transitively pulls in
 # ``tenseal`` for the default CKKS backend). If you have not installed
-# it, the example exits early with a helpful message rather than
-# erroring out the whole gallery build.
+# it, the remaining cells are skipped (guarded by ``HAS_FHE``) so the
+# gallery build still completes cleanly.
 
 try:
     from fairlearn_fhe import build_context, encrypt
-    from fairlearn_fhe.metrics import (
-        demographic_parity_difference as enc_dp,
-        equal_opportunity_difference as enc_eopp,
-        equalized_odds_difference as enc_eo,
-    )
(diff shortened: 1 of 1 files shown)


TamaraAtanasoska:
Hi @BAder82t! I don't think that the examples are a good place to feature functionalities that are part of a third-party package, although it is related. It adds maintenance load and keeping up with another library we officially do not integrate anywhere in the project. Has this maybe been discussed in some meeting I have missed @fairlearn/fairlearn-maintainers?

riedgar-ms:
I don't recall this coming up in any discussion (although my recent attendance at the community calls hasn't been great). This sounds like it should be an example for the `fairlearn-fhe` package?

BAder82t:
Thanks @riedgar-ms, that's fair. I'll port it to [fairlearn-fhe/examples/](https://github.com/BAder82t/fairlearn-fhe/tree/main/examples) and close this PR
Closing this; thanks for the look.

PR #1667, opened 2026-06-16
Fix API docs parameter rendering

romanlutz:
## Description
Raw `:code:` markup was leaking into rendered MetricFrame API docs because several numpydoc parameter declarations used `name:` instead of `name : type`. This fixes those malformed declarations in MetricFrame and related metric/example docstrings, then enables low-risk numpydoc parameter type cross-references so common Python, NumPy, and pandas parameter types link in generated API pages.

The xref configuration uses `numpydoc_xref_ignore = "all"` so unknown type words remain plain text instead of creating new unresolved references. The Python intersphinx key is renamed to `python` to match numpydoc's built-in aliases.

## Tests
- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

Built docs with `uv run python -m sphinx -b html -n -E -j auto docs docs\_build\html`; warning count stayed at baseline. Spot-checked `docs\_build\html\api_reference\generated\fairlearn.metrics.MetricFrame.html` for no raw `:code:` text and linked representative parameter types.

## Documentation

## Screenshots
N/A


TamaraAtanasoska (commented):
What happens in this case to the types of other libraries like `scikit-learn` for `narwhals`? Do I understand right that everything we want to cross-linked will need to be added here, or explicitly wrapped with `:obj:`? https://numpydoc.readthedocs.io/en/stable/install.html 

I suppose we are already mostly wrapping with the appropriate tags, but I think this needs to be added in that case to the documentation page in the how-to write docs when it is now so strictly enforced.

romanlutz:
@TamaraAtanasoska good point. I tried this for both scikit-learn and Narwhals.

For scikit-learn, we already had the intersphinx mapping, so I added targeted aliases for common parameter types we reference, including `BaseEstimator`, `TransformerMixin`, `Pipeline`, and `Bunch`.

For Narwhals, I verified that `https://narwhals-dev.github.io/narwhals/objects.inv` exists and is consumable by Sphinx, then added the intersphinx mapping plus aliases for `DataFrame`, `LazyFrame`, `Series`, `IntoDataFrame`, and `IntoDataFrameT`.

I kept `numpydoc_xref_ignore = "all"`, so unknown words and generic type descriptions like `estimator` stay plain text instead of creating unresolved cross-reference warnings. I also added a short section to the contributor documentation explaining that common Python/NumPy/pandas/scikit-learn/Narwhals types are configured, and that new third-party types should either use explicit roles or get a matching intersphinx mapping and alias.

I rebuilt the docs cleanly with `uv run python -m sphinx -b html -n -E -j auto docs docs\_build\html`; the warning count stayed at the existing baseline. I also spot-checked scikit-learn rendering in the generated HTML. The current generated pages do not expose a plain Narwhals type string that exercises the new aliases, but Sphinx consumed the Narwhals inventory without adding warnings.

docs/contributor_guide/contributing_documentation.rst
```diff
@@ -29,6 +29,26 @@ found in the files of the respective classes and methods. For example, take a lo
 of the :code:`CorrelationRemover` and the corresponding
 `generated page <file:///Users/tamara/FOSS/fairlearn/docs/_build/html/api_reference/generated/fairlearn.preprocessing.CorrelationRemover.html>`_.
```
TamaraAtanasoska:
```suggestion
`generated page <https://fairlearn.org/main/api_reference/generated/fairlearn.preprocessing.CorrelationRemover.html>`_.
```

Yikes, how did this pass review and how has it been there! Please accept it to fix it in this PR 🫨

docs/contributor_guide/contributing_documentation.rst
TamaraAtanasoska: This repeats in the documentation down. Is there any way to make Copilot sound a bit less prescriptive and more descriptive as our style in this document, and kind of merge these two paragraphs together? It might need some human shaping, I am happy to also help with specific edits.

TamaraAtanasoska (changes requested):
Code wise LGTM, but the documentation page needs a bit more work.

romanlutz:
@TamaraAtanasoska addressed both documentation comments in the latest push.

- Fixed the local `file://` generated-page link, and also updated the API reference index link above it to use `fairlearn.org/main`.
- Reworked the API docstring guidance so it is folded into the existing API reference explanation, removes the extra heading, and uses more descriptive wording around numpydoc parsing and xref aliases.

I rebuilt the docs with `uv run python -m sphinx -b html -n -E -j auto docs docs\_build\html`; it passed with the existing warnings.

docs/contributor_guide/contributing_documentation.rst
```diff
@@ -37,7 +52,8 @@ The instructions below are an addition to the instructions on the
 instructions on that page, and return to follow the further
 documentation specific guidelines.
 
-You may also need to install `pandoc <https://pandoc.org/installing.html>`_, `optipng <https://optipng.sourceforge.net>`_ and :code:`matplotlib`.
+You may also need to install `pandoc <https://pandoc.org/installing.html>`_,
+:code:`optipng`, and :code:`matplotlib`.
```
TamaraAtanasoska:
```suggestion
`optipng <https://optipng.sourceforge.net>`_, and :code:`matplotlib`.
```

TamaraAtanasoska: The link was deleted, but I think where is no need to.

TamaraAtanasoska (approved):
LGTM, except one suggestion about a link that was deleted, but it should stay. Thank you!

docs/contributor_guide/contributing_documentation.rst
romanlutz: Restored the OptiPNG link as suggested and rebuilt the docs with `uv run python -m sphinx -b html -n -E -j auto docs docs\_build\html`; the build passed with the existing warnings.

PR #1669, opened 2026-06-20
[DOC] add Examples sections to base fairness metrics

laurapiro17:
#### What does this implement/fix? Explain your changes.

The base metric functions in `fairlearn/metrics/_base_metrics.py` did not have
`Examples` docstring sections. This PR adds short, runnable `Examples` to all
seven public functions so users can see how each one is called and what it
returns directly from the API reference:

- `true_positive_rate`
- `true_negative_rate`
- `false_positive_rate`
- `false_negative_rate`
- `count`
- `mean_prediction`
- `selection_rate`

Each example uses the same small binary `y_true`/`y_pred` pair for consistency,
and follows the existing doctest style (using `.item()` for scalar results, as
in the `MetricFrame` docstring and the quickstart guide). All examples were run
and their outputs verified (doctests pass).

#### Does your contribution introduce a new dependency? If yes, which one?

No.

#### Any other comments?

Docstring-only change; no behaviour is modified and no test files are touched,
so no changelog entry is required by the changelog check. Happy to add one (or
to extend the same treatment to the higher-level fairness metrics in
`_fairness_metrics.py` in a follow-up) if reviewers prefer.

##### For all contributions
- [ ] I've added myself to the list of contributors. (Will use the @all-contributors bot after merge.)
- [x] The PR title starts with [DOC].

### DIFF
--- a/fairlearn/metrics/_base_metrics.py
+++ b/fairlearn/metrics/_base_metrics.py
@@ -108,6 +108,14 @@ def true_positive_rate(y_true, y_pred, sample_weight=None, pos_label=None) -> fl
     -------
     float
         The true positive rate for the data
+
+    Examples
+    --------
+    >>> from fairlearn.metrics import true_positive_rate
+    >>> y_true = [0, 1, 1, 0, 1, 1, 0, 1]
+    >>> y_pred = [0, 1, 0, 0, 1, 1, 1, 1]
+    >>> true_positive_rate(y_true, y_pred).item()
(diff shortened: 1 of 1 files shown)


TamaraAtanasoska:
Hi @laurapiro17, thank you for the contribution. For now, I won't merge the PR as I do not consider the examples in these particular docstrings necessary. Feel free to look through the issues and find something you would like to take on and work on from there.

PR #1671, opened 2026-07-03
MNT raise a clear error when X passed to BackendEngine is not 2D

### LINKED ISSUE #1656: MNT Adversarial BackendEngine silently assumes input X is 2D
## Context

[`fairlearn/adversarial/_backend_engine.py#L36`](https://github.com/fairlearn/fairlearn/blob/93b4d668c61270751c8dc19ba38731f2e552ba3a/fairlearn/adversarial/_backend_engine.py#L36) silently assumes `X` is two-dimensional:

```python
def __init__(self, base, X, Y, A):
    ...
    self.base = base

    n_X_features = X.shape[1]  # FIXME: what if X.ndim > 2?
    n_Y_features = base._y_transform.n_features_out_
    n_A_features = base._sf_transform.n_features_out_
```

If a caller passes a higher-dimensional `X` (e.g. an image batch shaped `(n_samples, height, width, channels)`) the assignment still succeeds but `n_X_features` is wrong, which leads to silently mis-sized predictor/adversary input layers downstream.

## Suggested action

Either:

- Add an explicit shape check up-front (raising a clear error when `X.ndim != 2`) so the failure mode is loud and early, or
- Flatten / accept higher-dimensional inputs intentionally and document the convention.

Today's behaviour is the worst of both worlds — it neither errors nor handles >2D input correctly.

Part of the broader adversarial-module cleanup tracked in #1480.

Surfaced during a repo-cleanup audit by @romanlutz.

NotArnav03:
## Description
Closes #1656.

`BackendEngine.__init__` reads `X.shape[1]` to figure out how many input features the predictor model needs, but it never checked how many dimensions `X` actually had. If someone passed in data with more than two dimensions, for example a batch of images shaped `(n_samples, height, width, channels)`, the code would still run and just take `shape[1]` as the feature count. That's the wrong number, so the predictor's input layer ends up the wrong size and the model trains on garbage without any warning.

I added a check right before that line that raises a `ValueError` with a clear message if `X.ndim != 2`. This matches how the rest of the adversarial module already documents `X` (the docstrings for `_raw_predict`/`predict` both say "two-dimensional numpy array"), and every existing test already only exercises 2D `X`, so this doesn't change behavior for any currently supported input, it just fails loudly instead of quietly doing the wrong thing.

## Tests
- [x] new tests added

Added `test_backend_engine_rejects_non_2d_X` in `test/unit/adversarial/test_adversarial_mitigation.py`, parametrized over both the torch and tensorflow fake backends, which fits a mitigator with a 3D `X` and checks that a `ValueError` mentioning "two-dimensional" is raised.

Ran the full adversarial test suite locally (`pytest test/unit/adversarial`): 295 passed.

## Documentation

Added a changelog entry under "Bug fixes" in `docs/user_guide/installation_and_version_guide/v0.15.0.rst`. I left the PR number as `TBD` in that entry since I didn't know it before opening this PR, happy to update it once the PR number is assigned, or if a maintainer prefers to fill it in during merge.

## Screenshots
Not applicable, no visualization changes.

### DIFF
--- a/docs/user_guide/installation_and_version_guide/v0.15.0.rst
+++ b/docs/user_guide/installation_and_version_guide/v0.15.0.rst
@@ -27,7 +27,7 @@ Bug fixes
   attribute :code:`constraints_`: :pr:`1632` by
   :user:`BALOGUN DAVID TAIWO <BALOGUN-DAVID>`.
 * Fixed :code:`fetch_diabetes_hospital` so that :code:`as_frame=False` returns numpy arrays instead of raising. The data is now always fetched as a pandas DataFrame internally and converted to numpy when needed. The skipped tests and the obsolete :code:`test_fetch_diabetes_hospital_as_ndarray_raises_value_error` test were removed: :pr:`1636` by :user:`Richard Ogundele <richardogundele>`.
-* :code:`BackendEngine` used to read :code:`X.shape[1]` without checking how many dimensions :code:`X` had, so passing data with more than two dimensions (for example an array of images) would silently build a predictor model with the wrong input size instead of failing. It now raises a :code:`ValueError` up front when :code:`X` is not two-dimensional: :pr:`TBD` by :user:`Arnav <NotArnav03>`.
+* :code:`BackendEngine` used to read :code:`X.shape[1]` without checking how many dimensions :code:`X` had, so passing data with more than two dimensions (for example an array of images) would silently build a predictor model with the wrong input size instead of failing. It now raises a :code:`ValueError` up front when :code:`X` is not two-dimensional: :pr:`1671` by :user:`Arnav <NotArnav03>`.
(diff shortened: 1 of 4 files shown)


NotArnav03:
While testing this I noticed the same class of bug also affects predict() and partial_fit() on an already-fitted model, since those paths don't go back through BackendEngine.__init__. Filed that separately as #1672 so it doesn't get missed, keeping this PR scoped to what #1656 asked for.

NotArnav03:
Fixed the pluralization Copilot flagged, the message now says "dimension(s)" so it reads correctly for X.ndim == 1 too.

riedgar-ms (approved):
Thanks!

romanlutz (approved):

PR #1673, opened 2026-07-06
MNT: clarify fitted-state check in adversarial mitigation

### LINKED ISSUE #1657: MNT Re-examine the check_is_fitted try/except pattern in _AdversarialFairness._validate_input
## Context

[`fairlearn/adversarial/_adversarial_mitigation.py#L690`](https://github.com/fairlearn/fairlearn/blob/93b4d668c61270751c8dc19ba38731f2e552ba3a/fairlearn/adversarial/_adversarial_mitigation.py#L690) wraps `check_is_fitted` in a `try`/`except NotFittedError` that has carried a `# TODO check this` for a long time:

```python
try:  # TODO check this
    check_is_fitted(self)
    is_fitted = True
except NotFittedError:
    is_fitted = False
```

The `try`/`except` is used here to *probe* fit state — `_validate_input` legitimately needs to know whether the estimator has been fit yet so it can decide whether to call `__setup` and re-initialize.

## Suggested action

Decide whether this probing pattern is appropriate and either remove the `# TODO` (with a brief justification) or refactor it. Options worth considering:

- Replace with a direct attribute check on the fitted attribute (e.g. `hasattr(self, "backendEngine_")`), which avoids the round-trip through an exception path.
- Keep the `try`/`except` but drop the stale `# TODO` and add a one-line explanation that we are intentionally probing the fit state.

Either resolution is small and self-contained.

Surfaced during a repo-cleanup audit by @romanlutz.

DresdenGman:
Resolves #1657 by replacing the `try/except NotFittedError` fit-state probe in `_AdversarialFairness._validate_input` with a direct `hasattr(self, 'backendEngine_')` check.

This keeps the existing behavior intact while removing the stale `# TODO check this` and avoiding the ambiguity of probing fit state through an exception path.

Tests: 288 passed (targeted adversarial tests, `python -m pytest test/unit/adversarial -q`)


riedgar-ms (approved):
Thanks!

Issue #1487, opened 2025-01-25
Plot the MetricFrame's Confidence Intervals


taharallouche:
### Description
This is a follow-up to #999.

Now that the MetricFrame class can compute the confidence intervals, we need to implement a function that would include their associated error-bars into the plot of the metrics.

### Note
I found [this plotting function](https://github.com/fairlearn/fairlearn/blob/main/fairlearn/metrics/_plotter.py#L117) that seems to handle confidence intervals, but it's not well documented (the user-guide's plotting section discusses only leveraging pandas' `.plot`).

Looking at how it is used in [these tests](https://github.com/fairlearn/fairlearn/blob/main/test/unit/metrics/test_plot_metricframe.py#L163), it doesn't seem to be compatible with the `MetricFrame`'s built-in ci feature.

DresdenGman:
Hi, is this still open for contribution? It looks like there hasn't been follow-up on the earlier discussion for a while.

If the issue is available, I'd be happy to take a stab at integrating `MetricFrame`'s built-in bootstrap confidence intervals into `plot_metric_frame`, while keeping the existing `conf_intervals` parameter as the explicit override path.

Happy to align on the preferred design direction first.

Anish8800:
Hi! Thanks for reaching out. Feel free to go ahead with it. Looking forward to seeing your implementation!

PR #1674, opened 2026-07-06
ENH auto-detect built-in CI in plot_metric_frame

DresdenGman:
Addresses #1487.

This PR lets `plot_metric_frame` use the confidence intervals already computed by `MetricFrame` when it was constructed with `n_boot` and `ci_quantiles`.

When `conf_intervals` is not provided, the plotting helper now detects the built-in `by_group_ci` results and converts them into the existing error-bar path. If users pass `conf_intervals` explicitly, that existing behavior continues to take precedence.

This keeps the change internal to the plotting bridge and does not add a new public API. I added tests for single-metric and multi-metric `MetricFrame` plots, the no-bootstrap fallback, and the explicit `conf_intervals` override path.

### DIFF
--- a/fairlearn/metrics/_plotter.py
+++ b/fairlearn/metrics/_plotter.py
@@ -114,6 +114,54 @@ def _plot_df(df, metrics, kind, subplots, legend_label, df_all_errors=None, **kw
     return axs
 
 
+def _get_conf_intervals_from_metric_frame(metric_frame):
+    """Build a ``{metric: ci_col}`` mapping from MetricFrame built-in bootstrap CI.
+
+    Returns an empty dict when no CI is available.
+    """
+    ci_quantiles = getattr(metric_frame, "ci_quantiles", None)
+    if ci_quantiles is None:
+        return {}
+
+    ci_data = getattr(metric_frame, "by_group_ci", None)
+    if ci_data is None:
+        return {}
+
+    # by_group_ci is a list indexed by ci_quantiles order
+    if len(ci_data) != len(ci_quantiles):
+        return {}
+
+    # Zip and sort by quantile value so lower < upper always
+    pairs = sorted(zip(ci_quantiles, ci_data), key=lambda x: x[0])
+
+    if len(pairs) > 2:
+        import warnings
+
+        warnings.warn(
+            f"ci_quantiles has {len(pairs)} entries; " "using the outermost pair for error bars.",
+            stacklevel=2,
+        )
+
+    lo_q, lo_data = pairs[0]
+    hi_q, hi_data = pairs[-1]
+
+    mapping = {}
+    if isinstance(lo_data, pd.DataFrame):
+        common_cols = set(lo_data.columns) & set(hi_data.columns)
+        for col in common_cols:
+            lower = lo_data[col].values
+            upper = hi_data[col].values
+            ci_col = f"__metricframe_ci_{col}"
+            mapping[col] = (ci_col, list(zip(lower, upper)))
+    else:
+        # Single metric: by_group_ci returns Series
+        col = lo_data.name or "metric"
+        ci_col = f"__metricframe_ci_{col}"
+        mapping[col] = (ci_col, list(zip(lo_data.values, hi_data.values)))
+
+    return mapping
+
+
 def plot_metric_frame(
     metric_frame: MetricFrame,
     *,
@@ -156,6 +204,10 @@ def plot_metric_frame(
     conf_intervals : list[str] | str | None
         The name of the confidence intervals to plot.
         Should match columns from the given :class:`fairlearn.metrics.MetricFrame`.
+
+        If ``None`` and the *metric_frame* was created with bootstrap
+        confidence intervals (``n_boot`` and ``ci_quantiles``), they are
+        auto-detected from :attr:`fairlearn.metrics.MetricFrame.by_group_ci`.
 
         Note:
             The return of the error function should be an array of the lower
@@ -207,6 +259,19 @@ def plot_metric_frame(
         for metric in list(df):
             if not _is_arraylike(df[metric].iloc[0]):
                 metrics.append(metric)
+
+    # When the user did not supply conf_intervals, try to auto-detect
+    # MetricFrame's built-in bootstrap confidence intervals.
+    if conf_intervals is None:
+        ci_map = _get_conf_intervals_from_metric_frame(metric_frame)
+        if ci_map:
+            conf_intervals = []
+            df = df.copy()
+            for m in metrics:
+                if m in ci_map:
+                    ci_col, ci_values = ci_map[m]
+                    df[ci_col] = [np.array(p) for p in ci_values]
+                    conf_intervals.append(ci_col)
 
     check_consistent_length(metrics, conf_intervals)
     if len(metrics) == 0:
--- a/fairlearn/metrics/_plotter.py
+++ b/fairlearn/metrics/_plotter.py
@@ -253,6 +253,11 @@ def plot_metric_frame(
 
     df = metric_frame.by_group
 
+    # Normalise to DataFrame so column iteration works for single-metric
+    # frames where by_group is a Series.
+    if isinstance(df, pd.Series):
+        df = df.to_frame()
+
     # only plot metrics that aren't arrays (filters out metric errors)
     if metrics is None:
         metrics = []
--- a/test/unit/metrics/test_plot_metricframe.py
+++ b/test/unit/metrics/test_plot_metricframe.py
@@ -207,7 +207,7 @@ def test_auto_ci_single_metric():
     for col, (ci_col, ci_values) in ci_map.items():
         assert col == "accuracy_score"
         assert ci_col == "__metricframe_ci_accuracy_score"
-        assert len(ci_values) == len(g_1.unique())
+        assert len(ci_values) == len(np.unique(g_1))
         for lower, upper in ci_values:
             assert isinstance(lower, float)
             assert isinstance(upper, float)
@@ -232,7 +232,7 @@ def test_auto_ci_multi_metric():
     # Verify the mapping has real CI values
     for col, (ci_col, ci_values) in ci_map.items():
         assert ci_col == f"__metricframe_ci_{col}"
-        assert len(ci_values) == len(g_1.unique())
+        assert len(ci_values) == len(np.unique(g_1))
 
 
 def test_no_ci_without_bootstrap():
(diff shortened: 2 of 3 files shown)


fairlearn/metrics/_plotter.py
```diff
@@ -201,6 +242,13 @@ def plot_metric_frame(
 
     df = metric_frame.by_group
 
+    # When the user did not supply conf_intervals, try to auto-detect
+    # MetricFrame's built-in bootstrap confidence intervals.
+    if conf_intervals is None:
+        df, conf_intervals = _get_conf_intervals_from_metric_frame(
+            metric_frame, df
+        )
```
romanlutz: Two correctness problems here, both stemming from the fact that the auto-detected `conf_intervals` are built in `df` column order and independently of the resolved `metrics` list:

1. **Wrong error bars when `metrics=` is reordered.** Further down we do `zip(metrics, conf_intervals)`. Since the helper emits CIs in column order, passing `metrics=["recall", "accuracy"]` on a bootstrapped multi-metric frame attaches recall's error bars to accuracy and vice-versa — silently, no error. I verified recall renders `(0.121, 0.055)` when its true errors are `(0.09, 0.128)`.

2. **Regression: a subset of `metrics` now crashes.** Auto-detect injects CIs for *all* metrics, so when the user selects a subset the lengths diverge and `check_consistent_length` raises: `plot_metric_frame(mf, metrics=["accuracy"])` → `ValueError: Found input variables with inconsistent numbers of samples: [1, 2]`. This worked before the PR.

Suggested fix: have the helper return a `{metric: ci_col}` mapping, then build `conf_intervals` *after* `metrics` is finalized (below), e.g. `conf_intervals = [ci_map[m] for m in metrics if m in ci_map]`, and skip metrics that don't have a CI rather than emitting a misaligned list.

fairlearn/metrics/_plotter.py
```diff
@@ -114,6 +114,47 @@ def _plot_df(df, metrics, kind, subplots, legend_label, df_all_errors=None, **kw
...
 
 
+def _get_conf_intervals_from_metric_frame(metric_frame, df):
+    """Construct conf_intervals from MetricFrame built-in bootstrap CI.
+
+    If the MetricFrame has ``by_group_ci`` in ``_result_cache``, inject
+    ``[lower, upper]`` columns into *df* and return corresponding names.
+
+    Returns (*df*, *conf_intervals*) where *conf_intervals* is ``None``
+    when no CI is available.
```
romanlutz: Since this adds user-visible behavior, two docs items are needed before merge:

- Update the public `plot_metric_frame` docstring to document that confidence intervals are now auto-detected from a bootstrapped `MetricFrame` when `conf_intervals` isn't passed. The current docstring still implies CIs only come from the `conf_intervals` argument.
- Add a "New features" entry to `docs/user_guide/installation_and_version_guide/v0.15.0.rst` with `:pr:`1674`` / `:user:` attribution (all other entries there follow that convention).

test/unit/metrics/test_plot_metricframe.py
```diff
@@ -188,3 +189,126 @@ def test_multi_ax_input(sample_metric_frame):
...
+        metrics=accuracy_score,
+        y_true=y_t, y_pred=y_p, sensitive_features=g_1,
+        n_boot=10, ci_quantiles=[0.025, 0.975], random_state=0,
+    )
+    df, conf_intervals = _get_conf_intervals_from_metric_frame(mf, mf.by_group)
+    assert conf_intervals is not None
+    assert len(conf_intervals) == 1
+    assert conf_intervals[0] in df.columns
+    ci_vals = df[conf_intervals[0]]
+    import numpy as np; assert all(isinstance(v, np.ndarray) and len(v) == 2 for v in ci_vals)
```
romanlutz: `ruff` flags this line: `E702 Multiple statements on one line (semicolon)`. Move the `import numpy as np` to the top of the module (it's already imported in several tests) and put the `assert` on its own line. Heads up more broadly: `black` reformats all three files in this PR and `isort` flags the import in `test_explicit_conf_intervals_not_overridden_by_auto`, so CI linting will fail until you run black/isort/ruff over the diff.

Separately, a coverage note: every end-to-end test here monkeypatches `_plot_df`, so the actual error-bar computation for the auto-CI path is never exercised. Worth one non-mocked test (or a `pytest-mpl` snapshot like the existing ones) plus cases for the subset-`metrics` and reordered-`metrics` bugs I flagged in `_plotter.py`.

fairlearn/metrics/_plotter.py
```diff
@@ -114,6 +114,47 @@ def _plot_df(df, metrics, kind, subplots, legend_label, df_all_errors=None, **kw
...
+    """
+    cache = getattr(metric_frame, "_result_cache", None)
```
romanlutz: Prefer the public API over reaching into `_result_cache`. `MetricFrame` exposes `by_group_ci` (list indexed by `ci_quantiles`) and `ci_quantiles` as public properties — using `metric_frame.ci_quantiles is not None` to gate and `metric_frame.by_group_ci` to read keeps this from breaking if the private cache layout changes.

fairlearn/metrics/_plotter.py
```diff
@@ -114,6 +114,47 @@ def _plot_df(df, metrics, kind, subplots, legend_label, df_all_errors=None, **kw
...
+    if cache is None or "by_group_ci" not in cache:
+        return df, None
+
+    ci_data = cache["by_group_ci"]
+    if not isinstance(ci_data, (list, tuple)) or len(ci_data) != 2:
+        return df, None
+
+    ci_lower, ci_upper = ci_data
```
romanlutz: Two edge cases in how the quantiles are interpreted:

- **`len(ci_data) != 2` silently drops the CI.** With `ci_quantiles=[0.025, 0.5, 0.975]` the user explicitly asked for bootstrapping but gets no error bars and no warning. We should handle >2 quantiles (e.g. pick the outer pair) or at least warn.
- **`ci_lower, ci_upper = ci_data` assumes the quantiles are sorted ascending.** `by_group_ci` is indexed by `ci_quantiles` in the order given, so `ci_quantiles=[0.975, 0.025]` makes lower/upper swap and later trips `_CONF_INTERVALS_FLIPPED_BOUNDS_ERROR` with a confusing message. Sorting by quantile value before splitting would make this robust.

fairlearn/adversarial/_adversarial_mitigation.py
```diff
@@ -683,11 +683,9 @@ def _validate_input(self, X, y, A, reinitialize=False):
...
                 )
 
-        try:  # TODO check this
-            check_is_fitted(self)
-            is_fitted = True
-        except NotFittedError:
-            is_fitted = False
+        # Probe whether the estimator has been fitted so _validate_input
+        # can decide whether to initialize the backend engine.
+        is_fitted = hasattr(self, 'backendEngine_')
```
romanlutz: This adversarial-mitigation change is unrelated to the plotting feature and would be better as its own PR (the removed `NotFittedError` import is correct, though — it's now unused).

On the change itself: it's functionally OK since `backendEngine_` and `_is_setup` are both set at the end of `__setup`. But the class's canonical fitted indicator is `__sklearn_is_fitted__`, which returns `hasattr(self, "_is_setup")` — that's what `predict()`'s `check_is_fitted` relies on. Probing `_is_setup` here instead of `backendEngine_` would keep the fitted-state definition in one place.

romanlutz (changes requested):
Thanks a lot for the contribution! I left a few comments. Let me know if you want me to help in addressing them.

DresdenGman:
Thanks for the thorough review — really helpful.

I've reworked the helper to use the public `by_group_ci` and `ci_quantiles` instead of reaching into `_result_cache`, and it now returns a `{metric: ci_col}` mapping so the auto-detect can be wired up after metrics are resolved. That should fix both the reorder bug and the subset crash.

Also handled the quantile edge cases — >2 quantiles now warns and picks the outermost pair, and reversed order is sorted before splitting. Added tests for all of those. Docstring and changelog are updated too. And I pulled out the adversarial mitigation change since it's already #1673.

fairlearn/metrics/_plotter.py
romanlutz: Verified the reorder and subset fixes — `metrics=["recall", "accuracy"]` now maps each metric to its own CI (recall renders `(0.09, 0.128)` as expected), and `metrics=["accuracy"]` on a multi-metric frame no longer raises. Thanks!

Moving the auto-detect block below the metric-inference loop introduced a new regression for single-metric frames, though. There `by_group` is a `Series`, so `list(df)` iterates over the metric *values* (e.g. `0.611...`) rather than names, and the later `df[metric]` lookup raises `KeyError: 0.6111111111111112`. This breaks all single-metric plotting (bootstrap or not), and `test_plot_metric_frame_auto_ci_single` currently fails because of it. The pre-refactor code sidestepped this by converting the Series to a frame before inferring metrics — normalizing `df` up front (e.g. `df = df.to_frame() if isinstance(df, pd.Series) else df`) before the loop should fix it. Leaving this thread open to track.

test/unit/metrics/test_plot_metricframe.py
romanlutz: The E702 line is gone and black/isort/ruff are all clean now, and the reordered/subset coverage was added — thanks. Two of the new tests fail as written, though: `test_auto_ci_single_metric` and `test_auto_ci_multi_metric` call `g_1.unique()`, but `g_1` is a NumPy array with no `.unique()` method → `AttributeError`. Use `len(np.unique(g_1))` (or `len(set(g_1))`). Separately, `test_plot_metric_frame_auto_ci_single` fails due to the single-metric regression noted on the `_plotter.py` thread. Leaving this open until the suite is green.

DresdenGman:
Good catch on both — I've added the `to_frame()` normalisation before the metrics loop so single-metric Series works again, and switched `g_1.unique()` to `np.unique(g_1)` in the tests.

fairlearn/metrics/_plotter.py
romanlutz: Confirmed fixed in `0870cc3` — normalizing `by_group` to a DataFrame (`df.to_frame()`) before metric inference resolves the single-metric `KeyError`. I re-ran the suite (17 passed) and verified single-metric real renders for normal quantiles, reversed quantiles, and the no-bootstrap path, plus that multi-metric reorder alignment is still correct. Resolving.

test/unit/metrics/test_plot_metricframe.py
romanlutz: Fixed in `0870cc3` — both tests now use `np.unique(g_1)` and pass, and the single-metric regression is resolved on the `_plotter.py` thread. Full suite is green (17 passed) and black/isort/ruff are clean. Resolving.

romanlutz (approved):

Issue #1658, opened 2026-06-12
ENH ThresholdOptimizer drops multi-column additional_group_data / additional_data


romanlutz:
## Context

`ThresholdOptimizer`'s data-marshalling helpers in [`fairlearn/postprocessing/_threshold_optimizer.py`](https://github.com/fairlearn/fairlearn/blob/93b4d668c61270751c8dc19ba38731f2e552ba3a/fairlearn/postprocessing/_threshold_optimizer.py) can only carry a *single* column of `additional_group_data` / `additional_data` through to the plotting and prediction paths. Multi-column inputs are either silently collapsed or rejected with a generic error. Four TODOs across the file flag the same gap:

- [L685-L687](https://github.com/fairlearn/fairlearn/blob/93b4d668c61270751c8dc19ba38731f2e552ba3a/fairlearn/postprocessing/_threshold_optimizer.py#L685-L687) — `_extract_group_data` only keeps `sensitive_feature_names[0]`:

  ```python
  # TODO: extend to multiple columns for additional group data
  # and name columns after original column names if possible
  # or store the original column names
  sensitive_feature_name = SENSITIVE_FEATURE_KEY
  if sensitive_feature_names is not None:
      ...
      sensitive_feature_name = sensitive_feature_names[0]
  ```

- [L723-L728](https://github.com/fairlearn/fairlearn/blob/93b4d668c61270751c8dc19ba38731f2e552ba3a/fairlearn/postprocessing/_threshold_optimizer.py#L723-L728) — `_reformat_data_into_dict` rejects multi-column `numpy.ndarray`:

  ```python
  if len(additional_data.shape) > 2 or (
      len(additional_data.shape) == 2 and additional_data.shape[1] > 1
  ):
      # TODO: extend to multiple columns for additional_group data
      raise ValueError(MULTIPLE_DATA_COLUMNS_ERROR_MESSAGE.format("sensitive_features"))
  ```

- [L731-L734](https://github.com/fairlearn/fairlearn/blob/93b4d668c61270751c8dc19ba38731f2e552ba3a/fairlearn/postprocessing/_threshold_optimizer.py#L731-L734) — multi-column `pandas.DataFrame` input is silently collapsed to the last column:

  ```python
  elif isinstance(additional_data, pd.DataFrame):
      # TODO: extend to multiple columns for additional_data by using column names
      for attribute_column in additional_data.columns:
          data_dict[key] = additional_data[attribute_column].values
  ```

[...]

PR #1679, opened 2026-07-11
ENH ThresholdOptimizer multicolumn sensitive feature data

BALOGUN-DAVID:
## Description

addresses #1658.

This updates `ThresholdOptimizer`’s internal data reformatting helpers so multi-column sensitive feature data is preserved instead of being silently collapsed or rejected. The helper now:

- retains DataFrame column names
- generates stable column names for multi-column ndarray inputs
- supports list-of-lists sensitive feature inputs
- groups on all preserved sensitive feature columns
- keeps score and label inputs single-column, since downstream postprocessing code expects internal `score` and `label` columns

This is scoped to the internal marshalling layer; the public `ThresholdOptimizer` path still has existing behavior around sensitive feature preprocessing.

## Tests

- [ ] no new tests required
- [x] new tests added
- [ ] existing tests adjusted

Ran:

```bash
python -m pytest test\unit\postprocessing -q
```

Result:

```text
1618 passed, 7 skipped, 192 deselected, 2 xfailed
```

## Documentation

Added a v0.14.0 release-note entry.


fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -714,33 +720,69 @@ def _reformat_data_into_dict(key, data_dict, additional_data):
...
     elif isinstance(additional_data, pd.DataFrame):
-        # TODO: extend to multiple columns for additional_data by using column names
-        for attribute_column in additional_data.columns:
-            data_dict[key] = additional_data[attribute_column].values
+        if allow_multiple_columns:
+            names = list(additional_data.columns) if column_names is None else column_names
+        else:
+            names = column_names
+        names = _validate_column_names(names, len(additional_data.columns))
+        return _add_columns(additional_data.to_numpy(), names)
```
romanlutz: Converting the whole DataFrame to one ndarray can coerce every sensitive-feature column to a common dtype and change group identities. For example, a frame containing a float column plus integer IDs `2**53` and `2**53 + 1` is converted to `float64`, where those IDs compare equal, so two sensitive groups are silently merged. Please extract each column independently (for example, via `additional_data.iloc[:, i].to_numpy()`) to preserve its dtype.

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -714,33 +720,69 @@ def _reformat_data_into_dict(key, data_dict, additional_data):
...
+
     elif isinstance(additional_data, pd.Series):
-        data_dict[key] = additional_data.values
+        names = _validate_column_names(column_names, 1)
+        data_dict[names[0]] = additional_data.values
+        return names
+
     elif isinstance(additional_data, list):
         if isinstance(additional_data[0], list):
```
romanlutz: This recognizes nested rows only when the first element is itself a `list`. A previously accepted input such as `[('a', 'x'), ('b', 'y')]` takes the other branch; its `(2, 2)` array is reshaped to `(4, 1)`, then fails against two labels/scores. Please normalize list and tuple rows consistently based on the resulting array's dimensionality, while explicitly rejecting ragged or higher-dimensional inputs.

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -714,33 +720,69 @@ def _reformat_data_into_dict(key, data_dict, additional_data):
...
+        if column_count == 1:
+            return [key]
+        return [f"{key}_{i}" for i in range(column_count)]
+
+    def _add_columns(values, names):
+        if len(names) > 1 and not allow_multiple_columns:
+            raise ValueError(MULTIPLE_DATA_COLUMNS_ERROR_MESSAGE.format(key))
+
+        for i, name in enumerate(names):
+            data_dict[name] = values[:, i] if len(names) > 1 else values.squeeze()
```
romanlutz: `squeeze()` removes the row dimension as well as the column dimension for singleton inputs. For example, `_reformat_and_group_data(["a", "a", "b", "b"], [0, 1, 0, 1], [0.5])` now stores `0.5` as a scalar, so pandas broadcasts it to all four rows instead of raising for inconsistent lengths. This can silently accept a malformed score output and train against repeated scores. Please always extract the column as `values[:, i]` (including the single-column case) so pandas retains the length check.

romanlutz (changes requested):

BALOGUN-DAVID:
Thanks for the careful review, @romanlutz. I’ve addressed the three issues you flagged
I also added regression tests for these cases.
Validated with the focused data-reformat tests and the nearby ThresholdOptimizer suites: `1535 passed, 192 deselected`.

fairlearn/postprocessing/_threshold_optimizer.py
```diff
@@ -714,33 +720,76 @@ def _reformat_data_into_dict(key, data_dict, additional_data):
...
+        if column_count == 1:
+            return [key]
+        return [f"{key}_{i}" for i in range(column_count)]
+
+    def _add_columns(columns, names):
+        if len(names) > 1 and not allow_multiple_columns:
+            raise ValueError(MULTIPLE_DATA_COLUMNS_ERROR_MESSAGE.format(key))
+
+        for name, values in zip(names, columns):
+            data_dict[name] = values
```
romanlutz, line 769: Could we guard against duplicate sensitive-feature column names before writing them into `data_dict`? Since the column name is the dictionary key here, an earlier column is silently overwritten. For example, `pd.DataFrame([["a", "x"], ["b", "x"]], columns=["group", "group"])` produces a single `("x", "x")` group and discards the first column, merging two distinct sensitive groups. Please either reject non-unique names or assign collision-free internal keys.

test/unit/postprocessing/test_threshold_optimizer_data_reformat.py
```diff
@@ -0,0 +1,132 @@
...
+            scores=[0.5],
+        )
+
+
+@pytest.mark.parametrize(
+    "sensitive_features",
+    [
+        [("a", "x"), ("b",)],
+        [[["a"]], [["b"]]],
+    ],
```
romanlutz, line 244: Please include a 3-D `np.ndarray` case here as well. ndarray dimensionality is handled by a separate branch from nested lists, and the current malformed cases exercise only the list branch.

test/unit/postprocessing/test_threshold_optimizer_data_reformat.py
```diff
@@ -0,0 +1,132 @@
...
+    sensitive_features = pd.DataFrame({"SF1": ["a", "a", "b", "b"], "SF2": ["x", "y", "x", "y"]})
+
+    grouped_data = _reformat_and_group_data(
+        sensitive_features=sensitive_features,
+        labels=[0, 1, 0, 1],
+        scores=[0.1, 0.2, 0.3, 0.4],
+    )
+
+    assert list(grouped_data.obj.columns) == ["SF1", "SF2", SCORE_KEY, LABEL_KEY]
+    assert sorted(grouped_data.groups.keys()) == [("a", "x"), ("a", "y"), ("b", "x"), ("b", "y")]
```
romanlutz, line 22: Could this test also assert that each group retains the expected score and label values? Since these inputs contain the full Cartesian product of group keys, checking only the key set can still pass if a sensitive-feature column or the score/label rows become misaligned. An assertion against `grouped_data.get_group(...)` would protect the marshalling contract.

test/unit/postprocessing/test_threshold_optimizer_data_reformat.py
```diff
@@ -0,0 +1,132 @@
...
+):
+    with pytest.raises(ValueError, match="single column in sensitive_feature"):
+        _reformat_and_group_data(
+            sensitive_features=sensitive_features,
+            labels=[0, 1],
+            scores=[0.1, 0.2],
+        )
+
+
+def test_reformat_and_group_data_rejects_reserved_sensitive_feature_names():
```
romanlutz, line 301: Could we extend the column-name validation coverage with mismatched name counts and, alongside the duplicate-name fix, both duplicate DataFrame labels and duplicate explicit `sensitive_feature_names`? These are distinct inputs to `_validate_column_names`, and either duplicate source currently risks dictionary-key overwrite.

test/unit/postprocessing/test_threshold_optimizer_data_reformat.py
```diff
@@ -0,0 +1,132 @@
...
+    )
+
+    assert list(grouped_data.obj.columns) == ["first", "second", SCORE_KEY, LABEL_KEY]
+    assert sorted(grouped_data.groups.keys()) == [("a", "x"), ("b", "y")]
+
+
+def test_reformat_and_group_data_rejects_multi_column_scores():
+    scores = pd.DataFrame({"score_1": [0.1, 0.2], "score_2": [0.3, 0.4]})
+
+    with pytest.raises(ValueError, match="single column in score"):
```
romanlutz: Could we parameterize this over both `scores` and `labels`? The helper's contract requires both auxiliary inputs to remain single-column, but this currently protects only the score path, so multi-column labels could regress unnoticed.

romanlutz (commented):
Thanks @BALOGUN-DAVID ! A couple more things then this looks ready to merge.

BALOGUN-DAVID:
Hi @romanlutz, thanks for the review 🤙🏾 I have fixed the changelog entry to include `:pr:`1679`` and Addressed the latest review.

romanlutz (approved):

PR #1683, opened 2026-07-24
DOC: document false positive rate parity and error rate parity

### LINKED ISSUE #1614: DOC: Expand and clarify documentation for fairness moments

Many fairness metrics (“moments”) are currently documented in the user guide, but only a few are covered and some explanations could be improved. 

This makes it harder for users to understand what each metric measures and whether it is used as a fairness criterion or objective.

Goal:
Document all available fairness moments in Fairlearn, clarify their meaning, and provide examples.

Proposed steps:
1. List all available fairness moments in Fairlearn.
2. For each metric, provide:
   - Definition (simple explanation)
   - Use case / scenario
   - Whether it’s used as a fairness criterion or objective
3. Ensure consistency and clarity across the user guide.
4. Optionally, add links to relevant papers or resources.

References:
- Related discussion: #1191
- Barocas, Hardt, Narayanan — Fairness and Machine Learning
- Jacobs & Wallach — Measurement and Fairness

SpiliosDimakopoulos:
Adds two missing fairness moments to the common fairness metrics guide: false positive rate parity and error rate parity, following the same structure (definition, use case, code example) as the existing sections. Updates the summary table accordingly.

Closes #1614

## Tests

- [ ] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

## Documentation

### DIFF
--- a/docs/user_guide/assessment/common_fairness_metrics.rst
+++ b/docs/user_guide/assessment/common_fairness_metrics.rst
@@ -304,7 +304,114 @@ However, by not considering whether false
 positive rates are equivalent across groups, equal opportunity does not
 capture the costs of missclassification disparities.
 
+.. _assessment_false_positive_rate_parity:
 
+False positive rate parity
+---------------------------
+
+False positive rate parity is the mirror image of equal opportunity:
+instead of requiring equal *true* positive rates across groups, it
+requires equal *false* positive rates. Another way of thinking about
+this metric is requiring equal treatment only within the subset of
+records belonging to the negative class. In a pre-trial risk
+assessment scenario, false positive rate parity requires that
+defendants in *group A* who would not have reoffended are just as
+likely to be incorrectly flagged as high-risk as defendants in
+*group B* who would not have reoffended.
+
+We mathematically define false positive rate parity using the
+following equation. A classifier :math:`h` satisfies false positive
+rate parity under a distribution over :math:`(X, A, Y)` if
+:math:`\E[h(X) \given A=a, Y=0] = \E[h(X) \given Y=0] \quad \forall a`.
+:footcite:cts:`hardt2016equality` Equivalently, the probability of a
+false positive prediction should not depend on group membership,
+among individuals who belong to the negative class.
+
+Just as equal opportunity does not capture the costs of
+misclassification disparities among the positive class, false
+positive rate parity does not capture disparities among the positive
+class. It should generally be considered alongside true positive
+rate parity (equal opportunity) rather than as a standalone
+criterion — used together, the two recover :ref:`equalized odds
+<assessment_equalized_odds>`. False positive rate parity on its own
+is most relevant when the *cost of a false positive* is the primary
+harm of concern, such as flagging someone for additional scrutiny,
+denying a benefit that would otherwise have been granted, or
+producing a false accusation.
+
+Fairlearn does not currently expose a dedicated
+:code:`false_positive_rate_parity_difference` convenience function,
+but the same result can be obtained directly from
+:class:`~fairlearn.metrics.MetricFrame` using
+:func:`~fairlearn.metrics.false_positive_rate` as the metric:
+
+.. doctest:: common_fairness_metrics_code
+    :options:  +NORMALIZE_WHITESPACE
+
+    >>> from fairlearn.metrics import MetricFrame, false_positive_rate
+    >>> mf = MetricFrame(metrics=false_positive_rate,
+    ...                  y_true=y_true,
+    ...                  y_pred=y_pred,
+    ...                  sensitive_features=sf_data)
+    >>> print(mf.difference())
+    1.0
+    >>> print(mf.ratio())
+    0.0
+
+As with equalized odds, a difference of 0 (or ratio of 1) indicates
+that false positive rate parity has been achieved.
+
+.. _assessment_error_rate_parity:
+
+Error rate parity
+------------------
+
+Error rate parity requires that a classifier's overall error rate —
+the proportion of predictions that are wrong, regardless of whether
+they are false positives or false negatives — be the same across
+groups. Unlike demographic parity, equalized odds, or equal
+opportunity, error rate parity does not distinguish between the
+*type* of mistake being made. It only asks whether one group is, on
+the whole, being predicted for less accurately than another.
+
+We mathematically define error rate parity using the following
+equation. A classifier :math:`h` satisfies error rate parity under a
+distribution over :math:`(X, A, Y)` if
+:math:`\E[\mathbb{1}[h(X) \neq Y] \given A=a] = \E[\mathbb{1}[h(X) \neq Y]] \quad \forall a`.
+:footcite:cts:`agarwal2018reductions`
+
+Error rate parity can surface quality-of-service harms: if a model
+is consistently less accurate for one group than another, that group
+is receiving a worse-performing system even when demographic parity
+or equalized odds happen to hold. However, because it collapses false
+positives and false negatives into a single number, error rate parity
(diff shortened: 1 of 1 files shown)


docs/user_guide/assessment/common_fairness_metrics.rst
```diff
@@ -304,7 +304,114 @@ However, by not considering whether false
...
+------------------
+
+Error rate parity requires that a classifier's overall error rate —
+the proportion of predictions that are wrong, regardless of whether
+they are false positives or false negatives — be the same across
+groups. Unlike demographic parity, equalized odds, or equal
+opportunity, error rate parity does not distinguish between the
+*type* of mistake being made. It only asks whether one group is, on
+the whole, being predicted for less accurately than another.
+
```
romanlutz, line 376: Nit: "being predicted for less accurately" reads awkwardly. Suggest: "…whether the model is, on the whole, less accurate for one group than another."

docs/user_guide/assessment/common_fairness_metrics.rst
```diff
@@ -410,31 +517,37 @@ which can be summed up as follows:
       are much less likely to cause harm than missed true positives, equal
       opportunity may be useful
 
-    * *Caveats:* If there are historical biases in the data, then the original labels
-      may hold little value. A large imbalance between the positive and negative classes
-      will also accentuate any statistical issues related to sensitive groups with low
-      membership
+    * *Caveats:* Same as equalized odds. In addition, since it ignores false
+      positives entirely, it should not be relied on alone in settings where
+      false positives carry meaningful costs
```
romanlutz, line 522: This bullet's *Caveats* used to read: "If there are historical biases in the data, then the original labels may hold little value. A large imbalance between the positive and negative classes will also accentuate any statistical issues related to sensitive groups with low membership." You've replaced it with "Same as equalized odds. In addition, …", which is unrelated to the PR's stated scope of *adding* two metrics and quietly drops information that was previously spelled out.

Could you either keep the original wording and just append the new "ignores false positives entirely" sentence, or — if you prefer the compressed style — apply it consistently and leave the substantive edits for a separate PR?

docs/user_guide/assessment/common_fairness_metrics.rst
```diff
@@ -304,7 +304,114 @@ However, by not considering whether false
...
+    >>> from fairlearn.metrics import MetricFrame, false_positive_rate
+    >>> mf = MetricFrame(metrics=false_positive_rate,
+    ...                  y_true=y_true,
+    ...                  y_pred=y_pred,
+    ...                  sensitive_features=sf_data)
+    >>> print(mf.difference())
+    1.0
+    >>> print(mf.ratio())
+    0.0
+
```
romanlutz, line 360: With the shared fixture at the top of the file, group `b` has exactly one negative-class sample and it's correctly classified, so its FPR is 0. That's why this doctest prints `difference == 1.0` and `ratio == 0.0` — technically correct, but pedagogically weak: the reader sees the worst-possible gap and what looks like a divide-by-zero, with no gradient between "no disparity" and "worst-case disparity."

Not blocking, but consider adding a sentence after the doctest noting the perfect `1.0`/`0.0` result reflects the small toy dataset — otherwise it reads like a bug. Modifying the shared fixture would ripple through every other doctest on the page and is out of scope here.

docs/user_guide/assessment/common_fairness_metrics.rst
```diff
@@ -410,31 +517,37 @@ which can be summed up as follows:
...
-.. [#f2] Epistemic trespassing is the process of taking expertise in one field and
-         applying it to another in which one does not have an equivalent (or any)
-         competence.
-         This is not an intrinsically bad thing - one could label all
-         interdisciplinary research a form of epistemic trespassing.
-         However, doing so successfully requires a willingness to learn the subtleties
-         of the new field.
+However, the fact these are common metrics does not make them applicable to any given
+situation.
+In particular, :ref:`demographic parity is often misapplied <assessment_four_fifths>`.
```
romanlutz, line 553: This PR silently removes the `.. topic:: References` / `.. footbibliography::` block and the `.. rubric:: Footnotes` with `[#f1]` and `[#f2]` that used to sit at the end of this file. Two consequences:

1. `[#f1]_` (line 441) and `[#f2]_` (line 469) are now unresolved footnote references — Sphinx will warn and, with `-W`, fail the build.
2. Every `:footcite:` citation on this page — including the two you add (`hardt2016equality`, `agarwal2018reductions`) plus the five pre-existing ones — no longer has a `footbibliography` directive to render into, so the bibliography section vanishes from the rendered page.

This looks unintentional (the file also lost its trailing newline, see other comment). Please restore the deleted block verbatim at the end of the file and rebuild the docs to confirm no `undefined label` / footnote / citation warnings appear.

docs/user_guide/assessment/common_fairness_metrics.rst
```diff
@@ -410,31 +517,37 @@ which can be summed up as follows:
...
 
-.. [#f2] Epistemic trespassing is the process of taking expertise in one field and
-         applying it to another in which one does not have an equivalent (or any)
-         competence.
-         This is not an intrinsically bad thing - one could label all
-         interdisciplinary research a form of epistemic trespassing.
-         However, doing so successfully requires a willingness to learn the subtleties
-         of the new field.
+However, the fact these are common metrics does not make them applicable to any given
+situation.
```
romanlutz, line 552: Nit but worth fixing while you're here: the diff reports `\ No newline at end of file`. Every other `.rst` in the tree ends with a newline — this is almost certainly a side effect of the same accidental truncation that dropped the References block above. Please re-add the trailing `\n`.

docs/user_guide/assessment/common_fairness_metrics.rst
```diff
@@ -304,7 +304,114 @@ However, by not considering whether false
...
+.. _assessment_false_positive_rate_parity:
+
+False positive rate parity
+---------------------------
+
+False positive rate parity is the mirror image of equal opportunity:
+instead of requiring equal *true* positive rates across groups, it
+requires equal *false* positive rates. Another way of thinking about
+this metric is requiring equal treatment only within the subset of
+records belonging to the negative class. In a pre-trial risk
```
romanlutz, line 316: Framing question, not blocking: FPR-equality isn't really a new concept at this point in the page. The equalized-odds section above (lines 219–229) already defines it as one of the two components (`P(h(X)=1 | Y=0)` equal across groups), and the equal-opportunity section explicitly calls out its absence as the relaxation. So this section reads a little oddly starting from scratch — as if FPR parity were unrelated to what came before.

Two options, both cleaner than the current structure:

1. **Reframe as a relaxation of equalized odds**, mirroring the equal-opportunity section — something like "False positive rate parity is a relaxed version of equalized odds that only considers conditional expectations with respect to negative labels, i.e., :math:`Y=0`. Another way of thinking about this metric is requiring equal outcomes only within the subset of records belonging to the negative class…" That parallels the equal-opportunity opening almost word-for-word and makes the TPR / FPR / EO relationship obvious.

2. **Reorder: put false positive rate parity *before* equal opportunity** (i.e., EO → FPR parity → equal opportunity, or EO → equal opportunity + FPR parity together as sibling relaxations). Then the two Y=1 / Y=0 relaxations sit side-by-side under equalized odds, which matches how the mitigation-side `reductions.rst` already groups True/False Positive Rate Parity together.

I'd lean toward option 1 for the smallest structural change, or option 2 (a joint "TPR / FPR parity" section) if you're open to reorganizing. Either way, the current standalone-and-later placement makes readers re-derive a connection the page already established.

romanlutz (changes requested):

PR #1698, opened 2026-08-10
fix: raise on missing values in MetricFrame sensitive and control features

### LINKED ISSUE #1155: More informative error message missing values in MetricFrame sensitive_features
#### Describe the bug
When a sensitive feature of containing strings in a `pd.Series` / `np.array` / `pd.DataFrame` has missing values, `MetricFrame` throws a rather uninformative error. 

The same error does not occur when you first convert the data to a list (`list(sf)`), in that case `nan` is just interpreted as a sensitive category.

#### Steps/Code to Reproduce

```python
import pandas as pd
import numpy as np
from fairlearn.metrics import MetricFrame, count

sf = pd.Series(['a', 'a', 'a', 'a', 'b', 'b', np.nan])

mf = MetricFrame(metrics={'count' : count},
                 y_true = [1, 2, 3, 5, 6, 7, 8],
                 y_pred = [1, 2, 3, 4, 5, 6, 7],
                 sensitive_features=sf,
                )
mf.by_group
```

#### Expected Results
An error message that hints towards having supplied a missing value.

#### Actual Results
A `TypeError: '<' not supported between instances of 'float' and 'str'`

```
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-24-bb38ac75fac0> in <module>
      5 sf = pd.Series(['a', 'a', 'a', 'a', 'b', 'b', np.nan])
      6 
----> 7 mf = MetricFrame(metrics={'count' : count},
      8                  y_true = [1, 2, 3, 5, 6, 7, 8],
      9                  y_pred = [1, 2, 3, 4, 5, 6, 7],

~/opt/anaconda3/lib/python3.8/site-packages/fairlearn/metrics/_metric_frame.py in compatible_metric_frame_init(self, metric, *args, **kwargs)
     90         # Call the new constructor with positional arguments passed as keyword arguments
     91         # and with the `metric` keyword argument renamed to `metrics`.
---> 92         new_metric_frame_init(self, **metric_arg_dict, **positional_dict, **kwargs)
     93 
     94     return compatible_metric_frame_init

~/opt/anaconda3/lib/python3.8/site-packages/fairlearn/metrics/_metric_frame.py in __init__(self, metrics, y_true, y_pred, sensitive_features, control_features, sample_params)
    296 
    297         # Now, prepare the sensitive features
--> 298         sf_list = self._process_features("sensitive_feature_", sensitive_features, y_t)
    299         self._sf_names = [x.name_ for x in sf_list]
    300 

~/opt/anaconda3/lib/python3.8/site-packages/fairlearn/metrics/_metric_frame.py in _process_features(self, base_name, features, sample_array)
    711         if isinstance(features, pd.Series):
    712             check_consistent_length(features, sample_array)
--> 713             result.append(GroupFeature(base_name, features, 0, None))
    714         elif isinstance(features, pd.DataFrame):
    715             for i in range(len(features.columns)):

~/opt/anaconda3/lib/python3.8/site-packages/fairlearn/metrics/_group_feature.py in __init__(self, base_name, feature_vector, index, name)
     49     def __init__(self, base_name: str, feature_vector, index: int, name: Optional[str]):
     50         """Help with the metrics."""
---> 51         self.classes_ = np.unique(feature_vector)
     52         self.raw_feature_ = feature_vector
     53 

<__array_function__ internals> in unique(*args, **kwargs)

~/opt/anaconda3/lib/python3.8/site-packages/numpy/lib/arraysetops.py in unique(ar, return_index, return_inverse, return_counts, axis)
    270     ar = np.asanyarray(ar)
    271     if axis is None:
--> 272         ret = _unique1d(ar, return_index, return_inverse, return_counts)
    273         return _unpack_tuple(ret)
    274 

~/opt/anaconda3/lib/python3.8/site-packages/numpy/lib/arraysetops.py in _unique1d(ar, return_index, return_inverse, return_counts)
    331         aux = ar[perm]
    332     else:
--> 333         ar.sort()
    334         aux = ar
    335     mask = np.empty(aux.shape, dtype=np.bool_)

TypeError: '<' not supported between instances of 'float' and 'str'
```

#### Versions
fairlearn 0.8.0

aiedwardyi:
## Description

Closes #1155.

A missing value in a sensitive or control feature was silently dropped from `by_group` while still counting towards `overall` - 7 rows in, 6 counted, no warning. `MetricFrame` now raises a `ValueError` naming the offending feature.

Picking up the thread on the issue: @riedgar-ms asked whether to warn or reject, and @hildeweerts said loud failure is fine as long as it is easy to diagnose, so this rejects. Happy to switch it to a warning if you'd rather not change behaviour for existing callers.

@hildeweerts also flagged that it should not matter whether the feature is a `Series` or a `list`. It did: `np.asarray(["a", "b", np.nan])` promotes to `<U32`, so a list-supplied missing value became a category literally named `'nan'`. The list branch now builds an object array, matching what the non-list branch already does, and `Series`, `list` and `ndarray` all behave the same.

## Tests
- [x] new tests added

Both new tests were confirmed to fail before the fix. Full unit suite: 6582 passed, 0 failed.

## Documentation

## Screenshots
n/a

Review fixes applied: 2c94729 - validate every list element before array conversion.


riedgar-ms (commented):
I'm happy to make this as a (technically) breaking change.

aiedwardyi:
Sounds good - keeping the ValueError.

aiedwardyi:
Thanks @riedgar-ms - anything else you'd like changed here, or is this good to go?

The one red check is a codecov upload failure, not a test failure: in that job the pytest step passed (6799 passed, 0 failed) and the job then failed on `Upload coverage reports to Codecov`. The Windows job was cancelled by matrix fail-fast as a knock-on.

riedgar-ms (approved):
I'm OK with this, although I see that @romanlutz has also been making some MetricFrame updates. Tagging @hildeweerts too

fairlearn/metrics/_metric_frame.py
```diff
@@ -991,7 +991,9 @@ def _process_features(self, base_name, features, sample_array) -> list[GroupFeat
                 result.append(GroupFeature(base_name, column, i))
         elif isinstance(features, list):
             if np.isscalar(features[0]):
-                f_arr = np.atleast_1d(np.squeeze(np.asarray(features)))
+                # dtype=object keeps a missing value from being cast to the
+                # string 'nan' and becoming a category of its own
+                f_arr = np.atleast_1d(np.squeeze(np.asarray(features, dtype=object)))
```
romanlutz, line 996: Could we use `pd.api.types.is_scalar(features[0])` here, or otherwise handle missing scalar sentinels before this branch? `np.isscalar(None)`, `np.isscalar(pd.NA)`, and `np.isscalar(pd.NaT)` are all `False`, so the result still depends on where the missing value appears:

```python
[None, "a", "b"]  # "Feature lists must be of scalar types"
["a", None, "b"]  # intended named missing-value error
```

That leaves the list behavior inconsistent. It may also be worth validating every element rather than only the first.

aiedwardyi: Fixed in 2c94729 - validate every list element before array conversion.

riedgar-ms (approved):
Thank you!

PR #1708, opened 2026-08-17
MNT Make OpenML tests optional for pull requests

romanlutz:
## Summary
- mark unit tests that download OpenML datasets and run them in a dedicated workflow
- make OpenML failures non-blocking on pull requests while keeping them strict on `main` and scheduled runs
- keep other-ML compatibility tests required by switching their smoke-test input to data bundled with scikit-learn
- defer the adversarial preprocessing OpenML download until the marked tests actually run

## Validation
- required cases in affected unit modules: 15 passed, 39 deselected
- OpenML marker collection: 39 tests collected
- LightGBM compatibility tests: 3 passed
- XGBoost compatibility tests: 3 passed
- PyTorch compatibility tests: 5 passed
- Ruff checks and formatting passed for changed Python files


.github/workflows/openml.yml
```diff
@@ -0,0 +1,41 @@
...
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.13"
+      - name: Cache OpenML datasets
+        uses: actions/cache@v4
+        with:
+          path: ${{ env.DATA_HOME }}
+          key: ${{ runner.os }}-openml-datasets-v1
```
riedgar-ms: Shouldn't the datasets be OS-agnostic? Or is dealing with that sort of thing a future PR?

riedgar-ms (approved):
No serious objections. Surprised that the cache isn't smoothing things out as it is, though.

.github/workflows/openml.yml
romanlutz: Yes, the dataset contents are OS-agnostic. This workflow currently runs only on Ubuntu, so `${{ runner.os }}` does not affect cache reuse today. If we expand it to an OS matrix, sharing the archive across platforms would also require `enableCrossOsArchive`; removing the OS prefix alone would not be sufficient. I’d handle that when adding another OS.

romanlutz: I dug into the failed runs. The cache was restored, but all scikit-learn versions shared the same immutable `openml-datasets-v1` key. Different versions use different cache paths—for example, scikit-learn 1.6 expects `download/1595261.gz`, while 1.9 expects `download/1595261/adult.arff.gz`. An exact cache hit cannot be updated, so missing files still triggered OpenML requests and failed with HTTP 504.

I’ve updated the key to include the resolved scikit-learn version and a hash of the dataset/test definitions, with a same-version fallback for refreshes. The optional workflow still protects PRs during cold-cache outages.

PR #1711, opened 2026-08-19
MNT enforce 2D inputs for fitted adversarial estimators

### LINKED ISSUE #1672: Adversarial predict()/partial_fit() don't catch non-2D X after the model is already fitted
## Context

#1656 / #1671 fixed the case where `BackendEngine.__init__` builds the predictor model using `X.shape[1]` without checking `X.ndim`, which caused a silently mis-sized model when `X` had more than two dimensions. That fix only covers the first call, when the model is being constructed.

There's a related gap on the paths that run after a model is already fitted:

- `predict()` -> `_raw_predict()` calls `validate_data(..., allow_nd=True, reset=False)`, then passes `X` straight to `self.backendEngine_.evaluate(X)`.
- `partial_fit()` on a model that's already been fit once goes through `_validate_input`, then calls `self.backendEngine_.train_step(X, y, A)` directly, bypassing `BackendEngine.__init__` (and therefore the new ndim check) entirely.

I checked what actually happens today if `X` passed to `predict()` has more dimensions than the `X` used for `fit()`:

- `sklearn`'s own feature-count check (`reset=False`) does not reliably reject non-2D arrays.
- In the PyTorch backend, `evaluate()` passes `X` straight into a `torch.nn.Linear`-based model. `Linear` operates on the last dimension, so in most cases a shape mismatch will raise a `RuntimeError` from PyTorch itself. But if the last dimension of the bad `X` happens to coincidentally match `n_features_in_`, `Linear` will silently apply itself across the extra dimensions instead of raising, producing an output with an unexpected shape rather than failing loudly.

## Suggested action

Decide on a single place to enforce "X must be 2D for the adversarial module" (or intentionally support richer input) and apply it consistently across `fit`, `partial_fit`, and `predict`, not just at model construction time. This is a bigger discussion than a single line fix since it touches how `allow_nd=True` is used across `_adversarial_mitigation.py`.

Found while working on #1656 / #1671.

romanlutz:
## Description

Closes #1672.

Already-fitted adversarial estimators could accept non-2D `X` through `predict()` and `partial_fit()` because those paths did not return to the `BackendEngine.__init__` guard added in #1671. This extracts that guard into a shared `_check_2d` helper and applies it consistently before backend evaluation or training, including when `skip_validation=True`.

The implementation preserves the Fairlearn-owned `_X_NOT_2D` message discussed in #1672 rather than relying on scikit-learn's different wording.

@SpiliosDimakopoulos @NotArnav03, your feedback would be especially helpful since this follows the approach discussed on the issue.

## Tests

- [ ] no new tests required
- [x] new tests added
- [ ] existing tests adjusted

Added regression coverage for fitted `predict()`, ordinary `partial_fit()`, and `partial_fit(skip_validation=True)`. The tests use a 3D input whose feature dimensions match the fitted model and verify that backend evaluation/training is never reached.

`293 passed, 5 skipped, 6 xfailed, 6 xpassed` in `test/unit/adversarial`.

## Documentation

Updated the v0.15.0 bug-fix notes.

## Screenshots

Not applicable.


NotArnav03:
Yes, this matches the approach. Thanks for picking it up, and apologies for the dangling thread: I said on #1672 that I'd folded the follow-up into #1671, but it never actually made it onto the branch, so #1671 merged with just the `__init__` guard. #1711 is the piece I'd described.

Confirming the design: a shared `_check_2d` raising `_X_NOT_2D` is what I had in mind, and placing it outside the `if not self.skip_validation:` block in `_validate_input` is the right call, since that's the one path with no other validation at all. Three suggestions before merge.

**1. `X.ndim` isn't safe on the `skip_validation=True` path.** That path never reaches `validate_data`, so `X` is whatever the caller passed, and `ndim` is an ndarray/DataFrame/sparse attribute rather than a general one:

```python
>>> [1, 2].ndim
AttributeError: 'list' object has no attribute 'ndim'
```

List input therefore raises `AttributeError` instead of the friendly `ValueError`, on exactly the path where there is no other validation to catch it. `np.ndim(X) != 2` covers all of these (`np.ndim([1, 2]) == 1`) and is a one-word change.

**2. Keep the rationale next to the helper.** `_check_2d` is deliberately independent of `ensure_2d`/`allow_nd`, which is the whole reason it exists rather than just flipping `allow_nd=False`. As written, `allow_nd=True` sits directly above `_check_2d(X)` in both `_raw_predict` and `_validate_input`, which reads like a contradiction, and I suspect someone will eventually delete one of them and silently reopen #1656. A short docstring would prevent that:

```python
def _check_2d(X):
    """Raise a clear error if ``X`` is not a two-dimensional array.

    Single source of truth for the "X must be 2D" requirement. Deliberately
    independent of scikit-learn's ``ensure_2d``/``allow_nd`` flags: those only
    reject 0D/1D input (and only reject >=3D when ``allow_nd`` is False), so the
    guarantee lives here rather than depending on a validation kwarg staying set
    a particular way.
    """
```

**3. Two small ones.** The new test parametrizes `fake_backend_env` with `["torch"]` only, while `test_backend_engine_rejects_non_2d_X` just above it covers `["torch", "tensorflow"]`, so those are worth keeping symmetric. And the changelog entry stops at the issue role, whereas the other `v0.15.0.rst` entries carry PR and author attribution:

```rst
... including when input validation is skipped: :pr:`1711` by :user:`Roman Lutz <romanlutz>`.
```

Happy to push any of these as a patch if that's easier than round-tripping.

romanlutz:
Accepted all suggestions. I switched the helper to NumPy dimensionality handling, documented why it is independent of scikit-learn validation, added list-input and symmetric backend coverage, and updated the changelog attribution.

NotArnav03:
Thanks, this all looks right to me. Good catch adding the guard in `partial_fit` before the `n_features_in_` comparison, I'd missed that `X.shape` would break there on list input.

One thing for a possible follow-up rather than this PR: on the `predict` path, `validate_data` still rejects 1D input with sklearn's wording before `_check_2d` runs, so 3D gets `_X_NOT_2D` while 1D gets "Expected 2D array, got 1D array instead". Not worth holding this up, and closing it properly means dropping `ensure_2d` so `_check_2d` is the sole gate, which is a bigger change than belongs here.

LGTM.

riedgar-ms (approved):
Looks reasonable

PR #1712, opened 2026-08-22
fix: enforce X.ndim == 2 in _AdversarialFairness.predict and repeated fit/partial_fit

### LINKED ISSUE #1672: Adversarial predict()/partial_fit() don't catch non-2D X after the model is already fitted

SpiliosDimakopoulos:
Fixes #1672

## Problem
#1656 / #1671 added a check in `BackendEngine.__init__` that raises a
clear `ValueError` when `X.ndim != 2` — but that check only runs when
the backend engine is *constructed*. Two paths bypass it entirely:

- `partial_fit()` on an already-fitted model (and a repeated, non
  warm-start call to `fit()`) reuse the existing `backendEngine_` and
  call `backendEngine_.train_step(X, y, A)` directly.
- `predict()` -> `_raw_predict()` validates `X` with `allow_nd=True`
  and passes it straight to `backendEngine_.evaluate(X)`.

In both cases a non-2D `X` (e.g. an image batch) either raises an
opaque backend-specific error, or — worse — is silently mis-applied if
the trailing dimension happens to match `n_features_in_`.

## Fix
Reuse the existing `_X_NOT_2D` message from #1671 and add the same
`ndim` check at the two remaining choke points:
- `_AdversarialFairness._validate_input` (shared by `fit`/`partial_fit`)
- `_AdversarialFairness._raw_predict` (used by `predict`)

## Tests
- Added `test_partial_fit_rejects_non_2d_X_on_second_call`
- Added `test_predict_rejects_non_2d_X`
- Full adversarial test suite passes locally (274 passed, 5 skipped,
  6 xfailed, 6 xpassed); `ruff check` / `ruff format --check` clean.

## Changelog
Added an entry to `docs/user_guide/installation_and_version_guide/v0.15.0.rst`
following the existing convention.

### DIFF
--- a/docs/user_guide/installation_and_version_guide/v0.15.0.rst
+++ b/docs/user_guide/installation_and_version_guide/v0.15.0.rst
@@ -37,6 +37,12 @@ Bug fixes
   :code:`FloatTransformer.inverse_transform` so that they raise
   :code:`NotFittedError` instead of :code:`AttributeError` when called before
   :code:`fit`: :pr:`1709` by :user:`SpiliosDimakopoulos`.
+* Extended the :code:`X` dimensionality check added in :pr:`1671` to
+  :code:`_AdversarialFairness.predict` and to repeated/:code:`partial_fit`
+  calls, which previously bypassed :code:`BackendEngine.__init__` (and
+  therefore the check) entirely and could silently mis-size predictions or
+  hand non-2D data straight to the backend model: :pr:`REPLACE_WITH_PR_NUMBER`
+  by :user:`REPLACE_WITH_YOUR_GITHUB_USERNAME`.
 
 Documentation
 -------------
--- a/fairlearn/adversarial/_adversarial_mitigation.py
+++ b/fairlearn/adversarial/_adversarial_mitigation.py
@@ -31,6 +31,7 @@ from ._constants import (
     _KWARG_ERROR_MESSAGE,
     _PREDICTION_FUNCTION_AMBIGUOUS,
     _PROGRESS_UPDATE,
+    _X_NOT_2D,
 )
 from ._preprocessor import FloatTransformer
 from ._pytorch_engine import PytorchEngine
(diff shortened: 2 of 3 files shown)


romanlutz:
Thanks for working on this. I compared this with #1711, and #1711 already covers the same fitted-model paths from #1672. It also handles skip_validation=True, checks dimensionality before partial_fit reads X.shape[1], supports array-like input, and verifies that invalid input never reaches the backend. I am closing this PR as a duplicate in favor of #1711.

PR #1715, opened 2026-08-24
FIX Handle linprog solver failures

romanlutz:
`solve_linprog` previously read `linprog` outputs without checking whether SciPy reported a successful solve. If either LP failed, invalid solver output could silently propagate into the reduction.

This change adds explicit success checks for both the primal and dual `linprog(method="highs-ds")` calls. On failure, `solve_linprog` now raises a `RuntimeError` that includes the problem type, method, solver status, and solver message. Successful solves keep the existing behavior.

Targeted tests monkeypatch `linprog` to return unsuccessful `OptimizeResult` values for both primal and dual failure paths.

Tested with:

```bash
uv run pytest test\unit\reductions\exponentiated_gradient\test_lagrangian.py -q
```

### DIFF
--- a/fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
+++ b/fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
@@ -26,6 +26,23 @@ logger = logging.getLogger(__name__)
 _MESSAGE_BAD_OBJECTIVE = (
     "Objective needs to be of the same type as constraints. Objective is {}, constraints are {}."
 )
+_MESSAGE_LINPROG_FAILED = (
+    "Linear programming failed while solving the {} problem with method {}. "
+    "Status: {}. Message: {}"
+)
+_LINPROG_METHOD = "highs-ds"
+
+
+def _raise_if_linprog_failed(result, problem):
+    if not result.success:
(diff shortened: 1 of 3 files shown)


fairlearn/reductions/_exponentiated_gradient/_lagrangian.py
```diff
@@ -26,6 +26,23 @@
...
+def _raise_if_linprog_failed(result, problem):
+    if not result.success:
+        raise RuntimeError(
+            _MESSAGE_LINPROG_FAILED.format(
+                problem,
+                _LINPROG_METHOD,
+                result.status,
+                result.message,
+            )
+        )
```
taharallouche, line 45:
```suggestion
def _raise_if_linprog_failed(result: OptimizeResult, problem: Literal["primal", "dual"]) -> None:
    if not result.success:
        raise RuntimeError(
            _MESSAGE_LINPROG_FAILED.format(
                problem,
                _LINPROG_METHOD,
                result.status,
                result.message,
            )
        )
```

taharallouche (approved):
Thank you @romanlutz 🫡 just one minor suggestion to add type annotations

PR #1716, opened 2026-08-24
CI: test against scikit-learn nightly builds

### LINKED ISSUE #965: Include a CI step against scikit-learn's nightly builds
I suggest we include a test which would be optional on PR merges, or even not at all on PRs, and rather a cron job, which tests against scikit-learn nightly builds, to be able to detect situations such as the ones we're having in https://github.com/fairlearn/fairlearn/issues/964 and https://github.com/fairlearn/fairlearn/issues/963

The nightly builds can be conveniently installed as:

    pip install --pre --extra-index https://pypi.anaconda.org/scipy-wheels-nightly/simple scikit-learn

@romanlutz @riedgar-ms WDYT?

quinnarnold:
## Description

Adds a scheduled CI job that runs the full Fairlearn test suite against the latest scikit-learn nightly wheel. The existing all-dependencies workflow gains an optional input so regular callers remain unchanged.

Closes #965.

AI-assisted: Yes.

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

Validated the nightly installation with scikit-learn 1.10.dev0 and ran the full non-OpenML suite on Python 3.14: 6,835 passed with no failures.

## Documentation

## Screenshots

Not applicable.

### DIFF
--- a/.github/workflows/test-all-deps.yml
+++ b/.github/workflows/test-all-deps.yml
@@ -48,6 +48,8 @@ jobs:
           pip install --upgrade --pre --extra-index
           https://pypi.anaconda.org/scientific-python-nightly-wheels/simple
           scikit-learn
+      - if: ${{ inputs.sklearnNightly }}
+        run: python -c "import sklearn; assert '.dev' in sklearn.__version__, sklearn.__version__"
       - run: pip install matplotlib
       - run: >
           pytest --mpl
--- a/.github/workflows/test-all-deps.yml
+++ b/.github/workflows/test-all-deps.yml
@@ -13,6 +13,10 @@ on:
         required: true
         type: string
         default: "3.13"
+      sklearnNightly:
+        required: false
+        type: boolean
+        default: false
 
 jobs:
   all-deps:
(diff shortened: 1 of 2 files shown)


quinnarnold:
This PR does not warrant a changelog entry because it only changes internal CI workflow configuration and does not affect Fairlearn user-facing behavior, APIs, or release functionality. The changelog check appears to be triggered because `.github/workflows/test-all-deps.yml` matches its broad test-file path check. Could a maintainer please apply the `No Changelog Needed` label?

.github/workflows/test-all-deps.yml
```diff
@@ -38,6 +42,12 @@ jobs:
           python-version: ${{ inputs.python }}
       - run: python scripts/install_requirements.py
       - run: pip install .
+      - name: Install scikit-learn nightly build
+        if: ${{ inputs.sklearnNightly }}
+        run: >
+          pip install --upgrade --pre --extra-index
+          https://pypi.anaconda.org/scientific-python-nightly-wheels/simple
+          scikit-learn
```
romanlutz, line 50: `--extra-index` still searches PyPI as well as the nightly index. If the nightly index is unavailable, has no compatible Python 3.14 wheel, or briefly trails a stable release, pip can keep or install stable scikit-learn and this job will still pass without testing the nightly build.

Could we verify the installed version after this step? For example:

```yaml
- run: python -c "import sklearn; assert '.dev' in sklearn.__version__, sklearn.__version__"
```

Let me know if I misunderstood how the nightly index is expected to be monitored.

quinnarnold: You understood correctly. I added the conditional post-install assertion in 7b27bdd so the job now fails if pip resolves a stable scikit-learn build. I also verified the assertion against the current Python 3.14 nightly wheel (`scikit-learn 1.10.dev0`).

romanlutz (approved):

PR #1719, opened 2026-08-26
DOC: add missing Raises sections to equalized_odds_difference and equalized_odds_ratio

SpiliosDimakopoulos:
## What does this change?

`equalized_odds_difference` and `equalized_odds_ratio` both raise a
`ValueError` when `agg` receives an invalid value, but this was not
documented in their docstrings. Users reading the API reference had no
way to know this exception could be raised.

Adds numpydoc-compliant `Raises` sections to both functions, consistent
with the pattern already used in `fairlearn/preprocessing/`.

## Files changed

- `fairlearn/metrics/_fairness_metrics.py`


taharallouche (approved):
Looks good to me 👍

PR #1721, opened 2026-08-30
DOC Improve ExponentiatedGradient guidance

### LINKED ISSUE #988: DOC Add user guide for ExponentiatedGradient

There is currently no user guide for `ExponentiatedGradient`.

Add the user guide for `ExponentiatedGradient`.

quinnarnold:
## Description

Closes #988.

Improve the ExponentiatedGradient user guide with a concise, intuition-first explanation of how the reduction builds a randomized mixture of predictors. Add practical epsilon guidance that distinguishes the fairness constraint's `difference_bound` from `ExponentiatedGradient.eps`, suggests a starting scale, and recommends validating the fairness-performance tradeoff on held-out data.

## Tests

- [x] no new tests required
- [ ] new tests added
- [ ] existing tests adjusted

Validated with a complete Sphinx dummy build using the repository development requirements.

## Documentation

## Screenshots

Not applicable; this is a prose-only documentation update.

### DIFF
--- a/docs/user_guide/mitigation/reductions.rst
+++ b/docs/user_guide/mitigation/reductions.rst
@@ -672,6 +672,10 @@ The :class:`ExponentiatedGradient` algorithm in Fairlearn is used to produce mod
 satisfy fairness constraints without needing access to sensitive features at deployment time.
 This algorithm creates a sequence of re-weighted datasets and retrains the
 wrapped classifier on each of these datasets.
+Intuitively, each round increases the emphasis on constraints that the current predictors
+violate, so the base estimator focuses on examples that help reduce those violations. The
+result is a randomized mixture of predictors: :meth:`ExponentiatedGradient.predict` samples
+from this mixture, and repeated calls may therefore produce different predictions.
 To instantiate an :class:`ExponentiatedGradient` model, we need to pass in a base estimator
 and fairness constraints. The fairness constraints are typically specified by providing
 an upper bound on the difference (or the ratio) between the largest and the smallest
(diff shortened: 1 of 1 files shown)


rabiya43:
Hi @quinnarnold, thanks for this. The `difference_bound` vs `eps` distinction is a real source of confusion, and the new "Choosing epsilon" section handles it well.

I ran a quick experiment on a small synthetic dataset (LogisticRegression + DemographicParity, ~4000 rows, results averaged over 30 `predict` draws) to see how the two parameters behave. A few optional observations:

1. **Your distinction holds up empirically.** `difference_bound` limits how far each group's selection rate can be from the overall rate, so I measured that quantity (the largest deviation from the overall rate). The unconstrained baseline sat at 0.192. With `eps` at its default, `difference_bound` = 0.1 / 0.05 / 0.02 / 0.005 gave 0.103 / 0.050 / 0.0225 / 0.0052, so the result tracks the bound closely. With `difference_bound=0.02` fixed, varying `eps` from 0.5 down to 0.005 changed it by less than 0.001 (0.0220 to 0.0225). So `difference_bound` is the knob users should tune, which supports your wording.
2. **The intro paragraph before the example** (not touched by this PR) still says "This bound is often referred to as epsilon", even though `eps` is a separate parameter, and the clarification only comes after the example. Maybe reword that sentence or point down to "Choosing epsilon"? The same sentence also describes the bound as being on the difference between the largest and smallest value across groups, whereas the `UtilityParity` docstring says the constraints compare each group with the overall population, so that might be worth aligning too.
3. **Reproducibility.** You mention that repeated `predict` calls can differ. It might be worth adding that `predict(X, random_state=<int>)` makes the output reproducible, and that the fitted `predictors_` and `weights_` can be inspected.
4. **`sample_weight_name`.** The example passes `sample_weight_name="classifier__sample_weight"` without explanation. A one-line note that a `Pipeline` needs this to route sample weights to its final step could help. Without it, `fit` raises `ValueError: Pipeline.fit does not accept the sample_weight parameter`.

Happy to hear that any of this is out of scope for this PR. I haven't looked at the rendered page, so this is just feedback on the text.

